PR #37, opened 2017-06-27
Send a version file to the robot on install.

Phasedd:
The version file contains the `git log -1` output to show the latest commit hash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,4 @@ tags
 .DS_Store
 .idea
 *.sublime-workspace
+version.txt
--- a/.gitignore
+++ b/.gitignore
@@ -74,4 +74,3 @@ tags
 .DS_Store
 .idea
 *.sublime-workspace
-version.txt
--- a/tools/install.py
+++ b/tools/install.py
@@ -6,10 +6,9 @@ import glob
 import hashlib
 import tempfile
 import b
-import subprocess
 
 from termcolor import cprint
-from subprocess import call, STDOUT
+from subprocess import call, STDOUT, Popen
 
 def register(command):
 
@@ -77,7 +76,7 @@ def run(ip_addr, hostname, config, scripts, user, toolchain, **kwargs):
 
     # Get list of config files.
     config_files = [os.path.relpath(c, build_dir) for c in b.cmake_cache['NUCLEAR_MODULE_DATA_FILES']]
-    
+
     # Get list of config files.
     if config in ['overwrite', 'o']:
         cprint('Overwriting configuration files on target', 'blue', attrs=['bold'])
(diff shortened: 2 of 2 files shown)


tayloryoung6396:
Seems to work...

tools/install.py
```diff
@@ -92,3 +93,9 @@ def run(ip_addr, hostname, config, scripts, user, toolchain, **kwargs):
 
     if config in ['ignore', 'i']:
         cprint('Ignoring configuration changes', 'blue', attrs=['bold'])
+        
+    # Pipe the git commit to file
+    versionFile = open("version.txt", "w")
```
TrentHouliston: You should match your variable style with the rest of the file. This should be version_file

tools/install.py
```diff
@@ -92,3 +93,9 @@ def run(ip_addr, hostname, config, scripts, user, toolchain, **kwargs):
...
+    subprocess.Popen("git log -1", stdout=versionFile, shell=True)
```
TrentHouliston: Don't use shell=true unless you really really need to. It can represent a security risk. Instead the first argument to this function should be an array of strings.

tools/install.py
```diff
@@ -6,6 +6,7 @@
 import hashlib
 import tempfile
 import b
+import subprocess
```
TrentHouliston: This was already imported below. Just fix that import line

tools/install.py
```diff
@@ -76,7 +77,7 @@ def run(ip_addr, hostname, config, scripts, user, toolchain, **kwargs):
 
     # Get list of config files.
     config_files = [os.path.relpath(c, build_dir) for c in b.cmake_cache['NUCLEAR_MODULE_DATA_FILES']]
-
+
```
TrentHouliston: Your sublime should be removing trailing white space? Are you editing in sublime with the project file open?

tools/install.py
TrentHouliston: Also you can just do this with subprocess call you don't need popen. In fact that might be dangerous as it's asynchronous.

.gitignore
```diff
@@ -74,3 +74,4 @@ tags
 .DS_Store
 .idea
 *.sublime-workspace
+version.txt
```
TrentHouliston: You should be generating this file in the build dir not here. So it shouldn't be git ignored.

tools/install.py
TrentHouliston: Also you should be generating this file in the build directory. You can get that off the b object.

Phasedd:
Should probably be tested on the actual robot before we merge (if the PR is accepted).

The original commit worked, and I tested outside of the install.py to test if the file was being created with the contents.

tools/install.py
```diff
@@ -92,3 +92,9 @@ def run(ip_addr, hostname, config, scripts, user, toolchain, **kwargs):
 
     if config in ['ignore', 'i']:
         cprint('Ignoring configuration changes', 'blue', attrs=['bold'])
+
+    # Pipe the git commit to file
+    version_file = open(os.path.join(build_dir, "version.txt"), "w")
```
TrentHouliston: You should use with syntax here rather than open and close

PR #39, opened 2017-06-28
Removes the by reference toupper function.

TrentHouliston:
Removing the by reference toupper function as it is very much a Cism.


tayloryoung6396:
You know the firmware installer doesn't work without this

Bidski:
@tayloryoung6396 this will probably actually work better ( @TrentHouliston has actually written the `toUpper` function correctly by the looks of it :P )

PR #40, opened 2017-06-28
Enable travis slack integration for NUbots

TrentHouliston:
And also disable emails


tayloryoung6396:
Does it work?

TrentHouliston:
It does

PR #43, opened 2017-07-07
Firmware Revert and Camera Reset

CptTony57:


TrentHouliston:
When doing a pull request, you really should only have it do one thing.
This PR is also doing a bunch of stuff with scripts and camera reset.
Are you good enough with git to separate out the individual commits into separate PRs?

CptTony57:
I'll give it a shot

PR #48, opened 2017-07-07
Script to find all unused modules

LukeFarrawell:
A python script that finds the following:

- Modules that are not being referenced in the roles
- Modules that are not being used in the roles due to being commented out


Phasedd:
not entirely sure where it should be located, so it is sitting inside the tools folder for now.

tools/unused_modules.py
```diff
@@ -0,0 +1,75 @@
...
+commentedModules = {}
+usedModules = []
+unusedButCommentedModules = []
+
+# a list of folders that could be within a module
+skip_list = ['src', 'data', 'config', 'tests', 'report']
+
+# find all the modules that are available
+for root, dirs, files in os.walk(modulesPath):
+	if len(files) != 0:
```
TrentHouliston: Should be indenting with spaces not tabs. You should open the NUbots project in sublime and it should ensure all this stuff for you.

tools/unused_modules.py
```diff
@@ -0,0 +1,75 @@
+import os
+import re
+
+rolesPath = os.path.abspath(os.path.join(os.path.dirname( __file__ ), '..', 'roles'))
```
TrentHouliston: The modules folder and roles folder have variables in cmake, and therefore the b script so grab them from there

tools/unused_modules.py
```diff
@@ -0,0 +1,75 @@
...
+basePath = os.path.abspath(os.path.join(os.path.dirname( __file__ ), '..', 'module')) + os.sep
+
+existingModules = {}
+missingModules = {}
+commentedModules = {}
+usedModules = []
+unusedButCommentedModules = []
+
+# a list of folders that could be within a module
+skip_list = ['src', 'data', 'config', 'tests', 'report']
```
TrentHouliston: A better way than using a skip list would be to look for the CMakeLists.txt file in the directory and finding a NUClearModule call inside it.

nuclear/tools/module.py
```diff
@@ -173,3 +183,74 @@ def generate_test(parts):
     return template.format()
 
 
+def module_unused(**kwargs):
+    rolesPath = os.path.abspath(os.path.join(os.path.dirname( __file__ ), '..', '..', 'roles'))
+    modulesPath = os.path.abspath(os.path.join(os.path.dirname( __file__ ), '..', '..', 'module'))
+    basePath = os.path.abspath(os.path.join(os.path.dirname( __file__ ), '..', '..', 'module')) + os.sep
```
Bidski: Dont use relative paths! Make use of the `cmake_cache`, thats what it is there for!

nuclear/tools/module.py
```diff
@@ -173,3 +183,74 @@ def generate_test(parts):
...
+
+    existingModules = {}
+    missingModules = {}
+    commentedModules = {}
+    usedModules = []
+    unusedButCommentedModules = []
```
Bidski: Maintain the already existing naming scheme

nuclear/tools/module.py
Phasedd: yep, that is the last thing I got to do. .`/b` wasn't working on windows, so I pushed so I could work on my laptop later.

nuclear/tools/module.py
```diff
@@ -173,3 +183,72 @@ def generate_test(parts):
...
+                    line = line.strip()
+
+                    if '#' in line: # a commented
+                        line = re.sub(r"##+", '#', line) # remove multiple # that are next to each other
+                        location = line.find('#')
+
+                        if location == 0:
+                            line = line[1:] # remove the comment and continue to process
+                            if '::' in line: # the commented line is a commented out module
+                                additionalCommentLoc = line.find('#')
```
Bidski: You missed one (variable naming scheme)

nuclear/tools/module.py
```diff
@@ -17,7 +18,16 @@ def register(command):
...
+    # Unused module subcommand
+    unused_command = subcommands.add_parser('unused', help='List all NUClear modules that are unused')
+
+def run(module_command, **kwargs):
+    if module_command == 'generate':
+        module_generate(**kwargs)
+    elif module_command == 'unused':
+        module_unused(**kwargs)
+
+def module_generate(**kwargs):
```
TrentHouliston: You should extract the path variable here like before

nuclear/tools/module.py
```diff
@@ -173,3 +183,72 @@ def generate_test(parts):
...
+    roles_path = b.cmake_cache["NUCLEAR_ROLES_DIR"]
+    modules_path = b.cmake_cache["NUCLEAR_MODULE_DIR"]
+    existing_modules = {}
+    missing_modules = {}
+    commented_modules = {}
+    used_modules = []
+    unused_commented_modules = []
+
+    # a list of folders that could be within a module
+    skip_list = ['src', 'data', 'config', 'tests', 'report']
```
TrentHouliston: Rather than doing it this way, just search for files named CMakeLists.txt that contain a call to NUClearModule(...)

nuclear/tools/module.py
```diff
@@ -173,3 +183,72 @@ def generate_test(parts):
...
+
+    # compare our roles to our existing modules
+    for root, dirs, files in os.walk(roles_path):
+        for file in files:
+            with open(os.path.join(roles_path, file), 'r') as role:
+                for line in role:
+                    line = line.strip()
+
+                    if '#' in line: # a commented
+                        line = re.sub(r"##+", '#', line) # remove multiple # that are next to each other
```
TrentHouliston: try to use single quotes everywhere

nuclear/tools/module.py
```diff
@@ -173,3 +183,72 @@ def generate_test(parts):
...
+
+                    if '::' in line: # the line is a module
+                        line = line.replace('::', os.sep)
+                        if line in existing_modules:
+                            existing_modules[line] += 1
+                            used_modules.append(line)
+                        else:
+                            missing_modules[line] = 1
+
+    print('Unused modules:\n')
```
TrentHouliston: Import termcolor and print using colors

nuclear/tools/module.py
```diff
@@ -173,3 +183,72 @@ def generate_test(parts):
...
+
+                                line = line.replace('::', os.sep)
+                                commented_modules[line.strip()] = 1
+                            continue
+                        else:
+                            # the comment was just a description, remove the comment and continue
+                            line = line[:location].strip()
+
+                    if '::' in line: # the line is a module
+                        line = line.replace('::', os.sep)
```
TrentHouliston: It's probably more reliable to do it the other way and convert the path to a namespace rather than namespace to path

ysims:
Superseded by #290.

PR #51, opened 2017-07-08
New protocol for Game controller 

CptTony57:


Bidski:
This looks fine to me, @TrentHouliston should probably give it a quick look just to make sure.

module/input/GameController/src/GameController.cpp
```diff
@@ -256,6 +264,7 @@ namespace input {
             if (newOwnPlayer.penaltyState != oldOwnPlayer.penaltyState
                 && newOwnPlayer.penaltyState != gamecontroller::PenaltyState::UNPENALISED) {
 
+                log("Penalised");
```
TrentHouliston: This shouldn't be here, this kind of log should come from behaviour instead.

module/support/configuration/GlobalConfig/data/config/darwin1/GlobalConfig.yaml
```diff
@@ -1,2 +1,2 @@
 playerId: 1
-teamId: 9
+teamId: 14
```
TrentHouliston: Move this from a per robot config to a normal one. Let the merging handle switching IDs.

PR #54, opened 2017-07-11
Removed function causing segmentation faults

CptTony57:
calculateCentreOfPressure has been causing segmentation faults in ForwardKinematics.h #456. It doesn't appear to be used so I removed it as I can't determine why it is crashing.


module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -650,8 +650,8 @@ namespace platform {
                     /************************************************
                      *                  CENTRE OF PRESSURE          *
                      ************************************************/
-                    sensors->centreOfPressure = convert<double, 3>(
-                        utility::motion::kinematics::calculateCentreOfPressure(kinematicsModel, *sensors));
+                    // sensors->centreOfPressure = convert<double, 3>(
+                    //    utility::motion::kinematics::calculateCentreOfPressure(kinematicsModel, *sensors));
```
Bidski: We should do better than this. What is the root cause of the seg fault.

TrentHouliston: Yeah, given that we have memory corruption issues, just because this function segfaults doesn't mean it's the thing causing it. So we should work out why this function is segfaulting before removing it.

PR #69, opened 2017-07-14
[WIP] Changed camera reset behaviour and updated information logs

CptTony57:


MattAmos:
Tony, is this okay to merge? We should really get rid of the logs asap

PR #71, opened 2017-07-14
Added condition to accept nubotsvm as a valid platform

MattAmos:


TrentHouliston:
@Bidski should review this one

shared/extension/Script.h
```diff
@@ -129,6 +129,10 @@ struct Script {
             return match[1].str();
         }
 
+        else if (hostname == "nubotsvm") {
```
Bidski: Remove this `else if` and change the regex to `"([a-z]+)([0-9]+)?"`

PR #76, opened 2017-07-16
Fix the nbs decoder to work with new nbs streams and NUsight

TrentHouliston:
This allows the decoder to parse nbs files using the new xxhash system.
It also adds the ability to decode NUsight<> wrapped messages

### DIFF
--- a/tools/decode.py
+++ b/tools/decode.py
@@ -27,7 +27,7 @@ def register(command):
     # Drone arguments
     command.add_argument('file'
         , metavar='file'
-        , help='The file to decode into a strint')
+        , help='The file to decode into a series of json objects')
 
 def run(file, **kwargs):
 
--- a/tools/decode.py
+++ b/tools/decode.py
@@ -55,82 +55,37 @@ def run(file, **kwargs):
 
     # Now open the passed file
     with gzip.open(file, 'rb') if file.endswith('nbz') or file.endswith('.gz') else open(file, 'rb') as f:
-        with gzip.open('output.pickle', 'wb') as of:
 
-            prev_goal_position = None
-            state = None
-            c_pass = []
+        # While we can read a header
+        while len(f.read(3)) == 3:
 
-            # While we can read a header
-            while len(f.read(3)) == 3:
+            # Read our size
+            size = struct.unpack('<I', f.read(4))[0]
+            print(size)
+            # Read our payload
+            payload = f.read(size)
 
-                # Read our size
-                size = struct.unpack('<I', f.read(4))[0]
+            # Read our timestamp
+            timestamp = struct.unpack('<Q', payload[:8])[0]
 
-                # Read our payload
-                payload = f.read(size)
+            # Read our hash
+            type_hash = payload[8:24]
 
-                # Read our timestamp
-                timestamp = struct.unpack('<Q', payload[:8])[0]
+            # If we know how to parse this type, parse it
+            if type_hash in parsers:
+                msg = parsers[type_hash][1].FromString(payload[24:])
 
-                # Read our hash
-                type_hash = payload[8:24]
+                # So the else can always exist
+                if False:
+                    pass
 
-                # If we know how to parse this type, parse it
-                if type_hash in parsers:
-                    msg = parsers[type_hash][1].FromString(payload[24:])
+                # If there is a special way to decode a particular message do it here
+                # if parsers[type_hash][0] == b'message.input.Sensors':
+                    # Read our servo
 
-                    if parsers[type_hash][0] == b'message.support.ServoHealthTestData':
-                        # Read our servo
-                        servo = msg.sensors.servo.rAnkleRoll
-
-                        if prev_goal_position == None:
-                            prev_goal_position = round(servo.goalPosition)
-                        elif prev_goal_position != round(servo.goalPosition):
-                            # Changing motion
-                            if state != None:
-                                pickle.dump((state, c_pass), of, protocol=pickle.HIGHEST_PROTOCOL)
-                                c_pass = []
-
-                            if prev_goal_position == 0:
-                                # Second half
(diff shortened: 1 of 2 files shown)


tools/decode.py
```diff
@@ -45,92 +53,51 @@ def run(file, **kwargs):
...
+
+            # If we know how to parse this type, parse it
+            if type_hash in parsers:
+                msg = parsers[type_hash][1].FromString(payload[16:])
+
+                # So the else can always exist
+                if False:
+                    pass
+
+                # If there is a special way to decode a particular message do it here
```
BrendanAnnable: Can you explain a situation where this would be needed? Is it worth having this empty block here?

tools/decode.py
```diff
@@ -45,92 +53,51 @@ def run(file, **kwargs):
...
         pb_type = message.__module__.split('.')[:-1]
         pb_type.append(message.__name__)
-        pb_type = '.'.join(pb_type).encode('utf-8')
-        pb_hash = '' # TODO swap to using XXHASH
-        # pb_hash = mmh3.hash_bytes(pb_type, 0x4e55436c, True)
-
+        pb_type = '.'.join(pb_type)
+        # Reverse to little endian
+        pb_hash = bytes(reversed(xxhash.xxh64(pb_type, seed=0x4e55436c).digest()))
         parsers[pb_hash] = (pb_type, message)
```
BrendanAnnable: The multi-lingo confused me, can this be named`decoders` to be consistent?

tools/decode.py
```diff
@@ -45,92 +53,51 @@ def run(file, **kwargs):
...
-    servo_health = {}
+        # We can also decode NUsight<> wrapped packets
+        nusight_type = 'NUsight<{}>'.format(pb_type)
+        # Reverse to little endian
+        nusight_hash = bytes(reversed(xxhash.xxh64(nusight_type, seed=0x4e55436c).digest()))
+        parsers[nusight_hash] = (nusight_type, NUsightDecoder(message))
+
 
     # Now open the passed file
     with gzip.open(file, 'rb') if file.endswith('nbz') or file.endswith('.gz') else open(file, 'rb') as f:
```
BrendanAnnable: It would be nice if there was either:

- a link to the definition of the file format
- or paste the definition I have in NUsight2 if one does not exist 😛

tools/decode.py
```diff
@@ -45,92 +53,51 @@ def run(file, **kwargs):
...
+
+                # If there is a special way to decode a particular message do it here
+                # if parsers[type_hash][0] == b'message.input.Sensors':
+                    # Read our servo
+
+                else:
+                    out = re.sub(r'\s+', ' ', MessageToJson(msg, True))
+                    out = '{{ "type": "{}", "timestamp": {}, "data": {} }}'.format(parsers[type_hash][0], timestamp, out)
+                    # Print as a json object
+                    print(out)
```
BrendanAnnable: (Optional) Do you think it would it be worth printing a summary? Maybe just the total number of messages and the number of each message type?

tools/decode.py
TrentHouliston: Some things like images you really don't want to decode as json but currently we don't have any decoders implemented for those here. It will end up as an if else if chain with the to json at the end. So currently that empty block allows an else until we get an if statement to replace it.

tools/decode.py
TrentHouliston: I was thinking it would be better printing out all json so it can be passed into another program to do further processing if needed

tools/decode.py
TrentHouliston: Done!

tools/decode.py
TrentHouliston: Pasted in your definition

tools/decode.py
TrentHouliston: I replaced the empty block with an if images so images won't print out

tools/decode.py
```diff
@@ -37,100 +45,62 @@ def run(file, **kwargs):
...
-                        })
-
-                    else:
-                        out = re.sub(r'\s+', ' ', MessageToJson(msg, True))
-                        out = '{{ "type": "{}", "timestamp": {}, "data": {} }}'.format(parsers[type_hash][0].decode('utf-8'), timestamp, out)
-                        # Print as a json object
-                        print(out)
 
+        # NBS File Format:
+        # 3 Bytes - NUClear radiation symbol header, useful for synchronisation when attaching to a existing stream.
```
BrendanAnnable: nit: _an_ existing stream (yes I have to update mine).

tools/decode.py
```diff
@@ -37,100 +45,62 @@ def run(file, **kwargs):
...
+        # 4 Bytes - The remaining packet length i.e. 16 bytes + N payload bytes
```
BrendanAnnable: nit: missing period.

tools/decode.py
```diff
@@ -37,100 +45,62 @@ def run(file, **kwargs):
...
+            # Read our hash
+            type_hash = payload[8:16]
+
+            # If we know how to parse this type, parse it
+            if type_hash in decoders:
+                msg = decoders[type_hash][1].FromString(payload[16:])
+
+                # Outputting an image as json would be bad
+                if decoders[type_hash][0] == b'message.input.Image':
+                    pass
```
BrendanAnnable: print without the data? otherwise we won't know they exist in the file. i.e.

```
out = re.sub(r'\s+', ' ', MessageToJson(msg, True))
out = '{{ "type": "{}", "timestamp": {} }}'.format(decoders[type_hash][0], timestamp)
print(out)
```

tools/decode.py
TrentHouliston: And by missing period you mean too many periods! (no other comments in the file have them)

tools/decode.py
TrentHouliston: fixed

PR #83, opened 2017-07-19
Implement semisimulated HardwareIO

TrentHouliston:
Implement semifake hardware that writes to real hardware but simulates its response.
Should help deal with timeouts. NOTE THIS SHOULD BE A TEMPORARY FIX NOT A PERMANENT SOLUTION.
FIX THE TIMEOUTS.


module/platform/darwin/HardwareIO/src/Convert.cpp
```diff
@@ -85,15 +85,15 @@ namespace platform {
         float Convert::servoPosition(const uint8_t id, const uint16_t value) {
             // offset and normalize the angle
             return utility::math::angle::normalizeAngle(
-                (((value - 2048) * POSITION_CONVERSION_FACTOR) + SERVO_OFFSET[id]) * SERVO_DIRECTION[id]);
+                (((value - 2048) * POSITION_CONVERSION_FACTOR) * SERVO_DIRECTION[id]) + SERVO_OFFSET[id]);
```
JakeFountain: will this require new parameters for darwin and igus?

module/platform/darwin/HardwareIO/src/HardwareIO.h
```diff
@@ -56,16 +59,29 @@ namespace platform {
             };
 
             struct ServoState {
+                // True if we need to write new values to the hardware
                 bool dirty = false;
 
+                // True if we simulate where we think the servos should be
+                // Note that we still write the commands to hardware
+                bool simulated = false;
```
JakeFountain: maybe change this variable name to "read_servo" or similar

module/platform/darwin/HardwareIO/src/darwin/Darwin.cpp
```diff
@@ -73,6 +74,20 @@ Darwin::Darwin(const char* name)
     uart.executeWrite(DarwinDevice::WriteCommand<uint8_t>(ID::BROADCAST, CM730::Address::RETURN_LEVEL, 1));
 }
 
+void Darwin::setConfig(const extension::Configuration& config) {
+
+    // Set servos to be enabled if they are not simulated
+    for (size_t i = 0; i < config["servos"].config.size(); ++i) {
+        enabledServoIds[i] = !config["servos"][i]["servo_direction"].as<bool>();
```
JakeFountain: should this be "simulated" rather than "servo_direction"?

TrentHouliston: Why yes it should!

module/platform/darwin/HardwareIO/src/HardwareIO.h
TrentHouliston: I think this one should be fine? Both are about the same I think.

module/platform/darwin/HardwareIO/src/Convert.cpp
TrentHouliston: It will but I already fixed it in the configuration files. Still wouldn't hurt to run in sensors test before running for real

PR #92, opened 2017-07-24
Updated field dimensions with 2017 rule changes

CptTony57:


BrendanAnnable:
@JosephusPaye / @Phasedd we should probably update this then https://github.com/NUbots/NUsight2/blob/master/src/shared/field/dimensions.ts#L72

JosephusPaye:
@BrendanAnnable Made a pull: https://github.com/NUbots/NUsight2/pull/112

PR #96, opened 2017-07-24
KickPlanner now uses game state as a factor for determining a kick

CptTony57:


module/behaviour/planning/KickPlanner/src/KickPlanner.cpp
```diff
@@ -92,12 +99,13 @@ namespace behaviour {
                 }
             });
 
-            on<Trigger<Ball>, With<Field>, With<FieldDescription>, With<KickPlan>, With<Sensors>>().then(
-                [this](const Ball& ball,
-                       const Field& field,
-                       const FieldDescription& fd,
-                       const KickPlan& kickPlan,
-                       const Sensors& sensors) {
+            on<Trigger<Ball>, With<Field>, With<FieldDescription>, With<KickPlan>, With<GameState>, With<Sensors>>()
```
TrentHouliston: Make the game state optional for this, we still want to be able to kick balls when testing without a game controller.

module/behaviour/planning/KickPlanner/src/KickPlanner.cpp
```diff
@@ -23,6 +23,8 @@
 
 #include "message/behaviour/KickPlan.h"
 #include "message/behaviour/ServoCommand.h"
+#include "message/input/GameEvents.h"
```
TrentHouliston: Don't think you need this header

TrentHouliston (dismissed):
Make sure this also disables it from kicking if we are in the free kick ready state. We don't want to get yellow cards.

JosephusPaye (approved):
Code LGTM

module/behaviour/planning/KickPlanner/src/KickPlanner.cpp
```diff
@@ -50,12 +52,18 @@ namespace behaviour {
         using message::behaviour::KickPlan;
         using KickType = message::behaviour::KickPlan::KickType;
         using message::behaviour::WantsToKick;
+        using message::input::GameState;
         using message::input::Sensors;
         using message::localisation::Ball;
         using message::localisation::Field;
+        using Phase         = message::input::GameState::Data::Phase;
+        using PenaltyReason = message::input::GameState::Data::PenaltyReason;
```
KipHamiltons, line 60: wouldn't `using message::input::GameState::Data::Phase` have the same effect as the explicit assignment of the alias? 
same for the line after
```suggestion
        using message::input::GameState::Data::Phase;
        using message::input::GameState::Data::PenaltyReason;
```

module/behaviour/planning/KickPlanner/src/KickPlanner.cpp
```diff
@@ -92,12 +100,20 @@ namespace behaviour {
                 }
             });
 
-            on<Trigger<Ball>, With<Field>, With<FieldDescription>, With<KickPlan>, With<Sensors>>().then(
-                [this](const Ball& ball,
-                       const Field& field,
-                       const FieldDescription& fd,
-                       const KickPlan& kickPlan,
-                       const Sensors& sensors) {
+            on<Trigger<ButtonMiddleDown>, Single>().then([this] { forcePlaying = true; });
```
KipHamiltons, line 103: How do you turn it off? could we do something like `forcePlaying = !forcePlaying`?

module/behaviour/planning/KickPlanner/src/KickPlanner.hpp
```diff
@@ -36,6 +36,7 @@ namespace behaviour {
 
         private:
             bool kickValid(const arma::vec3& ballPos);
```
KipHamiltons, line 38: what is this even doing? is it saying that if the arma::vec is null then it's false?? 
could be out of the scope of this PR, but this line makes no sense to me

module/behaviour/planning/KickPlanner/src/KickPlanner.cpp
JosephusPaye: Middle button always plays. Turning it off is red button.

KipHamiltons (approved):
LGTM

PR #101, opened 2017-07-25
Replace muParserX with Expression Toolkit to resolve segfaults

TrentHouliston:
MuparserX was causing segfaults due to it's terrible use of singletons (the most antiest of patterns)
Swap to using another library without this problem.


TrentHouliston:
Superseded by another PR

PR #115, opened 2017-07-26
Wait for camera to appear after reset before opening it

CptTony57:


MattAmos:
We need this fixed more than we need it to be good. LGTM!!

PR #146, opened 2018-03-16
Adds python module and OpenCL support

Bidski:


nuclear/module/NUClearModule.cmake
```diff
@@ -87,24 +98,35 @@ FUNCTION(NUCLEAR_MODULE)
...
                 COMMENT "Copying updated python file ${python_file}"
             )
-
         ENDFOREACH(python_file)
 
         FILE(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/src")
 
+        SET(ENV{PYTHONPATH} "${PYTHONPATH}:${PYTHON_SITE_PACKAGES}")
+        SET(ENV{NUCLEAR_MODULE_DIR} "${PROJECT_SOURCE_DIR}/${NUCLEAR_MODULE_DIR}")
+        SET(ENV{LD_LIBRARY_PATH} "/nubots/toolchain/${PLATFORM}/lib:/nubots/toolchain/lib")
```
TrentHouliston: This has to be smarter than this, This code is shared between all NUClear projects

nuclear/module/NUClearModule.cmake
```diff
@@ -87,24 +98,35 @@ FUNCTION(NUCLEAR_MODULE)
...
         ADD_CUSTOM_COMMAND(
             OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/src/${module_name}.h"
                    "${CMAKE_CURRENT_BINARY_DIR}/src/${module_name}.cpp"
             COMMAND ${CMAKE_COMMAND}
             ARGS -E env
-                PYTHONPATH="${PROJECT_BINARY_DIR}/python/nuclear/"
+                PYTHONPATH="${PROJECT_BINARY_DIR}/python/nuclear:${PYTHONPATH}:${PYTHON_SITE_PACKAGES}"
                 NUCLEAR_MODULE_DIR="${PROJECT_SOURCE_DIR}/${NUCLEAR_MODULE_DIR}"
+                LD_LIBRARY_PATH="/nubots/toolchain/${PLATFORM}/lib:/nubots/toolchain/lib"
```
TrentHouliston: Once again nubots/toolchain won't work here

nuclear/tools/module.py
```diff
@@ -107,15 +133,20 @@ class {className} : public NUClear::Reactor {{
         {closeNamespace}
 
         #endif  // {define}
-        """)
+        """
+    )
+
+    return template.format(
+        define='{}_H'.format('_'.join([p.upper() for p in parts])),
+        className=parts[-1],
```
TrentHouliston: A wild camelCase appeared

nuclear/tools/module.py
```diff
@@ -132,14 +163,19 @@ def generate_cpp(parts):
                 }});
             }}
         {closeNamespace}
-        """)
+        """
+    )
+
+    return template.format(
+        className=parts[-1],
```
TrentHouliston: More camelCase

nuclear/module/NUClearModule.cmake
```diff
@@ -171,9 +193,14 @@ FUNCTION(NUCLEAR_MODULE)
             COMMENT "Copying updated data file ${input_file}"
         )
 
-        SET(NUCLEAR_MODULE_DATA_FILES ${NUCLEAR_MODULE_DATA_FILES} ${output_file} CACHE INTERNAL "A list of all the data files that were generated by modules" FORCE)
+        LIST(APPEND NUCLEAR_MODULE_DATA_FILES ${output_file})
     ENDFOREACH(data_file)
 
+    # Remove duplicate entries from NUCLEAR_MODULE_DATA_FILES
```
TrentHouliston: This comment is wrong

nuclear/module/python/nuclear.py
```diff
@@ -152,11 +158,29 @@ def include_paths(self):
...
+    if not hasattr(item, "strip") and hasattr(item, "__getitem__") or hasattr(item, "__iter__"):
+        return True
+    else:
+        return False
+
+
+def bind_emit(self, msg):
+    # Iterable type
+    if is_sequence(msg):
+        return msg[0]._emit_vector(self._reactor_ptr, msg)
```
TrentHouliston: Still don't know how I feel about this emit vector thing, We're trying to get rid of it in the rest of the code too

nuclear/tools/module.py
```diff
@@ -154,22 +190,45 @@ def generate_readme(parts):
 
         ## Dependencies
 
-        """)
+        """
+    )
+
+    return template.format(
+        className=parts[-1], classNameTitle=len(parts[-1]) * '=', closeNamespace='\n'.join(['}' for x in parts[:-1]])
```
TrentHouliston: Even more!!

nuclear/tools/module.py
```diff
@@ -154,22 +190,45 @@ def generate_readme(parts):
...
+
+        from nuclear import Reactor, on, Trigger, Single, With, Every
+
+        @Reactor
+        class {className}(object):
+            def __init__(self):
+                # Constructor for {className}
+
+            @on(Configuration('{className}.yaml'))
+            def {className}_configfuration(self, config):
```
TrentHouliston: Did you actually get this code to work?

TrentHouliston (changes requested):
Some changes

nuclear/tools/module.py
Bidski: No, I should probably comment this out here for now. We need to create more bindings in general though (extensions and utilities)

nuclear/module/python/nuclear.py
Bidski: Well until it is completely removed from the rest of code this needs to stay

nuclear/tools/module.py
Bidski: Pretty sure these are your fault, I just formatted the code lol

nuclear/module/NUClearModule.cmake
Bidski: Hmmm .... this will take some thought

KipHamiltons:
@Bidski does this PR have stuff worth porting to the current codebase?

Bidski:
The python module support needs a lot more work to be properly usable (utilities need python wrappings). The Opencl support is mostly just an extension so you can do `on<OpenCL>` which is only useful if we think we will be implementing more OpenCL-based modules

Bidski:
We also need to see if we can figure out how to wrap extensions in python as well

KipHamiltons:
> The python module support needs a lot more work to be properly usable (utilities need python wrappings). 

what would that enable us to do?

Bidski:
> what would that enable us to do?

You mean apart from being able to write modules in python that integrate nicely with the C++ modules?

KipHamiltons:
No, that answers my question directly. That would be pretty nice!!!

PR #155, opened 2018-05-04
[WIP] Static Walk

ysims:


module/motion/walk/TorsoMovement/src/TorsoMovement.cpp
```diff
@@ -39,11 +39,6 @@ namespace motion {
             : Reactor(std::move(environment)), subsumptionId(size_t(this) * size_t(this) - size_t(this)) {
 
             on<Configuration>("TorsoMovement.yaml").then([this](const Configuration& config) {
-                double x = config["test"]["x"].as<double>();
```
TrentHouliston: Make sure you also remove it from the config file

ysims: Does the .yaml need to be deleted entirely?

TrentHouliston: No, just the values in it that you're not using

TrentHouliston: Although, if you don't have a configuration at all you can delete it
It might be better to leave an empty on<Configuration> and yaml file though, for when you want to add some configuration in the future

ysims: Yeah, it was all just test data.

module/motion/walk/TorsoMovement/src/TorsoMovement.cpp
```diff
@@ -0,0 +1,122 @@
...
+                    // Target given in support foot space
+                    Eigen::Affine3d Haf;
+                    Haf = target.Haf;
+
+                    // Create vector to target from support foot
+                    Eigen::Vector3d rAFf = -Haf.translation();
+                    // Position of target in torso space
+                    Eigen::Vector3d rATt = Htf * rAFf;
+                    // Create next torso target in torso space
+                    Eigen::Vector3d rT_tTt = rATt * 0.01;
```
Bidski: Should `0.01` be a configurable parameter?

shared/message/motion/FootMotionCommand.proto
```diff
@@ -1,57 +0,0 @@
-/*
```
Bidski: MIght not be a wise idea to delete this file just yet?

TrentHouliston: I think this was just used by the deleted modular walk engine

module/motion/walk/TorsoMovement/src/TorsoMovement.cpp
TrentHouliston: I think that's the speed of torso movement, which should be based on the difference between now and when the torso is supposed to reach it's position

module/motion/walk/TorsoMovement/src/TorsoMovement.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+                    // Create vector to target from support foot
+                    Eigen::Vector3d rAFf = -Haf.translation();
+                    // Position of target in torso space
+                    Eigen::Vector3d rATt = Htf * rAFf;
+                    rATt.normalize();
+
+                    // Find scale to reach target at specified time
+                    std::chrono::duration<double> time_left = target.timestamp - NUClear::clock::now();
+                    double distance = std::sqrt(std::pow(rATt.x(), 2) + std::pow(rATt.y(), 2) + std::pow(rATt.z(), 2));
```
Bidski: `double distance = rATt.norm()`

module/motion/walk/TorsoMovement/src/TorsoMovement.cpp
```diff
@@ -0,0 +1,136 @@
...
+                    }
+                    else {  // There is time left to complete
+                        scale = (distance / time_left.count()) * 0.01;
+                    }
+
+                    // Create next torso target in torso space
+                    Eigen::Vector3d rT_tTt = rATt * scale;
+                    // Create vector from torso to foot target
+                    Eigen::Vector3d rF_tFt = -rT_tTt;
+                    Eigen::Vector3d rF_tTt = Htf * rF_tFt;
```
Bidski: If this is the only place that `rF_tFt` is used then just change this `Eigen::Vector3d rF_tTt = Htf * (-rT_tTt);` and remove the line above (`Eigen::Vector3d rF_tFt = -rT_tTt;`)

PR #157, opened 2018-05-29
Updates toolchain and clang format

Bidski:
Updates some tool versions as well


puppet/modules/user_tools/manifests/init.pp
```diff
@@ -143,6 +143,7 @@
 
 10.1.1.11 i1 igus1
```
TrentHouliston: Now that you don't use the darwins anymore you could move these to the much coveted 10.1.1.robotnumber position

TrentHouliston:
Also you probably should rename this PR, since it's doing a lot more than just updating NUClear

puppet/modules/user_tools/manifests/init.pp
Bidski: I thought about that, figured we should maintain some legacy for a little longer

PR #158, opened 2018-05-30
Ginn/goal detection fix

danielginn:
Fix for goal detection segfault. You may not need to changes in .ignore, they are a carry over from folders I have in my vagrant for ORB-SLAM


.gitignore
```diff
@@ -74,3 +74,11 @@ tags
 .DS_Store
 .idea
 *.sublime-workspace
+
```
TrentHouliston: Please remove these gitignore changes from the PR

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -19,6 +19,7 @@
 
 #include "GoalDetector.h"
 
+#include <math.h>
```
TrentHouliston: `#include <cmath>`

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -591,9 +593,10 @@ namespace vision {
                         Eigen::Matrix3d rGCc_cov = convert<double, 3, 3>(arma::diagmat(
                             VECTOR3_COVARIANCE % covariance_amplifier));  // arma::diagmat(arma::vec3{0.01,0.01,0.001})
 
-
-                        it->measurement.push_back(
-                            Goal::Measurement(Goal::MeasurementType::CENTRE, rGCc_sphr, rGCc_cov));
+                        if (!isinf(rGCc_sphr[0]) && !isinf(rGCc_sphr[1]) && !isinf(rGCc_sphr[2])) {
```
TrentHouliston: `std::isinf`
If we're doing this kind of filtering, it might be good to check `std::isnan` too, also

Bidski: or maybe use `std::isfinite` instead

TrentHouliston: yes, std::isfinite is a much better option

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -19,6 +19,7 @@
 
 #include "GoalDetector.h"
 
+#include <cmath.h>
```
TrentHouliston: Just `#include <cmat>` does this even build?

.gitignore
Bidski: Isn't this just a newline at the end of the file?

TrentHouliston: Hmm, you might be right. In which case the IDE settings will be forcing it

PR #160, opened 2018-06-04
Add kick ability to keyboard walk

MattAmos:


TrentHouliston:
Seriously, go get the PS3 controller going, it's sooooo much more convenient

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.cpp
```diff
@@ -133,14 +136,25 @@ namespace behaviour {
             log("getup");
         }
 
-        void KeyboardWalk::kickRightForward() {
-            /*emit(std::make_unique<KickCommand>(KickCommand{
-                {-0.05, 0, 0}, //Ball is right of centre for right kick
-                {1, 0, 0}
-            }));*/
-            log("right forward kick");
+        void KeyboardWalk::kickLeft() {
```
Bidski: Is it worth making this a single function that takes in `LimbID` as a parameter?

roles/keyboardwalk.role
```diff
@@ -29,8 +28,9 @@ NUCLEAR_ROLE(
     # Behaviour
     behaviour::Controller
     behaviour::skills::Stand
-    #behaviour::skills::FallingRelax
+    behaviour::skills::FallingRelax
     behaviour::skills::Getup
+    behaviour::skills::KickScript
```
Bidski: Has this been tested?

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.cpp
MattAmos: Done!

roles/keyboardwalk.role
MattAmos: Not properly tested. Removed for this PR

TrentHouliston:
I'm not a big fan of removing falling relax, that's how things get broken

MattAmos:
Falling relax was removed before this PR. Will test in another branch and make another PR when it's stable.

PR #169, opened 2018-06-07
Fixed Classified Image Freeze during RoboCup

danielginn:


module/vision/LUTClassifier/src/GoalEnhancer.cpp
```diff
@@ -160,7 +160,7 @@ namespace vision {
             // Get the min and max for intersecting with top and bottom of the screen
 
             double jump = (maxTangent - minTangent) / double(GOAL_LINE_INTERSECTIONS);
-            for (auto d = minTangent; d <= maxTangent; d += jump) {
+            for (auto d = minTangent; d <= maxTangent && jump > 0; d += jump) {
```
TrentHouliston: Use an if instead, there is no point doing the for loop for one loop if jump is 0

danielginn: I wouldn't run through the loop if it failed on the conditions would it? Anyway, I have wrapped the for loop in an if statement.

TrentHouliston: The if makes more sense logically even if it would work this way

PR #174, opened 2018-06-14
Visual Mesh ball detection integration (and Leg Loads)

MattAmos:
This PR implements the Visual Mesh to detect balls rather than current LUT-based Ball Detection


module/vision/VisualMesh/src/VisualMesh.cpp
```diff
@@ -115,38 +115,44 @@ namespace vision {
...
+            std::vector<std::tuple<Eigen::Vector2i, Eigen::Vector2i, Eigen::Vector4d>,
+                        Eigen::aligned_allocator<std::tuple<Eigen::Vector2i, Eigen::Vector2i, Eigen::Vector4d>>>
+                lines;
 
 
-            // std::vector<std::array<int, 2>> pixel_coordinates = results.pixel_coordinates;
-            // std::vector<float> classification = results.classifications.front().second;
+            std::vector<std::array<int, 2>> pixel_coordinates;
+            std::vector<std::array<float, 2>> lazyConvert = results.pixel_coordinates;
+            pixel_coordinates.reserve(lazyConvert.size());
```
TrentHouliston: This may not be a segfault but it's basically one.
You want resize if you're doing it like this, or push_back on line 133 if you're doing reserve

module/vision/VisualMesh/src/VisualMesh.cpp
```diff
@@ -0,0 +1,174 @@
...
+            msg->indices = std::move(results.global_indices);
+
+            // Add our neighbourhood
+            msg->neighbourhood.reserve(results.neighbourhood.size());
+            for (const auto& n : results.neighbourhood) {
+                msg->neighbourhood.emplace_back(Eigen::Map<const Eigen::Matrix<int, 6, 1, Eigen::DontAlign>>(n.data()));
+            }
+
+            // Add our coordinates
+            msg->coordinates = results.pixel_coordinates.as<Eigen::Matrix<int, 2, 1, Eigen::DontAlign>>();
```
TrentHouliston: This is wrong here, you're basically doing a reinterpret cast on some floats into ints. That won't give your correct pixel coordinates.
it should be `results.pixel_coordinates.as<Eigen::Matrix<float, 2, 1, Eigen::DontAlign>>();`
but then you'll need to either make msg->coordinates floats or round the output

module/vision/VisualMesh/src/VisualMesh.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+
+            std::vector<std::array<int, 2>> pixel_coordinates;
+            std::vector<std::array<float, 2>> lazyConvert = results.pixel_coordinates;
+            pixel_coordinates.reserve(lazyConvert.size());
+            for (int i = 0; i < lazyConvert.size(); ++i) {
+                pixel_coordinates[i] = {int(lazyConvert[i][0]), int(lazyConvert[i][1])};
+            }
+
+            std::vector<float> classification = results.classifications.front().second;
```
TrentHouliston: Later you might want to change this to back to see the classifications rather than the image itself (also change below to output classes rather than colours)

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -314,185 +316,249 @@ namespace vision {
...
+                            continue;
+                        }
+
+                        balls->push_back(std::move(b));
                     }
 
-                    emit(utility::nusight::drawVisionLines(lines));
-                }
+                    /***********************************************
+                     *                  USTERS DRAW                *
```
MattAmos: This should be `CLUSTERS`

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -314,185 +316,249 @@ namespace vision {
...
-                                  << " Distance " << distance << " rBCc " << rBCc.t() << " screenAngular "
-                                  << b.visObject.screenAngular.transpose() << " angularSize "
-                                  << b.visObject.angularSize.transpose() << std::endl;
-                    }
+                        b.visObject.timestamp = NUClear::clock::now();
 
-                    balls->push_back(std::move(b));
-                }
+                        if (print_mesh_debug) {
+                            std::cout << "Gradient " << b.cone.gradient << " Center " << center.t() << " Radius "
```
MattAmos: This should be using NUClear log

PR #177, opened 2018-06-15
Fixes goal vision objects not calculating screen angular

MattAmos:


module/motion/HeadController/src/HeadController.cpp
```diff
@@ -89,10 +90,12 @@ namespace motion {
         on<Trigger<HeadCommand>>().then("Head Controller - Register Head Command", [this](const HeadCommand& command) {
             goalRobotSpace = command.robotSpace;
             if (goalRobotSpace) {
-                goalAngles = {command.yaw, command.pitch};
+                goalAngles = {utility::math::clamp(-float(M_PI_2), command.yaw, float(M_PI_2)),
```
Bidski: Maybe use the config values?

PR #178, opened 2018-06-15
[TESTING] Adds dribble behaviour based on current distance to goals 

MattAmos:


module/behaviour/planning/KickPlanner/src/KickPlanner.cpp
```diff
@@ -126,6 +128,17 @@ namespace behaviour {
                     // log("KickAngle",KickAngle);
                     // log("ballPosition",ballPosition);
                     // log("secondsSinceLastSeen",secondsSinceLastSeen);
+
+                    // Calculate the Manhattan distance to centre of goals (x = field_length / 2, y = 0) from current
+                    // field position
```
Bidski: Use euclidean distance?

ysims:
Old and unlikely to be completed

PR #184, opened 2018-06-19
Fixes selu softmax problem

tayloryoung6396:


module/vision/VisualMesh/src/mesh/VisualMesh.hpp
```diff
@@ -445,19 +445,22 @@ class VisualMesh {
                     constexpr const float lambda = 1.0507009873554804934193349852946;
                     constexpr const float alpha  = 1.6732632423543772848170429916717;
 
-                    // Apply selu
-                    if (vector_out) {
-                        std::string e = "in" + std::to_string(layer_no + 1);
+                    // If this is not our last layer, apply selu
+                    if (layer_no + 1 < conv.size()) {
```
TrentHouliston: This only checks the last layer, not the last layer in the last convolution

TrentHouliston:
Alex can approve since he can check that this actually works

TrentHouliston:
Also, for the record, it's not a problem. Just a difference in how the new networks were trained.
The old networks ran SELU after every layer, while the new networks don't run it on the final layer.

PR #186, opened 2018-06-27
Add the DSL and communication components of The Director

TrentHouliston:
This PR is purely the glue and API parts of the director without any of the algorithm logic that will be required (although it has some notes about how that algorithm will be implemented)


module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,105 @@
...
+        });
+
+        // Main director function
+        on<Every<90, Per<std::chrono::seconds>>>().then([this] {
+            // Sort tasks by priority
+            std::sort(global_tasks.begin(), global_tasks.end(), bypriority);
+
+            for (const auto& t : global_tasks) {
+
+                // We should have a current state tree (which initially may be empty), a transitional state tree and a
```
Bidski: This should appear in a separate document in the doc folder with some diagrams to explain the steps and conditions in this algorithm.

TrentHouliston: This is nowhere near fleshed out enough for that yet

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,238 @@
...
+                throw std::runtime_error("You cannot have multiple provider DSL words in a single on statement.");
+            }
+
+            auto provider = std::make_shared<Provider>(action, data_type, r);
+            group.providers.emplace_back(provider);
+            providers.emplace(r->id, provider);
+        }
+
+        /**
+         * Removes a provider from our list when it is unbound
```
KipHamiltons, line 62: I might be misunderstanding this function 
```suggestion
         * Unbinds a provider and removes it from our list
```

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,238 @@
...
+         *
+         * @param id the reaction id for the provider to be removed
```
KipHamiltons, line 64:
```suggestion
         * @param id the reaction id for the provider to be removed
         * @throws if the provider which the id refers to has not been loaded
```

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,238 @@
...
+            on<Configuration>("Director.yaml").then("Configure", [this](const Configuration& config) {
+                // clang-format off
+                std::string lvl = config["log_level"].as<std::string>();
+                if (lvl == "TRACE")      { this->log_level = NUClear::TRACE; }
+                else if (lvl == "DEBUG") { this->log_level = NUClear::DEBUG; }
+                else if (lvl == "INFO")  { this->log_level = NUClear::INFO;  }
+                else if (lvl == "WARN")  { this->log_level = NUClear::WARN;  }
+                else if (lvl == "ERROR") { this->log_level = NUClear::ERROR; }
+                else if (lvl == "FATAL") { this->log_level = NUClear::FATAL; }
+                // clang-format on
```
KipHamiltons, line 123:
```suggestion
                log_level = config["log_level"].as<NUClear::LogLevel>();
```

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,238 @@
...
+            });
+
+            // Removes all the providers for a reaction when it is unbound
+            on<Trigger<Unbind>, Sync<Director>>().then("Remove Provider", [this](const Unbind& unbind) {  //
+                remove_provider(unbind.id);
+            });
+
+            // Add a provider provider
+            on<Trigger<ProvidesReaction>, Sync<Director>>().then("Add Provider", [this](const ProvidesReaction& r) {
+                add_provider(r.type, r.reaction, r.action);
```
KipHamiltons, line 133: have you considered making `add_provider` take the `ProvidesReaction` object directly?

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,238 @@
...
+            });
+
+            // Add a when expression to this provider
+            on<Trigger<WhenExpression>, Sync<Director>>().then("Add When", [this](const WhenExpression& r) {
+                auto it = providers.find(r.reaction->id);
```
KipHamiltons, line 138: If you made this reaction and the one below it into functions like the reactions above, it could be easier to write tests for them. It also makes more sense if you added docstrings for them

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,238 @@
...
+                remove_provider(unbind.id);
+            });
+
+            // Add a provider provider
+            on<Trigger<ProvidesReaction>, Sync<Director>>().then("Add Provider", [this](const ProvidesReaction& r) {
+                add_provider(r.type, r.reaction, r.action);
+            });
+
+            // Add a when expression to this provider
+            on<Trigger<WhenExpression>, Sync<Director>>().then("Add When", [this](const WhenExpression& r) {
```
KipHamiltons, line 137: Are you concerned about the performance of the director, with nearly all of its reactions on the one `Sync` group?

KipHamiltons (commented):
Partial review. I'll continue when my brain feels a little less post-robocup fried

module/extension/Director/src/Director.cpp
TrentHouliston: If the number of tasks that are getting placed into the system exceeds the ability for a single core to tell other modules to do them, there's not a chance in hell that those modules themselves will be able to keep up :P

module/extension/Director/src/Director.cpp
TrentHouliston: I was considering that, Having the when and causing be added to specific functions probably makes it clearer

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,238 @@
...
+        struct StateUpdate {
+            StateUpdate(const uint64_t& provider_id, const std::type_index& type, const int& state)
+                : provider_id(provider_id), type(type), state(state) {}
+
+            /// The provider_id which was waiting on this type
+            uint64_t provider_id;
+            /// The enum type that this enum was listening for
+            std::type_index type;
+            /// The new value for the enum
+            int state;
```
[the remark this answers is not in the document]
TrentHouliston: No need to give them default values (and I shouldn't as there are no sensible defaults).
They're all set within the constructor as flags and there is no default constructor for this type.

If I don't set defaults here and I ever add a new field here that isn't set in the constructor it'll give me a helpful warning. If I set defaults I miss out on that warning.

module/extension/Director/src/Director.cpp
TrentHouliston: The comment that's there is correct. The "Unbind" of a provider emits something which calls this reaction.

So this is a response to being unbound, not an unbinding itself

module/extension/Director/src/Director.cpp
TrentHouliston: I did, but I decided I didn't want to do that as that makes it a specific function. I'd rather it this way in case I ever want to add a provider separate to these reactions

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,238 @@
+#include "Director.hpp"
+
+#include "extension/Configuration.hpp"
+
+namespace module {
```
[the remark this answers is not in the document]
TrentHouliston: Done

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,238 @@
...
+        using TaskPack = std::vector<std::shared_ptr<const DirectorTask>>;
+
+        /**
+         * This message gets emitted when a state we are monitoring is updated.
+         * When a `When` condition is being waited on a reaction will start monitoring that state.
+         * When it updates we check to see if this would now satisfy the conditions and if it does we emit
+         */
+        struct StateUpdate {
+            StateUpdate(const uint64_t& provider_id, const std::type_index& type, const int& state)
+                : provider_id(provider_id), type(type), state(state) {}
```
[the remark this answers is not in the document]
TrentHouliston: Do they really? I've never had that problem before with constructors and I'm pretty sure I have everything turned on

module/extension/Director/src/Director.cpp
TrentHouliston: However now that I think about it I think I will change it to that

module/extension/Director/src/Director.cpp
KipHamiltons: Is it normal to add information about when a function should be called to the function description?

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
KipHamiltons: I am pretty sure. [compiler explorer reckons gcc will](https://godbolt.org/z/bEe11oxh4), but it looks like clang won't.

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
KipHamiltons: Which warning is that? wouldn't you have to add a new field and give it a default value in the class?
Do you plan to delete the default constructor?

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+            std::shared_ptr<NUClear::threading::Reaction> reaction;
+            /// The enum type that this when expression claiming to manipulate
+            std::type_index type;
+            /// The resulting state this provider is claiming to provide if it is executed
+            int resulting_state;
+        };
+
+        /**
+         * Message to tell the Director that a provider task has finished executing
+         */
```
KipHamiltons, line 82: Have you considered making this a `proto` message? It seems like there's nothing too fancy about it, and it would fit that mould

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+        };
+
+        /**
+         * A task message that is to be sent to the Director for execution
+         */
+        struct DirectorTask {
+            DirectorTask(std::type_index type,
+                         uint64_t requester_id,
+                         uint64_t requester_task_id,
+                         std::shared_ptr<void> data,
```
KipHamiltons, line 99: are there any viable alternatives to this? `std::shared_ptr<void>` seems like a problem protobuf could maybe fix? it also rings some (loud) alarm bells in my mind :laughing:
maybe a template somehow??? idk lol. feels spicy

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+                // TODO make this return a special type that can be dereferenced into a T object
+                // TODO then you can add in extra information that can be passed to the provider (like when it is
+                // triggered due to a `Done` event)
+                // TODO here we need to get task information from the director
+                // TODO using r.id we can query the director if this provider is supposed to be running right now and
+                // also get the data for this provider
+                // If this provider is not supposed to be running right now we return a nullptr
+                // By returning nullptr that evaluates to something "falsy" it will prevent this reaction from running
+                // (unless the provider has an optional wrapper around it but then it's the users job to deal with that)
+                return nullptr;
```
KipHamiltons, line 162: The comment begs the question - what's the reason against using `std::optional` here?

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+            uint64_t requester_task_id;
+        };
+
+        /**
+         * A task message that is to be sent to the Director for execution
+         */
+        struct DirectorTask {
+            DirectorTask(std::type_index type,
+                         uint64_t requester_id,
+                         uint64_t requester_task_id,
```
KipHamiltons, line 98: placing parameters of the same type next to each other makes it easier to screw up using your interface

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+        template <typename DSL>
+        static inline void bind(const std::shared_ptr<NUClear::threading::Reaction>& reaction) {
+
+            // Tell the director about this when condition
+            reaction->reactor.powerplant.emit(std::make_unique<commands::WhenExpression>(
+                reaction,
+                typeid(State),
+                [](const int& v) { return expr<int>()(v, value); },
+                []() { return NUClear::dsl::operation::CacheGet<State>::get(); },
+                [](NUClear::Reactor& reactor,
```
KipHamiltons, line 229: there are a few moving parts here. Could you please add a couple of inline comments

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+                    return reactor.on<NUClear::dsl::word::Trigger<State>>().then(
+                        [fn](const State& s) { fn(static_cast<int>(s)); });
+                }));
+        }
+    };
+
+    /**
+     * This DSL word is used to create a promise that at some point when running this reaction the state indicated
+     * by T will be true. It may not be on any particular run of the reaction, in which case it will continue to run
+     * that reaction until it is true, or another graph jump becomes necessary.
```
KipHamiltons, line 240: Another graph jump? I don't understand that in this context.

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+    /**
+     * This is a special task that should be emitted when a provider finishes the task it was given.
+     * When this is emitted the director will re-execute the provider which caused this task to run.
+     *
+     * ```
+     * emit<Task>(std::make_unique<Done>());
+     * ```
+     */
+    struct Done {};
```
KipHamiltons, line 313: Another case where I am curious why you have elected against a protobuf message

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+}  // namespace extension::behaviour
+
+#endif  // EXTENSION_BEHAVIOUR_H
```
KipHamiltons, line 317:
```suggestion
#endif  // EXTENSION_BEHAVIOUR_HPP
```

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+     * task. These tasks form the root of the execution tree and their needs will be met on a highest priority first
+     * basis. These tasks will persist until the provider that they use emits a done task, or the task is re-emitted
+     * with a priority of 0.
+     *
+     * Subtasks:
+     * These are created when a provider task emits a task to complete. These tasks must be emitted each time that
+     * reaction is run to persist as if that reaction runs again and does not emit these tasks they will be
+     * cancelled. Within these tasks the priority is used to break ties between two subtasks that share the same
+     * root task. In these cases the subtask that requested the provider with the highest priority will have its
+     * task executed. The other subtask will be blocked until that task is no longer in the call queue.
```
KipHamiltons, line 271: I am confused about the last sentence. Here is my interpretation
```suggestion
     * task executed. The other subtask will be blocked until the first task is no longer in the call queue.
```

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+     *
+     * If a subtask is emitted with optional then it is compared differently to other tasks when it comes to priority.
+     * This task and all its decendents will be considered optional. If it is compared to a task that does not have
```
KipHamiltons, line 274:
```suggestion
     * This task and all its descendants will be considered optional. If it is compared to a task that does not have
```

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+     * optional in it's parantage, the non optional task will win. However decendents of this task that are not optional
```
KipHamiltons, line 275:
```suggestion
     * optional in it's parentage, the non optional task will win. However descendants of this task that are not optional
```

KipHamiltons (commented):
Complete review. I don't understand some of the NUClear idioms all too well, so you should have someone who does approve this.
I'll review again once you have added the docstrings and tags from my previous review to the other methods too.

shared/message/behaviour/state/Stability.proto
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright 2015 NUbots <nubots@nubots.net>
+ */
+
+syntax = "proto3";
+
+package message.behaviour.state;
+
+enum Stability {
+    // The robot has fallen on the ground
+    FALLEN = 0;
```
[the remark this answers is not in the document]
TrentHouliston: I'm not sure, I left this in as an example, Depending on how you implement getup it shouldn't be a problem to be Fallen.

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+         * Message used to tell the director about a new Causing relationship for this provider
+         */
+        struct CausingExpression {
+            CausingExpression(const std::shared_ptr<NUClear::threading::Reaction>& reaction,
+                              const std::type_index& type,
+                              const int& resulting_state)
+                : reaction(reaction), type(type), resulting_state(resulting_state) {}
+            /// The provider reaction that will cause this state
+            std::shared_ptr<NUClear::threading::Reaction> reaction;
+            /// The enum type that this when expression claiming to manipulate
```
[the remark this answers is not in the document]
TrentHouliston: The English words that are coding constructs thrown in do make it a little more confusing I'll quote them.

Basically, it's the type of the enum that this when expression is working with, so what it's watching for

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,317 @@
...
+            std::type_index type;
+            /// The resulting state this provider is claiming to provide if it is executed
```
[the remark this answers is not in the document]
TrentHouliston: Well, if someone handed you a function and said, "This function does X if you run it enough times" and you didn't have a chance to read and and had to hand it to someone else. I would say well the function claims it does X :P

shared/extension/Behaviour.hpp
TrentHouliston: When writing extensions I keep the messages that will go between the extension and the module that is processing for the extension together.
Nobody should ever be emitting these particular messages outside of this exact file, to do so would be an error.
Therefore I don't want to make it a public message. If I could I would hide them even more than I have by putting them in a subnamespace here.

shared/extension/Behaviour.hpp
TrentHouliston: If you think of one let me know, I hate using void shared pointers but unfortunately, as I have to store a bunch of director tasks for different types in the same list there is no way for me to have them with different templates. Using inheritance or otherwise to do it wouldn't help as I would just have moved the uncertainty in type from this pointer to the type itself.

Fortunately, in the getter component, it knows the type as it's on the other side of the template barrier so it knows what to cast this back to.

shared/extension/Behaviour.hpp
TrentHouliston: This getter will actually have an entire type that's there for it so the user can either decide to have a full context from the Director or just a message that they care about.
I haven't written it yet as its contents will depend on what the algorithm ends up doing.

shared/extension/Behaviour.hpp
TrentHouliston: True, but I still want both of these data types, and hopefully, the only person calling these internal extension messages in Behaviour will be the DSL words below. Nobody should be making these outside of this file in the emit<Task> extension.

shared/extension/Behaviour.hpp
TrentHouliston: As in, if A asked for this to run to make some condition true, but then something else happened that rendered the whole operation moot.
E.g. the kick wants the walk to cause a standing state.
But then the robot falls over before that happens.
Nobody cares about the walk getting to a standing state anymore as another graph jump has happened.

shared/extension/Behaviour.hpp
TrentHouliston: I prefer extensions to be self-contained and not include messages.
That way all the code for this extension is together.

Like how Configuration isn't done with a proto message even though it maybe could have been.

shared/extension/Behaviour.hpp
TrentHouliston: That would be the correct interpretation. But I think I would change it to be "the active task"

shared/extension/Behaviour.hpp
KipHamiltons: So the decision against proto is purely for encapsulation?

shared/extension/Behaviour.hpp
KipHamiltons: @CMurtagh-LGTM u have a good mind for such things. Can you think of any alternatives?

shared/extension/Behaviour.hpp
KipHamiltons: perfect

shared/extension/Behaviour.hpp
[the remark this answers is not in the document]
KipHamiltons: so there is uncertainty then? :p

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
KipHamiltons: also - muh trivial types :(

shared/extension/Behaviour.hpp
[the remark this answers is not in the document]
KipHamiltons:
```suggestion
            /// The enum type that this when expression claims to manipulate
```

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
TrentHouliston: By not defining a default constructor and defining another constructor that means there is no default constructor already

shared/extension/Behaviour.hpp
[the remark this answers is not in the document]
TrentHouliston: Oh 100% :P

shared/extension/Behaviour.hpp
TrentHouliston: And because we don't want other modules actually thinking these are things that should be emitted.

shared/extension/Behaviour.hpp
TrentHouliston: Documented them on the constructor itself and here

module/extension/Director/README.md
```diff
@@ -0,0 +1,14 @@
+Director
```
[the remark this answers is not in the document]
TrentHouliston: I'll fill this in when the director algorithm is done, for now I think it'll end up in way too much flux

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+         * @param data_ the data for this task
+         */
+        TaskInfo(std::shared_ptr<const T> data_) : data(data_) {}
+
+        /**
+         * Having a dereference operator allows NUClear to provide either the TaskInfo<T> object or a const T& object
+         *
+         * @return the data that is contained within this TaskInfo object
+         */
+        const T& operator*() {
```
KipHamiltons, line 62:
```suggestion
        [[nodiscard]] const T& operator*() {
```

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+            return *data;
+        }
+
+        /**
+         * This will control if NUClear should run the task
+         *
+         * @return true     there is data, NUClear should run the task
+         * @return false    there is no data, NUClear should not run the task
+         */
+        operator bool() {
```
KipHamiltons, line 72:
```suggestion
        [[nodiscard]] operator bool() {
```

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+         *
+         * @tparam DSL the NUClear DSL for the on statement
+         *
+         * @param reaction the reaction object that we are binding the Provider to
+         */
+        template <typename DSL>
+        static inline void bind(const std::shared_ptr<NUClear::threading::Reaction>& reaction) {
+
+            // Tell the Director
+            reaction->reactor.powerplant.emit(
```
Bidski, line 102: Can `reaction` ever be `nullptr` here?

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+    template <typename T>
+    struct Provides : public ProviderBase<T, commands::ProviderClassification::NORMAL> {};
+
+    /**
+     * The When DSL word is used to limit access to a Provider unless a condition is true.
+     * This will prevent a Provider from running if the state is false.
+     * However if there is a Entering<T> or Leaving reaction that has a Causing relationship for this When, it will
+     * be executed to ensure that the Provider can run.
+     *
+     * @tparam T the condition expression that must be true in order to execute this task.
```
Bidski, line 183: This does not line up with the actual template paramters

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+         *
+         * @tparam DSL the DSL from NUClear
+         *
+         * @param reaction the reaction that is having this when condition bound to it
+         */
+        template <typename DSL>
+        static inline void bind(const std::shared_ptr<NUClear::threading::Reaction>& reaction) {
+
+            // Tell the director about this when condition
+            reaction->reactor.powerplant.emit(std::make_unique<commands::WhenExpression>(
```
Bidski, line 199: Can `reaction` ever be `nullptr` here?

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+                reaction,
+                // typeindex of the enum value
+                typeid(State),
+                // Function that uses expr to determine if the passed value v is valid
+                [](const int& v) { return expr<int>()(v, value); },
+                // Function that uses get to get the current state of the reaction
+                []() { return NUClear::dsl::operation::CacheGet<State>::get(); },
```
Bidski, line 206:
```suggestion
                [](const int& v) -> bool { return expr<int>()(v, value); },
                // Function that uses get to get the current state of the reaction
                []() -> int { return NUClear::dsl::operation::CacheGet<State>::get(); },
```

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+                }));
+        }
+    };
+
+    /**
+     * This DSL word is used to create a promise that at some point when running this reaction the state indicated
+     * by T will be true. It may not be on any particular run of the Provider, in which case it will continue to run
+     * that reaction when new tasks arrive until it is true, or a change elsewhere changes which Provider to run.
+     *
+     * @tparam T the condition expression that this reaction will make true before leaving.
```
Bidski, line 221: This does not reflect the actual template parameters

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+     * reaction is run to persist as if that reaction runs again and does not emit these tasks they will be
+     * cancelled. Within these tasks the priority is used to break ties between two subtasks that share the same
+     * root task. In these cases the subtask that requested the Provider with the highest priority will have its
+     * task executed. The other subtask will be blocked until the active task is no longer in the call queue.
+     *
+     * If a subtask is emitted with optional then it is compared differently to other tasks when it comes to priority.
+     * This task and all its decendents will be considered optional. If it is compared to a task that does not have
+     * optional in it's parentage, the non optional task will win. However decendents of this task that are not optional
+     * will compare to eachother as normal. If two tasks both have optional in their parentage they will be compared as
+     * normal.
```
Bidski, line 263:
```suggestion
     * This task and all its descendants will be considered optional. If it is compared to a task that does not have
     * optional in it's parentage, the non optional task will win. However, descendants of this task that are not optional
     * will compare to each other as normal. If two tasks both have optional in their parentage they will be compared as
     * normal.
```

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+            }
+            // Erase from our list of Providers
+            providers.erase(id);
+        }
+        else {
+            throw std::runtime_error("Attempted to remove a Provider that was not loaded");
+        }
+    }
+
+    void Director::add_when(const WhenExpression& when) {
```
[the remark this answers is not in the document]
TrentHouliston: Nooo that's terrible! Docstrings only in the header. You should have a single source of truth. Imagine having two places documentation is available and if they get out of sync

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+     *
+     * @tparam T the Provider type that this task is for
+     */
+    template <typename T>
+    struct Task {
+
+        /**
+         * Emits a new task for the Director to handle.
+         *
+         * @param powerplant the powerplant object this emit is being called on
```
Bidski, line 273: We can have more than one `PowerPlant`? I thought this was a singleton object?

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+            // Add it to the list of when conditions
+            provider->when.emplace_back(when.type, when.validator, when.current, handle);
+        }
+        else {
+            throw std::runtime_error("When statements must come after a Provides, Entering or Leaving statement");
+        }
+    }
+
+    void Director::add_causing(const CausingExpression& causing) {
```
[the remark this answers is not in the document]
TrentHouliston: Likewise, docstrings go in the header, never in the c++ file. Unless there is no header for a c++ file

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+        template <typename DSL>
+        static inline void bind(const std::shared_ptr<NUClear::threading::Reaction>& reaction) {
+            // Tell the director
+            reaction->reactor.powerplant.emit(
+                std::make_unique<commands::CausingExpression>(reaction, typeid(State), value));
+        }
+    };
+
+    /**
+     * A Task to be executed by the director, and a priority with which to execute this task.
```
Bidski, line 242: What indicates best priority? Higher is best?

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+         * Emits a new task for the Director to handle.
+         *
+         * @param powerplant the powerplant object this emit is being called on
+         * @param data       the data element of the task
+         * @param name       a string identifier to help with debugging
+         * @param priority   the priority that this task is to be executed with
+         * @param optional   if this task is an optional task and can be taken over by lower priority non optional tasks
+         */
+        static void emit(NUClear::PowerPlant& powerplant,
+                         std::shared_ptr<T> data,
```
Bidski, line 280: `const&`? or maybe just `&`?

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+                         const std::string& name = "",
+                         int priority            = 1,
+                         bool optional           = false) {
```
Bidski, line 283:
```suggestion
                         const int& priority            = 1,
                         const bool& optional           = false) {
```

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+
+            // Work out who is sending the task so we can determine if it's a subtask
+            auto* task           = NUClear::threading::ReactionTask::get_current_task();
+            uint64_t reaction_id = task ? task->parent.id : -1;
+            uint64_t task_id     = task ? task->id : -1;
```
Bidski, line 288:
```suggestion
            uint64_t reaction_id = (task != nullptr) ? task->parent.id : -1;
            uint64_t task_id     = (task != nullptr) ? task->id : -1;
```

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+            auto g_it = std::find(group.providers.begin(), group.providers.end(), provider);
+            if (g_it != group.providers.end()) {
+                group.providers.erase(g_it);
+            }
+
+            // Now we need to deal with the cases where this Providers was in use when it was unbound
+            if (provider->active) {
+                if (group.providers.empty()) {
+                    // This is now an error, there are no Providers to service the task
+                    log<NUClear::ERROR>("The last Provider for a type was removed while there were still tasks for it");
```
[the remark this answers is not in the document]
TrentHouliston: I did consider throwing there, but I didn't want the flow of the program to end. Plus there's a chance this is getting called from a destructor and throwing in a destructor makes for terminate called recursively

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+        on<Trigger<CausingExpression>, Sync<Director>>().then("Add Causing", [this](const CausingExpression& causing) {
+            add_causing(causing);
+        });
+
+        // A task has arrived, either it's a root task and send it off, or build up our pack for when it's done
+        on<Trigger<DirectorTask>, Sync<DirectorTask>>().then(
+            "Director Task",
+            [this](std::shared_ptr<const DirectorTask> task) {
+                // Root level task, make the TaskPack immediately and send it off to be executed as a root task
+                if (providers.count(task->requester_id) == 0) {
```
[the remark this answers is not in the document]
TrentHouliston: The requester_id is the reaction_id of whoever is emitting this. So it's not just if the requester_id is 0, it's if I have no idea who the requester is (i.e. they are not another provider) So they'll either be a random reaction somewhere, or they'll be a reactionless emit from some other non NUClear thread

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+                                       "cannot be executed as the Provider",
+                                       providers[task->requester_id]->reaction->identifier[0],
+                                       "is not active and cannot make subtasks");
+                }
+            });
+
+        // This reaction runs when a Provider finishes to send off the task pack to the main director
+        on<Trigger<ProviderDone>, Sync<DirectorTask>>().then("Package Tasks", [this](const ProviderDone& done) {
+            // Get all the tasks that were emitted by this provider and send it as a task pack
+            auto e = pack_builder.equal_range(done.requester_task_id);
```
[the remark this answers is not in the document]
TrentHouliston: Somewhat irrelevant since it's returning an std::pair with two iterators in it.
Also it'll be an rvalue not an lvalue that is returned so making it a const reference will work because of a technicality in C++, but it's not a reference it'll actually be an rvalue in disguise

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+                tasks.emplace_back(it->second);
+            }
+
+            // Sort the task pack so highest priority tasks come first
+            std::sort(tasks.begin(), tasks.end(), [](const auto& a, const auto& b) {
+                return a->priority < b->priority;
+            });
+
+            // Emit the task pack
+            emit(std::make_unique<TaskPack>(TaskPack(std::move(tasks))));
```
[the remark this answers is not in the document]
TrentHouliston: Good point, no point having the TaskPack constructor here when we are already doing make_unique

shared/extension/behaviour/commands.hpp
```diff
@@ -0,0 +1,197 @@
...
+ * This namespace holds all of the communication primitives that are used by the Behaviour header to send messages to
+ * the Director algorithm. You shouln't be emitting any of these messages outside of the behaviour header.
+ */
+namespace extension::behaviour::commands {
+
+    /**
+     * The classification type of the Provider
+     */
+    enum class ProviderClassification {
+        /// NORMAL Providers are the usual Providers that execute in a loop if needed
```
Bidski, line 37: Not sure I like `NORMAL`

shared/extension/behaviour/commands.hpp
```diff
@@ -0,0 +1,197 @@
...
+                         const std::type_index& type_,
+                         const ProviderClassification& classification_)
+            : reaction(reaction_), type(type_), classification(classification_) {}
+
+        /// The reaction for this Provider
+        std::shared_ptr<NUClear::threading::Reaction> reaction;
+        /// The data type that this Provider services
+        std::type_index type;
+        /// The action type that this Provider is using (normal, entering, leaving)
+        ProviderClassification classification;
```
Bidski, line 67:
```suggestion
        std::shared_ptr<NUClear::threading::Reaction> reaction{};
        /// The data type that this Provider services
        std::type_index type{};
        /// The action type that this Provider is using (normal, entering, leaving)
        ProviderClassification classification = ProviderClassification::NORMAL;
```

shared/extension/behaviour/commands.hpp
```diff
@@ -0,0 +1,197 @@
...
+         * @param validator_    a function that will determine if the passed value satisfies this when condition
+         * @param current_      a function that will get the current state of the reaction
+         * @param binder_       a function that can be used to bind a reaction that monitors when the state changes
+         */
+        WhenExpression(const std::shared_ptr<NUClear::threading::Reaction>& reaction_,
+                       const std::type_index& type_,
+                       bool (*validator_)(const int&),
+                       int (*current_)(),
+                       NUClear::threading::ReactionHandle (*binder_)(NUClear::Reactor&,
+                                                                     std::function<void(const int&)>))
```
Bidski, line 89: Since when did you learn C? :stuck_out_tongue:

shared/extension/behaviour/commands.hpp
```diff
@@ -0,0 +1,197 @@
...
+            : reaction(reaction_), type(type_), validator(validator_), current(current_), binder(binder_) {}
+
+        /// The Provider reaction that is contingent on this when condition
+        std::shared_ptr<NUClear::threading::Reaction> reaction;
+        /// The enum type that this when expression is looking at
+        std::type_index type;
```
Bidski, line 95:
```suggestion
        std::shared_ptr<NUClear::threading::Reaction> reaction{};
        /// The enum type that this when expression is looking at
        std::type_index type{};
```

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+        auto& group = groups[provides.type];
+
+        // Check if we already inserted this element as a Provider
+        if (providers.count(provides.reaction->id) != 0) {
+            throw std::runtime_error("You cannot have multiple Provider DSL words in a single on statement.");
+        }
+
+        auto provider = std::make_shared<Provider>(provides.classification, provides.type, provides.reaction);
+        group.providers.emplace_back(provider);
```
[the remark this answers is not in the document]
TrentHouliston: It's a valid point, just hope clang-tidy doesn't yell at me for it

shared/extension/behaviour/commands.hpp
```diff
@@ -0,0 +1,197 @@
...
+        CausingExpression(const std::shared_ptr<NUClear::threading::Reaction>& reaction,
+                          const std::type_index& type,
+                          const int& resulting_state)
+            : reaction(reaction), type(type), resulting_state(resulting_state) {}
+        /// The Provider reaction that will cause this state
+        std::shared_ptr<NUClear::threading::Reaction> reaction;
+        /// The enum type that this `When` expression claims to manipulate
+        std::type_index type;
+        /// The resulting state this Provider is claiming to provide if it is executed
+        int resulting_state;
```
Bidski, line 125:
```suggestion
        std::shared_ptr<NUClear::threading::Reaction> reaction{};
        /// The enum type that this `When` expression claims to manipulate
        std::type_index type{};
        /// The resulting state this Provider is claiming to provide if it is executed
        int resulting_state = 0;
```

shared/extension/behaviour/commands.hpp
```diff
@@ -0,0 +1,197 @@
...
+         *
+         * @param requester_id      the reaction_id of the NUClear reaction that finished executing
+         * @param requester_task_id the reaction_task_id of the NUClear reaction task that finished executing
+         */
+        ProviderDone(const uint64_t& requester_id, const uint64_t& requester_task_id)
+            : requester_id(requester_id), requester_task_id(requester_task_id) {}
+        /// The reaction_id of the Provider that finished
+        uint64_t requester_id;
+        /// The specific task_id of the Provider that finished
+        uint64_t requester_task_id;
```
Bidski, line 144:
```suggestion
        uint64_t requester_id = 0;
        /// The specific task_id of the Provider that finished
        uint64_t requester_task_id = 0;
```

shared/extension/behaviour/commands.hpp
```diff
@@ -0,0 +1,197 @@
...
+    };
+
+    /**
+     * A task message that is to be sent to the Director for execution
+     */
+    struct DirectorTask {
```
Bidski, line 150:
```suggestion
    template <typename T>
    struct DirectorTask {
```

shared/extension/behaviour/commands.hpp
```diff
@@ -0,0 +1,197 @@
...
+         * @param requester_task_id     the task_id of the reaction task that emitted this
+         * @param data                  the task data to be sent to the provider
+         * @param name                  a string name for this task for use in debugging
+         * @param priority              the priority that this task is to run with
+         * @param optional              whether this task is optional or not
+         */
+        DirectorTask(std::type_index type_,
+                     uint64_t requester_id_,
+                     uint64_t requester_task_id_,
+                     std::shared_ptr<void> data_,
```
Bidski, line 166:
```suggestion
        DirectorTask(const std::type_index& type_,
                     const uint64_t& requester_id_,
                     const uint64_t& requester_task_id_,
                     const std::shared_ptr<T>& data_,
```

shared/extension/behaviour/commands.hpp
```diff
@@ -0,0 +1,197 @@
...
+        /// The reaction task id of the requester (if it is a Provider later a ProviderDone will be emitted)
+        uint64_t requester_task_id;
+        /// The data for the command, (the data that will be given to the Provider)
+        std::shared_ptr<void> data;
+        /// A name for this task to be shown in debugging systems
+        std::string name;
+        /// The priority with which this task is to be executed
+        int priority;
+        /// If this task is optional
+        bool optional;
```
Bidski, line 191:
```suggestion
        std::type_index type{};
        /// The reaction id of the requester (could be the id of a Provider)
        uint64_t requester_id = 0;
        /// The reaction task id of the requester (if it is a Provider later a ProviderDone will be emitted)
        uint64_t requester_task_id = 0;
        /// The data for the command, (the data that will be given to the Provider)
        std::shared_ptr<T> data{};
        /// A name for this task to be shown in debugging systems
        std::string name = "";
        /// The priority with which this task is to be executed
        int priority = 0;
        /// If this task is optional
        bool optional = false;
```

shared/message/behaviour/state/Stability.proto
```diff
@@ -0,0 +1,35 @@
...
+    // The robot is not on the ground, but is not dynamically stable
+    FALLING = 1;
+    // The robot is dynamically stable, provided continued control input it should not fall
+    DYNAMIC = 2;
+    // The robot is statically stable, if control input ceased immediately it would not fall
+    STATIC = 3;
+    // The robot is standing upright on both feet
+    STANDING = 4;
+}
```
Bidski, line 35:
```suggestion
    // The robot is standing upright on both feet
    STANDING = 4;
    // The robot is doing a backflip
    BACKFLIP = 5;
}
```

Bidski (changes requested):

shared/extension/behaviour/commands.hpp
TrentHouliston: And if I were to do this, how would I later on<Trigger<DirectorTask>>?

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+    struct StateUpdate {
+        StateUpdate(const uint64_t& provider_id_, const std::type_index& type_, const int& state_)
+            : provider_id(provider_id_), type(type_), state(state_) {}
+
+        /// The reaction id of the Provider which was waiting on this type
+        uint64_t provider_id;
+        /// The enum type that this enum was listening for
+        std::type_index type;
+        /// The new value for the enum
+        int state;
```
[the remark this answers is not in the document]
TrentHouliston: By having a non-default constructor the default constructor is deleted by default. I don't need to explicitly delete it

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+    };
+
+    /**
+     * Adds a Provider to the list of active Providers
+     *
+     * @param data_type the data type that this Provider services
+     * @param r         the reaction that is used to run this Provider
+     * @param action    the action type for this Provider (entering, leaving, normal)
```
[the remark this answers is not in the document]
TrentHouliston: Fixed

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
TrentHouliston: Have you considered looking in the header? Or maybe hovering your mouse over the function?

shared/extension/behaviour/commands.hpp
TrentHouliston: No default for these otherwise I don't get to take advantage of compiler warnings.
Since I only have a non-default constructor if I were not to set a new field I made in the constructor it would yell at me.

If I set defaults for everything you can get bugs that the compiler would otherwise warn you about.
Only set defaults if you have a default constructor.

shared/extension/behaviour/commands.hpp
TrentHouliston: No defaults if no default constructor

shared/message/behaviour/state/Stability.proto
TrentHouliston: Is doing a backflip considered more stable than standing? :P

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
TrentHouliston: They're in the header, you don't double document things there should be a single source of truth.
If you're in vscode you can just hover the function name to get it's docs

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    Director::Director(std::unique_ptr<NUClear::Environment> environment) : Reactor(std::move(environment)) {
+
+        on<Configuration>("Director.yaml").then("Configure", [this](const Configuration& config) {
+            log_level = config["log_level"].as<NUClear::LogLevel>();
+        });
+
+        // Removes all the Providers for a reaction when it is unbound
+        on<Trigger<Unbind>, Sync<Director>>().then("Remove Provider", [this](const Unbind& unbind) {  //
```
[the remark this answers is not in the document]
TrentHouliston: To stop clang-format wrapping this onto a weird looking line

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+            });
+
+        // This reaction runs when a Provider finishes to send off the task pack to the main director
+        on<Trigger<ProviderDone>, Sync<DirectorTask>>().then("Package Tasks", [this](const ProviderDone& done) {
+            // Get all the tasks that were emitted by this provider and send it as a task pack
+            auto e = pack_builder.equal_range(done.requester_task_id);
+            TaskPack tasks;
+            for (auto it = e.first; it != e.second; ++it) {
+                tasks.emplace_back(it->second);
+            }
```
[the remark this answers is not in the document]
TrentHouliston: That doesn't work, this isn't a thing that has begin/end, it's a std::pair with a first/second. So that kind of foreach loop doesn't work

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
TrentHouliston: fine, `s/e/range/`

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+            // Sort the task pack so highest priority tasks come first
+            std::sort(tasks.begin(), tasks.end(), [](const auto& a, const auto& b) {
+                return a->priority < b->priority;
+            });
```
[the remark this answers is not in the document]
TrentHouliston: I want higher priority (bigger first).
So yeah this looks wrong, should be >

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
TrentHouliston: Originally TaskPack was a struct, then I made it a using for an std::vector but this was left behind

module/extension/Director/src/director/Provider.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+namespace module::extension {
+
+    /**
+     * A Provider instance which is able to execute a specific task type
+     */
+    struct Provider {
+
+        /**
+         * The data held by a when condition in order to monitor when the condition will be met, as well as to
```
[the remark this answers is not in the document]
TrentHouliston: I agree, I have noticed that

module/extension/Director/src/director/Provider.hpp
```diff
@@ -0,0 +1,75 @@
...
+         * compare it to causing statements to determine if those causing statements would satisfy this Provider
+         */
+        struct WhenCondition {
+            WhenCondition(const std::type_index& type_,
+                          bool (*validator_)(const int&),
+                          int (*current_)(),
```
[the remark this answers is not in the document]
TrentHouliston: Maybe!

module/extension/Director/src/director/Provider.hpp
```diff
@@ -0,0 +1,75 @@
...
+                          NUClear::threading::ReactionHandle handle_)
+                : type(type_), validator(validator_), current(current_), handle(std::move(handle_)) {}
+
+            /// The type of state that this condition is checking
+            std::type_index type;
```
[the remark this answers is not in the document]
TrentHouliston: No, no default constructor and no default values

module/extension/Director/src/director/ProviderGroup.hpp
```diff
@@ -0,0 +1,63 @@
...
+        /**
+         * The current state of the Provider group
+         */
+        enum State {
+            /// This Provider currently isn't doing anything
+            IDLE,
+            /// This Provider is currently in an "Entering" state
+            ENTERING,
+            /// This Provider group is currently in a "Normal" state
+            NORMAL,
```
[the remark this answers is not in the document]
TrentHouliston: It's that it's an on<Provides> not an on<Leaving> or on<Entering>

shared/extension/Behaviour.hpp
```diff
@@ -0,0 +1,313 @@
...
+     *
+     * @tparam T the type of task that is stored in this task_info object
+     */
+    template <typename T>
+    struct TaskInfo {
+
+        /**
+         * Default constructor, as data will be null NUClear will be told not to execute this task.
+         */
+        TaskInfo() {}
```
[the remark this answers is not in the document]
TrentHouliston: Clang-tidy yells at me for doing that because that's what the default constructor already does

shared/extension/Behaviour.hpp
TrentHouliston: NUClear would explode if it were possible :P

shared/extension/Behaviour.hpp
TrentHouliston: No

shared/extension/Behaviour.hpp
TrentHouliston: It is, probably could change the comment here to describe it's the NUClear context you have in this function

shared/extension/Behaviour.hpp
TrentHouliston: Higher is best, although the algorithm hasn't been written yet :P

module/extension/Director/src/Director.cpp
```diff
@@ -212,19 +211,19 @@ namespace module::extension {
         // This reaction runs when a Provider finishes to send off the task pack to the main director
         on<Trigger<ProviderDone>, Sync<DirectorTask>>().then("Package Tasks", [this](const ProviderDone& done) {
             // Get all the tasks that were emitted by this provider and send it as a task pack
-            auto e = pack_builder.equal_range(done.requester_task_id);
-            TaskPack tasks;
+            auto e     = pack_builder.equal_range(done.requester_task_id);
+            auto tasks = std::make_unique(tasks);
             for (auto it = e.first; it != e.second; ++it) {
```
Bidski:
```suggestion
            auto task_pack     = pack_builder.equal_range(done.requester_task_id);
            auto tasks = std::make_unique(tasks);
            for (auto it = task_pack.first; it != task_pack.second; ++it) {
```

Bidski (changes requested):

shared/extension/behaviour/commands.hpp
TrentHouliston: No default constructor therefore no default values

shared/extension/behaviour/commands.hpp
TrentHouliston: No default constructor therefore no default values

module/extension/Director/src/Director.cpp
TrentHouliston: I chose range

module/extension/Director/src/director/ProviderGroup.hpp
```diff
@@ -0,0 +1,63 @@
...
+        /// The current state of this Provider group
+        State state = State::IDLE;
+        /// List of individual Providers that can service tasks for this type
+        std::vector<std::shared_ptr<Provider>> providers;
+        /// The current task that is running on this Provider
+        std::shared_ptr<const ::extension::behaviour::commands::DirectorTask> active_task;
+        /// The queue of tasks waiting to run if the situation changes
+        std::vector<std::shared_ptr<const ::extension::behaviour::commands::DirectorTask>> task_queue;
+        /// List of subordinate provider groups
+        std::vector<std::map<std::type_index, ProviderGroup>::iterator> subordinates;
```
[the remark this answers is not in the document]
TrentHouliston: These are only needed fo things that don't have default constructors like raw structs

module/extension/Director/src/director/Provider.hpp
```diff
@@ -0,0 +1,75 @@
...
+        /// The data type this Provider is for
+        std::type_index type;
+        /// The reaction object to run in order to run this Provider
+        std::shared_ptr<NUClear::threading::Reaction> reaction;
+        /// Any when conditions that are required by this Provider
+        std::vector<WhenCondition> when;
+        /// A list of types and states that are caused by running this Provider
+        std::map<std::type_index, int> causing;
+        /// A boolean that is true if this Provider is currently active and running (can make subtasks)
+        bool active;
```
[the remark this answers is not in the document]
TrentHouliston: No default constructor therefore no default values

shared/extension/Behaviour.hpp
TrentHouliston: Indeed

shared/extension/Behaviour.hpp
TrentHouliston: Why put the return types in?

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+        auto& group = groups[provides.type];
+
+        // Check if we already inserted this element as a Provider
+        if (providers.count(provides.reaction->id) != 0) {
+            throw std::runtime_error("You cannot have multiple Provider DSL words in a single on statement.");
+        }
+
+        auto provider = std::make_shared<Provider>(provides.classification, provides.type, provides.reaction);
+        group.providers.push_back(provider);
```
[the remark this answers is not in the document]
Bidski: I think it should be mentioned that `push_back` will only accept objects of the expected type, or objects which can be casted to said type, and the new element in the vector with either be copy or move constructed (via the appropriate constructor). `emplace_back` will pass its arguments to a constructor of the expected type and will therefore "construct in place". Having said this, the copy and move constructors are also constructors that `emplace_back` can call and so I don't see any reason why we should use `push_back` in this instance?

module/extension/Director/src/Director.cpp
```diff
@@ -0,0 +1,254 @@
...
+        if (it != providers.end()) {
+            auto provider = it->second;
+
+            // Add a reaction that will listen for changes to this state and notify the director
+            auto id               = when.reaction->id;
+            auto type             = when.type;
+            auto validator        = when.validator;
+            ReactionHandle handle = when.binder(*this, [this, id, type, validator](const int& state) {  //
+                if (validator(state)) {
+                    emit(std::make_unique<StateUpdate>(id, type, state));
```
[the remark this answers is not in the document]
TrentHouliston: Nah, then the pointer wouldn't be unique

shared/extension/Behaviour.hpp
TrentHouliston: NUClear doesn't do that and I'm copying it... There might even be a reason why...
Although I don't remember right now :(

shared/extension/behaviour/commands.hpp
TrentHouliston: Come up with a sensible alternative

shared/extension/behaviour/commands.hpp
TrentHouliston: No default constructor, no default values

shared/extension/Behaviour.hpp
TrentHouliston: I mean I guess sure I can put them in anyway...
Also you have drawn my attention to the fact that that get is wrong as it returns a shared_ptr

Bidski: consistency (with the lambda two lines down), and the raw function pointers you are using (in `WhenExpression`) explicitly specify what the return type should be, so why not?

Bidski (changes requested):

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
TrentHouliston: @KipHamiltons

module/extension/Director/src/Director.cpp
KipHamiltons, line 92: If you're worried about throwing in this because of destructors, then should you reconsider the throw at the end of this function?

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
KipHamiltons: I agree completely. You should probably remove the other docstrings from this .cpp then :p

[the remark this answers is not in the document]
TrentHouliston: I did that :P didn't realise I had double documented

module/extension/Director/src/Director.cpp
TrentHouliston: Nah in that case I want it to explode.
That line should be unreachable unless something very bad and strange has happened.

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
KipHamiltons: ahh true. That makes sense

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
KipHamiltons: I think he got caught out by the same thing as me. There are docstrings for other functions in here, but not this one. They should be removed.

Bidski (approved):
Looks good enough to me. I guess this is going to change (and bugs will be uncovered) once we actually start implementing things anyway

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
KipHamiltons: Because we're not using `emplace_back` as the constructor. My understanding (stolen shamelessly from man with big brain  [here](https://www.youtube.com/watch?v=jKS9dSHkAZY) is that `push_back` is good for when you have already constructed the object, whereas `emplace_back` is when you're constructing in place. There's certainly a case to be made here for removing the binding to `provider` and constructing it in the `emplace_back`, but as it stands, if you're going to construct it first, I think it should be `push`

KipHamiltons (approved):
NURecord for most comments on a PR :grin: 
Sorry @LachlanCourt

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
CMurtagh-LGTM: Usually the compiler will optimise and omit the copy anyway.

PR #187, opened 2018-07-03
Updates kinematics information for the igus

Bidski:
- Corrects knee offset value so that hip pitch, knee, and ankle pitches all align vertically in zombie.
- Provides accurate measurements for limb lengths and distances between servos
- Adds an extra offset from hip roll to hip pitch in the forward kinematics calculations (does something need to happen in IK too?)
- Renames arm kinematics in config file and adds comments to for each parameter
- Removes darwin configs and makes the igus config the default one


module/platform/darwin/HardwareIO/data/config/HardwareIO.yaml
```diff
@@ -1,7 +1,7 @@
 battery:
```
JakeFountain: I dont know about these values

module/motion/KinematicsConfiguration/src/KinematicsConfiguration.cpp
```diff
@@ -114,37 +114,25 @@ namespace motion {
...
-        model.arm.SHOULDER_LENGTH            = objShoulder["length"].as<float>();
-        model.arm.SHOULDER_WIDTH             = objShoulder["width"].as<float>();
-        model.arm.SHOULDER_HEIGHT            = objShoulder["height"].as<float>();
-        model.arm.UPPER_ARM_LENGTH           = objUpperArm["length"].as<float>();
-        model.arm.UPPER_ARM_Y_OFFSET         = objUpperArm["offset"].as<arma::vec2>()[0];
-        model.arm.UPPER_ARM_X_OFFSET         = objUpperArm["offset"].as<arma::vec2>()[1];
-        model.arm.LOWER_ARM_LENGTH           = objLowerArm["length"].as<float>();
-        model.arm.LOWER_ARM_Y_OFFSET         = objLowerArm["offset"].as<arma::vec2>()[0];
-        model.arm.LOWER_ARM_Z_OFFSET         = objLowerArm["offset"].as<arma::vec2>()[1];
+        arma::vec3 shoulder         = objArm["shoulder"].as<arma::vec3>();
```
JakeFountain: LGTM

shared/utility/motion/ForwardKinematics.h
```diff
@@ -157,6 +157,8 @@ namespace motion {
                 return positions;
             }
 
+            // Translate to hip pitch joint
+            runningTransform = runningTransform.translateZ(model.leg.HIP_PITCH_OFFSET);
```
JakeFountain: This needs to happen before the hip roll transform is set.

module/platform/darwin/HardwareIO/data/config/HardwareIO.yaml
Bidski: Most of these changes relate to the old default configuration file being for the darwin. The only values of interest here are the knee offset which was updated from `pi / 8` to `0.45` which we determined from `scripttuner` and zombie

tayloryoung6396:
This should be merged

PR #190, opened 2019-01-25
Updates PS3 Walk

Bidski:
PS3 walk now functions more correctly.

PS3Walk is now able to
- Control walking (triangle + left joystick) and kicking (L1 and R1).
- Run the stand script (circle)
- Run a defined set of scripts in a specific order (cross + config file)

Head control doesn't seem to work, but thats a problem for future me.


module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
```diff
@@ -29,28 +32,43 @@
...
+            on<Configuration>("PS3Walk.yaml").then([this](const Configuration& config) {
+                joystick.connect(config["controller_path"], config["accelerometer_path"]);
+
+                actions.clear();
+                for (const auto& action : config["action_scripts"].as<std::vector<std::string>>()) {
+                    actions.push_back(action);
+                }
+            });
 
             on<Every<1, std::chrono::milliseconds>, Single>().then([this] {
```
TrentHouliston: Make this on<IO> for the joystick instead

shared/message/input/JoystickEvent.proto
```diff
@@ -0,0 +1,87 @@
...
+        UNKNOWN = 0;
+        BUTTON  = 1;    // Button press/release
+        AXIS    = 2;    // Joystick movement
+        INIT    = 128;  // Initial state of device
+    }
+
+    enum Axis {
+        LEFT_JOYSTICK_HORIZONTAL  = 0;
+        LEFT_JOYSTICK_VERTICAL    = 1;
+        L2                        = 2;
```
tayloryoung6396: Why's L2 an Axis? Should i not be a button?

Bidski: Probably because the driver reports the L2/R2 events as being "axis" events.

TrentHouliston: L2 and R2 are variable, not just on off

shared/message/input/JoystickEvent.proto
```diff
@@ -0,0 +1,87 @@
...
+     * The timestamp of the event
+     */
+    google.protobuf.Timestamp timestamp = 1;
+
+    /**
+     * The value associated with this joystick event.
+     * For buttons this will be either 1 (down) or 0 (up).
+     * For axes, this will range between -32768 and 32767.
+     */
+    int32 value = 2;
```
TrentHouliston: You should make this a float between -1 and 1

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
```diff
@@ -17,60 +17,203 @@
...
+                    }
+                }
+            });
+
+            // Trigger when the joystick accelerometer has an event to read
+            on<IO>(joystick_acc_fd, IO::READ).then([this] {
+                // Accelerometer comes in here .... if you know which axis is which
+            });
+
+            on<Trigger<JoystickEvent>>().then([this](const JoystickEvent& event) {
```
TrentHouliston: You should move this to another module so that people can still use the controller to test various other things without having to have an entire walk/wall controller coming along with it.

shared/message/input/JoystickEvent.proto
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * The event type.
+     */
+    EventType type = 3;
+
+    /**
+     * The axis/button number.
+     */
+    Axis   axis   = 4;
```
TrentHouliston: The axis event and button event should probably be separate messages. Maybe a separate message per button type/axis type.

module/behaviour/strategy/PS3Walk/src/PS3Walk.h
```diff
@@ -23,54 +23,32 @@
...
         private:
-            Joystick joystick;  // TODO: make configurable
+            size_t id;
             arma::vec strafe{0, 0};
-            float rotationalSpeed = 0;
-            bool moving           = false;
-            bool headLocked       = false;
-            float headPitch       = 0;
-            float headYaw         = 0;
+            arma::vec prevStrafe{0, 0};
```
MattAmos: Is it worth Eigen-ifying this module?

module/motion/ScriptEngine/data/scripts/igus/LiftSign.yaml
```diff
@@ -0,0 +1,328 @@
+- duration: 1000
```
MattAmos: We can probably delete the LiftSign scripts

module/motion/ScriptEngine/data/scripts/igus/LiftSign1.yaml
```diff
@@ -0,0 +1,246 @@
+- duration: 1000
```
MattAmos: Delete

module/motion/ScriptEngine/data/scripts/igus/StandSign.yaml
```diff
@@ -0,0 +1,82 @@
+- duration: 3000
```
MattAmos: Delete

module/behaviour/strategy/PS3Walk/src/PS3Walk.h
TrentHouliston: There should be very little change required

module/input/PS3Controller/src/PS3Controller.cpp
```diff
@@ -0,0 +1,244 @@
...
+            // Timestamp: uint32_t
+            // Value:     int16_t
+            // Type:      uint8_t
+            // Number:    uint8_t
+            constexpr size_t JOYSTICK_EVENT_SIZE =
+                sizeof(uint32_t) + sizeof(int16_t) + sizeof(uint8_t) + sizeof(uint8_t);
+
+            // Try to read a JoystickEvent worth of data
+            std::array<uint8_t, JOYSTICK_EVENT_SIZE> buffer;
+            auto num_bytes = read(controller_fd, buffer.data(), buffer.size());
```
TrentHouliston: This only reads a single packet, you should keep reading these until there are no more. The way it works (you’re reading from a kernel driver not a serial device) you’ll always get exactly this size packet. This also means you can just read into a pragma packed struct to avoid the later reinterpret casts

module/input/PS3Controller/src/PS3Controller.cpp
```diff
@@ -0,0 +1,244 @@
...
+            if (num_bytes > -1) {
+                for (const uint8_t& byte : buffer) {
+                    event_buffer.push_back(byte);
+                }
+            }
+
+            while (event_buffer.size() >= JOYSTICK_EVENT_SIZE) {
+                // Extract timestamp from event buffer in milliseconds and convert it to a NUClear::clock::time_point
+                auto timestamp = std::chrono::time_point<NUClear::clock, std::chrono::milliseconds>(
```
TrentHouliston: Are we sure this is the right clock for this? It might be steady clock vs system clock etc. it might be that a duration makes more sense

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
```diff
@@ -17,123 +17,143 @@
...
+                    NUClear::log("Requesting Right Front Kick");
+                    emit(std::make_unique<KickScriptCommand>(
+                        KickScriptCommand(Eigen::Vector3d(1, 0, 0),  // vector pointing forward relative to robot
+                                          LimbID::RIGHT_LEG)));
+                }
+            });
+
+            on<Trigger<LeftJoystick>>().then([this](const LeftJoystick& joystick) {
+                switch (joystick.direction.value) {
+                    case LeftJoystick::Direction::HORIZONTAL:
```
TrentHouliston: Do we get separate events for movement in the horizontal and vertical directions?

module/input/PS3Controller/src/PS3Controller.cpp
```diff
@@ -0,0 +1,244 @@
...
+                }
+
+                // Erase axis/button from buffer
+                event_buffer.erase(event_buffer.begin(), std::next(event_buffer.begin(), sizeof(uint8_t)));
+            }
+        });
+
+        // Trigger when the joystick accelerometer has an event to read
+        on<IO>(accelerometer_fd, IO::READ).then([this] {
+            // Accelerometer comes in here .... if you know which axis is which
```
TrentHouliston: Should we read this? I think there is also gyroscope information in these packets too

module/input/PS3Controller/src/PS3Controller.cpp
```diff
@@ -0,0 +1,244 @@
...
+                    type &= ~static_cast<uint8_t>(EventType::INIT);
+                    is_init = true;
+                }
+                switch (static_cast<EventType>(type)) {
+                    case EventType::BUTTON: is_button = true; break;
+                    case EventType::AXIS: is_axis = true; break;
+                    case EventType::INIT: /* This is handled above. */ break;
+                }
+                // Erase type from buffer
+                event_buffer.erase(event_buffer.begin(), std::next(event_buffer.begin(), sizeof(uint8_t)));
```
TrentHouliston: Far better to just loop as you read structs. No need for this buffer

module/input/PS3Controller/src/PS3Controller.cpp
```diff
@@ -0,0 +1,244 @@
...
+                        case Axis::JOYSTICK_RIGHT_HORIZONTAL:
+                            emit(std::make_unique<RightJoystick>(
+                                timestamp, RightJoystick::Direction::HORIZONTAL, normalised, is_init));
+                            break;
+                        case Axis::JOYSTICK_RIGHT_VERTICAL:
+                            emit(std::make_unique<RightJoystick>(
+                                timestamp, RightJoystick::Direction::VERTICAL, normalised, is_init));
+                            break;
+                        case Axis::R2: emit(std::make_unique<R2Trigger>(timestamp, normalised, is_init)); break;
+                        case Axis::ACCELEROMETER_X: /* TODO: Handle this. */ break;
```
TrentHouliston: This comes from the other io right? So unless you’re moving this into its own function to process genetic joystick event structs (not a bad idea) you should just be able to delete these

shared/message/input/PS3Controller.proto
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright 2013 NUbots <nubots@nubots.net>
+ */
+
+syntax = "proto3";
+
+package message.input.PS3Controller;
```
TrentHouliston: Packages don’t have things like this! This should probably either be snake case, or make PS3Controller a message.

shared/message/input/PS3Controller.proto
```diff
@@ -0,0 +1,246 @@
...
+
+message CrossButton {
+    // The timestamp of the event
+    google.protobuf.Timestamp timestamp = 1;
+
+    // The state of the button.
+    bool pressed = 2;
+
+    // Is this an initialisation message?
+    bool init = 3;
```
TrentHouliston: Do we even need to know about the init messages? Are they useful outside of the module that creates them?

shared/message/input/PS3Controller.proto
```diff
@@ -0,0 +1,246 @@
...
+
+syntax = "proto3";
+
+package message.input.PS3Controller;
+
+import "google/protobuf/timestamp.proto";
+
+message CrossButton {
+    // The timestamp of the event
+    google.protobuf.Timestamp timestamp = 1;
```
TrentHouliston: If you’re going to include these timestamps you need to make sure the clocks make sense. Otherwise when future person does now - timestamp it’s going to give nonsense

module/input/PS3Controller/src/PS3Controller.h
```diff
@@ -0,0 +1,32 @@
...
+    private:
+        void connect();
+        void disconnect();
+
+        int controller_fd;
+        int accelerometer_fd;
+        std::string controller_path;
+        std::string accelerometer_path;
+
+        std::vector<uint8_t> event_buffer;
```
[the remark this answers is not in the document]
Bidski: To allow us to receive partial events from the "joystick"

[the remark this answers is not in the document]
Bidski: Probably a rare occurrence, but probably worth allowing for

module/input/PS3Controller/src/PS3Controller.cpp
Bidski: you tell me, all I know is its a timestamp (`event timestamp in milliseconds`)

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
Bidski: yes

module/input/PS3Controller/src/PS3Controller.cpp
Bidski: I havent been able to figure out the data yet, and we dont use it, so I left it as it is for further use

module/input/PS3Controller/src/PS3Controller.cpp
Bidski: I think I put them there to shut the compiler up about not handle enum cases

shared/message/input/PS3Controller.proto
Bidski: what?

shared/message/input/PS3Controller.proto
Bidski: Init messages are the initial state of the system. The controller sends these (one for each button) when it connects I think

shared/message/input/PS3Controller.proto
Bidski: then tell me what the clock is

module/input/PS3Controller/src/PS3Controller.h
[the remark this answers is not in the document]
TrentHouliston: It's impossible to receive partial events from the joystick. It's not like a serial channel it's more of a kernel message thing.

module/input/PS3Controller/src/PS3Controller.cpp
TrentHouliston: Well then it would be wrong to use it directly as a NUClear clock time point as it's misleading. Unless you know for sure that the time is measured in the same clock you'll have to do a time conversion somehow or just send it as a duration of some sort.

module/input/PS3Controller/src/PS3Controller.cpp
TrentHouliston: Ahh, the old "exercise for the reader"

module/input/PS3Controller/src/PS3Controller.cpp
TrentHouliston: `default:` ?

shared/message/input/PS3Controller.proto
TrentHouliston: You've made a namespace `message::input::PS3Controller`, the last component looks like a class but it's a namespace. You should either make it a class (message) or make it conform to the namespace (make it `ps3controller` over `PS3Controller`

shared/message/input/PS3Controller.proto
TrentHouliston: So you get an initial value for each button type, but since the person didn't press the button manually you want to know this was an "initialisation" message. It still might be worthwhile ignoring the init messages as this is basically a "controller connected" message

module/input/PS3Controller/src/PS3Controller.cpp
TrentHouliston: I retract my previous comment. It's probably a good idea in case other buttons are added. But I'd remove the TODO comment since we don't want to todo it here :P

shared/message/input/PS3Controller.proto
TrentHouliston: Who knows! either some c clock, or maybe a clock on the device itself

module/input/PS3Controller/src/PS3Controller.cpp
Bidski: yep, one of those

shared/message/input/PS3Controller.proto
Bidski: it may be important to be able to differentiate between the initial state of the controller and a future state, i think it should remain

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
```diff
@@ -17,123 +17,144 @@
...
+                        log<NUClear::INFO>("Start walking");
+                        moving = true;
                     }
                 }
-                // If it's closed then we should try to reconnect
-                else if (!joystick.valid()) {
-                    joystick.reconnect();
+            });
+
+            on<Trigger<HeadControlEvent>>().then([this](const HeadControlEvent& event) {
```
TrentHouliston: I'm not sure exactly how I feel about naming these messages things like `HeadControlEvent`. On one hand it means if you change what event triggers this it's one place to change. On the other hand, looking at this code I have no idea what button to push to make it happen without scrolling to the top of the file.

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
```diff
@@ -17,123 +17,144 @@
...
+                        KickScriptCommand(Eigen::Vector3d(1.0, 0.0, 0.0), LimbID::RIGHT_LEG)));
+                }
+            });
+
+            on<Trigger<WalkVelocityEvent>>().then([this](const WalkVelocityEvent& event) {
+                switch (event.direction.value) {
+                    case WalkVelocityEvent::Direction::HORIZONTAL:
+                        // y is left relative to robot
+                        // strafe[1] = -joystick.value;
+                        walkCommand.z() = -event.value;
```
TrentHouliston: Is there supposed to be some limit or scaling on this?

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
```diff
@@ -17,123 +17,144 @@
...
+                        KickScriptCommand{Eigen::Vector3d(1.0, 0.0, 0.0), LimbID::LEFT_LEG}));
+                }
+            });
+
+            on<Trigger<RightKickEvent>>().then([this](const RightKickEvent& event) {
+                if (event.pressed) {
+                    log<NUClear::INFO>("Requesting Right Front Kick");
+                    // vector pointing forward relative to robot
+                    emit(std::make_unique<KickScriptCommand>(
+                        KickScriptCommand(Eigen::Vector3d(1.0, 0.0, 0.0), LimbID::RIGHT_LEG)));
```
TrentHouliston: Lol, this one is using a constructor and the other is using uniform initialisation (`{}` on 127 and `()` on 136)

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
```diff
@@ -17,123 +17,144 @@
...
+                    case WalkVelocityEvent::Direction::VERTICAL:
+                        // x is forward relative to robot
+                        walkCommand.x() = -event.value;
+                        break;
+                }
+            });
+
+            on<Trigger<HeadDirectionEvent>>().then([this](const HeadDirectionEvent& event) {
+                switch (event.direction.value) {
+                    case HeadDirectionEvent::Direction::HORIZONTAL: headYaw = -event.value; break;
```
TrentHouliston: Is there supposed to be some scaling etc on these values? Also instead of having a headYaw and headPitch value you could just make a HeadCommand object and throw that around

module/input/PS3Controller/src/PS3Controller.cpp
```diff
@@ -0,0 +1,239 @@
...
+                connect();
+            }
+        });
+
+        on<Shutdown>().then([this] { disconnect(); });
+
+        // Details on Linux Joystick API
+        // https://www.kernel.org/doc/Documentation/input/joystick-api.txt
+        // Trigger when the joystick has an data to read
+        on<IO>(controller_fd, IO::READ).then([this] {
```
TrentHouliston: Currently, if the joystick fails to connect in the `on<Configuration>` or reconnects in the on every this `on<IO>` won't pick it up. You should move this on statement into the `connect` function and .unbind() the previous reaction on reconnection. Otherwise it won't work.

module/input/PS3Controller/src/PS3Controller.cpp
```diff
@@ -0,0 +1,239 @@
...
+            // Try to read a JoystickEvent worth of data
+            JoystickEvent event;
+            while (read(controller_fd, &event, sizeof(JoystickEvent)) == sizeof(JoystickEvent)) {
```
TrentHouliston: `while (read(controller_fd, &event, sizeof(JoystickEvent)) == sizeof(JoystickEvent))`
`for (JoystickEvent event; read(controller_fd, &event, sizeof(JoystickEvent)) == sizeof(JoystickEvent);)`
¯\\_(ツ)_/¯

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
Bidski: There's just no pleasing you, is there

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
Bidski: scaling happens when we send the message

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
Bidski: scaling happens when we send the message

module/input/PS4Controller/src/PS4Controller.cpp
```diff
@@ -0,0 +1,294 @@
...
+                        log<NUClear::WARN>(fmt::format("Timestamp: {}", event.timestamp));
+                        log<NUClear::WARN>(fmt::format("Value....: {}", event.value));
+                        log<NUClear::WARN>(
+                            fmt::format("Type.....: {}",
+                                        is_init ? event.type | static_cast<uint8_t>(EventType::INIT) : event.type));
+                        log<NUClear::WARN>(fmt::format("Number...: {}", static_cast<int>(event.number)));
+                    }
+                }
+            });
+            controller_err_reaction = on<IO>(controller_fd, IO::ERROR).then([this] { disconnect(); });
```
TrentHouliston: You can combine this reaction with the other one by just having
```cpp
on<IO>(controller, IO::READ | IO::ERROR).then([this] (const IO::Event& event) {
    if (event.events & IO::ERROR) {
        disconnect();
    }
});
```

KipHamiltons:
What would we have to do to make ps3 walk a thing again? It's my understanding that it's currently broken because ps3 controller's are painful to connect?

Bidski:
IIRC this PR was pretty much ready to go, but there was talk about getting a different controller so we held off on merging this until we got said new controller and then that never happened. At this point, if we want PS3Walk to be a thing, we should source a new controller and update this branch to use it

tayloryoung6396:
@Bidski It's worth merging this, the PR does what it says with or without us currently owning a PS3 controller. A future issue should be created to source and update the controller

Bidski:
> @Bidski It's worth merging this, the PR does what it says with or without us currently owning a PS3 controller. A future issue should be created to source and update the controller

Someone should probably test this then and make sure its still operational (assuming testing is actually a possibility). Also looks like some conflicts need resolving

ysims:
Closing thiis as superceded by #1236.

PR #194, opened 2019-04-23
Updates travis to use xenial

Bidski:


TrentHouliston:
Can we instead upgrade travis to xenial and use the regular deb etc?

PR #196, opened 2019-06-13
Applies corrections for CM740

tayloryoung6396:


module/platform/darwin/HardwareIO/src/HardwareIO.cpp
```diff
@@ -312,6 +312,7 @@ namespace platform {
 
                     // Read our data
                     Darwin::BulkReadResults data = darwin.bulkRead();
+                    std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
Bidski: this should be removed I think

tayloryoung6396: no

MattAmos: Tested and still works

PR #203, opened 2019-06-16
[TESTING] Performs Low-Pass Filtering on servo positions during walk cycle

MattAmos:


module/motion/OldWalkEngine/data/config/igus1/OldWalkEngine.yaml
```diff
@@ -70,6 +70,11 @@ walk_cycle:
   single_support_phase:
     start: 0.3 #0.3
     end: 0.6 #0.6
+  # Second Order Low Pass Filter difference equation:
```
Bidski: If this is the same for every robot then only put it in the top-level config file. Keep robot specific config files for instances where the parameters deviate per robot

module/motion/OldWalkEngine/src/OldWalkEngine.cpp
```diff
@@ -168,8 +168,13 @@ namespace motion {
         //     }
         // }));
 
-        on<Startup, Trigger<KinematicsModel>>().then("Update Kin Model",
-                                                     [this](const KinematicsModel& model) { kinematicsModel = model; });
+        on<Startup, Trigger<KinematicsModel>>().then("Update Kin Model", [this](const KinematicsModel& model) {
+            kinematicsModel = model;
+            // Initialise previous positions for servo (20) LPFs
+            for (uint i = 0; i < 20; ++i) {
```
Bidski: `previousPositions.fill(std::vector<float>(2, 0.0f));`

module/platform/darwin/HardwareIO/src/HardwareIO.h
```diff
@@ -22,6 +22,8 @@
 
 #include <nuclear>
 
+#include <fstream>
```
Bidski: Is this needed?

module/motion/OldWalkEngine/src/OldWalkEngine.h
```diff
@@ -23,6 +23,8 @@
 #include <armadillo>
 #include <nuclear>
 
+#include <fstream>
```
Bidski: Is this needed?

module/platform/darwin/HardwareIO/data/config/HardwareIO.yaml
```diff
@@ -1,7 +1,7 @@
 battery:
-  charged_voltage: 12.6
```
Bidski: Why are these changes in this PR?

module/platform/darwin/HardwareIO/data/config/igus1/HardwareIO.yaml
```diff
@@ -7,6 +7,11 @@ bus_reset_wait_time_us: 0
 byte_wait: 1200
 packet_wait: 100000
 
+# Second Order Low Pass Filter difference equation:
```
Bidski: Are these meant to be here?

module/platform/darwin/HardwareIO/data/config/igus2/HardwareIO.yaml
```diff
@@ -7,6 +7,11 @@ bus_reset_wait_time_us: 0
 byte_wait: 1200
 packet_wait: 100000
 
+# Second Order Low Pass Filter difference equation:
```
Bidski: Are these meant to be here?

module/motion/OldWalkEngine/data/config/igus1/OldWalkEngine.yaml
MattAmos: The config params are intended to be configured for each robot independently. Not sure how uniquely the robots will have to be tuned w.r.t the LPF, but putting them in the config for future use if necessary. Can remove them from robot-specific config files if you think it's better convention.

module/platform/darwin/HardwareIO/data/config/HardwareIO.yaml
MattAmos: The Darwin voltages were overwritten when copy & pasting, will revert

module/motion/OldWalkEngine/data/config/igus1/OldWalkEngine.yaml
Bidski: The correct convention is to use the per robot configs only when there are per robot changes (we have been doing this really poorly up until now). But if you think there is a really high likelihood that we will have per robot filtering parameters then only put them in the per robot files and not in the top-level file (I feel like this could be a bad idea though).

MattAmos: Okay, will remove

ysims:
Closing since it's unlikely to have any more work done on it

PR #225, opened 2019-07-02
Corrects missing line in FootDownNetwork

MattAmos:
Enhances the not-quite-perfect FootDownNetwork.yaml


TrentHouliston:
If only someone had kept that valuable number

PR #230, opened 2019-07-02
Image compressor centre variable missing

tayloryoung6396:


module/output/ImageCompressor/src/ImageCompressor.cpp
```diff
@@ -44,6 +44,7 @@ namespace output {
         msg->lens.projection   = int(image.lens.projection);
         msg->lens.focal_length = image.lens.focal_length;
         msg->lens.fov          = image.lens.fov;
+        msg->lens.centre       = image.lens.centre;
```
Bidski: *sighs*

PR #238, opened 2019-07-05
Transpose the VM rays so they can be drawn in NUsight fast

TrentHouliston:
NUsight needs the rays as columns so it can draw them, so we need to transpose them over here


roles/robocup.role
```diff
@@ -3,51 +3,50 @@ NUCLEAR_ROLE(
     extension::FileWatcher
     support::SignalCatcher
     support::logging::ConsoleLogHandler
+
```
Bidski: Why is this file changed?

TrentHouliston: Shouldn't have been

PR #239, opened 2019-07-06
Prevent measuring goal posts with only one edge detected

Bidski:


module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -305,7 +307,9 @@ namespace vision {
                         log<NUClear::DEBUG>(fmt::format("Found {} goal posts", goals->goals.size()));
                     }
 
-                    emit(std::move(goals));
+                    if (goals->goals.size() > 0) {
```
TrentHouliston: Should still emit empty goals, that's a measurement

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -148,80 +148,82 @@ namespace vision {
...
-                            utility::math::geometry::sort_by_theta(cluster.begin(), right, rays, world_offset);
-                            utility::math::geometry::sort_by_theta(right, other, rays, world_offset);
-                            left_side =
-                                rays.col(*std::next(cluster.begin(), std::distance(cluster.begin(), right) / 2));
-                            right_side = rays.col(*std::next(right, std::distance(right, other) / 2));
-                            for (auto it = cluster.begin(); it != cluster.end(); it = std::next(it)) {
-                                const Eigen::Vector3f& p0(rays.col(*it));
-                                if (p0.z() < bottom_point.z()) {
-                                    bottom_point = p0;
+                        if ((std::distance(cluster.begin(), right) != 0) && (std::distance(right, other) != 0)) {
```
TrentHouliston: Should you also check that there is a base point to use?

TrentHouliston: Checking if other is not empty may let us know if we can't actually see the bottom of the goals (we're looking in the sky and the goals cut off before the base of the image)

Bidski: This is an issue for a future PR. Not being able to see the base implies that we also have not detected a green horizon, in which case, how do we know that the goal post belongs to our field? `other` also contains points that may be at top of the posts and so a non-zero quantity in `other` does not necessarily mean that we can see the base either.

PR #241, opened 2019-07-07
Adds green horizon to robocup role

ysims:


TrentHouliston:
This was already finished somehow

PR #242, opened 2019-07-07
Send `is_left` with Image and CompressedImage

JosephusPaye:
So we can order the cameras correctly in NUsight.


TrentHouliston:
We want to limit the amount of information in an image to a minimum.

PR #243, opened 2019-07-09
The great purge of 2019

TrentHouliston:
NUbots has many many modules that are no longer used and often don't even work properly anymore.
This PR removes most of the codebase that has fallen into disuse and will likely never be revived.
The result of this should be a simpler to maintain codebase going forward.
image


tools/extract_images.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
```
Bidski: Doesn't belong here

TrentHouliston: oops... this is what you get from git commit -am

BrendanAnnable:
This seems to have broken a lot of roles:

```
error while loading shared libraries: libsupportNUbugger.so: cannot open shared object file: No such file or directory
```

The commit before this one works.

TrentHouliston:
Are the roles that are broken ones that have been deleted but you previously built?

BrendanAnnable:
Oh you're totally right — I downloaded the VM for the first time in forever and assumed the build directory was fresh — nope years old.

...in which case I need some roles back! (particularly nusighttest  😛)

PR #245, opened 2019-07-10
Fix execute permissions on the b script

TrentHouliston:
B script lost its execute permissions in the purge (#243), restore them


BrendanAnnable:
I was wondering why I needed to `chmod +x b`...

PR #246, opened 2019-07-10
Swap to using libuv for file watching instead of libfswatch

TrentHouliston:


TrentHouliston:
Oops, duplicated this with #270, and that one works

PR #247, opened 2019-07-10
Convert to using Eigen for the particle and UKF filters

TrentHouliston:
Swapping to eigen all at once is crazy, this converts the filters and those that use them to use Eigen instead of armadillo


TrentHouliston:
Closed in favour of #267

PR #248, opened 2019-07-10
Unpurge the nusighttest role

BrendanAnnable:
I want my debug waves!


TrentHouliston:
You could just move the relevant parts to within NUsight proper, and just hide it behind one of the reaction configs like the other types

BrendanAnnable:
> > You could just move the relevant parts to within NUsight proper, and just hide it behind one of the reaction configs like the other types

I think that would be mixing test/harness code with real code — ideally there would be very a clean separation. I can imagine when working on the other modules (e.g. localisation), I'll want a very specific harness to run a very specific set of data 
with a very limited set of modules — and I probably don't want to pollute the global roles list with my internal module harnesses — this feels like something we need to add 1st-class support for.

PR #250, opened 2019-07-25
Fix Aravis Buffer on Invalid Image from Camera.

wongosaurus:
Please check if this breaks your robot. We had issues where cameras stopped sending images after some timeouts, after close investigation, the camera was running out of buffers to insert images.

### DIFF
--- a/module/input/Camera/src/AravisCamera.cpp
+++ b/module/input/Camera/src/AravisCamera.cpp
@@ -176,8 +176,6 @@ namespace input {
         else {
             // In the case where it fails, reinsert the Buffer back into the stream, so we don't run out.
             if (buffer != NULL) {
-                // std::cout << "Buffer Failed: " << context->cameraID << " " << arv_buffer_get_status(buffer) <<
-                // std::endl;
                 arv_stream_push_buffer(stream, buffer);
             }
         }
--- a/module/input/Camera/src/AravisCamera.cpp
+++ b/module/input/Camera/src/AravisCamera.cpp
@@ -152,30 +152,30 @@ namespace input {
         ArvBuffer* buffer;
         buffer = arv_stream_try_pop_buffer(stream);
 
-        if ((buffer != NULL) && (arv_buffer_get_status(buffer) == ARV_BUFFER_STATUS_SUCCESS)) {
-            int width, height;
-            size_t buffSize;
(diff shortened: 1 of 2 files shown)


module/input/Camera/src/AravisCamera.cpp
```diff
@@ -173,6 +173,14 @@ namespace input {
             context->reactor.emit<Scope::DIRECT>(msg);
             arv_stream_push_buffer(stream, buffer);
         }
+        else {
+            // In the case where it fails, reinsert the Buffer back into the stream, so we don't run out.
+            if (buffer != NULL) {
+                // std::cout << "Buffer Failed: " << context->cameraID << " " << arv_buffer_get_status(buffer) <<
```
TrentHouliston: Don't commit commented code!

wongosaurus: Done

module/input/Camera/src/AravisCamera.cpp
```diff
@@ -173,6 +173,12 @@ namespace input {
             context->reactor.emit<Scope::DIRECT>(msg);
             arv_stream_push_buffer(stream, buffer);
         }
+        else {
```
Bidski: I think I would prefer to this to be
```
if (buffer != NULL) {
    if (arv_buffer_get_status(buffer) == ARV_BUFFER_STATUS_SUCCESS) {
        // send message
        arv_stream_push_buffer(....); 
    }
    else {
        arv_stream_push_buffer(....);
    }
}
```

Whats your opinion @TrentHouliston ?

TrentHouliston: I agree it should be changed to that, but I also think the compiler would probably optimise it to that anyway :P

Bidski: readability concerns are more important than potential compiler optimisations?

wongosaurus: Changes made, it matches the same format as the examples given in the documentation:
https://aravisproject.github.io/docs/aravis-0.4/ArvCamera.html

TrentHouliston: > readability concerns are more important than potential compiler optimisations?

I said I agree!

PR #253, opened 2019-07-26
Combines HardwareIO and HardwareSimulator

MattAmos:
Allows for periodic polling of the CM730 to determine whether to simulate or use real hardware.


KipHamiltons:
@Bidski (and even @MattAmos?) are we keen to update this PR or discard it?

Bidski:
Im guessing HardwareSimulator probably needs some updating, and webots may be a viable alternative

MattAmos:
Yeah it sounds like Webots would be a better alternative now. 👍 To close the PR

PR #254, opened 2019-07-30
Add initial support for BuildKite

JosephusPaye:
This PR adds initial BuildKite support for building NUbots code. https://buildkite.com/nubots/nubots

I've added a `Dockerfile` defining an image based on Ubuntu 16.04 with the NUbots build toolchain installed. This built image is stored on DockerHub at https://hub.docker.com/r/nubots/buildkite

Also setup a build agent on ~one~ six of the iMacs in the lab. When an agent receives a new build from BuildKite, it checks out the code, runs a new container based on the image, mounts the code in the container, and runs the build.

~Currently it's just calling the build scripts in `.travis/`, but we can rework this to be independent with no references to Travis when we remove Travis.~ Uses a combination of scripts in `.buildkite/scripts` and the Dockerfile.


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,18 @@
...
+  - label: "Build for NUC"
+    command: "/code/NUbots/.travis/build_code.sh"
+    plugins:
+      - docker#v3.3.0:
+          image: "nubots/buildkite:0.2"
+          workdir: "/code/NUbots"
+          propagate-uid-gid: true
+          environment:
+            - "PLATFORM=nuc7i7bnh"
+            - "CCACHE_DIR=/code"
```
TrentHouliston: Does this cache dir persist through different builds? Otherwise, ccache is useless

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,18 @@
+steps:
+  - label: "Validate Clang Format"
+    command: "/code/NUbots/.travis/validate_format.sh"
```
TrentHouliston: If we are going to transition we should probably swap over from travis to buildkite entirely

Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM ubuntu:16.04
```
TrentHouliston: If docker works well enough it's possible that we might be able to look into it again for use instead of the NUbots vm in vagrant. Docker has a few advantages over it.

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN rm -rf /NUbots-build \
+    && apt-get autoremove --purge
+
+# Create the directory that will be used for checkouts and builds
+RUN mkdir -p /code/NUbots
+
+# Create the nubots user and group with id 1000. This will be used to map to
+# the user with id 1000 on the host (typically the first non-root user)
+# Used to fix permission issues
+RUN groupadd -g 1000 nubots \
```
TrentHouliston: This seems suspicious... Do other examples do this?

MattAmos:
Also worth mentioning that initial builds using buildkite reduced the build time down to around 6 minutes rather than 30-40 👍

TrentHouliston:
> Also worth mentioning that initial builds using buildkite reduced the build time down to around 6 minutes rather than 30-40 👍

And that's without any of the potential distcc compilations that are possible. Although those will be much tricker to setup with docker

.buildkite/pipeline.yml
JosephusPaye: `ccache` is still useful for local builds right? E.g. in the case where your build fails halfway through, you fix an error, run the build again and it doesn't have to start from scratch.

If it isn't ccache that's providing this then yeah, we should probably remove it.

.buildkite/pipeline.yml
JosephusPaye: I feel like that decision is above my paygrade 😜. But yeah we can easily switch to removing all references to Travis if we decide that.

Dockerfile
JosephusPaye: Yeah @Bidski has mentioned this. After this is merged I will be running some tests to see if it works well on Windows for general development. That will probably be a different PR though.

Dockerfile
JosephusPaye: Yeah, from what I could find, this is normal [[1]](https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b) [[2]](https://jtreminio.com/blog/running-docker-containers-as-current-host-user/#ok-so-what-actually-works) [[3]](https://www.reddit.com/r/docker/comments/cihjtf/how_do_i_make_sure_that_files_created_by/). The problem it's trying to solve:

- The docker daemon that runs builds, containers, etc runs as root (there's no way around this, as far as I can tell)
- Root in the container is the same as root on the host -- with reduced privileges -- (since the kernel is shared, user and group ids are also shared, even if the associated user or group names are different between host and container)
- As such files that are created by processes running in the container by default will be owned by root, leading to permission issues when a process in the host wants to change/remove those files. As can be seen with [this failed build](https://buildkite.com/nubots/nubots/builds/5#b5178e6e-d916-4c45-be6b-0458051928d7/3-9), this lead to "permission denied" errors when cleaning out old build files for a new build.
- The [BuildKite recommendation](https://github.com/buildkite-plugins/docker-buildkite-plugin#propagate-uid-gid-optional-boolean) for this is to propagate the user/group id to the container, so files created in the container are owned by the corresponding host user. For this to work properly there needs to be a known user in the container to map to from the host.

There are [other ways](https://buildkite.com/docs/agent/v3/docker#permissions-errors-when-using-docker) to solve this problem, but this is one of the recommended  options.

.buildkite/pipeline.yml
TrentHouliston: The way ccache works is whenever you compile a file, it takes a hash of the content that was compiled. Then if you ever compile the same file again it just gives you the compiled object. But to do this it needs a persistent storage location so that it can store this cache. For what we're doing we should probably remove it.

.buildkite/pipeline.yml
TrentHouliston: Well, if buildkite works and is 6 minutes to travis' 40 minutes we are going to use buildkite. There's no point waiting around for 40 minutes when we already have a CI done. We're going to use either buildkite or travis, but never both.

Dockerfile
TrentHouliston: Agreed that it should be a separate PR. However, if you do that it also means that we will replace puppet with docker.

Bidski:
```
docker run -it --rm --init --volume /var/lib/buildkite-agent/builds/imac-5-1/nubots/nubots:/code/NUbots --workdir /code/NUbots -u 999:999 --env BUILDKITE_JOB_ID --env BUILDKITE_BUILD_ID --env BUILDKITE_AGENT_ACCESS_TOKEN --volume /usr/bin/buildkite-agent:/usr/bin/buildkite-agent --label com.buildkite.job-id=e9e06231-3d01-4184-aee2-a2b25b10e6ee nubots/buildkite:0.2 /bin/sh -e -c /code/NUbots/.travis/validate_format.sh
--
  | docker: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.40/containers/create: dial unix /var/run/docker.sock: connect: permission denied.
``
Looks like we have some permissions errors on the macs

TrentHouliston:
The macs we configured first were working, maybe it's one of the ones that were configured later

.buildkite/pipeline.yml
JosephusPaye: Removed `ccache`.

.buildkite/pipeline.yml
JosephusPaye: Removed references to Travis except puppet stuff which is being sorted in #255.

Dockerfile
JosephusPaye: In progress in #255.

JosephusPaye:
Added requested changes, fixed permission errors with agents. This should be ready to merge now. Further docker changes will land in #255.

PR #255, opened 2019-08-04
Add docker as our build system and replace vagrant and puppet

TrentHouliston:
This PR removes vagrant and puppet as the system that is used for building the codebase.
As this is a significant change to how the NUbots codebase is built and used there should be feedback from as many team members as possible.

The readme describes the new method of building, running, and installing to the robots.
https://github.com/NUbots/NUbots/blob/houliston/docker/README.md
If this readme is missing anything that is needed add it.

Things still left to to
- [x] Ensure that the optimisation flags for a platform are passed through (toolchain files should handle this correctly now)
- [x] Ensure that code built and installed using this method works as expected 9(vision sending data to nusight has been tested)
- [x] Test more robot systems using code built using this method (vision sending data to nusight has been tested, `keyboardwalk` or similar should also be tested to ensure servo communication happens as expected)
- [x] Test gazebo communication
- [x] Run ldconfig after installing toolchain to the robot
- [x] Build docker image on a mac (and/or any CPU that is older than a 7th gen i7) to ensure all cross-compilation issues are resolved (if there are any)
- [x] Add a method to pull config/scripts from the robot and map back to source folder (for a later PR)
- [x] Only install robot-specific config/scripts for the selected robot and not all robots (for a later PR [install.py)](https://gist.github.com/Bidski/a27a6f1a2823f09e8ced9c3d96a0f2f0)
- [ ] Cache docker images (needs to happen after merge to master)


tools/format.py
```diff
@@ -12,10 +11,18 @@ def register(command):
...
-    for d in [os.path.join(b.project_dir, p) for p in ["module", "shared", "tests"]]:
-        for dirpath, dnames, fnames in os.walk(d):
-            for f in fnames:
-                if f.endswith((".h", ".c", ".cc", ".cxx", ".cpp", ".hpp", ".ipp", ".proto")):
-                    print("Formatting", f)
-                    call(["clang-format-6.0", "-i", "-style=file", os.path.join(dirpath, f)])
+    # Change into the project directory
+    os.chdir(b.project_dir)
+
+    # Use git to get all of the files that are committed to the repository
```
Bidski: Check for binaries before trying to format the files, or run in docker

tools/run.py
```diff
@@ -0,0 +1,27 @@
...
+def run(args, **kwargs):
+
+    # Change into the build directory
+    os.chdir(os.path.join(b.project_dir, "..", "build"))
+
+    # Add 'bin/` to the command (first argument)
+    args[0] = os.path.join("bin", args[0])
+
+    # Run the command
+    exit(pty.spawn(args) << 8)
```
Bidski: Wrong shifties

tools/format.py
TrentHouliston: Running in docker!

tools/run.py
TrentHouliston: `>>`

TrentHouliston:
302 Commits! This was getting rather large

Bidski:
> 302 Commits! This was getting rather large

302 commits of you writing epically broken code!!! Get good son!

PR #258, opened 2019-08-19
Improve the nbs decoder so it does not need the VM to run

TrentHouliston:
Improve the nbs decoder so that it can run without the aid of the VM. This opens up more opportunities to use it for manipulating nbs files with less effort.
Also add on code to synchronize on packet headers for malformed or partial nbs streams.

### DIFF
--- a/nuclear/message/Neutron.cmake
+++ b/nuclear/message/Neutron.cmake
@@ -164,10 +164,8 @@ FOREACH(proto ${protobufs})
     ADD_CUSTOM_COMMAND(
         OUTPUT "${outputpath}/${file_we}.pb.cc"
                "${outputpath}/${file_we}.pb.h"
-               "${outputpath}/${file_we}_pb2.py"
         COMMAND ${PROTOBUF_PROTOC_EXECUTABLE}
         ARGS --cpp_out=lite:${message_binary_include_dir}
-             --python_out=${message_binary_include_dir}
              -I${message_binary_include_dir}
              -I${CMAKE_CURRENT_SOURCE_DIR}/proto
              "${outputpath}/${file_we}.proto"
--- a/tools/nbs_decoder/__init__.py
+++ b/tools/nbs_decoder/__init__.py
@@ -105,10 +105,10 @@ def decode(path):
                 c = f.read(1)
                 # No bytes to read we are EOF
                 if len(c) == 0:
-                    sync_state = state.EOF
+                    sync_state = state.eof
 
                 # We can jump straight to lock1 from any state except the eof state
-                if sync_state != state.EOF and c == b"\xE2":
+                if sync_state != state.eof and c == b"\xE2":
                     sync_state = state.header_lock_1
                 elif sync_state == state.header_lock_1 and c == b"\x98":
                     sync_state = state.header_lock_2
--- a/tools/nbs_decoder/__init__.py
+++ b/tools/nbs_decoder/__init__.py
@@ -7,6 +7,7 @@ import sys
 import os
 import pkgutil
 import glob
+import enum
 import b
 import google.protobuf.message
 from google.protobuf.json_format import MessageToJson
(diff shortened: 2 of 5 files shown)


tools/nbs_decoder/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+                        packet = (decoders[type_hash][0], timestamp, decoders[type_hash][1].FromString(payload[16:]))
+                        yield packet
+                    except:
+                        pass
+            else:
+                c = f.read(1)
+                # No bytes to read we are EOF
+                if len(c) == 0:
+                    sync_state = state.eof
+                elif c == b"\xE2":
```
Bidski: State machine does not require a header to exist in its entirety

tools/nbs_decoder/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+
+proto_files = [
+    *glob.glob(os.path.join(nuclear_proto_dir, "**", "*.proto"), recursive=True),
+    *glob.glob(os.path.join(user_proto_dir, "**", "*.proto"), recursive=True),
+]
+
+# Build the protocol buffers
+if (
+    os.system(
+        "protoc --python_out={protobuf_path} -I{nuclear_dir} -I{user_dir} {proto_files}".format(
```
Bidski: We should do a version check on protoc to ensure it meets a minimum version requirement (if we have one)

nuclear/message/Neutron.cmake
```diff
@@ -164,10 +164,8 @@ FOREACH(proto ${protobufs})
     ADD_CUSTOM_COMMAND(
         OUTPUT "${outputpath}/${file_we}.pb.cc"
                "${outputpath}/${file_we}.pb.h"
-               "${outputpath}/${file_we}_pb2.py"
         COMMAND ${PROTOBUF_PROTOC_EXECUTABLE}
         ARGS --cpp_out=lite:${message_binary_include_dir}
-             --python_out=${message_binary_include_dir}
```
Bidski: Why is removed? This will be needed for python modules (eventually)

Bidski: Nevermind this one

tools/nbs_decoder/__init__.py
TrentHouliston: Your critique here is invalid as it was never a state machine in the first place.
On an unrelated note, I've made it a state machine now so it actually works :P

tools/nbs_decoder/__init__.py
TrentHouliston: That's what the if is for here. If protoc can't parse it then it'll return an error code which means we can't use it. The shell is not redirected so you'll see whatever protoc complained about in your terminal.

Bidski: Its not a question of whether or not protoc will error on a "bad" file, but whether the error protoc will give will be ambiguous in the context

tools/nbs_decoder/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+                        packet = (decoders[type_hash][0], timestamp, decoders[type_hash][1].FromString(payload[16:]))
+                        yield packet
+                    except:
+                        pass
+            else:
+                c = f.read(1)
+                # No bytes to read we are EOF
+                if len(c) == 0:
+                    sync_state = state.eof
+                elif c == b"\xE2":  # No need to check the state here as we can step to lock2 from any state
```
Bidski: step to lock1?

tools/nbs_decoder/__init__.py
TrentHouliston: I feel like it's just an extra check that's not really needed for a simple tool

TrentHouliston:
> Add an extra format file to the roles folder so we can specialise the formatting there

I think this was for #259

PR #259, opened 2019-08-19
Add a cmake format file, format cmake files and get buildkite to test they stay formatted

TrentHouliston:
cmake-format is able to auto format our cmake files.
This PR adds the style guide, formats the files and ensures that they stay formatted in the future via buildkite


TrentHouliston:
The docker image needs to have `cmake-format` installed via pip and gnu `parallel` installed via apt before buildkite will be happy with this

.buildkite/scripts/validate_clang_format.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
Bidski: It is probably prudent to maintain posix shell compliance so that we can be as flexible as possible. 

This means `export -f` wont exist so we need to rework this script (and the other one).

Is it worth having two separate scripts/build processes to do the validation? Why not just merge all validation into a single script?

PR #262, opened 2019-08-22
Amalgamate the nbs tools into a single subcommand and add a stats tool

TrentHouliston:
As more and more tools come about to work with nbs files it clutters up the main namespace.
This PR moves them all into a `./b nbs {}` subcommand.
It also adds a new command `./b nbs stats {}` which will print some high-level statistics about the contents and data rates of nbs files`

It also has some minor fixes for the `./b` tooling in general


tools/footdown.py
```diff
@@ -62,7 +62,7 @@ def displacement(fk):
 def dataset(path, left_state, right_state, servos, fields, lr_duplicate, foot_delta, accelerometer, gryoscope):
     xs = []
     ys = []
-    for type_name, timestamp, msg in tqdm(nbs_decoder.decode(path), dynamic_ncols=True, unit="packet"):
+    for type_name, timestamp, msg, raw in tqdm(nbs_decoder.decode(path), dynamic_ncols=True, unit="packet"):
```
Bidski: what is `raw`? is it used anywhere?

tools/nbs_tools/extract_images.py
```diff
@@ -22,7 +20,7 @@ def run(in_file, out_path, **kwargs):
     output_path = os.path.join(out_path, os.path.basename(os.path.splitext(in_file)[0]))
     os.makedirs(output_path, exist_ok=True)
 
-    for type_name, timestamp, msg in nbs_decoder.decode(in_file):
+    for type_name, timestamp, msg, raw in decoder.decode(in_file):
```
Bidski: replace `raw` with `_`?

tools/nbs_tools/json.py
```diff
@@ -1,22 +1,20 @@
...
-    # Drone arguments
-    command.add_argument("path", metavar="path", help="The file to decode into a series of json objects")
+    # Command arguments
+    command.add_argument("path", metavar="path", help="The nbs file to convert to json")
 
 
 def run(path, **kwargs):
 
-    for type_name, timestamp, msg in nbs_decoder.decode(path):
+    for type_name, timestamp, msg, raw in decoder.decode(path):
```
Bidski: unused `raw`

tools/footdown.py
TrentHouliston: Raw is the raw bytes from the file. It's used in stats to get a size in bytes, and I figured it could be useful one day if you're ever just "passing through" protobufs and reencoding them. The other option would be to return the size instead of the raw bytes. and if you want to renencode you need to actually re serialise the protobufs

PR #265, opened 2019-09-23
Create an improved Aravis camera module

TrentHouliston:
Creates an improved Aravis camera module that has better time synchronisation, and the ability to access all settings exposed by the genicam interface.

Things still to do:

- [x] Improve the settings interface as currently there is lots of code duplication
- [x] Synchronise the sensor packets from SensorFilter to the Hcw matrix for the correct time. This will mean storing a sequence of past sensor packets and then finding the lower bound for the best fit for the time.
- [x] Store the offset from the head kinematics to the camera in each configuration file rather than an "is_left" option


module/input/Camera/src/settings.h
```diff
@@ -0,0 +1,187 @@
...
+    template <>
+    struct SettingsFunctions<ArvGcInteger> {
+        static int64_t read(ArvGcInteger* setting, GError** error) {
+            return arv_gc_integer_get_value(setting, error);
+        }
+        static void write(ArvGcInteger* setting, const int64_t& v, GError** error) {
+            arv_gc_integer_set_value(setting, v, error);
+        }
+        static bool valid(ArvGcInteger* setting, const int64_t& v) {
+            GError* error = nullptr;
```
Bidski: `std::shared_ptr` with `g_error_free` as deleter?

module/input/Camera/src/settings.h
```diff
@@ -0,0 +1,187 @@
...
+            }
+            std::vector<std::string> options;
+            for (unsigned int i = 0; i < len; ++i) {
+                options.emplace_back(values[i]);
+            }
+            g_free(values);
+
+            if (!options.empty() && std::find(options.begin(), options.end(), v) == options.end()) {
+                std::stringstream s;
+                s << "Value must be one of [";
```
Bidski: `{ ... }` is really the better option here

module/input/Camera/src/time_sync.cpp
```diff
@@ -0,0 +1,106 @@
...
+            int64_t local;
+            int64_t cam;
+            bool valid;
+        };
+
+        // This is what we return if we can't work out the timestamp offset now
+        CameraContext::TimeCorrection output;
+        output.live   = true;
+        output.offset = 0;
+        output.kf.p   = 10e6 * 10e6;    // Initial stddev of about 10 milliseconds
```
Bidski: This seems wrong

module/input/Camera/src/Camera.cpp
```diff
@@ -1,129 +1,364 @@
...
-        on<Configuration>("Camera.yaml").then("Camera Module Configuration", [this](const Configuration& config) {
-            dumpImages = config["dump_images"].as<bool>();
-        });
+    using utility::support::Expression;
+    using utility::vision::fourcc;
 
-        on<Configuration>("Cameras").then("Camera driver loader", [this](const Configuration& config) {
-            // Monitor camera config directory for files.
-            // Each file MUST define a "driver", we use this driver to load the appropriate handler for the camera.
+    /// The amount of time to observe after recalibrating to work out how long image transfer takes
```
Bidski: units???

module/input/Camera/src/Camera.cpp
```diff
@@ -1,129 +1,364 @@
...
+            arv_stream_set_emit_signals(stream.get(), true);
+        });
+
+        on<Shutdown>().then([this] {
+            for (auto& camera : cameras) {
+                // Stop the video stream.
+                arv_camera_stop_acquisition(camera.second.camera.get());
+                arv_stream_set_emit_signals(camera.second.stream.get(), false);
             }
+            arv_debug_shutdown();
```
Bidski: is this needed?

Bidski: maybe wrap this in a config param (along with the thing that enables debugging) so we can enable it if needed

module/input/Camera/src/Camera.cpp
```diff
@@ -1,129 +1,364 @@
...
-    }
+                auto msg        = std::make_unique<Image>();
+                msg->format     = context->fourcc;
+                msg->dimensions = Eigen::Matrix<unsigned int, 2, 1>(width, height);
+                // TODO(trent) use an std::vector here to avoid the copy
+                msg->data.insert(msg->data.end(), buff, buff + buffSize);
+                msg->camera_id = context->camera_id;
+                msg->name      = context->name;
+                msg->timestamp = NUClear::clock::time_point(nanoseconds(ts));
+                // TODO attach Hcw once we have access to sensors
```
Bidski: This needs to exist for this PR to be able to merge

module/input/Camera/src/Camera.cpp
```diff
@@ -1,129 +1,364 @@
...
+                // Put the buffer back on the queue and show an error message
+                arv_stream_push_buffer(stream, buffer);
+                NUClear::log<NUClear::WARN>("Failed getting buffer from", context->name, "with error", msg);
+            }
         }
+    }
 
-        V4L2Cameras.clear();
+    void Camera::control_lost(ArvGvDevice*) {
+        NUClear::log<NUClear::WARN>("Control of a camera has been lost ");
```
Bidski: Maybe upgrade this to `NUCLEAR::FATAL` and also get camera details and remove the offending camera from our list (and print out details of which camera is lost)

module/input/Camera/src/settings.h
TrentHouliston: Doesn't work as it's a pointer to a pointer

module/input/Camera/src/time_sync.cpp
TrentHouliston: It's in nanoseconds, added a comment

module/input/Camera/src/Camera.cpp
TrentHouliston: Added units

module/input/Camera/src/Camera.cpp
TrentHouliston: Upgraded to fatal, remember to downgrade this to error one day when we can recover from it

module/input/Camera/src/Camera.cpp
TrentHouliston: Working on it

module/input/Camera/src/Camera.cpp
TrentHouliston: Removed it, it's not very useful anyway

TrentHouliston:
This should be mostly ready to test, However someone needs to go into the configuration for the cameras and add in the relative offsets for the cameras from the head pitch servo

module/input/Camera/src/Camera.cpp
```diff
@@ -47,7 +50,7 @@ namespace input {
              ************************************************************************/
             if (it == cameras.end()) {
                 // Strip the .yaml off the name of the file to get the name of the camera
-                std::string name = ::basename(config.path.substr(0, config.path.find_last_of('.')).c_str());
+                std::string name = ::basename(config.fileName.substr(0, config.fileName.find_last_of('.')).c_str());
```
Bidski: `snake_case`?

module/input/Camera/src/Camera.cpp
```diff
@@ -231,13 +234,27 @@ namespace input {
             arv_stream_set_emit_signals(stream.get(), true);
         });
 
+        on<Trigger<Sensors>>().then("Buffer Sensors", [this](const Sensors& sensors) {
```
Bidski: Why not use a `Last<N>`?

module/input/Camera/src/Camera.cpp
```diff
@@ -327,8 +345,46 @@ namespace input {
...
                 msg->timestamp = NUClear::clock::time_point(nanoseconds(ts));
-                // TODO attach Hcw once we have access to sensors
+
+                Eigen::Affine3d Hcw;
+
+                /* Mutex Scope */ {
+                    std::lock_guard<std::mutex> lock(reactor.sensors_mutex);
+
+                    Eigen::Affine3d Hcp = context->Hcp;
+                    Eigen::Affine3d Hpw;
```
Bidski:
```
Eigen::Affine3d Hpw = Eigen::Affine3d::Identity();
if (!reactor.Hpws.empty()) {
    /// .....
}
```

module/input/Camera/src/Camera.cpp
```diff
@@ -327,8 +345,46 @@ namespace input {
...
+                    if (reactor.Hpws.empty()) {
+                        Hpw = Eigen::Affine3d::Identity();
+                    }
+                    else {
+                        // Find the first time that is not less than the target time
+                        auto Hpw_it = std::lower_bound(reactor.Hpws.begin(),
+                                                       reactor.Hpws.end(),
+                                                       std::make_pair(msg->timestamp, Eigen::Affine3d::Identity()),
+                                                       [](const auto& a, const auto& b) { return a.first < b.first; });
```
Bidski: Pretty sure we can lose the `make_pair` here
```
auto Hpw_it = std::lower_bound(reactor.Hpws.begin(),
                                                       reactor.Hpws.end(),
                                                       msg->timestamp,
                                                       [](const auto& a, const NUClear::clock::time_point& b) { return a.first < b; });
```

module/input/Camera/src/Camera.cpp
```diff
@@ -357,7 +413,7 @@ namespace input {
     }
 
     void Camera::control_lost(ArvGvDevice*) {
-        NUClear::log<NUClear::WARN>("Control of a camera has been lost ");
+        NUClear::log<NUClear::FATAL>("Control of a camera has been lost ");
```
Bidski: Should still add in the context pointer

module/input/Camera/src/Camera.cpp
TrentHouliston: Because I wouldn't have a way of passing that to the static function

module/input/Camera/src/Camera.cpp
TrentHouliston: Not my fault, the Horus one is path, your configuration is fileName

Bidski: You should update it then

module/input/Camera/src/Camera.cpp
TrentHouliston: I've often had trouble with these signatures... but maybe this time it'll be different!

module/input/Camera/src/Camera.cpp
TrentHouliston: Added, and added the name to the message

module/input/Camera/src/Camera.cpp
```diff
@@ -1,129 +1,418 @@
...
 
-                if (dumpImages) {
-                    utility::vision::saveImage(fmt::format("image-{}.ppm", count++), *msg);
+                // Apply our timestamp offset
+                ts = ts + context->time.offset;
+
+                auto msg        = std::make_unique<Image>();
+                msg->format     = context->fourcc;
+                msg->dimensions = Eigen::Matrix<unsigned int, 2, 1>(width, height);
+                // TODO(trent) use an std::vector here to avoid the copy
```
Bidski: You might as well fix this since it seems like a simple enough fix

module/input/Camera/src/Camera.cpp
Bidski: Fine, fix it later then

PR #268, opened 2019-10-21
Upgrade the DataLogging module and allow it to split files as it writes and generate less code

TrentHouliston:


module/support/logging/DataLogging/src/DataLogging.cpp
```diff
@@ -0,0 +1,138 @@
...
+                            }
+                        }
+
+                        // Unbind any recorders we didn't save
+                        for (auto& handle : handles) {
+                            log<NUClear::INFO>("Data logging for type", handle.first, "disabled");
+                            handle.second.unbind();
+                        }
+
+                        // Add any recorders new ones that we don't have yet
```
Bidski: You what now?

module/support/logging/DataLogging/src/DataLogging.cpp
```diff
@@ -0,0 +1,138 @@
...
+                        for (auto& setting : cfg["messages"].config) {
+                            std::string name = setting.first.as<std::string>();
+                            bool enabled     = setting.second.as<bool>();
+
+                            // If it was enabled and we are keeping it enabled, keep it and remove it from the old list
```
Bidski: You sure that is what this does?

shared/utility/support/math_string.cpp
```diff
@@ -40,6 +40,11 @@ namespace support {
         table.add_constants();
         table.add_constant("auto", std::numeric_limits<double>::infinity());
 
+        // File size constants
+        table.add_constant("KiB", 1024);
+        table.add_constant("MiB", 1048576);
+        table.add_constant("GiB", 1073741824);
```
Bidski: Is it worth replacing the raw values with `std::pow(2, 10)`, etc? Or maybe as a [hex floating point literal](https://en.cppreference.com/w/cpp/language/floating_literal)?
```
0x1p10 = 1024
0x1p20 = 1048576
0x1p30 = 1073741824
```

module/support/logging/DataLogging/src/DataLogging.cpp
TrentHouliston: I speak well the english!

shared/utility/support/math_string.cpp
TrentHouliston: Maybe, but these are pretty safe

module/support/logging/DataLogging/src/DataLogging.cpp
TrentHouliston: I am sure that this is not what it does? That's kinda the same...

PR #270, opened 2019-11-07
Add a libuv based filesystem watcher

TrentHouliston:
Adds a libuv based filesystem watcher that backs the configuration system. Removes the need for libfswatch which wasn't as reliable.


module/extension/FileWatcher/src/FileWatcher.cpp
```diff
@@ -27,142 +29,205 @@ namespace extension {
...
-
-                if (reactions != handler->second.end()) {
-                    for (auto& reaction : reactions->second) {
-                        uint flags = 0;
-                        for (auto& flag : event.get_flags()) {
-                            flags |= flag;
-                        }
+        // Work out what path we are watching
+        std::array<char, 512> pathbuff;
+        pathbuff.fill('\0');
```
Bidski: `std::array<char, 512> pathbuff = {0};`?

module/extension/FileWatcher/tests/FileWatcherTestDirectory.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+
+TEST_CASE(
+    "Given a directory containing two files, when the file contents are modified, then change events are triggered for "
+    "each file.") {
+
+    INFO("Start Test");
+
+    NUClear::PowerPlant::Configuration config;
+    config.thread_count = 1;
```
Bidski: Since we know we used to have some sort of race condition is it worth adding extra threads to try and make sure it doesn't happen?

module/extension/FileWatcher/src/FileWatcher.cpp
TrentHouliston: I don't think that actually works in c++

module/extension/FileWatcher/tests/FileWatcherTestDirectory.cpp
TrentHouliston: Where the race condition was this won't make a difference

PR #273, opened 2019-11-12
Remove Armadillo from all modules in the support meta-module

Bidski:


module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -20,6 +20,8 @@
 #include "GoalDetector.h"
 
 #include <fmt/format.h>
+#include <Eigen/Core>
```
KipHamiltons: Not in module/support

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
```diff
@@ -592,7 +592,7 @@ namespace behaviour {
                 return scaledResults;
             }
 
-            Quad boundingBox = getScreenAngularBoundingBox(fixationObjects);
+            Quad<Eigen::Vector2d> boundingBox = getScreenAngularBoundingBox(fixationObjects);
```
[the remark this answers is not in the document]
Bidski: The quad class got rewritten to allow templating, this change is mostly unavoidable

shared/utility/math/geometry/Quad.h
```diff
@@ -43,7 +240,16 @@ namespace math {
              * @param right    The right x pixel value.
              * @param bottom   The bottom y pixel value.
              */
-            void set(double left, double top, double right, double bottom);
+            void set(const Scalar& left, const Scalar& top, const Scalar& right, const Scalar& bottom) {
```
KipHamiltons: Should this be inline keyworded? If not, how come this implementation is on the header file /*legit question, because I thought that's against standard practice */ ?

Bidski: [https://en.cppreference.com/w/cpp/language/inline](https://en.cppreference.com/w/cpp/language/inline)

> A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function.

KipHamiltons: Ahh, it's implicit. Thanks!

TrentHouliston: That page isn't actually correct. I'll fix it, it's implicit because it's a template not because it's a class

module/support/optimisation/WalkOptimiser/src/WalkOptimiser.cpp
```diff
@@ -19,52 +19,55 @@
...
 
-#include "message/input/ServoID.h"
+#include <fmt/format.h>
+
+#include "message/support/SaveConfiguration.h"
+#include "utility/input/ServoID.h"
 #include "utility/math/angle.h"
-#include "utility/math/optimisation/PGAoptimiser.h"
-#include "utility/support/yaml_armadillo.h"
+// #include "utility/math/optimisation/PGAoptimiser.h"
```
TrentHouliston: Commented code?

module/support/optimisation/WalkOptimiser/src/WalkOptimiser.cpp
```diff
@@ -19,52 +19,55 @@
...
                 }
 
                 walk_command.segments.clear();
-                for (auto& segment : config["segments"]) {
+                for (auto& segment : config["segments"].config) {
 
                     walk_command.segments.push_back(FixedWalkCommand::WalkSegment());
-                    walk_command.segments.back().direction   = segment["direction"].as<arma::vec>();
+                    walk_command.segments.back().direction   = segment["direction"].as<Expression>();
                     walk_command.segments.back().curvePeriod = segment["curvePeriod"].as<double>();
```
TrentHouliston: Should this be as Expression too?

module/support/optimisation/WalkOptimiser/src/WalkOptimiser.cpp
```diff
@@ -86,27 +89,27 @@ namespace support {
                     // Start optimisation
                     std::cerr << "Optimiser command" << std::endl;
                     // Get samples
-                    samples = utility::math::optimisation::PGA::getSamples(
-                        getState(walkConfig), parameter_sigmas, number_of_samples);
+                    // samples = utility::math::optimisation::PGA::getSamples(
+                    //     getState(walkConfig), parameter_sigmas, number_of_samples);
```
TrentHouliston: More commented code?

module/support/optimisation/WalkOptimiser/src/WalkOptimiser.cpp
```diff
@@ -129,36 +132,34 @@ namespace support {
...
+                    setWalkParameters(getWalkConfig(samples.row(currentSample).transpose()));
+                    // Now wait for WalkConfigSaved
+                }
+            });
 
             on<Trigger<OptimisationComplete>, Sync<WalkOptimiser>>().then("Record Results", [this] {
                 // Combine samples
-                arma::vec result = utility::math::optimisation::PGA::updateEstimate(samples, fitnesses);
+                Eigen::VectorXd result;
+                // result = utility::math::optimisation::PGA::updateEstimate(samples, fitnesses);
```
TrentHouliston: Pretty sure you've entirely broken this module.
So either this porting is not yet finished or you need to delete this module entirely

shared/utility/math/angle.h
```diff
@@ -85,6 +86,10 @@ namespace math {
             return d;
         }
 
+        inline double vectorToBearing(const Eigen::Vector2d& dirVec) {
+            return std::atan2(dirVec.y(), dirVec.x());
+        }
```
TrentHouliston: Really?...

shared/utility/math/geometry/Quad.h
```diff
@@ -28,13 +29,209 @@ namespace utility {
 namespace math {
     namespace geometry {
 
+        // *********************
+        // * ARMADILLO VERSION *
+        // *********************
```
TrentHouliston: If we're trying to move away from armadillo, rather than reimplementing things like the Quad class, should we look into bringing in a computational geometry library?

shared/utility/math/matrix/transform.h
```diff
@@ -0,0 +1,67 @@
...
+            Eigen::Rotation2Dd R(local.linear());
+            Eigen::Vector2d newPos = R * reference.translation();
+            Eigen::Affine2d result;
+            result.translation() = local.translation() + newPos;
+            // do not use normalizeAngle here, causes bad things when turning! TODO: unsure on cause
+            result.linear() = Eigen::Rotation2Dd(Eigen::Rotation2Dd(local.linear()).angle()
+                                                 + Eigen::Rotation2Dd(reference.linear()).angle())
+                                  .toRotationMatrix();
+            return result;
+        }
```
TrentHouliston: Don't the Affine2d already do this by multiplying them together?

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
```diff
@@ -592,7 +592,7 @@ namespace behaviour {
                 return scaledResults;
             }
 
-            Quad boundingBox = getScreenAngularBoundingBox(fixationObjects);
+            Quad<arma::vec2> boundingBox = getScreenAngularBoundingBox(fixationObjects);
```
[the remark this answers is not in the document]
Bidski: Yes, this is intentional. I rewrote `Quad` to let it be templated and this forced me to do this. It will be fixed in a later PR

module/support/SoccerSimulator/src/SoccerSimulator.cpp
```diff
@@ -66,16 +67,17 @@ namespace support {
...
     using utility::math::angle::vectorToBearing;
     using utility::math::coordinates::cartesianToSpherical;
-    using utility::math::matrix::Transform2D;
+    using utility::math::transform::localToWorld;
+    using utility::math::transform::worldToLocal;
     using utility::motion::kinematics::calculateRobotToIMU;
-    using utility::nusight::drawArrow;
-    using utility::nusight::drawSphere;
+    // using utility::nusight::drawArrow;
+    // using utility::nusight::drawSphere;
```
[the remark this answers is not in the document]
Bidski: Should probably be removed I think, pretty sure they don't exist any more

module/support/optimisation/WalkOptimiser/src/WalkOptimiser.cpp
Bidski: Intentional as an example of how to use this module since you purged all of our optimisers in the thanos finger snap.

module/support/optimisation/WalkOptimiser/src/WalkOptimiser.cpp
Bidski: I guess it could be

module/support/optimisation/WalkOptimiser/src/WalkOptimiser.cpp
Bidski: No, you broke this when you purged all of optimisers

shared/utility/math/geometry/Quad.h
Bidski: That is a much bigger overhaul than what we are planning right now. This armadillo version is only surviving until the behaviour modules are converted to Eigen

shared/utility/math/matrix/transform.h
Bidski: will have to look into that

module/support/optimisation/WalkOptimiser/src/WalkOptimiser.cpp
JosephusPaye: Probably leave a comment saying that then.

shared/utility/math/angle.h
Bidski: Not sure what you are asking here

Bidski: Not sure what you are asking here

TrentHouliston: That indeed was a terrible statement with no context.
What I was meaning is that this function just calls another function. There is no complexity being masked so I think this function is unecessary. Whoever is calling it could just use atan2 themselves

module/support/VirtualCamera/src/VirtualCamera.cpp
```diff
@@ -32,11 +32,9 @@ namespace support {
     using extension::Configuration;
 
     using message::input::Image;
-    using message::vision::LookUpTable;
 
     using utility::support::Expression;
-
-    using FOURCC = utility::vision::FOURCC;
+    using utility::vision::FOURCC;
```
TrentHouliston: There is a new fourcc thing that came in with the new camera module that we should probably use instead of these constants, although maybe that is for another PR. (or maybe we should delete this whole module as i'm not sure it even works)

module/support/SoccerSimulator/src/SoccerSimulator.cpp
```diff
@@ -340,11 +360,14 @@ namespace support {
...
-
-            emit(drawSphere("ball", {world.ball.position(0), world.ball.position(1), 0}, 0.1, 0));
+            Eigen::Vector2d bearingVector = world.robotPose.rotation() * Eigen::Vector2d::UnitX();
+            Eigen::Vector3d robotHeadingVector(bearingVector.x(), bearingVector.x(), 0.0);
+            // emit(drawArrow("robot",
+            //                Eigen::Vector3d(world.robotPose.translation().x(), world.robotPosetranslation()..y(),
+            //                0.0), 1.0, robotHeadingVector, 0.0));
+
+            // emit(drawSphere("ball", Eigen::Vector3d(world.ball.position.x(), world.ball.position.y(), 0.0), 0.1,
+            // 0.0));
```
TrentHouliston: Commented code?

module/support/SoccerSimulator/src/SoccerSimulator.cpp
```diff
@@ -66,16 +67,14 @@ namespace support {
...
     using utility::motion::kinematics::calculateRobotToIMU;
-    using utility::nusight::drawArrow;
-    using utility::nusight::drawSphere;
     using utility::nusight::graph;
     using utility::support::Expression;
 
     double triangle_wave(double t, double period) {
-        auto a = period;  // / 2.0;
-        auto k = t / a;
+        double a = period;  // / 2.0;
```
TrentHouliston: What does this comment mean?

module/support/configuration/SoccerConfig/src/SoccerConfig.cpp
```diff
@@ -19,18 +19,17 @@
...
 
-using extension::Configuration;
-using message::support::FieldDescription;
-
 namespace module {
 namespace support {
     namespace configuration {
 
+        using extension::Configuration;
+
```
TrentHouliston: Should there be this gap between the usings here?

module/support/optimisation/WalkOptimiser/src/WalkOptimiser.cpp
```diff
@@ -19,59 +19,63 @@
...
+// Comment remains as an example of usage for this module.
+// #include "utility/math/optimisation/PGAoptimiser.h"
+#include "utility/support/yaml_expression.h"
 
 namespace module {
 namespace support {
     namespace optimisation {
 
+        using extension::Configuration;
+
```
TrentHouliston: Should there be this gap? Doesn't it not sort the usings if there is a gap?

shared/utility/math/matrix/transform.h
```diff
@@ -0,0 +1,54 @@
...
+            double angle       = utility::math::angle::normalizeAngle(Eigen::Rotation2Dd(reference.linear()).angle()
+                                                                - Eigen::Rotation2Dd(world.linear()).angle());
+            diff.linear()      = Eigen::Rotation2Dd(angle).toRotationMatrix();
+            diff.translation() = reference.translation() - world.translation();
+
+            Eigen::Affine2d result;
+            result.translation() = world.linear() * diff.translation();
+            result.linear()      = diff.linear();
+            return result;
+        }
```
TrentHouliston: Is this all that's left of the transform? Isn't this just the same as world.inverse() * reference or something? We have been trying to get away from this world local etc with the `Htw` etc code. So this should just be some multiplication of either `world * reference` or `world.inverse() * reference`

module/support/SoccerSimulator/src/SoccerSimulator.cpp
Bidski: Haven't a clue, it was there originally. I guess we used to use a half period at some point?

TrentHouliston: I suppose it's just as confusing as the rest of this module

module/support/VirtualCamera/src/VirtualCamera.cpp
Bidski: If it doesn't work then it should be fixed. A fake camera module has its uses. Whatever we do, we can make that a later PR.

shared/utility/math/matrix/transform.h
Bidski: `worldToLocal` seems to have no connection to `world * reference` or `world.inverse() * reference`, or any permutation of those that I can think of.

TrentHouliston: the intention of worldToLocal was that you'd pass a world reference and get it back as a local reference. Then the corresponding localToWorld would do the opposite. The idea being that the world transform you give is actually a world to local transform itself. So if it's not doing that's a bad sign

Bidski: It appears that `worldToLocal` should be equivalent to `world.inverse() * reference`. This will need further investigating once we get to a module that actually matters (like the walk engine)

PR #274, opened 2019-11-22
Update Arch Linux install scripts

Bidski:


doc/ArchInstall/arch-chroot_install.sh
```diff
@@ -7,14 +7,18 @@ HOME="/home/${USER}"
 HOST="nugus"
 HOSTNAME="${HOST}${ROBOT_NUMBER}"
 IP_ADDR="10.1.1.${ROBOT_NUMBER}"
+ETHERNET_INTERFACE="${ETHERNET_INTERFACE:-eth0}"
+WIFI_INTERFACE="${WIFI_INTERFACE:-wlan0}"
```
TrentHouliston: Is there a way to enforce that these go to the hardware based names rather than these names?

doc/ArchInstall/arch-chroot_install.sh
```diff
@@ -80,7 +85,8 @@ pacman -S --noconfirm --needed \
 	linux-headers \
 	bluez \
 	bluez-utils \
-	rsync
+	rsync \
+	linux-firmware
```
TrentHouliston: This normally comes in via pacstrap

doc/ArchInstall/arch-chroot_install.sh
```diff
@@ -234,10 +240,10 @@ EOF
 
 # Make sure the system checks /usr/local for libraries
 echo -e "/usr/local/lib\n/usr/local/lib64" > /etc/ld.so.conf.d/usrlocal.conf
+ldconfig
 
 # Make sure python checks /usr/local for packages
-sed "s/^\(PREFIXES\s=\s\)\[\([^]]*\)\]/\1[\2, '\/usr\/local']/" -i /usr/lib/python3.7/site.py \
-ldconfig
+sed "s/^\(PREFIXES\s=\s\)\[\([^]]*\)\]/\1[\2, '\/usr\/local']/" -i /usr/lib/python3.7/site.py
```
TrentHouliston: Is this wrong now with the 3.8 changes?

TrentHouliston:
Do we think it would be a good idea to make this into a custom ISO builder via a b script tool?

PR #277, opened 2019-11-26
Update Dockerfile so it no longer has hardcoded python versions

TrentHouliston:
Arch updated to python 3.8 so we need to fix up this code so that it matches


TrentHouliston:
You know, this should probably be worked out automatically rather than constantly bumping this number

Bidski:
`python -c "import sys; print("/usr/lib/python{}.{}/site.py".format(sys.version_info.major, sys.version_info.minor))"`

PR #278, opened 2019-11-27
Upgrade the NBS tools and add a video decoder

TrentHouliston:
This PR is required for the camera calibration code to work


tools/nbs_tools/decode_image/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+from .decode_bayer import decode_bayer
+from .decode_jpeg import decode_jpeg
+
+from .fourcc import fourcc
+
+
+def decode_image(data, format):
+
+    if format in [fourcc("JPBG"), fourcc("JPRG"), fourcc("JPGR"), fourcc("JPGB")]:
+        return decode_bayer(data, format)
```
Bidski: This should decode the jpeg before calling `decode_bayer`. This function should also handle raw bayer formats

tools/nbs_tools/decode_image/decode_bayer.py
```diff
@@ -0,0 +1,52 @@
...
+import numpy as np
+import tensorflow as tf
+
+from .fourcc import fourcc
+
+lookups = {}
+
+
+def decode_bayer(data, format, factor=2):
+    raw_img = tf.io.decode_jpeg(data, channels=1)
```
Bidski: This function should only decode bayer formats. Jpeg decompression should happen external to this

tools/nbs_tools/video.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python3
+
+import multiprocessing
+import multiprocessing.pool
+import os
+
+import numpy as np
+import tensorflow as tf
```
Bidski: what are you doing?????

tools/nbs_tools/decode_image/video_recorder.py
```diff
@@ -0,0 +1,106 @@
...
+        self.video_path = os.path.join(output_path, "{}.mp4".format(camera_name))
+        self.buffer_size = buffer_size
+
+        self.timecode = open(self.timecode_path, "w")
+        self.frames = []
+        self.start_time = None
+
+        self.encoder = Popen(
+            [
+                "ffmpeg",
```
Bidski: Has ffmpeg been added to the docker image? Or has this dependency been introduced in some other manner?

TrentHouliston: this runs on the host, not on docker, so presumably, you'd need to have FFmpeg installed to run `./b nbs video`

tools/nbs_tools/video.py
TrentHouliston: I have no idea what you mean

tools/nbs_tools/decode_image/decode_bayer.py
TrentHouliston: It's actually not even decoding bayer, It's just repermuting the mosaic. The actual debayering is happening in the ffmpeg call

tools/nbs_tools/decode_image/__init__.py
TrentHouliston: It actually just permutes bayer, so in theory, you don't even have to decode a raw bayer using this

PR #283, opened 2019-12-11
Fix boundary condition in CalculateGroundSpace()

ysims:
Clamps Rwf(2,2) to prevent nan as a result of acos being given a value out of bounds.


shared/utility/motion/ForwardKinematics.h
```diff
@@ -644,7 +645,7 @@ namespace motion {
             MatrixType Rwf(Hwt.rotation() * Rtf);
 
             // Dot product of foot z (in world space) with world z
-            Scalar alpha = std::acos(Rwf(2, 2));
+            Scalar alpha = std::acos(utility::math::clamp(-1.0, Rwf(2, 2), 1.0));
```
TrentHouliston: In angle.h there is an acos_clamped function that already does this. Although looking at that code it shouldn't use fmax and fmin, just max and min

TrentHouliston: The fix in angle.h should look something like this (also fix asin_clamped) while you're at it
```cpp
        template <typename T>
        inline T acos_clamped(const T& a) {
            return std::acos(std::max(std::min(a, T(1.0)), T(-1.0)));
        }
```

shared/utility/motion/ForwardKinematics.h
```diff
@@ -31,6 +31,7 @@
 #include "utility/input/LimbID.h"
 #include "utility/input/ServoID.h"
 #include "utility/math/angle.h"
+#include "utility/math/comparison.h"
```
TrentHouliston: Don't need this anymore

PR #285, opened 2020-01-12
Make buildkite push up new images to dockerhub when merged to master

TrentHouliston:


JosephusPaye:
Looks good, just uncomment the `master` scoping thingy and can merge.

.buildkite/pipeline.yml
```diff
@@ -1,4 +1,17 @@
 steps:
+  # Push up new images to DockerHub when they get to master
```
JosephusPaye: Would be nice to say why we push the images: for faster during development/CI where nothing changes in the build system.

JosephusPaye: Maybe also where the credentials are stored: exposed as env vars in the buildkite agent hook on each agent.

PR #288, opened 2020-02-11
Remove util-linux as it is poision (makes machines unbootable)

TrentHouliston:
Also fix the accidentally downgraded glib


tools/dockerise/__init__.py
```diff
@@ -44,113 +44,6 @@ def build_platform(platform):
...
-        except KeyError:
-            print(event["status"])
-        except ValueError:
-            print(event["status"])
-
-    # Clean up all the bars
-    for k, v in bars.items():
-        v["bar"].reset()
-        v["bar"].close()
-
```
Bidski: Did you intend this?

TrentHouliston: I did not, it doesn't work on my work machine so I delete it

PR #290, opened 2020-02-19
Command to find unused modules

KelseyBkah:
Adds command to find unused modules


tools/unused.py
```diff
@@ -0,0 +1,86 @@
...
+def register(command):
+    # Install help
+    command.help = "Creates a list of unused or commented out modules"
+
+
+# @run_on_docker
+def run(**kwargs):
+
+    source_path = b.cmake_cache["NUbots_SOURCE_DIR"]
+    roles_path = source_path + "/roles"
```
TrentHouliston: When joining paths in python you should use `os.path.join('a', 'b', 'c', 'd')`
```python
    roles_path = os.path.join(source_path, "/roles")
```
That said the roles path should be another variable in the cmake_cache. However, it must have been missed and should be added to `nuclear/CMakeLists.txt` and used in `nuclear/roles/CMakeLists.txt`
and then you can use
```suggestion
    roles_path = os.path.join(source_path + b.cmake_cache['NUCLEAR_ROLE_DIR'])
```

tools/unused.py
```diff
@@ -0,0 +1,86 @@
...
+    modules_path = source_path + "/modules"
```
TrentHouliston: This one is in the cmake_cache already
```suggestion
    modules_path = os.path.join(source_path, b.cmake_cache['NUCLEAR_MODULE_DIR'])
```

tools/unused.py
```diff
@@ -0,0 +1,86 @@
...
+    # print(roles_path, modules_path)
```
TrentHouliston: Don't leave commented code
```suggestion
```

tools/unused.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
+
+import os
+import re
+import b
+
+# @run_on_docker
```
TrentHouliston: Don't leave commented code (and this module 100% should not be running on docker)
```suggestion
```

tools/unused.py
```diff
@@ -0,0 +1,86 @@
...
+def register(command):
+    # Install help
+    command.help = "Creates a list of unused or commented out modules"
+
+
+# @run_on_docker
```
TrentHouliston:
```suggestion
```

tools/unused.py
```diff
@@ -0,0 +1,86 @@
...
+    # print(roles_path, modules_path)
+    existing_modules = {}
+    missing_modules = {}
+    commented_modules = {}
+    used_modules = []
+    unused_commented_modules = []
+
+    # a list of folders that could be within a module
+    skip_list = ["src", "data", "config", "tests", "report"]
+    skip_list = [s + os.sep for s in skip_list]
```
TrentHouliston: No need to do a skip list, Just search every single folder for one that matches what a module would look like
```python
re.search(r'(?:[^/]+/)*(\w+)/src/(\w+).hp{2}?$')`
```

tools/unused.py
```diff
@@ -0,0 +1,86 @@
...
+    # find all the modules that are available
+    for root, dirs, files in os.walk(modules_path):
+        if len(files) != 0:
+            if not any(substring in (root + os.sep) for substring in skip_list):
+                existing_modules[
+                    root.replace(modules_path + os.sep, "")
+                ] = 0  # make sure we remove the base path from the module
+
+    # compare our roles to our existing modules
+    for root, dirs, files in os.walk(roles_path):
```
TrentHouliston: `os.listdir` is plenty for the roles directory, we're not going to have roles in subdirectories

tools/unused.py
```diff
@@ -0,0 +1,86 @@
...
+        for file in files:
+            with open(os.path.join(roles_path, file), "r") as role:
+                for line in role:
+                    line = line.strip()
+
+                    if "#" in line:  # a commented
```
TrentHouliston: Rather than doing it this way, a regular expression that finds modules and commented modules would work well.
Like this to find modules
```python
re.findall(r'(\w+::(?:\w+:{2}?)*', line)
```
And then check if the start of this line is preceded by the first index of a '#' character

TrentHouliston (dismissed):
Some suggestions to improve the code

tools/unused.py
LukeFarrawell: Previously you had suggested the following:

> A better way than using a skip list would be to look for the CMakeLists.txt file in the directory and finding a NUClearModule call inside it.

> Rather than doing it this way, just search for files named CMakeLists.txt that contain a call to NUClearModule(...)

tools/unused.py
```diff
@@ -0,0 +1,83 @@
...
+import b
+
+def register(command):
+    # Install help
+    command.help = "Creates a list of unused or commented out modules"
+
+
+def run(**kwargs):
+
+    source_dir = b.cmake_cache[b.cmake_cache["CMAKE_PROJECT_NAME"] + "_SOURCE_DIR"]
```
LukeFarrawell: Using `'` instead of `"` to be consistent with the rest of the files / other python files.

tools/unused.py
```diff
@@ -0,0 +1,83 @@
...
+                                    line = line[:additional_comment_loc]
+
+                                line = line.replace("::", os.sep)
+                                commented_modules[line.strip()] = 1
+                            continue
+                        else:
+                            # the comment was just a description, remove the comment and continue
+                            line = line[:location].strip()
+
+                    if "::" in line:  # the line is a module
```
LukeFarrawell: previously @TrentHouliston suggested:

> It's probably more reliable to do it the other way and convert the path to a namespace rather than namespace to path

Can't remember if we ever discussed what scenarios in which my implementation would not work.

tools/unused.py
```diff
@@ -0,0 +1,83 @@
...
+                        line = line.replace("::", os.sep)
+                        if line in existing_modules:
+                            existing_modules[line] += 1
+                            used_modules.append(line)
+                        else:
+                            missing_modules[line] = 1
+
+    print("Unused modules:\n")
```
LukeFarrawell: @TrentHouliston wanted to use termcolour to colour the output, not sure if this is still something that is wanted.

tools/unused.py
TrentHouliston: That's a way better idea! Good job past me and bad present me for forgetting...

tools/unused.py
Bidski: Our auto-formatting forces `"` now

KipHamiltons:
This now builds and LGTM

Bidski (approved):
Looks OK to me

PR #291, opened 2020-02-19
Improve the b scripts logic so that it can find nested tools

TrentHouliston:
The b script currently needs to have subtools like nbs that need to be updated each time that a new tool is added. Also when using one of these tools, it imports all of them so if your system does not have the dependencies required to run one of them the import will fail.

This PR instead makes it look through the files recursively to find the tool you want and load just that unless you provide an invalid tool. Then it will try to load everything to provide you with the help information


nuclear/b.py
```diff
@@ -85,50 +85,69 @@
...
+        for dirpath, dnames, fnames in os.walk(path):
+            for f in fnames:
+                if f != "__init__.py" and f.endswith(".py"):
+
+                    # Check if this is the tool for the job
+                    components = os.path.relpath(os.path.join(dirpath, f[:-3]), path).split(os.sep)
+                    if sys.argv[1 : len(components) + 1] == components:
+
+                        # Load the module
+                        fqn = ".".join(components)
```
Bidski: Whats `fqn` for?

TrentHouliston: It's for the next few lines where I instead of using fqn just use the same thing again!

PR #292, opened 2020-02-25
Remove (almost all) build warnings

KipHamiltons:
Each commit should be a different type of error.

I am not sure about the casts I did to change signedness for comparison of integer warnings: is there a better cast for that situation than (int)<var>?

There are two warnings from the openCL engine used by visualmesh which I haven't got rid of, but they are in untracked files unfortunately.


module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -72,7 +71,7 @@ namespace vision {
                 // Partition the indices such that we only have the ball points that dont have ball surrounding them
                 auto boundary = utility::vision::visualmesh::partition_points(
                     indices.begin(), indices.end(), neighbours, [&](const int& idx) {
-                        return idx == indices.size() || (cls(BALL_INDEX, idx) >= config.confidence_threshold);
+                        return idx == (int) indices.size() || (cls(BALL_INDEX, idx) >= config.confidence_threshold);
```
TrentHouliston: Never use c style casts, they're always the wrong choice. Depending on the circumstance you should be using either `reinterpret_cast`, `static_cast` or a functional cast `int(indices.size())`

module/extension/FileWatcher/src/FileWatcher.cpp
```diff
@@ -90,6 +90,8 @@ namespace extension {
                 }
             }
         }
+        // Shut compiler up about unused argument by pretending to use it
+        (void) status;
```
TrentHouliston: Rather than having an unused argument used like this, you should either be removing the variable or using it for something. So, in this case, change the function signature to
```cpp
void FileWatcher::file_watch_callback(uv_fs_event_t* handle, const char* filename, int events, int /*status*/) {
```

module/localisation/BallLocalisation/src/BallLocalisation.cpp
```diff
@@ -57,7 +57,7 @@ namespace localisation {
 
         /* Run Time Update */
         on<Every<15, Per<std::chrono::seconds>>, Sync<BallLocalisation>, With<FieldDescription>, With<Sensors>>().then(
-            "BallLocalisation Time", [this](const FieldDescription& field, const Sensors& sensors) {
+            "BallLocalisation Time", [this]() {
```
TrentHouliston: When you remove all the arguments you don't need the parenthesis (although I think this is a personal choice)
```cpp
"BallLocalisation Time", [this] {
```
is equally fine.

module/behaviour/strategy/SoccerStrategy/src/SoccerStrategy.cpp
```diff
@@ -338,7 +338,7 @@ namespace behaviour {
             emit(std::move(reset));
         }
 
-        void SoccerStrategy::penaltyShootoutLocalisationReset(const FieldDescription& fieldDescription) {
+        void SoccerStrategy::penaltyShootoutLocalisationReset() {
```
TrentHouliston: I find it odd that we don't use this variable. It implies that we are doing something wrong and the warning is valid and should be fixed, not silenced.

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -76,7 +76,7 @@ namespace vision {
                 // Partition the indices such that we only have the goal points that dont have goal surrounding them
                 auto boundary = utility::vision::visualmesh::partition_points(
                     indices.begin(), indices.end(), neighbours, [&](const int& idx) {
-                        return idx == indices.size() || (cls(GOAL_INDEX, idx) >= config.confidence_threshold);
+                        return idx == (int) indices.size() || (cls(GOAL_INDEX, idx) >= config.confidence_threshold);
```
TrentHouliston: c style cast

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -125,7 +125,8 @@ namespace vision {
                             cluster.end(),
                             neighbours,
                             [&](const int& idx) {
-                                return idx == indices.size() || (cls(GOAL_INDEX, idx) >= config.confidence_threshold);
+                                return idx == (int) indices.size()
```
TrentHouliston: C style cast

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -134,7 +135,8 @@ namespace vision {
...
```
TrentHouliston: C style cast

module/vision/GreenHorizonDetector/src/GreenHorizonDetector.cpp
```diff
@@ -50,7 +50,7 @@ namespace vision {
                 indices.end(),
                 neighbours,
                 [&](const int& idx) {
-                    return idx == indices.size()
+                    return idx == (int) indices.size()
```
TrentHouliston: C style cast

module/vision/VisualMesh/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 # Add the VisualMesh project to be built
 add_subdirectory(VisualMesh)
 
+add_compile_definitions(CL_TARGET_OPENCL_VERSION=120)
```
TrentHouliston: The new version of the visual mesh should fix up a lot of the warnings from that code

nuclear/cmake/Modules/NUClearCompilerSettings.cmake
```diff
@@ -24,3 +24,6 @@ if(NOT MSVC)
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
   set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 endif()
+
+# Set OpenCL Target version
+add_compile_definitions(CL_TARGET_OPENCL_VERSION=120)
```
TrentHouliston: This shouldn't go here

nuclear/message/generator/Enum.py
```diff
@@ -175,9 +175,15 @@ def generate_cpp(self):
...
-            }}"""
+            namespace {namespace} {{
+
+                std::ostream& operator<< (std::ostream& out, const {fqn}& val) {{
+                    return out << static_cast<std::string>(val);
+                }}
+            
+            }}
+
+            """
```
TrentHouliston: What did this do?

shared/utility/vision/visualmesh/VisualMesh.h
```diff
@@ -88,7 +88,7 @@ namespace vision {
                     }
                 }
                 // Only add cluster to list if it meets minimum size requirment
-                if (cluster.size() >= min_cluster_size) {
+                if ((int) cluster.size() >= min_cluster_size) {
```
TrentHouliston: C style cast

module/vision/BallDetector/src/BallDetector.cpp
KipHamiltons: Which is correct in this circumstance?

module/behaviour/strategy/SoccerStrategy/src/SoccerStrategy.cpp
KipHamiltons: I will remove the change then?

module/vision/VisualMesh/CMakeLists.txt
KipHamiltons: Is that on a branch I can merge into this branch?

nuclear/cmake/Modules/NUClearCompilerSettings.cmake
KipHamiltons: Where should it go?

module/vision/BallDetector/src/BallDetector.cpp
TrentHouliston: Either functional or static cast, Personally I prefer functional as it's shorter.

nuclear/message/generator/Enum.py
KipHamiltons: The code this generated created a warning that the operator was not declared in the namespace, and that it was only there as a friend. This removed that warning. Although I am not exactly sure why this is different to the compiler than 
`std::ostream& {namespace}operator<< (std ...)`
but it got rid of it.

module/behaviour/strategy/SoccerStrategy/src/SoccerStrategy.cpp
TrentHouliston: Either that or work out why we're not using the description of the field when we reset for a penalty.

nuclear/cmake/Modules/NUClearCompilerSettings.cmake
TrentHouliston: I'm not sure, but this is setting the property for all files in the entire codebase. It's something that should be fixed by the visual mesh

nuclear/message/generator/Enum.py
TrentHouliston: Interesting... I'm glad you found a way to get rid of that warning I couldn't find one.
You'll need to make sure that printing out enums still works.
Remove the excess whitespace around the namespace here as well.

module/localisation/BallLocalisation/src/BallLocalisation.cpp
Bidski: You could also do 
`"BallLocalisation Time", [this](const FieldDescription& /*field*/, const Sensors& /*sensors*/)`

Not that it makes a lot of difference

module/behaviour/strategy/SoccerStrategy/src/SoccerStrategy.cpp
Bidski: Do we even call this function anywhere? If we do then removing that argument would surely cause and error?

module/extension/FileWatcher/src/FileWatcher.cpp
KipHamiltons: I tried that, but it says that the function is used by /usr/local/include/uv.h:1586:48: when I try to build it, so it looks to me like the function is referenced by another library, and requires that function signature

module/vision/VisualMesh/CMakeLists.txt
TrentHouliston: No somebody has to go and update the visual mesh code to work with the new architecture which is still somewhat in progress

module/behaviour/strategy/SoccerStrategy/src/SoccerStrategy.cpp
TrentHouliston: It's called on 229

nuclear/cmake/Modules/NUClearCompilerSettings.cmake
Bidski: Did having this in the ImageCompressor and VisualMesh modules not resolve this warning?

module/extension/FileWatcher/src/FileWatcher.cpp
TrentHouliston: Don't remove the last argument, just the variable name that it's going in. So leave the int part only

module/behaviour/strategy/SoccerStrategy/src/SoccerStrategy.cpp
KipHamiltons: It's called once in the same file. I removed that argument from the call.

module/localisation/BallLocalisation/src/BallLocalisation.cpp
TrentHouliston: Actually, if we're not using them why don't we just remove the With<> parts?

nuclear/cmake/Modules/NUClearCompilerSettings.cmake
KipHamiltons: I changed the flag inside the module/VisualMesh and it was still giving the error from  nuclear/roles/CMakeFiles/balllocalisation.dir/balllocalisation.cpp.o and similar others from nuclear/roles/CMakeFiles/...

TrentHouliston: It's something that needs to be fixed by the VisualMesh rather than this code

module/extension/FileWatcher/src/FileWatcher.cpp
KipHamiltons: Roger that

module/localisation/BallLocalisation/src/BallLocalisation.cpp
Bidski: This is also a valid point

module/behaviour/strategy/SoccerStrategy/src/SoccerStrategy.cpp
Bidski: In that case we should put it back and figure out why we aren't using it

nuclear/cmake/Modules/NUClearCompilerSettings.cmake
Bidski: Yea, I think you need to define it in the header file itself, not as a command line define as it spreads to things that include it

module/extension/FileWatcher/src/FileWatcher.h
```diff
@@ -69,7 +69,7 @@ namespace extension {
         /// @brief Called by the powerplant to build and setup the FileWatcher reactor.
         explicit FileWatcher(std::unique_ptr<NUClear::Environment> environment);
         ~FileWatcher();
-        static void file_watch_callback(uv_fs_event_t* handle, const char* filename, int events, int status);
+        static void file_watch_callback(uv_fs_event_t* handle, const char* filename, int events, int /*status*/);
```
TrentHouliston: I don't think you need to remove it from the header to get rid of the warning

nuclear/cmake/Modules/NUClearCompilerSettings.cmake
```diff
@@ -23,7 +23,4 @@ if(NOT MSVC)
   set(CMAKE_CXX_STANDARD 17)
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
   set(CMAKE_POSITION_INDEPENDENT_CODE ON)
-endif()
-
-# Set OpenCL Target version
-add_compile_definitions(CL_TARGET_OPENCL_VERSION=120)
\ No newline at end of file
+endif()
```
TrentHouliston: No newline at end of file (although I'm sure `cmake-format` is yelling at you for this already)

cmake/Modules/FindAravis.cmake
```diff
@@ -1,5 +1,5 @@
 include(ToolchainLibraryFinder)
-ToolchainLibraryFinder(
+toolchainlibraryfinder(
```
TrentHouliston: `cmake-format` shouldn't have done this. Have you merged in master recently?

KipHamiltons: I have just merged master in, although I only branched off master a few days ago

module/extension/FileWatcher/src/FileWatcher.h
KipHamiltons: Roger that

cmake/Modules/FindAravis.cmake
TrentHouliston: https://github.com/NUbots/NUbots/blob/hamiltons/fix_build_warnings/.cmake-format.py#L233
That line in `cmake-format` should prevent it from messing with the case of ToolchainLibraryFinder. The fact that it either means you're using the wrong version of the docker image (run `./b target generic`) or something else has gone wrong

KipHamiltons: that worked. Had to rebuild a bunch, but ./b format fixed the capitalisation back I think

TrentHouliston: Need to commit the changes to undo. It will just not change the case, not fix it back to the correct case

KipHamiltons:
enum.py confirmed to not give build errors when the << operator was used.

PR #293, opened 2020-03-19
Added extract_balls.py to nbs

danielginn:


tools/nbs.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import nbs_tools.extract_images
+import nbs_tools.extract_balls
```
TrentHouliston: Pretty sure this file should be gone? Or have I forgotten to make a PR for that upgrade

ysims: You forgot to make a PR for that upgrade? :P 
The ball extraction was added because a mechatronics student needed the data. 
I feel like it should be easier to extract data though? It seems strange to have to make a new tool for every type of message we want to decode to json.

danielginn: Yes, we re-created the ball extractor for the mecha student. Do I need to make any changes?

TrentHouliston: That PR seems like it went in #291 so there shouldn't be an nbs.py but it might have been missed deleting this file.
Also why do you need a tool to extract things to json? There already exists the json tool

danielginn: It is just a custom tool for this mecha guy. The json tool is just a screen dump of everything. So do I need to do anything?

JosephusPaye: If I recall correctly the JSON tool just dumps all recorded messages? He wanted a subset of the messages with specific data. But perhaps something like https://stedolan.github.io/jq/ could be used?

TrentHouliston: At the very least, instead of making a new tool for every single thing you want to dump with JSON, add a command-line argument to the JSON dumper that makes it only dump out specific message types

ysims:
I'm gonna delete this since it isn't something we want in the main codebase.
Branch will stay alive until the mecha student is finished or someone updates the tool to allow command line arguments for specific messages.

PR #295, opened 2020-04-07
Improve use of curses windows and log messages

Bidski:
Creates two curses windows on the screen.

 - A 5 line window at the top of the screen to display current walk command and status.
 - A second window occupying the rest of the screen to display log messages.

Log messages should hopefully display with the appropriate colour as they normally would.

The log messages window should scroll off the bottom correctly, but any lines that scroll off the top will be lost. FileLogHandler should be coupled with these changes if log messages want to be kept.

image


JosephusPaye:
This is nice! How would you enable file logging with this if you want to keep log messages?

Bidski:
FileLogHandler module, but I have had issues using it with docker, needs some more playing with

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.cpp
```diff
@@ -77,64 +120,166 @@ namespace behaviour {
...
+
+            on<Trigger<LogMessage>, Sync<KeyboardWalk>>().then([this](const LogMessage& packet) {
+                // Where this message came from
+                std::string source = "";
+
+                // If we know where this log message came from, we display that
+                if (packet.task) {
+                    // Get our reactor name
+                    std::string reactor = packet.task->identifier[1];
+
```
Bidski, line 138: Currently this line causes a set fault. Not sure why.

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.cpp
```diff
@@ -77,64 +120,166 @@ namespace behaviour {
...
+            });
+
+            on<Trigger<LogMessage>, Sync<KeyboardWalk>>().then([this](const LogMessage& packet) {
+                // Where this message came from
+                std::string source = "";
+
+                // If we know where this log message came from, we display that
+                if (packet.task) {
+                    // Get our reactor name
+                    std::string reactor = packet.task->identifier[1];
```
Bidski, line 137: Currently this line causes a set fault. Not sure why.

TrentHouliston: Not every log line has a reactor name... not sure if that would cause this though

Bidski: It seems like the `task` pointer is being deallocated i between the if statement and `std::string reactor = packet.task->identifier[1];` 

Either that or its already an invalid pointer by the time we reach the if statement

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.cpp
```diff
@@ -77,64 +120,166 @@ namespace behaviour {
...
                     case 'q': quit(); return;
-                    default: log("Unknown Command");
+                    default:
+                        log<NUClear::ERROR>("Unknown Command");
+                        printStatus();
+                        break;
+                }
+            });
+
+            on<Trigger<LogMessage>, Sync<KeyboardWalk>>().then([this](const LogMessage& packet) {
```
TrentHouliston, line 130: The Sync<KeyboardWalk> is what is causing this segfault, you can't use a reschedule operation here.

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.h
```diff
@@ -43,6 +46,12 @@ namespace behaviour {
             float headYaw   = 0;
             float headPitch = 0;
 
+            std::shared_ptr<WINDOW> command_window, log_window;
```
TrentHouliston, line 49: Separate these out into separate lines

module/support/logging/FileLogHandler/data/config/docker/FileLogHandler.yaml
```diff
@@ -0,0 +1 @@
+log_file: /home/nubots/NUbots/log
```
TrentHouliston, line 1: Why is this a part of this PR?

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.cpp
```diff
@@ -236,51 +188,81 @@ namespace behaviour {
             wrefresh(log_window.get());
         }
 
+        void KeyboardWalk::update_window(const UpdateWindow& packet) {
```
TrentHouliston, line 191: Seems a bit excessive to have this as a struct now

module/support/logging/FileLogHandler/data/config/docker/FileLogHandler.yaml
Bidski: Probably isn't very necessary, unless I also add a keyboardwalkfake role

Bidski:
@ysims want to test this before I merge it?

PR #297, opened 2020-04-16
Minor Docker fixes

Bidski:
Changes introduced in this PR
 - Fixes the meson pkg-config issue 
 - `./b format` was failing because isort and cmake-format were being installed in `/home/nubots/.local/bin` 
 - Adds nuc8 toolchain


docker/usr/local/bin/install-from-source
```diff
@@ -224,6 +224,7 @@ elif [ "${METHOD}" = 'install-meson-from-source' ]; then
         --cross-file="${PREFIX}/meson.cross" \
         --buildtype=release \
         --prefix="${PREFIX}" \
+        --pkg-config-path="${PKG_CONFIG_PATH}" \
```
JosephusPaye, line 227: Is this a fix for the issue @ysims was having?

Bidski: Yep

tools/format.py
```diff
@@ -18,16 +18,20 @@ def run(**kwargs):
     # Change into the project directory
     os.chdir(b.project_dir)
 
+    # Prepend the users local bin directory to the path
+    env = os.environ
+    env["PATH"] = os.pathsep.join([os.path.join(env["HOME"], ".local", "bin"), env["PATH"]])
```
TrentHouliston, line 23: We need to fix this properly, make sure pip installs it in the correct location

PR #305, opened 2020-05-07
Split up the docker image pushing to multiple agents

TrentHouliston:
Split the docker image pushing so it goes to multiple agents, this should help improve time for image pushing


Bidski:
How do we know that this has worked?

PR #306, opened 2020-05-18
Fix linking errors introduced by changing to GCC10

Bidski:
GCC10 has changed a default from `-fcommon` to `-fno-common`. This will result in multiple definition linker errors for some libraries. More information can be [found here](https://wiki.gentoo.org/wiki/Gcc_10_porting_notes/fno_common).


docker/Dockerfile
```diff
@@ -237,7 +240,10 @@ RUN install-from-source ftp://ftp.alsa-project.org/pub/lib/alsa-lib-1.2.1.2.tar.
     --without-debug
 RUN install-from-source https://github.com/espeak-ng/pcaudiolib/archive/1.1.tar.gz
 COPY --chown=nubots:nubots usr/local/package/espeak.sh /usr/local/package/espeak.sh
-RUN /usr/local/package/espeak.sh https://github.com/espeak-ng/espeak-ng/releases/download/1.50/espeak-ng-1.50.tgz
+# GCC10 changed the default from -fcommon to -fno-common.
+# This causes multiple definition errors during linking for espeak
+RUN CFLAGS="-fcommon" CXXFLAGS="-fcommon" \
```
TrentHouliston, line 245: Maybe just apply the patch https://github.com/espeak-ng/espeak-ng/commit/812ea8131904a6e171209220feb75842f6453d69
Or if they have a new release that includes it use that

docker/usr/local/toolchain/generate_toolchains.py
```diff
@@ -107,11 +107,11 @@ def generate_toolchain_script(target):
...
-        export CFLAGS="{flags}"
-        export CXXFLAGS="${{CFLAGS}}"
-        export CPPFLAGS="${{CFLAGS}}"
-        export FFLAGS="${{CFLAGS}}"
-        export FCFLAGS="${{CFLAGS}}"
+        export CFLAGS="${{CFLAGS}} {flags}"
+        export CXXFLAGS="${{CXXFLAG}} ${{CFLAGS}}"
+        export CPPFLAGS="${{CPPFLAGS}} ${{CFLAGS}}"
+        export FFLAGS="${{FFLAGS}} ${{CFLAGS}}"
+        export FCFLAGS="${{FCFLAGS}} ${{CFLAGS}}"
```
TrentHouliston, line 114: So this just forwards the environment along with everything else? Does this only work for autotools or cmake etc as well

docker/Dockerfile
```diff
@@ -143,7 +143,10 @@ RUN install-from-source https://github.com/KhronosGroup/OpenCL-CLHPP/archive/mas
 
 # OpenCL loader library
 RUN install-package ruby
-RUN install-from-source https://github.com/OCL-dev/ocl-icd/archive/v2.2.12.tar.gz
+# GCC10 changed the default from -fcommon to -fno-common.
+# This causes multiple definition errors during linking for espeak
+RUN CFLAGS="-fcommon" CXXFLAGS="-fcommon" \
```
TrentHouliston, line 148: I assume newer versions of this don't fix the problem?

docker/usr/local/toolchain/generate_toolchains.py
Bidski: My understanding that cmake uses `CMAKE_<LANG>_FLAGS_INIT` and `<LANG>FLAGS` to initialise `CMAKE_<LANG>_FLAGS`. So I think it combines both of those and so it should.

We have no need to test any of the other build tools (cmake, meson, boost), so we wont know until we need to.

PR #307, opened 2020-05-21
Added descriptions to protobuf messages

KipHamiltons:
Mission is to add brief descriptions to the proto messages, ready for the upcoming NUbots docu-lution.
Descriptions I didn't write:
~`SoccerObjectPriority.proto` has one message, which I have no idea how to describe.~
`Subsumption.proto` has one message and I am not sure about what it really does.
~`KickCommand.proto` has one `enum` `KickCommandType`, and I don't know what it means for a kick to be `NORMAL` vs `POWER` (also, we don't ever seem to do `POWER` kicks?)~

@TrentHouliston suggested I add descriptions to some of the nested objects. I have done this for `Behaviour.proto`, `FieldTarget.proto`, `FixedWalkCommand.proto`, `KickPlan.proto` so far. Was this along the lines you were thinking?

### DIFF
--- a/shared/message/Geometry.proto
+++ b/shared/message/Geometry.proto
@@ -24,21 +24,33 @@ package message;
 import "Vector.proto";
 import "Matrix.proto";
 
+/**
+ * A normal vector and the distance to the line.
+ */
 message Line {
     fvec3 normal   = 1;
     float distance = 2;
 }
 
+/**
+ * The circle's centre position and radius.
+ */
 message Circle {
     double radius = 1;
     vec2   centre = 2;
 }
 
+/**
+ * A 3D matrix representation of the Ellipse.
+ */
 message Ellipse {
     // See http://en.wikipedia.org/wiki/Matrix_representation_of_conic_sections
     mat3 ellipse = 1;
 }
 
+/**
+ * Four corner points of a quadrilateral.
+ */
 message Quad {
     vec2 tl = 1;
     vec2 tr = 2;
@@ -46,10 +58,16 @@ message Quad {
     vec2 br = 4;
 }
 
+/**
+ * Contains the polygon's location.
+ */
 message Polygon {
     vec2 point = 1;
 }
 
+/**
+ * Contains the positions of the four corners of a frustrum.
+ */
 message Frustum {
     vec3 tl = 1;
     vec3 tr = 2;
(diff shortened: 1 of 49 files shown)


ysims:
`POWER` is probs the penalty kicks. They're designed to kick the ball in one go from about... I think it's halfway between the goals and center... into the goal, with no consideration for the welfare of the robot.

KipHamiltons:
> `POWER` is probs the penalty kicks. They're designed to kick the ball in one go from about... I think it's halfway between the goals and center... into the goal, with no consideration for the welfare of the robot.

I don't think it was ever implemented. I can't find anywhere in the code which checks what value it has. The only time it's ever mentioned I could find is in `KickPlanner.cpp` where in both cases a command is created with it set to `NORMAL`.

ysims:
> > `POWER` is probs the penalty kicks. They're designed to kick the ball in one go from about... I think it's halfway between the goals and center... into the goal, with no consideration for the welfare of the robot.
> 
> I don't think it was ever implemented. I can't find anywhere in the code which checks what value it has. The only time it's ever mentioned I could find is in `KickPlanner.cpp` where in both cases a command is created with it set to `NORMAL`.

Yeah I don't remember it being used. At Sydney we just used ScriptRunner. It would be good to use in the future, but that would require writing some behaviour to go with it. Like 'when you have a clear shot of the goals, POWER KICK!'

KipHamiltons:
> > > `POWER` is probs the penalty kicks. They're designed to kick the ball in one go from about... I think it's halfway between the goals and center... into the goal, with no consideration for the welfare of the robot.
> > 
> > 
> > I don't think it was ever implemented. I can't find anywhere in the code which checks what value it has. The only time it's ever mentioned I could find is in `KickPlanner.cpp` where in both cases a command is created with it set to `NORMAL`.
> 
> Yeah I don't remember it being used. At Sydney we just used ScriptRunner. It would be good to use in the future, but that would require writing some behaviour to go with it. Like 'when you have a clear shot of the goals, POWER KICK!'. Although it can hurt the robot.

I'll add a comment for it, now I know what the purpose is

TrentHouliston:
Power was implemented as a way to swap between our nice gentler kicks and our full power kick. I'm not sure it ever was integrated super well. SoccerObjectPriority is designed to prioritise what the head behaviour should look at. So if it's looking for ball first then goal, or goal then ball etc

KipHamiltons:
> SoccerObjectPriority is designed to prioritise what the head behaviour should look at. So if it's looking for ball first then goal, or goal then ball etc

Added.

PR #308, opened 2020-06-05
Extend servo reading to handle current

Bidski:
This is untested, but it should do the job.


Bidski:
I am in agreement with @TrentHouliston . Testing is left to the user as an exercise.

JosephusPaye:
LOL, started something similar in [paye/servo-current](https://github.com/NUbots/NUbots/compare/paye/servo-current): https://github.com/NUbots/NUbots/commit/09079861a92320ecb8e6fd68548f975fda83efe0

Yours is a bit more complete though.

ysims:
This was originally done for a mechatronics student after current readings.
I assume if it came up once, it'll come up again. Is there a better solution for this?

JosephusPaye:
Perhaps reading smaller chunks in two separate requests would be faster than reading everything in one big request.

That said, we should test this and measure how much slower it is before doing any optimisations.

Bidski:
> Perhaps reading smaller chunks in two separate requests would be faster than reading everything in one big request.
> 
> That said, we should test this and measure how much slower it is before doing any optimisations.

This PR was only created because there was going to be a mecha FYP that wanted to use current for something.

The better option is to upgrade our comms to Dynamixel V2 which contains functionality for indirect addresses. There are also a lot of other potential improvements that come from using the V2 protocol

tayloryoung6396:
This PR should be deleted. An issue should be made for version 2 protocol with a suggestion to use a custom read table

ysims:
> This PR should be deleted. An issue should be made for version 2 protocol with a suggestion to use a custom read table

Can you make an issue, ideally with some info on what needs to be done before we use v2.

tayloryoung6396:
> > This PR should be deleted. An issue should be made for version 2 protocol with a suggestion to use a custom read table
> 
> Can you make an issue, ideally with some info on what needs to be done before we use v2.

I think i'll leave it as an exercise for the owner of the PR

PR #312, opened 2020-06-17
Added comments to motion messages

KipHamiltons:
Tell me if I should add a different reviewer.
Most of this is probably wrong of course.

PR #314, opened 2020-06-17
Added comments to behaviour messages

KipHamiltons:
Let me know if I should add a different reviewer.
Most of this is probably wrong of course.


shared/message/behaviour/Subsumption.proto
```diff
@@ -21,6 +21,9 @@ syntax = "proto3";
 
 package message.behaviour;
 
+/**
+ *
+ */
```
Bidski, line 26: Subsumption is the system we use to grant different modules access to different limbs of the robot. Each module only requests the limbs that it needs to function and a priority for how important the module is.

This message appears to only be used for sending subsumption information to NUsight. Check out the [Controller module](https://github.com/NUbots/NUbots/tree/master/module/behaviour/Controller) for how this is used. 

@TrentHouliston can also provide details since he wrote the system.

shared/message/behaviour/WalkPath.proto
```diff
@@ -23,6 +23,9 @@ package message.behaviour;
 import "message/behaviour/MotionCommand.proto";
 import "Vector.proto";
 
+/**
+ * Information used to make a path, to complete a MotionCommand which is included in the message.
+ */
 message WalkPath {
     // Sequence of robot states that form a path:
```
Bidski, line 30: Keep the formatting of comments consistent

shared/message/behaviour/Nod.proto
```diff
@@ -20,6 +20,13 @@ syntax = "proto3";
...
 package message.behaviour;
 
+/**
+ * Boolean indicating to nod.
+ */
 message Nod {
+
+    /**
+     * Boolean indicating to nod.
+     */
```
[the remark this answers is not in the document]
KipHamiltons: roger that

shared/message/behaviour/SoccerObjectPriority.proto
```diff
@@ -29,7 +36,7 @@ message SoccerObjectPriority {
         GOAL_RIGHT              = 4;
         GROUND_LEFT             = 5;
         GROUND_RIGHT            = 6;
-        OTHE                    = 7;
+        OTHER                   = 7;
```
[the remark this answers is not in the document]
KipHamiltons: Look, I am not sure. I can't see any code ever checking or referencing that option for the enum, which I think is why I made the change from memory.

shared/message/behaviour/Look.proto
```diff
@@ -23,6 +23,9 @@ package message.behaviour;
 import "google/protobuf/duration.proto";
 import "Vector.proto";
 
+/**
+ * Details of where and when to look
+ */
 message Look {
```
ysims, line 29: I feel like this file could do with more comments. I'm not really sure what angle and arcSize mean in this context. The angle I guess is the target angle for the head... But not sure how arcSize fits in.
From what I can tell, fixation, saccade and pan are different head movement types. Fixation being 'go to this position and stay', saccade 'go to this position but stop on the way for a set period of time the continue' but not sure how accurate that is, and pan being move towards the goal at a slow even pace? Only a guess tho.

shared/message/behaviour/SoccerObjectPriority.proto
```diff
@@ -29,7 +34,7 @@ message SoccerObjectPriority {
...
     }
 
     int32      ball       = 1;
```
ysims, line 40: What do these even mean

ysims (commented):
Just some more areas I feel have gaps

TrentHouliston (commented):
So many of these PRs have documentation comments that don't add anything. Comments should only be used when they're needed to add information that's not already there.

For example if you have

```
/// A measurement
message Measurement {
...
}
```

It's really a comment that doesn't add any information to the message, and just adds extra noise that someone will have to maintain.

KipHamiltons:
> So many of these PRs have documentation comments that don't add anything. Comments should only be used when they're needed to add information that's not already there.
> 
> For example if you have
> 
> ```
> /// A measurement
> message Measurement {
> ...
> }
> ```
> 
> It's really a comment that doesn't add any information to the message, and just adds extra noise that someone will have to maintain.

That's a fair criticism; I really wasn't sure how far to go, and what to add for things that really felt "self-documenting". I have updated this PR by removing the most egregious comments of that type, and I'll update the others now. Let me know if there are any left that you think still aren't worth keeping

shared/message/behaviour/Look.proto
KipHamiltons: This file doesn't seem to be actually used in the code base. @Bidski is this file worth keeping?

shared/message/behaviour/SoccerObjectPriority.proto
KipHamiltons: Not sure. `GOAL_SEARCH` and `LOST` are the ones used in the codebase, but they seem fairly self-explanatory

ysims: Sorry, maybe should have put this comment a bit lower. I meant the ball, goal and line ints.

shared/message/behaviour/Look.proto
KipHamiltons: Without knowing the ideas behind how this was implemented, and without any references in the codebase of how it is used, I don't feel confident enough to add comments about this.

ysims: Yeah, it's a tough one. I think behaviour needs a really good clean out. From what I understand behaviour is always left until RoboCup, and then it gets hacked together real quick, so no surprise it's a bit over the place. 
It can be something for the behaviour group to go through.

KipHamiltons: @ElderNoSpace :p

ElderNoSpace: @KipHamiltons  Maybe after the director.

Bidski: I think this was stuff Jake was using for head behaviour. If it isnt being referenced anywhere in the codebase then remove it. If we need it again it will still be in the commit history

shared/message/behaviour/MotionCommand.proto
```diff
@@ -23,30 +23,25 @@ package message.behaviour;
...
 
 /**
- * Details of which walk behaviour to execute and where on the field to do it.
+ * Details of which walk behaviour to execute and where on the field to do it
  */
 message MotionCommand {
-    // Defines the possible types of motion command:
     enum Type {
-        // Stop moving and just stand still, (equivalent to a VelocityCommand of {0} m/s).
+        /// Stop moving and just stand still, (equivalent to a VelocityCommand of {0} m/s)
```
Bidski, line 30:
```suggestion
        /// Stop moving and just stand still (equivalent to a VelocityCommand of 0 m/s)
```

shared/message/behaviour/ServoCommand.proto
```diff
@@ -22,41 +22,20 @@ package message.behaviour;
...
 
-/**
- * A command for a servo to execute.
- */
 message ServoCommand {
-
-    /**
-     * The subsumption which created the command.
-     */
+    /// The subsumption which created the command
```
Bidski, line 26:
```suggestion
    /// The subsumption ID which created the command
```

Bidski (changes requested):

shared/message/behaviour/Look.proto
KipHamiltons: Submitted #411

shared/message/behaviour/MotionCommand.proto
```diff
@@ -22,28 +22,26 @@ package message.behaviour;
...
 import "Vector.proto";
 
+/**
+ * Details of which walk behaviour to execute and where on the field to do it
+ */
 message MotionCommand {
-    // Defines the possible types of motion command:
     enum Type {
-        // Stop moving and just stand still, (equivalent to a VelocityCommand of {0} m/s).
+        /// Stop moving and just stand still (equivalent to a VelocityCommand of 0 m/s)
```
ysims, line 30:
```suggestion
        /// Stop moving and just stand still (equivalent to disabling the walk engine)
```
More accurate, might have been different in the past but QuinticWalk just disables it.

KipHamiltons:
@Bidski could I please grab a review on this one

shared/message/behaviour/MotionCommand.proto
```diff
@@ -27,21 +27,21 @@ import "Vector.proto";
  */
 message MotionCommand {
     enum Type {
-        /// Stop moving and just stand still, (equivalent to a VelocityCommand of {0} m/s)
-        StandStill = 0;
+        /// Stop moving and just stand still (equivalent to disabling the walk engine)
```
Bidski, line 30: I dont think this is true. The walk engine could still be enabled (and active) but the robot can be standing still. A veloctiy command of 0m/s and 0rad/s is more correct

Bidski (changes requested):

shared/message/behaviour/MotionCommand.proto
ysims: I remember looking into this because of this comment, and the walk engine disables the main reactor in this case. It doesn't get a 0 walk command.

ysims: Hmmmm or maybe it does?

ysims: I'm wrong, Alex is right, it's just a 0 walk command. I'm not sure what I'm 'remembering' then 😆

Bidski: I think, to a certain degree, we should avoid comments in the messages that are dependent on the current implementation in the modules. Just because the current walk engine disables itself for `STAND_STILL` (assume it does for the sake of argument), doesn't mean another walk engine we implement will do the same thing, but if this comment says that it does then it will be confusing later on. In this case, I think the correct comment should be `Stop moving and just stand still`, leave it up to the walk engine how it wants to implement that behaviour

KipHamiltons: Great point. I agree completely. I'll update the comment as such.

ysims (dismissed):
Looks fine.

shared/message/motion/KickCommand.proto
```diff
@@ -32,8 +32,11 @@ enum KickCommandType {
 }
 
 message KickCommand {
-    vec3            target            = 1;  // The point to kick
-    vec3            direction         = 2;  // force is the magnitude
+    /// The point to kick
```
Bidski: What?

Bidski (changes requested):

shared/message/motion/KickCommand.proto
KipHamiltons: This was from a merge commit. This file isn't edited in this PR. 

In #312 The comment got moved from next to the field to above it with the triple slash, as you can see here.

tayloryoung6396 (approved):
I mean, lgtm. Some of them seem a little in depth

PR #316, opened 2020-06-17
Added comments to output messages

KipHamiltons:
Let me know if I should add a different reviewer.
Most of this is probably wrong of course.


shared/message/output/CompressedImage.proto
```diff
@@ -25,27 +25,85 @@ import "google/protobuf/timestamp.proto";
...
-    uvec2                     dimensions = 2;
-    bytes                     data       = 3;
-    uint32                    camera_id  = 4;
-    string                    name       = 5;
-    google.protobuf.Timestamp timestamp  = 6;
-    mat4                      Hcw        = 7;
-    Lens                      lens       = 8;
+    /**
+     * Encoding of colour channels, e.g. BGGR, PRG8, RGBA, etc.
+     * see module::output::compressor::turbojpeg::Compressor::compress for full list.
```
TrentHouliston, line 71: This isn't the best place to find a full list. This is actually a fourcc code, so a code made up of 4 numbers. So any four characters would define "a format" (if it's valid or means anything is up to the user) but in theory, you could start adding things like `h264` etc the definition is not so limited

shared/message/output/CompressedImage.proto
```diff
@@ -25,27 +25,85 @@ import "google/protobuf/timestamp.proto";
...
+     */
+    bytes data = 3;
+
+    /**
+     * Unique identifying number for the camera which took the image.
+     */
+    uint32 camera_id = 4;
+
+    /**
+     * Identifying name of the image.
```
TrentHouliston, line 91:
```suggestion
     * Name of the camera that took the image.
```

shared/message/output/CompressedImage.proto
```diff
@@ -25,27 +25,85 @@ import "google/protobuf/timestamp.proto";
...
+     */
+    uvec2 dimensions = 2;
+
+    /**
+     * The actual image pixel data.
+     */
+    bytes data = 3;
+
+    /**
+     * Unique identifying number for the camera which took the image.
```
TrentHouliston, line 86:
```suggestion
     * Unique identifier for the camera which took the image.
```

shared/message/output/Say.proto
```diff
@@ -21,10 +21,12 @@ syntax = "proto3";
...
- * TODO document
- *
- * @author Trent Houliston
+ * Is a string for the robot to say.
  */
 message Say {
+
+    /**
+     * The string to say.
+     */
```
TrentHouliston, line 30: When documenting fields just use a triple slash rather than having a full block, you've done this throughout and it greatly increases the number of lines which makes it harder to read
```suggestion
    /// The string to say.
```

shared/message/output/CompressedImage.proto
```diff
@@ -25,27 +25,85 @@ import "google/protobuf/timestamp.proto";
 import "Matrix.proto";
 import "Vector.proto";
 
+/**
+ * Contains lens and image output information.
```
TrentHouliston, line 29: In general, I personally try to avoid using full stops in a comment like this unless you actually intend to make a new sentence. Many tools that use these sorts of comments to generate documentation work better when the text doesn't end in a full stop
```suggestion
 * Contains lens and image output information
```

TrentHouliston (dismissed):
Try to reduce the number of lines you're adding with the comments. For messages etc you can use `/** */` block comments, but for individual fields use a triple slash style `/// Text`

shared/message/output/CompressedImage.proto
KipHamiltons: Is there a best place to find a full list? Should I just refer to this list as the ones which are likely to be used?

shared/message/output/CompressedImage.proto
```diff
@@ -25,27 +25,45 @@ import "google/protobuf/timestamp.proto";
...
             RECTILINEAR = 1;
             EQUIDISTANT = 2;
             EQUISOLID   = 3;
         }
-
-        Projection projection   = 1;
-        float      focal_length = 2;
-        float      fov          = 3;
-        fvec2      centre       = 4;
+        /// The type of projection this lens is using
```
ysims, line 43:
```suggestion
        /// Lens projection type
```

shared/message/output/CompressedImage.proto
```diff
@@ -25,27 +25,45 @@ import "google/protobuf/timestamp.proto";
...
 
+/**
+ * Contains lens and image output information
+ */
 message CompressedImage {
+
+    /// The lens type and details about the images it makes
     message Lens {
+
+        /// Types of projection of our lenses
```
ysims, line 36:
```suggestion
        /// Lens projection types
```

shared/message/output/CompressedImage.proto
ysims: Maybe just say 'see module::output::compressor::turbojpeg::Compressor::compress for currently supported formats'?

ysims (approved):
I don't see any issues.

KipHamiltons:
@TrentHouliston unless you object, I am going to dismiss your review

shared/message/output/CompressedImage.proto
```diff
@@ -33,20 +33,28 @@ message CompressedImage {
...
             EQUISOLID   = 3;
         }
-
-        Projection projection   = 1;
-        float      focal_length = 2;
-        float      fov          = 3;
-        fvec2      centre       = 4;
-        fvec2      k            = 5;
+        Projection projection = 1;
+        /// Normalised focal length: focal length in pixels / image width
```
JosephusPaye, line 37:
```suggestion
        /// Normalised focal length, in pixels / image width
```

shared/message/output/CompressedImage.proto
```diff
@@ -33,20 +33,28 @@ message CompressedImage {
...
+        float focal_length = 2;
+        /// The angular diameter that the lens covers (the area that light hits on the sensor). In radians
```
JosephusPaye, line 39:
```suggestion
        /// Angular diameter that the lens covers (the area that light hits on the sensor), in radians
```

shared/message/output/CompressedImage.proto
```diff
@@ -33,20 +33,28 @@ message CompressedImage {
...
+        float fov = 3;
+        /// Normalised image centre offset: pixels from centre to optical axis / image width
```
JosephusPaye, line 41:
```suggestion
        /// Normalised image centre offset, in pixels from centre to optical axis / image width
```

shared/message/output/CompressedImage.proto
```diff
@@ -33,20 +33,28 @@ message CompressedImage {
...
+    uvec2  dimensions = 2;
+    bytes  data       = 3;
+    uint32 id         = 4;
+    /// Name of the camera that took the image
+    string name = 5;
+    /// Time when the image was taken
+    google.protobuf.Timestamp timestamp = 6;
+    /// World to camera Affine3D transform
+    mat4 Hcw = 7;
+    /// The lens information of the camera which took the image
```
JosephusPaye, line 58:
```suggestion
    /// Lens information of the camera that took the image
```

JosephusPaye (commented):
Looks good, just a few minor suggestions

shared/message/output/CompressedImage.proto
KipHamiltons: I think that's less clear. what about this:
```Normalised image centre offset - the number of pixels from centre to optical axis / image width```

shared/message/output/CompressedImage.proto
JosephusPaye: For these it's about deciding if we're using articles (_the_ lens..., _an_ object..., etc) or not (lens..., object..., etc). Went with the latter but it doesn't matter as long as we're consistent here and in other places.

shared/message/output/CompressedImage.proto
JosephusPaye: Fair

JosephusPaye (approved):
LGTM

KipHamiltons:
@TrentHouliston unless you say otherwise soon, I'll be dismissing your review :p

PR #318, opened 2020-06-30
Refactor Sensors.proto and add foot down transforms

Bidski:


module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -422,8 +422,9 @@ namespace platform {
                     /************************************************
                      *            Foot down information             *
                      ************************************************/
-                    sensors->right_foot_down = true;
-                    sensors->left_foot_down  = true;
+                    sensors->feet.resize(2);
+                    sensors->feet[ServoSide::RIGHT].down = true;
+                    sensors->feet[ServoSide::LEFT].dowtrue;
```
TrentHouliston, line 427: Wut?

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -517,6 +518,8 @@ namespace platform {
                         else if (!foot_down) {
                             previous_foot_down[side] = false;
                         }
+
+                        sensors->feet[side].Hwf = footlanding_Hwf[side];
```
TrentHouliston, line 522: So when our foot is not on the ground here I assume that means it'll just contain whatever the last position it had on the ground was

Bidski: Every sensors update will have the most recent Hwf for both feet

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
Bidski: Wut?

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -458,8 +441,8 @@ namespace platform {
                         }
                     }
 
-                    sensors->right_foot_down = feet_down[0];
-                    sensors->left_foot_down  = feet_down[1];
+                    sensors->feet[ServoSide::RIGHT].down = feet_down[0];
+                    sensors->feet[ServoSide::LEFT].down  = feet_down[1];
```
TrentHouliston, line 445: Should these 0/1 be changed to ServoSide?

shared/message/input/Sensors.proto
```diff
@@ -43,46 +42,37 @@ message Sensors {
...
-    repeated FSR    fsr                = 5;
+    repeated Foot   feet               = 5;
     repeated Servo  servo              = 6;
     repeated Button button             = 7;
     repeated LED    led                = 8;
     float           voltage            = 9;
     float           battery            = 10;
-    vec3            centre_of_pressure = 11;
-    mat2            robot_to_IMU       = 12;
+    repeated mat4   forward_kinematics = 15 [(array_size) = 20];
```
TrentHouliston, line 70: While we're here do we want to rename this to what it actually is? People always have to work it out and look it up.

shared/message/input/Sensors.proto
```diff
@@ -43,46 +42,37 @@ message Sensors {
...
+    vec4            centre_of_mass     = 17;
```
TrentHouliston, line 71: Likewise we should rename this to `rMTt` or whatever it actually is

shared/message/input/Sensors.proto
```diff
@@ -43,46 +42,37 @@ message Sensors {
...
+    mat4            Hgt                = 18;
+    mat3            inertial_tensor    = 20;
```
TrentHouliston, line 73: I don't know if our convention handles inertia tensors, nor do I know how to use them but it would be good to know what space etc this is measured in

shared/message/input/Sensors.proto
TrentHouliston: After looking at the code it seems there are `Htp` where `p` is part

Bidski: Particle, but yes

shared/message/input/Sensors.proto
Bidski: `rMTt`? torso to centre of mass in torso space? seems accurate

module/input/Camera/src/Camera.cpp
```diff
@@ -299,9 +299,9 @@ namespace input {
                                       })));
 
             // Get torso to head, and torso to world
-            Eigen::Affine3d Htp(sensors.Htp[ServoID::HEAD_PITCH]);
+            Eigen::Affine3d Htx(sensors.Htx[ServoID::HEAD_PITCH]);
```
TrentHouliston, line 302: This shouldn't be Htx here, this should be something real like Hth (for head), or it looks like before it was Htp (which is why it was all replaced here)

module/input/Camera/src/Camera.cpp
```diff
@@ -299,9 +299,9 @@ namespace input {
                                       })));
 
             // Get torso to head, and torso to world
-            Eigen::Affine3d Htp(sensors.forward_kinematics[ServoID::HEAD_PITCH]);
+            Eigen::Affine3d Htx(sensors.Htx[ServoID::HEAD_PITCH]);
             Eigen::Affine3d Htw(sensors.Htw);
-            Eigen::Affine3d Hpw = Htp.inverse() * Htw;
+            Eigen::Affine3d Hpw = Htx.inverse() * Htw;
```
TrentHouliston, line 304:
```suggestion
            Eigen::Affine3d Hpw = Htp.inverse() * Htw;
```

PR #320, opened 2020-07-06
Fix foot down calculation

Bidski:


shared/message/input/Sensors.proto
```diff
@@ -51,8 +52,9 @@ message Sensors {
         uint32 colour = 2;
     }
     message Foot {
-        bool down = 1;
-        mat4 Hwf  = 2;
+        bool            down = 1;
+        mat4            Hwf  = 2;
+        motion.BodySide side = 3;
```
TrentHouliston, line 57: Not 100% sure this is needed here as the index already implies it...

TrentHouliston: Actually, 100% remove this, people will think changing side == changing the meaning. But others will use it as an index and we have a problem.

Bidski:
You saw the nbs file @TrentHouliston , you tell me of it worked

TrentHouliston:
Probably kinda maybe?

Bidski:
Good enough for me

PR #323, opened 2020-07-07
Make DataPlayback cleaner and better able to handle multiple files

TrentHouliston:
This also makes the configuration no longer generated so it's easier to work with these files in docker

Pairs with [NUbook PR](https://github.com/NUbots/NUbook/pull/65)


module/support/logging/DataPlayback/data/config/DataPlayback.yaml
```diff
@@ -0,0 +1,19 @@
+# A list of files to load, if this list is empty it will instead look at the arguments on the command line
+# So you can call ./binary a.nbs b.nbs c.nbs etc
+files: []
+
+# Once the NBS file has finished playing should we start again at the beginning
+loop_playback: true
+
+# If the system should shut down once it finishes playing the file
+shutdown_on_end: false
```
Bidski, line 9: I assume these two are mutually exclusive?

module/support/logging/DataPlayback/data/config/DataPlayback.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# The amount of time into the future to buffer, should be at least 1000ms
+buffer_time: 1500
+
+# You can put any messages that is emitted in the system here and it will be logged
+# The below ones are just the most common and here as an example
```
Bidski, line 15:
```suggestion
# You can put any messages that is emitted in the system here and it will be played back
# The below ones are just the most common and here as an example
```

module/support/logging/DataPlayback/data/config/DataPlayback.yaml
TrentHouliston: They are

module/support/logging/DataPlayback/data/config/DataPlayback.yaml
```diff
@@ -2,18 +2,22 @@
...
+on_end: LOOP
 
 # If the system should shut down once it finishes playing the file
 shutdown_on_end: false
 
 # The amount of time into the future to buffer, should be at least 1000ms
 buffer_time: 1500
 
-# You can put any messages that is emitted in the system here and it will be logged
+# You can put any messages that is emitted in the system here and it will be played back
```
Bidski, line 18:
```suggestion
# You can put any messages that are emitted in the system here and they will be played back
```

module/support/logging/DataPlayback/src/DataPlayback.cpp
```diff
@@ -52,12 +58,15 @@ namespace support {
                     catch (const std::exception& ex) {
 
                         // If we reached the end of the file we either terminate or loop depending on the setting
-                        if (input_file.eof()) {
+                        if (!input_file || input_file->eof()) {
```
Bidski, line 61:
```suggestion
                        if (!input_file || !input_file->good()) {
```

module/support/logging/DataPlayback/src/DataPlayback.cpp
```diff
@@ -84,19 +97,26 @@ namespace support {
...
                                 playback_handle.disable();
-                                buffered = true;
+                                input_file = nullptr;
+                                buffered   = true;
                             }
                         }
                         // We tried to read something funny (possibly tried to read too many bytes)
-                        else if (input_file.bad() || input_file.fail()) {
-                            input_file.clear();
+                        else if (input_file->bad() || input_file->fail()) {
```
Bidski, line 118:
```suggestion
                        else if (!input_file->good()) {
```

module/support/logging/DataPlayback/src/DataPlayback.cpp
TrentHouliston: In this case, I specifically want to check fo see if we have reached the end of the file

Bidski: Dont we also care if a non-recoverable error has occurred on the file we are currently reading?

Bidski: I guess the real question is, why is branch special? If we take the other branch, next loop around we will be back at this branch anyway?

ysims:
Can confirm it works.
Needs some info in NUbook. Some information on how it works would be good, in the 'System' section under the 'Subsystem' chapter. It would probably fit in a new page called 'Support'. I'm in the process of writing a guide for recording and playback as well.

module/support/logging/DataPlayback/README.md
```diff
@@ -2,13 +2,20 @@ DataPlayback
 ============
 
 ## Description
+DataPlayback is a module which will allow you to play back `.nbs` files as if they were happening in realtime.
+This can be very useful for when you are trying to diagnose a problem for which you have a recording as you can continuously play back the same data.
 
+The module will buffer up packets from the future and emit them at the same rate that they were emitted originally.
 
 ## Usage
+In order to play back the messages, you need to create a role with this module that has the modules you wish to test. You then need to enable the message types that are the input to these modules by setting them to true in the configuration (e.g. `message.input.sensors: true`). Make sure the `.nbs` file contains the messages you are trying to emit (you can check with `./b nbs stats file`).
```
Bidski, line 11:
```suggestion
In order to play back the messages, you need to create a role with this module that has the modules you wish to test. You then need to enable the message types that are the input to these modules by setting them to true in the configuration (e.g. `message.input.Sensors: true`). Make sure the `.nbs` file contains the messages you are trying to emit (you can check with `./b nbs stats file`).
```

Bidski: You also need to make sure that any modules that emit a message you are playing back are removed from the role (e.g. is `message.input.Sensors` is being played back `platform::darwin::SensorFilter` and/or `platform::darwin::HardwareIO`/`platform::darwin::HardwareSimulator` should **not** be in the role as it could create conflicting messages in the system).

module/support/logging/DataPlayback/README.md
```diff
@@ -2,13 +2,20 @@ DataPlayback
...
 
+There are two methods to provide nbs files to this module, via the configuration file or via the command line. If the configuration file has arguments those are used first. If there are no files provided in the configuration file, the arguments from the command line are used instead. This is so that if you have another module in your role that relies on command line arguments you have an alternate way to provide these files.
```
Bidski, line 13: If both config file and command line arguments are provided will it play config file first, followed by command line arguments, or only config file?

.vscode/settings.json
```diff
@@ -6,10 +6,17 @@
...
     "editor.rulers": [
         120
     ],
+    "protoc": {
+        "options": [
+            "-I${workspaceFolder}/nuclear/message/proto"
+        ]
+    },
     "python.formatting.provider": "black",
+    "yaml.format.printWidth": 120,
```
Bidski, line 19: Why are these in here?

module/support/logging/DataPlayback/README.md
TrentHouliston: Only config file

.vscode/settings.json
TrentHouliston: Because they should be! And I don't want to do another PR to do it :P

Bidski: whats the protoc stuff?

TrentHouliston: Those are so that the compilation/errors for protobuf works properly in vscode. To do that they need the include path

PR #325, opened 2020-07-10
Use docker buildkit for docker image building

Bidski:


tools/dockerise/__init__.py
```diff
@@ -170,12 +81,18 @@ def platforms():
 
 
 def get_selected_platform():
+    # Get information about the selected image
     try:
-        selected = client.images.get("{}:selected".format(repository))
+        img_info = json.loads(
+            subprocess.Popen(
+                ["docker", "image", "inspect", "{}:selected".format(repository)], stdout=subprocess.PIPE
+            ).communicate()[0],
```
TrentHouliston, line 89: No need for all of this here, you can just use subprocess.check_output
```suggestion
            subprocess.check_output(["docker", "image", "inspect", "{}:selected".format(repository)]),
```

tools/dockerise/__init__.py
```diff
@@ -243,23 +164,48 @@ def run(rebuild, clean, platform, **kwargs):
...
-                except docker.errors.NotFound:
                     build_volume = None
 
                 # If we don't have a volume, make one
                 if build_volume is None:
-                    build_volume = client.volumes.create(build_volume_name)
+                    try:
+                        subprocess.Popen(
+                            ["docker", "volume", "create", build_volume_name], stdout=subprocess.DEVNULL
+                        ).communicate()
```
TrentHouliston, line 204: This also could probably just be an if statement with `subprocess.call` rather than a try-catch block

tools/dockerise/__init__.py
```diff
@@ -243,23 +164,48 @@ def run(rebuild, clean, platform, **kwargs):
...
+                    ).communicate()
+
+                    build_volume = build_volume_name
+                except subprocess.CalledProcessError:
+                    build_volume = None
+
+                # If we are cleaning, remove this volume so we can recreate it
+                if build_volume and clean:
+                    try:
+                        subprocess.Popen(
```
TrentHouliston, line 190: if statement with `subprocess.call` here too

tools/dockerise/__init__.py
```diff
@@ -243,23 +164,48 @@ def run(rebuild, clean, platform, **kwargs):
...
+                            ).communicate()
+                        except subprocess.CalledProcessError:
+                            cprint("docker image tag returned a non-zero exit code", "red", attrs=["bold"])
+                            exit(1)
 
                 # Find the volume for this platform
                 build_volume_name = "{}_{}_build".format(repository, platform)
                 try:
-                    build_volume = client.volumes.get(build_volume_name)
+                    subprocess.Popen(
```
TrentHouliston, line 179: subprocess.call with an if statement

tools/dockerise/__init__.py
```diff
@@ -243,23 +164,48 @@ def run(rebuild, clean, platform, **kwargs):
                     build_platform(platform)
                     # If we were building the selected platform we have to move our selected tag up
                     if selected_platform:
-                        client.images.get(tag).tag(repository, "selected")
+                        try:
+                            subprocess.Popen(
```
TrentHouliston, line 168: `subprocess.call` with an if statement

tools/dockerise/__init__.py
```diff
@@ -233,8 +154,8 @@ def run(rebuild, clean, platform, **kwargs):
                 # Check if the image we want exists
                 tag = "{}:{}".format(repository, platform)
                 try:
-                    client.images.get(tag)
-                except docker.errors.ImageNotFound:
+                    subprocess.Popen(["docker", "image", "inspect", tag], stdout=subprocess.DEVNULL).communicate()
```
TrentHouliston, line 157: `subprocess.call` with an if statement

tools/dockerise/__init__.py
```diff
@@ -225,6 +144,8 @@ def run(rebuild, clean, platform, **kwargs):
                 func(rebuild=rebuild, clean=clean, platform=platform, **kwargs)
             # Otherwise go and re-run the b script in docker
             else:
+                pty = WrapPty()
```
TrentHouliston, line 147: Do you even use this? Also, should you be using it for some of the commands?

.buildkite/docker-compose.yml
```diff
@@ -7,12 +7,16 @@ services:
...
+        BUILDKIT_INLINE_CACHE: 1
     volumes:
       - type: bind
         source: ./..
         target: /home/nubots/NUbots
         read_only: true
         consistency: cached
+    environment:
+      COMPOSE_DOCKER_CLI_BUILD: 1
+      DOCKER_BUILDKIT: 1
```
TrentHouliston, line 19: Do these environment variables only apply within the docker image or outside as well?

tools/dockerise/__init__.py
Bidski: It doesn't suppress the output if we just use check_output?

TrentHouliston: I thought it would capture output, is there some stderr output that's not getting captured?

Bidski: Also, check_output wont let me set the stdout argument

tools/dockerise/__init__.py
Bidski: pty wont suppress stdout or stderr, which we want most of the time. Also, it is used later on

tools/dockerise/__init__.py
Bidski: I decided not to capture stderr so we can see the errors if there are any, but we dont want the json output of docker inspect dumping into the terminal every time you run a b command

PR #327, opened 2020-07-17
Simscape udp

CMurtagh-LGTM:
This is the half of the SimscapeUDP project that works in the codebase, the other half is in a [fork](https://github.com/ElderNoSpace/Simscape/tree/murtagh/UDP) of Simscape (most of the code is in the UDP folder).

I still have problems with the simulation, so in my [fork](https://github.com/ElderNoSpace/Simscape/tree/murtagh/UDP) I have removed gravity. This has enabled me to test what I have wrote.

I basically sending servotargets messages over UDP, then from the simulation receiving a darwinSensors/Servos message made from the servo stores.


module/platform/SimscapeUDP/README.md
```diff
@@ -0,0 +1,14 @@
...
+    Middleware for matlab and NUbots codebase
+
+## Usage
+    Add to a Role to pass ServoTargets messages to the Matlab NUbots/Simscape which then simulates the robot. The simulation also generates sensor data from the Servo and emits them as Servos messages.
+
+## Emits
+    DarwinSensors/Servos
+    ServoTargets in the UDP scope
+
+## Dependencies
```
KipHamiltons, line 14: Should we include MATLAB as a dependency?

Also, this only worked with ubuntu 20.04 right? that should probably be on here somewhere I think

module/platform/SimscapeUDP/src/SimscapeUDP.h
```diff
@@ -0,0 +1,44 @@
...
+            std::string ip;
+            uint inPort;
+            uint outPort;
+        } config;
+    };
+
+}  // namespace platform
+}  // namespace module
+
+#endif
```
KipHamiltons, line 44: `#endif // MODULE_PLATFORM_SimscapeUDP_H`

KipHamiltons: actually SimscapeUDP_H should be all caps I think

module/platform/SimscapeUDP/src/SimscapeUDP.h
```diff
@@ -0,0 +1,44 @@
...
+namespace platform {
+
+    class SimscapeUDP : public NUClear::Reactor {
+    public:
+        explicit SimscapeUDP(std::unique_ptr<NUClear::Environment> environment);
+
+    private:
+        struct {
+            std::string ip;
+            uint inPort;
```
KipHamiltons, line 36: should this be a specific int size? I don't think this matters, but I don't know for sure

TrentHouliston: Technically for a port it would be a `uint16_t`. But it doesn't really matter unless you care about where you're storing it

module/platform/SimscapeUDP/README.md
ElderNoSpace: MATLAB is not a dependency for this part of the project, but I will hyperlink to the part that is dependent on it.

module/platform/SimscapeUDP/src/SimscapeUDP.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+        /*
+         * Receive a packet from the sim and emit it for the robot
+         */
+        on<UDP>(config.inPort).then([this](const UDP::Packet& packet) {
+            if (packet) {
+                std::vector<char> tmp = packet.payload;
+
+                while (tmp.back() == '\0') {
+                    tmp.pop_back();
```
KipHamiltons, line 66: Could use a comment here about why you need to remove the zero pad perhaps?

KipHamiltons:
I think this has been largely superseded by webots. Do you want to keep this PR open @CMurtagh-LGTM?

CMurtagh-LGTM:
Without Matt, noone has knowledge of simscape, if someone wants to do matlab stuff they can reopen this and @ me.

PR #328, opened 2020-07-18
Remove id from servos

TrentHouliston:
As the servos list is always looked up by id anyway. If this field contained anything other than it's index it would cause bugs. Therefore it shouldn't exist in the first place


Bidski:
Is the ID field seriously not used anywhere else?

Bidski:
I don't know how I feel about this. The argument for keeping the field is that there is no reason why ID and index can't be different, and if we are not using the ID field then we should be.

TrentHouliston:
The problem is if we use both methods it just gets confusing. 99% of the time we use the array using the enum. The other time in HeadBehaviour we do an entire loop through every servo just to try and find one we are looking for.

PR #329, opened 2020-07-22
Add a tool for filtering out message types from nbs files

TrentHouliston:
This tool will allow you to take an nbs file, and filter out specific message types from it. This can be useful for example if you have an NBS file with images/sensors in it but don't want to carry around a gigantic file. You could strip out the image messages and just keep the sensor data.

### DIFF
--- a/tools/nbs/filter.py
+++ b/tools/nbs/filter.py
@@ -59,6 +59,6 @@ def run(files, keep, remove, output, **kwargs):
                 if valid:
                     out.write(packet.timestamp, packet.msg)
 
-            # Update the progress bar
-            progress.n = decoder.bytes_read()
-            progress.update(0)
+                # Update the progress bar
+                progress.n = decoder.bytes_read()
+                progress.update(0)
--- a/tools/nbs/filter.py
+++ b/tools/nbs/filter.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+
+import fnmatch
+
+from .nbs import Decoder, Encoder
+
+
+def register(command):
+    command.help = (
+        "Filter an nbs stream, removing messages from it. ",
+        "The command will first remove any patterns provided with -r ",
+        "(or everything if no remove patterns are given) ",
+        "and then keep any of those removed that match the -k patterns",
+    )
+
+    # Command arguments
+    command.add_argument("files", metavar="files", nargs="+", help="The nbs files to filter")
+    command.add_argument("-k", "--keep", action="append", help="A message pattern to keep")
+    command.add_argument("-r", "--remove", action="append", help="A message pattern to remove")
+    command.add_argument("-o", "--output", default="out.nbs", help="The output file to store the filtered nbs in")
+
+
+def run(files, keep, remove, output, **kwargs):
+
+    # If no remove was provided, we assume we remove all
+    if remove is None:
+        if keep is None:
+            print("You must provide either something to remove or keep")
+            exit(1)
+        else:
+            remove = ["*"]
+
+    # If we didn't provide any keep arguments we at least need to ignore them
+    if keep is None:
+        keep = []
+
+    with Encoder(output) as out:
+
+        decoder = Decoder(*files)
+        with tqdm(total=len(decoder), unit="B", unit_scale=True, dynamic_ncols=True) as progress:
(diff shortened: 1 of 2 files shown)


tools/nbs/filter.py
```diff
@@ -0,0 +1,64 @@
...
+        "Filter an nbs stream, removing messages from it. ",
+        "The command will first remove any patterns provided with -r ",
+        "(or everything if no remove patterns are given) ",
+        "and then keep any of those removed that match the -k patterns",
+    )
+
+    # Command arguments
+    command.add_argument("files", metavar="files", nargs="+", help="The nbs files to filter")
+    command.add_argument("-k", "--keep", action="append", help="A message pattern to keep")
+    command.add_argument("-r", "--remove", action="append", help="A message pattern to remove")
```
Bidski, line 21: Probably useful to include an example of how to use these? Is it just regex? How do we handle multiple search patterns for `-r` and `-k`? Is it possible to specify `-r` or `-k` more than once?

TrentHouliston: I suppose, and yes you can have as many k/r as you want. It's not regex though, it's the same as shell uses with * and ?. I could go to regex though... thoughts?

JosephusPaye: I think regex would be good 👍, would be the least surprising thing.

Also, by message pattern, do you mean a pattern to match the message name? Can it match nested message names as well? Do those have a `.`?

Bidski: Shell globbing is probably fine, but regex is probably more understood (if people actually understand regex). Either way we need to be careful that the shell doesn't screw up the pattern anyway. 

I honestly don't really see us using too much in the way of wildcards for this, probably more likely that we will be listing all the messages we want/don't want. If we are using wildcards its probably to avoid typing `message.` all the time (`*Image` rather than `message.input.Image`). Is it worth perhaps changing to a comma-separated list for `-k` and `-r`, we can split on the comma and then do regex/globbing on the splits.

TrentHouliston: The main reason I went with shell globbing over regex was that if you did regex, all the dots in message.input.Sensors would technically need to be escaped (although it would probably work regardless and `message\.input\.Sensors` just looks overly complicated

Bidski: `.*Sensors` although that gets you DarwinSensors too

tools/nbs/filter.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def register(command):
+    command.help = (
+        "Filter an nbs stream, removing messages from it. ",
+        "The command will first remove any patterns provided with -r ",
+        "(or everything if no remove patterns are given) ",
+        "and then keep any of those removed that match the -k patterns",
+        "for example to keep only Sensors and DarwinSensors ",
+        "`./b nbs filter -k message.input.Sensors -k message.platform ",
```
TrentHouliston, line 17:
```suggestion
        "`./b nbs filter -k message.input.Sensors -k message.platform.darwin.DarwinSensors ",
```

tools/nbs/filter.py
```diff
@@ -0,0 +1,67 @@
...
+
+def register(command):
+    command.help = (
+        "Filter an nbs stream, removing messages from it. ",
+        "The command will first remove any patterns provided with -r ",
+        "(or everything if no remove patterns are given) ",
+        "and then keep any of those removed that match the -k patterns",
+        "for example to keep only Sensors and DarwinSensors ",
+        "`./b nbs filter -k message.input.Sensors -k message.platform.darwin.DarwinSensors ",
+        "`or to remove all CompressedImages `./b nbs filter -r message.output.CompressedImage",
```
Bidski, line 18:
```suggestion
        "Filter an nbs stream, removing messages from it. ",
        "The command will first remove any patterns provided with -r ",
        "(or everything if no remove patterns are given) ",
        "and then keep any of those removed that match the -k patterns",
        "for example to keep only Sensors and DarwinSensors ",
        "`./b nbs filter -k message.input.Sensors -k message.platform.darwin.DarwinSensors` ",
        "or to remove all CompressedImages `./b nbs filter -r message.output.CompressedImage` ",
        "or to keep all vision messages `./b nbs filter -k message.vision.*`",
```

PR #330, opened 2020-07-22
Fix Docker mount error in WSL 2

JosephusPaye:
The current `run_in_docker()` implementation detects if it's running in WSL and translates the source bind path to a WSL path. This doesn't works in WSL 2, and causes the mount to fail; leading to this error when running the `b` script:

```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"./b\": stat ./b: no such file or directory": unknown.
```

This PR ensures that the path translation is done only in WSL 1.


ysims:
Can confirm this branch works in Windows 😄

PR #332, opened 2020-07-29
Remove gradient from the cone as as it's duplicate information

TrentHouliston:
Radius and gradient are the same number and having both is terrible

### DIFF
--- a/module/vision/BallDetector/src/BallDetector.cpp
+++ b/module/vision/BallDetector/src/BallDetector.cpp
@@ -149,7 +149,6 @@ namespace vision {
                         // Ball cam space info
                         b.cone.axis     = horizon.Hcw.topLeftCorner<3, 3>().cast<float>() * axis;
                         float proj      = 1.0f / radius;
-                        b.cone.gradient = std::sqrt(proj * proj - 1.0f);
                         b.cone.radius   = radius;
 
                         // https://en.wikipedia.org/wiki/Angular_diameter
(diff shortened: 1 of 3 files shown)


module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -147,10 +147,9 @@ namespace vision {
                         }
 
                         // Ball cam space info
-                        b.cone.axis     = horizon.Hcw.topLeftCorner<3, 3>().cast<float>() * axis;
-                        float proj      = 1.0f / radius;
-                        b.cone.gradient = std::sqrt(proj * proj - 1.0f);
-                        b.cone.radius   = radius;
+                        b.cone.axis   = horizon.Hcw.topLeftCorner<3, 3>().cast<float>() * axis;
+                        float proj    = 1.0f / radius;
```
Bidski, line 151: Is `proj` still used?

shared/message/Geometry.proto
```diff
@@ -58,7 +58,7 @@ message Frustum {
 }
 
 message Cone {
-    fvec3 axis     = 1;
-    float gradient = 2;
-    float radius   = 3;
+    fvec3 axis = 1;
+    // Cos(theta) of the cone angle
+    float radius = 3;
```
Bidski, line 63: Should we update the index?

TrentHouliston: Would mean old NBS files won't play properly... Do we have any we care about?

module/vision/BallDetector/src/BallDetector.cpp
TrentHouliston: Removed

shared/message/Geometry.proto
Bidski: good point, we probably do have some

PR #335, opened 2020-08-13
Add an image decompressor to allow playback of nbs files with compressed images

TrentHouliston:


module/input/ImageDecompressor/src/ImageDecompressor.cpp
```diff
@@ -0,0 +1,139 @@
...
+                        std::rethrow_exception(eptr);
+                    }
+                    else {
+                        // Successful compression!
+                        ++decompressed;
+                        return;
+                    }
+                }
+            }
+            // We failed to compress this image
```
Bidski, line 123:
```suggestion
            // We failed to find an available decompressor this image
```

module/input/ImageDecompressor/src/ImageDecompressor.h
```diff
@@ -0,0 +1,57 @@
...
+
+namespace module {
+namespace input {
+
+    class ImageDecompressor : public NUClear::Reactor {
+    private:
+        struct DecompressorContext {
+            struct Decompressor {
+                /// This atomic bool gets turned on/off depending on if it is in use
+                std::unique_ptr<std::atomic<bool>> active;
```
Bidski, line 20: Does it really need to be a unique pointer?

TrentHouliston: It does have to be a unique pointer due to the way atomics work

PR #337, opened 2020-10-05
Removes Darwin chest location config value and leg joints functions

KipHamiltons:
Removes the config value `TEAMDARWINCHEST_TO_ORIGIN` referenced in #312 
A yaml file which set the value was also modified to remove it too.

This made one function obsolete, but it was not used anywhere. It was removed.
Another related function `calculateLegJointsTeamDarwin(..)` which used that removed function was also removed.


ysims (approved):
Yeah I don't see any reason to keep this, we should be getting rid of Darwin references in the codebase, and the nugus certainly doesn't use it.

PR #338, opened 2020-10-28
Make the ./b format script be multithreaded

TrentHouliston:
Means that you spend less time waiting for `./b format` to finish


tools/format.py
```diff
@@ -20,14 +41,7 @@ def run(**kwargs):
...
-        elif f.endswith(("CMakeLists.txt", ".cmake", ".role")):
-            print("Formatting {} with cmake-format".format(f))
-            call(["cmake-format", "-i", f])
-        elif f.endswith((".py")):
-            print("Formatting {} with isort and black".format(f))
-            call(["isort", f])
-            call(["black", f])
+
+    with multiprocessing.Pool(multiprocessing.cpu_count()) as pool:
+        for r in pool.imap_unordered(_do_format, files.splitlines()):
```
KipHamiltons, line 46: I was curious about whether changing the chunk size would have a large impact on the performance, so I unscientifically benched it on my pc and was only able to squeeze out an approx 1.5 s (~10%) run time improvement, with a chunk size of 20.

KipHamiltons (approved):
This is way faster. Nice job. LGTM

tools/format.py
TrentHouliston: Interesting... I wouldn't have thought it would make much of a difference

PR #339, opened 2020-11-05
Make dockerise copy over the current users UID to avoid permissions errors

TrentHouliston:
Currently, if you are not the first user on your computer (your UID is not 1000) then when you run in your docker image you'll get permissions errors. This sets the user id to the same as your current user when generating images.

### DIFF
--- a/tools/dockerise/__init__.py
+++ b/tools/dockerise/__init__.py
@@ -74,6 +74,8 @@ def build_platform(platform):
             "BUILDKIT_INLINE_CACHE=1",
             "--build-arg",
             "platform={}".format(platform if platform != "buildkit" else "generic"),
+            "--build-arg",
+            "user_uid={}".format(os.getuid()),
             "-t",
             local_tag,
         ],


tools/dockerise/__init__.py
```diff
@@ -74,6 +74,8 @@ def build_platform(platform):
             "BUILDKIT_INLINE_CACHE=1",
             "--build-arg",
             "platform={}".format(platform if platform != "buildkit" else "generic"),
+            "--build-arg",
+            "user_uid={}".format(os.getuid()),
```
JosephusPaye, line 78: [`os.getuid()`](https://docs.python.org/3/library/os.html#os.geteuid) is not available on Windows (natively). Though I suppose we can start assuming WSL?

TrentHouliston: If you want to add some code you could just set it up so that on windows which is not wsl, you can just ignore this whole build argument (or set it to the default of 1000)

JosephusPaye: Yeah, plus there are other things in the code that won't work without WSL. We can fix those all in the future if necessary.

JosephusPaye (approved):

PR #340, opened 2020-11-11
Always ensure latest archlinux-keyring package is installed

Bidski:


docker/Dockerfile
```diff
@@ -22,11 +22,16 @@ RUN pacman -Syyuu --noconfirm --needed --overwrite \*
 # Add a script that installs packages
 COPY usr/local/bin/install-package /usr/local/bin/install-package
 
+# Install latest archlinux-keyring package so we have the most up-to-date keys
+RUN install-package wget
+RUN wget https://archive.archlinux.org/repos/last/core/os/x86_64/archlinux-keyring-20201028-1-any.pkg.tar.zst \
```
JosephusPaye, line 27: Is this URL something we'll have to update regularly? Or is there something like a /latest.pkg.tar.zst we could use?

JosephusPaye (commented):
The branch works for me with Docker on Windows. Just one question but LGTM otherwise.

docker/Dockerfile
Bidski: We might be able to use 
```
RUN wget https://www.archlinux.org/packages/core/any/archlinux-keyring/download/ -O /var/tmp/archlinux-keyring.pkg.tar.zst
RUN pacman -U /var/tmp/archlinux-keyring.pkg.tar.zst --noconfirm
```

Haven't tested this though

JosephusPaye: I added it and tested locally and it works

KipHamiltons: Just tested this. was able to build on my machine without issue

JosephusPaye:
This is failing due to [an error downloading from sourceforge](https://buildkite.com/nubots/nubots/builds/891#ae90b562-bac7-467c-8789-47324c8eef51/6-60):

screenshot of error, see the link above for text version

KipHamiltons (approved):
LGTM. 

(@JosephusPaye looks like the arma source website is back up I suppose. it built and passed)

JosephusPaye:
Yeah it finally built.

JosephusPaye (approved):

PR #341, opened 2020-11-18
Allow passing container network as arg to `./b run`

JosephusPaye:
Pairs with https://github.com/NUbots/NUsight2/pull/335.

This allows for connecting the container to a custom docker network, for communication with NUsight when it's run in a container.

A custom network for the containers can be created like so:

```sh
docker network create nuclearnet
```

And usage of `./b run` can be adjusted like so:

```sh
./b run keyboardwalkfake --network nuclearnet
```

This is the only way to get the two to talk on Windows when NUbots is running in a container: run NUsight in a container too, and connect both containers to the same Docker network.

The other options:

- `host` networking (the current default): [not available on Windows](https://docs.docker.com/network/network-tutorial-host/#prerequisites), though it works on Linux ~(and probably macOS too)~
- `bridge` networking: [also doesn't work on Windows](https://docs.docker.com/docker-for-windows/networking/#there-is-no-docker0-bridge-on-windows)

### DIFF
--- a/tools/dockerise/__init__.py
+++ b/tools/dockerise/__init__.py
@@ -148,9 +148,7 @@ def run_on_docker(func):
                 nargs="?",
                 help="The image to use for the docker container",
             )
-            command.add_argument(
-                "--network", dest="network", help="Run the container on the specified docker network"
-            )
+            command.add_argument("--network", dest="network", help="Run the container on the specified docker network")
 
             func(command)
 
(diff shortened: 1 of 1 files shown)


ysims (approved):
Tested and it works. Python code looks fine, can't see any way it'd mess anything up.

TrentHouliston (approved):
LGTM although for the record host doesn't really work on OSX either as it also runs inside a hypervisor

PR #345, opened 2021-01-11
Allow for piping output of `./b` commands

JosephusPaye:
This enables things like:

- `./b format > format-report.txt` - save the formatting report to a file
- `./b format | grep some_file.cpp -A 5` - see the results of formatting a specific file


KipHamiltons (approved):
Beautiful!

Interestingly, when I tried `./b build > build.log`, it piped compiler errors to the log too, which I would think should be printed to stderr. Seems like docker might divert the pipe to stdout for my terminal? Interested to hear what you think is happening

JosephusPaye:
Yeah, I think it currently directs both stdout and stderr to stdout on the host.

PR #347, opened 2021-01-14
Updates arch version and fixes issues that added

KipHamiltons:
Many builds of LLVM later, it's done. Not exactly sure what was different in this build which worked, but if the shoe fits...


docker/Dockerfile
```diff
@@ -10,8 +10,8 @@
 ### The date in the Server command must either match, or be more recent than,
 ### the date in the FROM command
 ### Date format is ISO year/month/day
-FROM archlinux:20200505
-RUN echo "Server=https://archive.archlinux.org/repos/2020/05/05/\$repo/os/\$arch" | tee /etc/pacman.d/mirrorlist
+FROM archlinux:base
+RUN echo "Server=https://archive.archlinux.org/repos/month/\$repo/os/\$arch" | tee /etc/pacman.d/mirrorlist
```
TrentHouliston, line 14: What's changed here? I would recommend keeping a date versioned release as otherwise, we will have the same problem in the future that caused us to have to add this. One day arch will update base and it'll break something and fixing it will be way more urgent

docker/Dockerfile
```diff
@@ -22,22 +22,19 @@ RUN pacman -Syyuu --noconfirm --needed --overwrite \*
...
     sudo \
     python \
     python-pip \
     base-devel \
     ninja \
     cmake \
     meson \
-    git
+    git \
+    ncurses
```
TrentHouliston, line 37: Why do we need to install ncurses here?

docker/Dockerfile
```diff
@@ -93,46 +90,67 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr/local
 RUN install-from-source https://www.zlib.net/zlib-1.2.11.tar.gz
 
 # LLVM and Clang
-RUN BUILD_FILE_DIR='llvm' install-cmake-from-source \
-    https://github.com/llvm/llvm-project/archive/llvmorg-9.0.1.tar.gz \
-    -DLLVM_ENABLE_PROJECTS="clang"
+RUN BUILD_FILE_DIR='llvm' install-cmake-from-source https://github.com/llvm/llvm-project/archive/llvmorg-11.0.1.tar.gz \
+    -DLLVM_ENABLE_PROJECTS="clang" \
```
TrentHouliston, line 94: You should put clang-tools-extra here so you get clang-tidy
```suggestion
    -DLLVM_ENABLE_PROJECTS="clang;clang-tools-extra" \
```

docker/Dockerfile
```diff
@@ -93,46 +90,67 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr/local
...
+RUN install-cmake-from-source https://github.com/intel/media-driver/archive/intel-media-20.4.5.tar.gz \
     -DINSTALL_DRIVER_SYSCONF=OFF \
     -DLIBVA_DRIVERS_PATH="/usr/local/lib/dri" \
     -DMEDIA_RUN_TEST_SUITE=OFF \
     -DBUILD_TESTING=OFF \
     -DBUILD_KERNELS=ON \
     -DENABLE_KERNELS=ON \
     -DENABLE_NONFREE_KERNELS=ON \
-    -DBUILD_TYPE=Release
+    -DCMAKE_INSTALL_PREFIX=/usr \
```
TrentHouliston, line 147: Why does this have to go in `/usr` this will cause problems when you try to install onto a robot

docker/Dockerfile
```diff
@@ -93,46 +90,67 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr/local
...
+    -DCMAKE_INSTALL_LIBDIR=lib
 
 # Install intel compute runtime (OpenCL implementation)
 COPY --chown=root:root etc/OpenCL/vendors/intel.icd /etc/OpenCL/vendors/intel.icd
 COPY usr/local/package/intel-compute-runtime.sh /usr/local/package/intel-compute-runtime.sh
-RUN /usr/local/package/intel-compute-runtime.sh https://github.com/intel/compute-runtime/archive/20.01.15264.tar.gz \
-    -DNEO_DRIVER_VERSION=19.40.14409 \
-    -DSKIP_ALL_ULT=ON \
-    -DSKIP_UNIT_TESTS=ON \
-    -DIGDRCL__IGC_LIBRARY_PATH="/usr/local/lib"
```
TrentHouliston, line 135: I would keep these out here rather than hiding them in the script.
Things that get hidden in the script lose visibility compared to updating the install-from-source itself.
You make a custom script when there is a weird nonstandard way of installing things in terms of steps.

docker/Dockerfile
```diff
@@ -212,6 +230,7 @@ RUN install-from-source https://gitlab.gnome.org/GNOME/glib/-/archive/2.63.3/gli
     -Ddefault_library=both \
     -Dinternal_pcre=true \
     && cp /usr/local/lib/glib-2.0/include/glibconfig.h /usr/local/include/glibconfig.h
+RUN install-package libusb
```
TrentHouliston, line 233: Why are we installing a package for this rather than building it.
In general, we would only want to install a package if it's used TO build something. Not as a dependency of a library itself.

docker/usr/local/package/intel-graphics-compiler.sh
```diff
@@ -0,0 +1,60 @@
...
+    -DCMAKE_INSTALL_LIBDIR=lib \
+    -DIGC_OPTION__ARCHITECTURE_TARGET='Linux64' \
+    -DIGC_PREFERRED_LLVM_VERSION='11.0.0' \
+    -DVC_INTRINSICS_SRC="../../vc-intrinsics" \
+    -DINSTALL_SPIRVDLL=0 \
+    -DINSTALL_GENX_IR=ON \
+    -Wno-dev
+
+# Build the package
+ninja -j `nproc`
```
TrentHouliston, line 54: Never use `command` style when executing commands. use $(command) instead
```suggestion
ninja -j $(nproc)
```

docker/usr/local/package/opencl-clang.sh
```diff
@@ -40,14 +40,14 @@ mkdir -p build
 cd build
 
 # Configure using cmake
-cmake .. \
+cmake .. "$@" \
```
TrentHouliston, line 43: I would leave the standard arguments out in the docker file and continue to pass them in like this.
Hiding things in the script files makes it harder to read

TrentHouliston (changes requested):

docker/Dockerfile
KipHamiltons: I added that because [arch has it listed as a dependency of intel-graphics-compiler](https://archlinux.org/packages/community/x86_64/intel-graphics-compiler/)

docker/Dockerfile
KipHamiltons: I was planning to add that in a future PR where I pull in the proper `.clang-tidy` file and make the changes necessary for it to run

docker/Dockerfile
KipHamiltons: In this example I copied over the flags from the arch PKGBUILD verbatim. I can try it with `/usr/local` or something else if you like?

docker/usr/local/package/intel-graphics-compiler.sh
KipHamiltons: is that a reason for that? I have found it common in build scripts that I have seen

docker/Dockerfile
TrentHouliston: Everything we build that needs to go on the robot should be in /usr/local

docker/Dockerfile
TrentHouliston: Fair enough

docker/Dockerfile
TrentHouliston: If it's a dependency then we should build it ourselves. Anything that'll actually run on the robot should be built by us.

docker/usr/local/package/intel-graphics-compiler.sh
TrentHouliston: @Bidski You know more about this stuff than I do, why do we use $() over ``

docker/Dockerfile
KipHamiltons: Roger that. I had a stupid reason, but on reflection, you're right.

docker/usr/local/package/intel-graphics-compiler.sh
Bidski: [See this](http://mywiki.wooledge.org/BashFAQ/082)

docker/Dockerfile
Bidski: Just to kick the horse while it is down. We should always maintain a specific date, this is mainly to enable us to keep the robots in sync with the docker images

docker/Dockerfile
```diff
@@ -93,46 +90,67 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr/local
 RUN install-from-source https://www.zlib.net/zlib-1.2.11.tar.gz
 
 # LLVM and Clang
-RUN BUILD_FILE_DIR='llvm' install-cmake-from-source \
-    https://github.com/llvm/llvm-project/archive/llvmorg-9.0.1.tar.gz \
-    -DLLVM_ENABLE_PROJECTS="clang"
+RUN BUILD_FILE_DIR='llvm' install-cmake-from-source https://github.com/llvm/llvm-project/archive/llvmorg-11.0.1.tar.gz \
+    -DLLVM_ENABLE_PROJECTS="clang" \
+    -DCMAKE_INSTALL_PREFIX=/usr \
```
Bidski, line 95: Always install to `/usr/local`

Also, `install-from-source` already sets `CMAKE_INSTALL_PREFIX` for you

docker/Dockerfile
```diff
@@ -93,46 +90,67 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr/local
...
+    -DFFI_LIBRARY_DIR="$(pkg-config --variable=libdir libffi)" \
+    -DLLVM_ENABLE_SPHINX=OFF \
+    -DLLVM_BUILD_LLVM_DYLIB=ON \
+    -DLLVM_LINK_LLVM_DYLIB=ON \
+    -DLLVM_BINUTILS_INCDIR=/usr/include
+
+# SPIRV LLVM Translator
+RUN install-cmake-from-source https://github.com/KhronosGroup/SPIRV-LLVM-Translator/archive/v11.0.0.tar.gz \
+    -DBUILD_SHARED_LIBS=ON \
+    -DCMAKE_INSTALL_PREFIX=/usr \
```
Bidski, line 109: Always install to `/usr/local`

Also, `install-from-source` already sets `CMAKE_INSTALL_PREFIX` for you

docker/Dockerfile
```diff
@@ -93,46 +90,67 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr/local
...
 
-# Libpci access and libdrm for communicating with graphics hardware
+# intel graphics compiler
+COPY usr/local/package/intel-graphics-compiler.sh /usr/local/package/intel-graphics-compiler.sh
+RUN /usr/local/package/intel-graphics-compiler.sh https://github.com/intel/intel-graphics-compiler/archive/igc-1.0.5435.tar.gz
+
+# intel compute runtime dependencies
+# gmmlib
+RUN install-cmake-from-source https://github.com/intel/gmmlib/archive/intel-gmmlib-20.4.1.tar.gz \
+    -DCMAKE_INSTALL_PREFIX=/usr \
```
Bidski, line 126: Always install to `/usr/local`

Also, `install-from-source` already sets `CMAKE_INSTALL_PREFIX` for you

docker/Dockerfile
Bidski: Also, `install-from-source` already sets `CMAKE_INSTALL_PREFIX` for you

docker/usr/local/package/intel-graphics-compiler.sh
Bidski: TL;DR `` are legacy and can do some weird things, $() is the new-fancy and does less weird things

docker/usr/local/package/opencl-clang.sh
```diff
@@ -40,14 +40,14 @@ mkdir -p build
...
     -DCMAKE_BUILD_TYPE="Release" \
     -DCMAKE_TOOLCHAIN_FILE="${PREFIX}/toolchain.cmake" \
     -DLINUX_RESOURCE_LINKER_COMMAND:PATH="${PREFIX}/bin/linux_resource_linker" \
+	-DCMAKE_INSTALL_PREFIX=/usr \
```
Bidski, line 47: `/usr/local`

docker/usr/local/package/opencl-clang.sh
```diff
@@ -40,14 +40,14 @@ mkdir -p build
...
     -DLLVMSPIRV_INCLUDED_IN_LLVM=OFF \
-    -DSPIRV_TRANSLATOR_DIR=/usr/local \
+    -DSPIRV_TRANSLATOR_DIR=/usr \
```
Bidski, line 49: `/usr/local`

Bidski (changes requested):

docker/Dockerfile
KipHamiltons: `install-from-source` didn't set it for CMake builds, but I have added that because you make a good point that it really should.

docker/usr/local/package/intel-graphics-compiler.sh
KipHamiltons: spicy! I'll keep that in mind

docker/Dockerfile
TrentHouliston: Sure it does, via the toolchain file https://github.com/NUbots/NUbots/blob/master/docker/usr/local/toolchain/generate_toolchains.py#L34

Bidski: My bad, the [toolchain file](https://github.com/NUbots/NUbots/blob/master/docker/usr/local/toolchain/generate_toolchains.py#L34) sets it

KipHamiltons: ahh my bad too. I forgot about those args

docker/usr/local/bin/install-from-source
```diff
@@ -154,7 +154,7 @@ elif [ "${METHOD}" = 'install-cmake-from-source' ]; then
         "$@" \
         -DCMAKE_BUILD_TYPE="Release" \
         -DCMAKE_TOOLCHAIN_FILE="${PREFIX}/toolchain.cmake" \
-        -DCMAKE_INSTALL_PREFIX"=/usr/local" \
+        -DCMAKE_INSTALL_PREFIX="${PREFIX}" \
```
TrentHouliston, line 157: This shouldn't be needed since it's in the toolchain file

TrentHouliston (changes requested):

docker/Dockerfile
```diff
@@ -93,46 +89,76 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr/local
...
+    -DLLVM_LINK_LLVM_DYLIB=ON \
+    -DLLVM_BINUTILS_INCDIR=/usr/include
+
+# SPIRV LLVM Translator
+RUN install-cmake-from-source https://github.com/KhronosGroup/SPIRV-LLVM-Translator/archive/v11.0.0.tar.gz \
+    -DBUILD_SHARED_LIBS=ON \
+    -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+    -DCMAKE_SKIP_RPATH=ON \
+    -DLLVM_INCLUDE_TESTS=ON \
+    -DLLVM_EXTERNAL_LIT=/usr/bin/lit
```
Bidski, line 110: `/usr/local/bin/lit`? This comes from LLVM right?

Bidski (changes requested):

docker/Dockerfile
KipHamiltons: That's one of the flags from the arch PKGBUILD which I grabbed. Now that I think I have a fix for the intel compute runtime building issues I found before, should I remove those flags which I added which weren't necessary? Other examples include all of these 4 flags in this snippet, and things like `-DLLVM_BUILD_LLVM_DYLIB=ON`

If not, in this specific case, I am not sure. It would seem to me that we could get by without that flag or the one above it which includes tests.

TrentHouliston: I would remove it, and probably the test one too. Building llvm is painful enough without building the tests

Bidski: If you think they are unnecessary then remove them.

Including tests is something we almost always turn off to speed up the builds. 

We need to be careful with our builds, specifically the NUC builds. You need to assume that the NUC build is not capable of running on the host machine. So you should assume that running `lit` from a LLVM that was built for the NUC is going to fail. In this case, we should be installing LLVM/Clang from pacman and using `-DLLVM_EXTERNAL_LIT=/usr/bin/lit` to ensure that it will work

KipHamiltons: I have removed those two and all the others I added which weren't required or didn't remove stuff we didn't use

KipHamiltons: @Bidski could you clarify that a little please? when exactly should we install llvm/clang with pacman?

Bidski (dismissed):
Nothing seems obviously wrong

docker/Dockerfile
Bidski: There should be a system installation of LLVM/Clang, so that when you build SPIRV-LLVM-Translator you can direct it to link against the LLVM/Clang that we built, but use `lit` (and other binaries) while it is building.

This needs to be the case for any library that needs to run a binary that it compiles as part of its build process (this is the massive headache that is cross-compilation). For instance, in `opencl-clang.sh` we first build `linux_resource_linker` without any special compiler flags so that `opencl-clang` can safely run that binary while it is building.

KipHamiltons: Jeeez  that sounds like an absolute headache! I'll note that for a future lil' project for the build team :wink: 
Are there other packages which could use similar treatment?

Bidski: This is mostly a trial and error thing, unless you study the build process of every library we build.

We have covered most of the libraries we install already, but we may have missed some. At this point, I would just wait until some poor unlucky soul gets a bizarre build issue :P

KipHamiltons: hahahaha roger that

TrentHouliston (commented):
Get it building first otherwise, my approval will just get reset

KipHamiltons:
I was going to do the formatting in another PR :crying_cat_face:

TrentHouliston (approved):

PR #349, opened 2021-01-25
Fix the SmoothSpline unit test.

ysims:
The SmoothSpline unit test was incorrect because time values were not consecutive.

The error was not picked up because the test failed rarely. It failed rarely because the allowable error was (incorrectly) too large. A missing negative sign gave an error threshold of 1,000,000 instead of 0.000001 :grimacing: 

This PR fixes both the error threshold and the issue with consecutive time values.

### DIFF
--- a/shared/tests/utility/motion/TestSplines.cpp
+++ b/shared/tests/utility/motion/TestSplines.cpp
@@ -236,13 +236,23 @@ TEST_CASE("Test Smooth Spline", "[utility][motion][splines][SmoothSpline]") {
         size_t noPoints = rand() % 6;
         std::vector<Eigen::Vector4d> points;
 
+        // This value will keep track of the last t value - we want them to be consecutive, so we will check against
+        // this. Start it negative since the generated t values will not be negative
+        double point_t = -1;
+
         // Add our points to the spline
         for (size_t i = 0; i < noPoints; i++) {
             Eigen::Vector4d point = Eigen::Vector4d::Random();  // random 4d array of numbers between -1 and 1
             point *= 10;                                        // values are between -10 and 10
-            point.x() += (i + 1) * 10;  // this will ensure no t values are negative and they are always consecutive
+            point[0] += 10;                                     // this will ensure no t values are negative
+
+            // If the last t and this t are not consecutive, add them together... plus 1 incase point_t is 0
+            point[0] = (point[0] < point_t) ? (point[0] + point_t + 1) : point[0];
+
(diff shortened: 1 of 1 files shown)


shared/tests/utility/motion/TestSplines.cpp
```diff
@@ -220,7 +221,7 @@ TEST_CASE("Test Polynom", "[utility][motion][splines][Polynom]") {
     }
 }
 
-const float ERROR = 1e6;  // Error for testing floats
+const float ERROR = 1e-6;  // Error for testing floats
```
Bidski, line 224:
```suggestion
const float ERROR = 1e-6f;  // Error for testing floats
```

shared/tests/utility/motion/TestSplines.cpp
```diff
@@ -229,20 +230,30 @@ TEST_CASE("Test Smooth Spline", "[utility][motion][splines][SmoothSpline]") {
...
+        // This value will keep track of the last t value - we want them to be consecutive, so we will check against
+        // this. Start it negative since the generated t values will not be negative
+        double point_t = -1;
+
         // Add our points to the spline
         for (size_t i = 0; i < noPoints; i++) {
             Eigen::Vector4d point = Eigen::Vector4d::Random();  // random 4d array of numbers between -1 and 1
             point *= 10;                                        // values are between -10 and 10
-            point.x() += (i + 1) * 10;  // this will ensure no t values are negative and they are always consecutive
+            point[0] += 10;                                     // this will ensure no t values are negative
```
Bidski, line 248: `std::abs()`? If all you care about is it being positive?

shared/tests/utility/motion/TestSplines.cpp
```diff
@@ -229,20 +230,30 @@ TEST_CASE("Test Smooth Spline", "[utility][motion][splines][SmoothSpline]") {
...
+    for (int j = 0; j < 100; j++) {
         spline.reset();
 
         // Get a random number of points for this spline
         size_t noPoints = rand() % 6;
         std::vector<Eigen::Vector4d> points;
 
+        // This value will keep track of the last t value - we want them to be consecutive, so we will check against
+        // this. Start it negative since the generated t values will not be negative
+        double point_t = -1;
```
Bidski, line 242:
```suggestion
        double point_t = -1.0;
```

Bidski (changes requested):

Bidski (approved):
LGTM

PR #352, opened 2021-02-03
Fix the formatting in NUsight2 files added when making a monorepo

TrentHouliston:


KipHamiltons:
How come this has changed the formatting for so many non-nusight files?
It looks like buildkite's check doesn't like those changes

TrentHouliston:
Seems I forgot to update my docker image before running format

JosephusPaye (approved):
Nothing looks out of place

PR #353, opened 2021-02-03
Swap from buildkite to github actions

TrentHouliston:
Currently using the online runners. We can swap to self-hosted runners before we merge


JosephusPaye:
I think it's fair to say we're sticking with Buildkite for now.

KipHamiltons:
This PR is ready for review once again. Here's a list of things I have just added:

- Removed the remaining buildkite files/hooks
- Added action to each job which cancels previous running jobs if a new one supersedes it (styfle/cancel-workflow-action@0.9.1)
- Consolidated the dependabot configs for nusight and gh actions
- Added action for caching the yarn packages for nusight
- Use node14 container (the one we officially support) instead of node12
- Disable CI_BUILD and ENABLE_CLANG_TIDY cmake flags, because we aren't ready for them yet :'(

Did I miss anything we wanted to change/add?

.github/workflows/nubots.yaml
```diff
@@ -0,0 +1,179 @@
+# Continuous Integration tests
+name: NUbots CI
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
+  push:
+    branches: [main]
```
Bidski, line 8: Do we need this given we cant even push to main?

.github/workflows/nubots.yaml
```diff
@@ -0,0 +1,179 @@
...
+        name: Output the created image
+        run: echo "::set-output name=image::nubots/nubots:${{ hashFiles('docker/**') }}"
+
+  # Build the codebase
+  build_nubots:
+    name: "Build and test NUbots"
+
+    # The type of runner that the job will run on
+    runs-on: ubuntu-20.04
+    needs: build_docker
```
Bidski, line 69: Does this mean we will build the generic image and push it to dockerhub before we build the code?

.github/workflows/nubots.yaml
```diff
@@ -0,0 +1,179 @@
...
+
+      - name: Configure the code
+        run: |
+          mkdir -p build
+          cd build
+          cmake .. -GNinja \
+                   -DCMAKE_BUILD_TYPE=Release \
+                   -DCMAKE_TOOLCHAIN_FILE=/usr/local/toolchain.cmake \
+                   -DBUILD_TESTS=ON \
+                   -DCI_BUILD=ON
```
Bidski, line 94: Why are we not using the b script?

CMakeLists.txt
```diff
@@ -42,9 +42,11 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
 endif()
 
-# Default not to run the clang-tidy checks, default to whatever our CI_BUILD is
+# Default not to run the clang-tidy checks
```
Bidski, line 45: Why is this file changed?

Bidski (dismissed):
Not entirely sure I am qualified to review this PR, but here you go

.github/workflows/nubots.yaml
KipHamiltons: My assumption was that it does a build after PRs are merged. 
They do it in the docs [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on) so I think that's correct??

.github/workflows/nubots.yaml
KipHamiltons: I believe so. @TrentHouliston was telling me that each PR will indeed have its own image pushed to dockerhub. I think that happens at the end of `build_docker` stage.

.github/workflows/nubots.yaml
KipHamiltons: Good question. I'll try it with the `b` script.

.github/workflows/nubots.yaml
Y0sh1dk: Why push a image on PR? is that so later jobs can use the pushed image?

CMakeLists.txt
KipHamiltons: If we enable CI_BUILD without changing the file, then the builds will fail until all warnings are fixed and clang-tidy has been integrated with its warnings fixed. Rather than not using the `CI_BUILD` variable in a CI build, I have changed this file, noting that we should run clang-tidy and -Werror once those issues have been addressed properly. #431 will add clang-tidy to the CI_BUILD once the clang-tidy stuff has been fixed. We can do something similar for the warnings once #465 is merged.

.github/workflows/nubots.yaml
Bidski: That doesn't answer the question of main being protected?

.github/workflows/nubots.yaml
KipHamiltons: That worked. Just had to move the build directory :+1:

.github/workflows/nubots.yaml
KipHamiltons: Yessir

Bidski: If we want to keep this behaviour then we can add a step to this PR to delete those pushed images

CMakeLists.txt
Bidski: I dont think this is the right way to do this

.github/workflows/nubots.yaml
KipHamiltons: Did you want to add that to this PR rather than follow up? There will be follow-up PRs for not building nusight or nubots if either hasn't changed, and trying to cache C++ builds.
Personally, I am keen to avoid scope creep on this PR, but if you reckon it's worth waiting on this PR, I'll have a crack at doing it

.github/workflows/nubots.yaml
```diff
@@ -0,0 +1,171 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+
+      # Build the docker image
+      - name: 🐳 Build the docker image 🐳
+        uses: docker/build-push-action@v2
+        with:
+          pull: true
+          tags: "nubots/nubots:${{ hashFiles('docker/**') }}"
```
Y0sh1dk, line 50: Can we just use the branch name here instead of a hash?
```suggestion
          tags: "nubots/nubots:${{ GITHUB_REF#refs/heads/ }}"
```

Y0sh1dk:
This seems to be missing steps for building and pushing our NUC and generic images. This would need to be done in a separate workflow that only triggers on changes to master.

.github/workflows/nubots.yaml
TrentHouliston: Even better, use the PR number. But the idea behind this was that you only choose a different image if the actual files in the docker folder change (not just if it's a different branch). In theory two branches could have the same docker files and therefore share an image.

The biggest problem with any of this is cleaning up the images once we are done with the PR

TrentHouliston:
> This seems to be missing steps for building and pushing our NUC and generic images. This would need to be done in a separate workflow that only triggers on changes to master.

A separate `Images.yaml` workflow which builds each of the images and uploads them would be good.

.github/workflows/nubots.yaml
Y0sh1dk: Ah ok makes sense, yeah PR number would be better. With the hash I dont think different PR's with the same dockerfiles could share images because of `cache-from: type=registry,ref=nubots/nubots:generic` ?

Y0sh1dk: Actually nvm, It would cache the common layers between the generic and hash image because of the `pull: true` I believe.

KipHamiltons: I have done something similar. It probably won't work though :laughing:

CMakeLists.txt
KipHamiltons: What do you propose?

.github/workflows/nubots.yaml
KipHamiltons: Main is pushed to when PRs are merged to main, is my understanding. The protection just means that that's the only time main is pushed to

.github/workflows/nubots.yaml
TrentHouliston: That is the idea, that it will always use the common layers in the nubots:generic (so if you don't change the image like 99% of PRs it'll just entirely cache)

.github/workflows/images.yaml
```diff
@@ -0,0 +1,84 @@
...
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push events for the main branch
+  push:
+    branches: [main]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # Build the nuc8i7beh image
+  build_nuc8i7beh:
```
TrentHouliston, line 13: Use a build with variables here rather than splitting up like this. All images are the same with a different tag

.github/workflows/images.yaml
```diff
@@ -0,0 +1,84 @@
+# NUC image creation and push to DockerHub
```
Y0sh1dk, line 1: Will also need to build and push `generic` in here.

KipHamiltons: True

.github/workflows/images.yaml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+
+      # Build the docker image
+      - name: 🐳 Build the docker image 🐳
+        uses: docker/build-push-action@v2
+        with:
+          pull: true
+          tags: "nubots/nubots:${{ matrix.platform }}"
```
Y0sh1dk, line 44: Similar here, It might be nice to read in all references to our docker registry from repository environment/secrets instead of having it hardcoded in the workflow. This would be more relevent for when we add a RoboCup image workflow as we could easily change where the image is pushed by changing the repository environment variable.

.github/workflows/images.yaml
```diff
@@ -0,0 +1,51 @@
...
+    branches: [main]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # Build and push all images
+  build_and_push:
+    name: "Build and Push Docker Images"
+
+    # The type of runner that the job will run on
+    runs-on: ubuntu-20.04
```
Y0sh1dk, line 17: Reading in the `runs-on` parameter from repository environment/secrets would allow us to switch from hosted to self-hosted runners and vice versa easily by just changing the value in our repo settings. Might be handy.

CMakeLists.txt
Bidski: dont edit this file and get the other PRs merged in?

.github/workflows/nubots.yaml
TrentHouliston: This runs whenever main is updated, A push to main happens when you merge a PR into main

.github/workflows/images.yaml
TrentHouliston: You really should try to avoid having a situation where an environment outside of the code can influence the building success of the code.

It's better to have this configuration in the code here so that if changes are needed in order to make the code build, they go through the code review process. The reason that things like the docker password aren't in here is you really don't want to commit the passwords to the code.

.github/workflows/images.yaml
TrentHouliston: I would consider the RoboCup image (if you mean automatically upload the current RoboCup image to the competition during the competition time) to be an entirely separate workflow from this. Lots of very different steps.

That said, I think if you are in a contest it would be very cool to have a branch that automatically uploads the image to the competition. That way you will never again forget to push up the contest, it'll always be whatever is on that branch

.github/workflows/images.yaml
Y0sh1dk: Yeah ok makes sense, I was just thinking of a situation where we would suddenly want to switch where all our runners are running (Uni loses net/power etc), but I guess merging a PR into main, and merging main into whatever other branches is the better solution.

.github/workflows/images.yaml
Y0sh1dk: Thats the plan :+1: , and yeah definetly a separate workflow. I guess its better for changes to the registry address to go through a review process rather than just changing a repo variable.

.github/workflows/nubots.yaml
```diff
@@ -0,0 +1,171 @@
+# Continuous Integration tests
+name: NUbots CI
+
+# Controls when the action will run.
+on:
```
CMurtagh-LGTM, line 5: Could we break this up into [multiple files](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run) and use [path matching](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths)? So we don't have to build the docker image when we do a change to a cpp file, or build the codebase when we change a b tool. Probably for a future PR.

CMurtagh-LGTM: We could even send a [slack message](https://github.com/archive/github-actions-slack) on push to master when the dockerfile is changed. :thinking:

.github/workflows/images.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+    strategy:
+      matrix:
+        platform: ["generic", "nuc8i7beh", nuc7i7bnh]
```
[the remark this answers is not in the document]
KipHamiltons: Well spotted. thanks

.github/workflows/nubots.yaml
KipHamiltons: Great ideas. I'll make an issue for github actions next steps, like the documentation generation next steps one

CMurtagh-LGTM (commented):
LGTM, other than the cmake lists stuff.

KipHamiltons:
@Bidski @CMurtagh-LGTM I have reverted the CMake changes on this PR. There is still a loooot to do for clang-tidy warnings before they're resolved. 
@TrentHouliston, if it bothers you enough that we don't have the `CI_BUILD` variable enabled on CI, I'll make a follow-up PR which just deals with restoring those semantics

TrentHouliston:
It does bother me that the CI PR doesn't enable the CI flag. It's something that will get easily missed in the future. My option is that you should have the flag in this PR since this PR is about CI. It's not GitHub actions fault it doesn't build with that flag. It's not a bug in this code. It's a bug in the code that's building. Now we have a known bug in the CI and a known bug in the cmake build, rather than just one in cmake

.github/workflows/nubots.yaml
Bidski: It seems like a natural action that should happen to he honest. This way, from the moment this PR gets merged, we wont have any "mess" left that needs a manual clean up later.

KipHamiltons: Added the functionality in `dockerhub-cleanup.yaml` workflow :+1:

.github/workflows/images.yaml
```diff
@@ -0,0 +1,51 @@
...
+      # Build the docker image
+      - name: 🐳 Build the docker image 🐳
+        uses: docker/build-push-action@v2.7.0
+        with:
+          pull: true
+          tags: "nubots/nubots:${{ matrix.platform }}"
+          file: docker/Dockerfile
+          context: docker
+          build-args: |
+            platform=${{ matrix.platform }}
```
CMurtagh-LGTM, line 48: I think we need `BUILDKIT_INLINE_CACHE=1` to use the buildkit inline cache exporter, as the docker one is virtually nonexistent.

.github/workflows/images.yaml
```diff
@@ -48,4 +48,7 @@ jobs:
             platform=${{ matrix.platform }}
             BUILDKIT_INLINE_CACHE=1
           push: true
-          cache-from: type=registry,ref=nubots/nubots:${{ matrix.platform }}
+          cache-from: |
+            type=registry,ref=nubots/nubots:pull-request-${{ github.event.number }}
```
TrentHouliston, line 52: The `images.yaml` workflow doesn't run on a PR so this is wrong isn't it?.

.github/workflows/images.yaml
```diff
@@ -48,4 +48,7 @@ jobs:
...
+            type=registry,ref=nubots/nubots:generic
```
TrentHouliston, line 53: You should be caching from the type of this image you're building, not generic all the time. The nuc7i7bnh won't have much in common with the generic image, so caching is pointless. Cache from the correct image

.github/workflows/images.yaml
```diff
@@ -46,6 +46,7 @@ jobs:
           context: docker
           build-args: |
             platform=${{ matrix.platform }}
+            BUILDKIT_INLINE_CACHE=1
```
TrentHouliston, line 49: This is not needed and shouldn't make a difference since we are using buildx to build. Buildkit is used always by buildx

.github/workflows/nubots.yaml
```diff
@@ -52,11 +52,9 @@ jobs:
           context: docker
           build-args: |
             platform=generic
+            BUILDKIT_INLINE_CACHE=1
```
TrentHouliston, line 55: Remove this, it's not needed since we use buildx

.github/workflows/nubots.yaml
```diff
@@ -52,11 +52,9 @@ jobs:
...
           push: true
-          cache-from: |
-            type=registry,ref=nubots/nubots:pull-request-${{ github.event.number }}
-            type=registry,ref=nubots/nubots:generic
-            type=inline
+          cache-from: type=inline
```
TrentHouliston: I'm pretty sure you should keep caching from the dockerhub image here

.github/workflows/images.yaml
CMurtagh-LGTM: Buildx uses the docker cache by default, which leaves lots to be desired. This enables the buildkit cache, which is much better.

CMurtagh-LGTM (approved):
LGTM

PR #354, opened 2021-02-05
Swap to using Indexed NBS files and refactor the NBS tools to make them simpler

TrentHouliston:


ElderNoSpace (approved):
LGTM, but of course it does, most of the code is mine :P.

PR #358, opened 2021-02-05
Change how signals are handled through docker so ./b scripts can be piped

TrentHouliston:


JosephusPaye:
I already ported this across in https://github.com/NUbots/NUbots/pull/345.

PR #360, opened 2021-02-05
Make the module generate make hpp files and have a log level setting by default

TrentHouliston:


ElderNoSpace (approved):
LGTM, but having the whole log_level thing is a bit clunky.

PR #361, opened 2021-02-05
Add a tool to the b script to help with some yarn tasks

TrentHouliston:


tools/utility/yarn.py
```diff
@@ -0,0 +1,39 @@
...
+
+    print(
+        colored(
+            "Cannot find eslint, run `./b yarn` first if you need to format typescript files", "red", attrs=["bold"]
+        )
+    )
+    return None
+
+
+def find_package_json(project_root=".", package_name="NUsight"):
```
JosephusPaye, line 23: This function is probably doing unnecessary work. We can start by just checking that `nusight2/package.json` exists. It's unlikely that it won't be there.

KipHamiltons:
Could you please add information about using this command to NUbook?

It'd probably go in the [build system page](https://nubook.nubots.net/system/foundations/build-system) with the current taxonomy of NUbook, but it seems to me that there's definitely scope for a `b script` page, now that we are expanding it to do more non-build related stuff (tests, and now yarn stuff)

JosephusPaye:
Will do. And yes, it makes since to have a dedicated page for the b scripts now.

tools/utility/yarn.py
Bidski: That is literally the first thing this function does

tools/yarn.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
+
+import os
+
+import b
+from dockerise import WrapPty, run_on_docker
```
Bidski, line 6: Since we dont actually run this on docker
```suggestion
from dockerise import WrapPty
```

Bidski (changes requested):

tools/utility/yarn.py
JosephusPaye: You mean after this commit? [`cafae04` (#361)](https://github.com/NUbots/NUbots/pull/361/commits/cafae049a6dc94b80722c4605a218855d843076c) Or before?

Bidski: Oh, nevermind then :P

Bidski (dismissed):
Looks OK to me

KipHamiltons:
fix the formatting and merge this bad boy @JosephusPaye :sunglasses:

Bidski (approved):
LGTM

PR #363, opened 2021-02-05
Upgrade the camera calibration code for higher accuracy

TrentHouliston:
This upgrades the camera calibration code and archives higher accuracy in the intrinsics and extrinsics.
However it is much much slower than the current method and as is, will only run on a computer with significant power.
Up to you guys what you do with this one.


JosephusPaye:
Perhaps we can have both, and add this as a different tool.

PR #367, opened 2021-02-08
Optimise line projection shader

JosephusPaye:
Moves the work from the fragment shader to the vertex shader, projecting vertices in the vertex shader instead of checking line coincidence for every pixel in the fragment shader. This is ~likely~ reduces GPU computation by several orders of magnitude.


TrentHouliston (approved):
I like it

PR #368, opened 2021-02-15
Fix bug in proto_matrix.hpp where memory was not resized properly

TrentHouliston:
The object was reserved meaning that there was enough space to copy over, however, the size was not updated resulting in protobuf not serializing any bytes


JosephusPaye (approved):
LGTM

PR #369, opened 2021-02-16
Adds coordinate utility unit tests

Crafty15:
- Added unit tests for the coordinate functions 
- uses catch2's BDD style macros. 
- Plz guys it's been 2 years :rofl:


shared/tests/utility/math/docs/cartesianToSpherical.m
```diff
@@ -0,0 +1,17 @@
+function conversion = cartesianToSpherical(inputVec)
```
KipHamiltons, line 1: Ahh, there has been a miscommunication from my end! When I was saying we want to use matlab for our ground truth, what I meant was we should use the matlab builtin functions which do these. So instead of implementing a function which does the same thing on matlab, use their standard library versions [(here's this example)](https://au.mathworks.com/help/matlab/ref/sph2cart.html) for these functions.
Our reasoning is that matlab should certainly have rock solid implementations for these functions (because they're common functions), so we want to match _their_ implementation closely.

My bad for poorly communicating that to you!

shared/tests/utility/math/docs/genValsInRange.m
```diff
@@ -0,0 +1,6 @@
+function randVals = genValsInRange(min, max, total, cols)
+%This function generates random values in the given range
+%params: min = lower val, max = upper val, total = total numbers generated
+% cols = number of columns (use this for vectors etc)
+randVals = (max - min).*rand(total,cols) + min;
```
KipHamiltons, line 5: If we're not giving matlab a random seed, won't this produce different values each time? I think it would be good if this set the seed, or at least commented, noting the seed which produced the values used in the final tests.

shared/tests/utility/math/docs/README.txt
```diff
@@ -0,0 +1,14 @@
+The files in this folder contain the same input data used in TestCoordinates.cpp along with some matlab scripts that perform the same calculations found in coordinates.h
+The output from matlab is used to validate the coordinates.h calculations in the NUBOTS codebase.
+Test input data can be found in both matlab and c++ formats. The matlab format data is passed into the functions starting with calc_****.m to produce the output data.
+Test output data provided is in a cpp format.
+
+NOTE: During initial testing when values at the extremes of the data types domain, DBL_MAX and DBL_MIN, were found to cause inf return values when passed to the functions
```
KipHamiltons, line 6: did `DBL_MIN` give `inf` return values? wtf?

shared/tests/utility/math/docs/README.txt
```diff
@@ -0,0 +1,14 @@
...
+being tested. While normally some measures would have been put in place to handle this, it was determined that the performance impact would be to great, and that these values won't be
+possible during normal operation.
```
KipHamiltons, line 8: Noting here as well that it's not that the values aren't possible, it's more that we're assuming they won't happen.

shared/tests/utility/math/docs/README.txt
```diff
@@ -0,0 +1,14 @@
+The files in this folder contain the same input data used in TestCoordinates.cpp along with some matlab scripts that perform the same calculations found in coordinates.h
+The output from matlab is used to validate the coordinates.h calculations in the NUBOTS codebase.
+Test input data can be found in both matlab and c++ formats. The matlab format data is passed into the functions starting with calc_****.m to produce the output data.
+Test output data provided is in a cpp format.
+
+NOTE: During initial testing when values at the extremes of the data types domain, DBL_MAX and DBL_MIN, were found to cause inf return values when passed to the functions
+being tested. While normally some measures would have been put in place to handle this, it was determined that the performance impact would be to great, and that these values won't be
```
KipHamiltons, line 7: Name the `functions being tested` so that future code readers know what this refers to.

shared/tests/utility/math/docs/README.txt
```diff
@@ -0,0 +1,14 @@
...
+possible during normal operation.
+The maximum test values that were determined to be safe are:
+TEST_DBL_MAX    = 5.9667e+153;
+TEST_DBL_MIN    = 1.4853e-307;
+
+The values returned compared to the pre-calculated values should be within the error threshold:
+ERROR_THRESHOLD = 1e-6;
```
KipHamiltons, line 14: Didn't this change to being a percentage threshold, rather than an absolute threshold?

shared/tests/utility/math/docs/cartesianToSpherical.m
KipHamiltons: Also, the scripts that remain should be placed in a subfolder of this one I think

shared/tests/utility/math/docs/genValsInRange.m
```diff
@@ -0,0 +1,6 @@
+function randVals = genValsInRange(min, max, total, cols)
+%This function generates random values in the given range
+%params: min = lower val, max = upper val, total = total numbers generated
```
KipHamiltons, line 3: I would suggest putting the params on separate lines in this comment. On one line, it's a little harder for me to parse than I think it could be

KipHamiltons (changes requested):
First impressions. I haven't looked at the actual test file yet, but there is some stuff to look at otherwise.

A question as well: what's the thinking behind having the text files duplicating the test data?

shared/utility/math/coordinates.hpp
```diff
@@ -17,8 +17,8 @@
  * Copyright 2013 NUbots <nubots@nubots.net>
  */
 
-#ifndef UTILITY_MATH_COORDINATES_HPP
-#define UTILITY_MATH_COORDINATES_HPP
+#ifndef UTILITY_MATH_COORDINATES_H
+#define UTILITY_MATH_COORDINATES_H
```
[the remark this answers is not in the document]
Crafty15: I changed this to resolve a conflict on advice from Ysobel. The .h files were changed to .hpp while i was working on my branch.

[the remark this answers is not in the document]
ysims: This isn't a change as such, as it is something that didn't get fixed after the merge. You'll want to change it from H to HPP in all cases. Just an annoying thing from the H to HPP change, that is in all branches older than that commit.

shared/tests/utility/math/docs/genValsInRange.m
Crafty15: Hmmm, yes. I didn't think of using a seed. I was only accounting for if someone wanted to regen a new set. The extra copies of the test data were included in the docs folder in case of an accidental deletion or something. If we want it to make the same data each time i'm happy to work on it further though.

shared/tests/utility/math/docs/cartesianToSpherical.m
Crafty15: That makes sense! Shouldn't be a big change. I'll sort that out.

shared/tests/utility/math/docs/genValsInRange.m
KipHamiltons: Since you have it committed to git, there's no need to worry about accidental deletion. And, now that a commit with the data been pushed to the remote repo (and thus to the github servers), the information will live on as long as the repo does, accessible basically forever, through that commit.

shared/tests/utility/math/docs/calc_spherToCart.m
```diff
@@ -0,0 +1,9 @@
+%%
+% 
+%  PREFORMATTED
```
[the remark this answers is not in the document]
Crafty15: I think that happens when matlab generates a function template. I'll tidy all this up as i implement the build in functions.

shared/tests/utility/math/docs/README.txt
Crafty15: Yep. That change happened further down where instead of just comparing the difference between the pre-calc value and the method result in the REQUIRE statement, i used Approx(cart_compare.x()).epsilon(ERROR_THRESHOLD).

KipHamiltons: I think this comment in this readme could reflect that we have opted to use a percentage threshold (and also even any reasoning behind that threshold or why we used percentages, if you think that could be useful to future maintainers?)

shared/tests/utility/math/docs/README.txt
Crafty15: I'm going to have to double check this.

shared/tests/utility/math/docs/README.txt
```diff
@@ -0,0 +1,14 @@
+The files in this folder contain the same input data used in TestCoordinates.cpp along with some matlab scripts that perform the same calculations found in coordinates.h
```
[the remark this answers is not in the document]
Crafty15: Good call! I'll do that.

shared/tests/utility/math/docs/cartesianToSpherical.m
Crafty15: For the record, pretty sure you did say "built in functions" but i was thinking of trig functions.

.devcontainer.json
```diff
@@ -4,7 +4,8 @@
   "image": "nubots:generic",
```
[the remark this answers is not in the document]
TrentHouliston: Not only that but these changes are a move backwards

shared/tests/utility/math/TestCoordinates.cpp
```diff
@@ -0,0 +1,1029 @@
...
+ * Test value domain to stay between the values of TEST_DBL_MAX = 5.9667e+153, and TEST_DBL_MIN = 1.4853e-307
+ * This means the usual domain edge cases - DBL_MAX and DBL_MIN - have been left out or replaced. It has been determined
+ * that inputs approaching these edge cases will not be possible in normal operation. See coordinates.h comments for
+ * more details.
+ * @author Liam Craft
+ */
+
+static constexpr double ERROR_THRESHOLD = 1e-6;
+static constexpr double TEST_DBL_MAX    = 5.9667e+153;
+static constexpr double TEST_DBL_MIN    = 1.4853e-307;
```
TrentHouliston, line 43: For each of these constants, you should try to get them from limits if you can.
e.g. `std::numeric_limits<double>::max()` and `std::numeric_limits<double>::lowest()` etc.
Also when you're doing tests to make sure floating point error hasn't grown enough instead of using a fixed number like 1e-6 for the error threshold, you should be checking against machine epsilon.

Have a look at https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon
Basically, the error you should expect in a set of floating-point operations is related to the size of the two numbers, as well as how many operations you are doing on them.

shared/tests/utility/math/TestCoordinates.cpp
```diff
@@ -0,0 +1,1029 @@
...
+
+
+/** Cartesian test coords */
+static const std::array<Eigen::Vector3d, 200> cart_coords = {
+    Eigen::Vector3d(-0.492109181440401e+19, -0.080404385128274e+19, -1.673196359027018e+19),
```
TrentHouliston, line 48: If you have large sets of data like this in tests, it's generally best practice to separate the "logic" from the "data".
Put these constant sets into other files and include them. That way the logic of the test is clear in one file without getting hidden with large amounts of data like this.

shared/tests/utility/math/docs/README.txt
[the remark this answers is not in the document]
TrentHouliston: If you want to see what markdown is supported within here you should look at github flavoured markdown, as that's what it will be rendered with https://github.github.com/gfm/

shared/tests/utility/math/docs/README.txt
TrentHouliston: The extremely large and small numbers will often cause numerical instability. I'm not sure it makes sense to use them in lots of these tests as you will get nonsensical results. You should mainly make sure that values that are within our domain and range end up with sensible answers.

shared/tests/utility/math/docs/README.txt
TrentHouliston: Read the information about https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon which will help you write about how you should be testing. It shouldn't be a percentage error or an absolute error per say. It should be an error based on the machine epsilon (which is based on the relative magnitude of the numbers in the operation)

shared/tests/utility/math/docs/calc_spherToCart.m
```diff
@@ -0,0 +1,9 @@
...
+%  TEXT
+% 
+function out = calc_spherToCart(spher_coords)
+for i = 1 : length(spher_coords)
+    disp(sphericalToCartesian(spher_coords(i,:)));
+end
```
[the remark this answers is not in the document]
TrentHouliston: Our vscode should be putting that in if you save the file in vscode :)

shared/tests/utility/math/docs/cart_coords_in_cpp.txt
```diff
@@ -0,0 +1,200 @@
+Eigen::Vector3d(-0.492109181440401e+19, -0.080404385128274e+19, -1.673196359027018e+19),
```
TrentHouliston, line 1: What is the purpose of this file?
I'd recommend either making it a csv file or not having it at all.
likewise for the others

shared/tests/utility/math/docs/spher_coords_cpp.txt
```diff
@@ -0,0 +1,256 @@
+Eigen::Matrix<double, 3, 1>(-5, 0, 0),
```
TrentHouliston, line 1: I'd recommend removing these files as they clutter up the repo and don't add much. The scripts that made them are an important part. In general, you don't ever want to commit anything that was generated as it's fragile.

TrentHouliston (commented):
Some review comments

shared/tests/utility/math/docs/README.txt
KipHamiltons: This begs the question - how do we define or find out what our domain and range are for a given module?

shared/tests/utility/math/TestCoordinates.cpp
KipHamiltons: two questions I have, having read your comment and that page:

1. How would you suggest to take into account how many operations are being done on the floats?
2. In that cppreference example, it has multiplies by a ULP==2. Looking at the [wiki page for ULP](https://en.wikipedia.org/wiki/Unit_in_the_last_place), it has examples which are more like 2^{-51} or 2^{-53}, which makes a lot of sense based on that definition of ULP. Using 2^{-51} where cppreference has used what they're calling `ulp` would certainly cause underflow though, because machine epsilon * 2^{-53} will obviously underflow. Should we disregard that term from the calculation entirely, or simply always use "2" as in their example, or some other approach entirely? It seems like I am probably misunderstanding what cppreference means when it says `ulp`

shared/tests/utility/math/docs/cart_coords_in_cpp.txt
KipHamiltons: Would it be worth making a utility for parsing csv files of test data? This will probably be a recurring theme as we develop more test files

shared/tests/utility/math/TestCoordinates.cpp
TrentHouliston: Sometimes it's hard to know exactly how many operations are being done on a float. But what you'll see is that each operation will have a particular error in regards to units in the last place. Most of the time the error will be <= 0.5 ULP. If you look up documentation you can find out what exactly the expected ULP error is for any operation you're doing.

As for what it means in general since you're dealing with floating-point numbers there isn't a single "error" value that you can reliably use. You can think of floating-point numbers being scientific notation with a particular number of decimal places. So every time you do a maths operation the amount of error you gain is related to the relative magnitudes of numbers. So as an example, if you had a "float" that held 5 significant figures you might have 5.0000e10 So checking your error is less than 1e-6 doesn't make sense as the smallest possible number difference would be 5.0001e10. 1e-6 is smaller than the smallest increment you could do to that number. That's what the machine epsilon is designed to help represent, the smallest "step" you can possibly take based on the magnitude of the numbers involved. So when you're checking for error and how accurate your calculations are, you should use epsilon in order to work out how many units in the last place you want to be accurate to. From our example, that's equivalent to how many significant figures you want to be accurate too. Because you're dealing with binary though it's more like how many bits of error you want to be able to have.

This is also why if you're ever doing some maths operations and you have some large numbers and small numbers, sometimes it's beneficial to rearrange your functions so you only ever do operations on numbers of similar magnitude because large and small numbers together can result in numerical instability due to the operation being unable to represent the small changes in a large number.

shared/tests/utility/math/docs/cart_coords_in_cpp.txt
TrentHouliston: It might be a good idea, the hardest part of all this is where to put it all. I'm not convinced that the doc folder belongs here with the code in the first place

shared/tests/utility/math/TestCoordinates.cpp
TrentHouliston: Also as one more example as to where this can get weird. Imagine you were testing a `sin` function and you decided that 1e-6 was a good threshold for checking. Because of how ULP works, if you stuck a number that was smaller than 1e-6 into the sin function (say 1-e10), your error is much much lower than 1e-6 because the relative scales of the two numbers is smaller. Likewise, when you stick a big number into sin your error will be larger

KipHamiltons: Roger that. We'll roll with 2 for now 👍

KipHamiltons: wouldn't that sine example be covered by the `std::fabs(input)` term in their example though?

TrentHouliston: So for a more concrete example if you had a ground truth for v for `sin(1e-10) == v`, your error in those two would be much less than 1e-6. But then if you did `sin(5) == v` you could have a much higher error as the magnitude of the numbers is larger. When you're testing floating point error all you have is ULP because smaller numbers have the same number of ULP error even though they have smaller absolute error.

shared/tests/utility/math/TestCoordinates.cpp
```diff
@@ -0,0 +1,1029 @@
...
+        Eigen::Vector3d cart_input   = cart_coords[i];
+        Eigen::Vector3d cart_compare = cartToSpher_results[i];
+        Eigen::Vector3d spher_result = utility::math::coordinates::cartesianToSpherical(cart_input);
+        Approx epsilon_x             = Approx(cart_compare.x()).epsilon(ERROR_THRESHOLD);
+        Approx epsilon_y             = Approx(cart_compare.y()).epsilon(ERROR_THRESHOLD);
+        Approx epsilon_z             = Approx(cart_compare.z()).epsilon(ERROR_THRESHOLD);
+        //***Output table***
+        INFO(fmt::format("{:*^50}", "Calculating Spherical coordinates for the origin"));
+        INFO(fmt::format("{:-^50}", ""));
+        INFO(fmt::format("Input:\n x = {}\n y = {}\n z = {}", cart_input.x(), cart_input.y(), cart_input.z()));
```
Bidski, line 967: 99% sure that if you use [fmt ostream](https://fmt.dev/latest/api.html#ostream-api) you cant print `Eigen` vectors and matrices directly (might need to add a transpose as well)
```suggestion
        INFO(fmt::format("Input: {}", cart_input.tranpose()));
```

shared/tests/utility/math/TestCoordinates.cpp
```diff
@@ -0,0 +1,1029 @@
...
+}
+
+/** Test sphericalToCartesian conversion */
+TEST_CASE("Test coordinate conversion - Spherical to cartesian.", "[utility][math][coordinates]") {
+    INFO(fmt::format("{:*^50}", "Calculating cartesian coordinates for the origin"));
+
+    for (size_t i = 0; i < spher_coords.size(); i++) {
+        Eigen::Matrix<double, 3, 1> spher_input   = spher_coords[i];
+        Eigen::Matrix<double, 3, 1> spher_compare = spherToCart_results[i];
+        Eigen::Matrix<double, 3, 1> cart_result   = Eigen::Matrix<double, 3, 1>(0, 0, 0);
```
Bidski, line 996:
```suggestion
        Eigen::Vector3d spher_input   = spher_coords[i];
        Eigen::Vector3d spher_compare = spherToCart_results[i];
        Eigen::Vector3d cart_result   = Eigen::Vector3d::Zero();
```

shared/tests/utility/math/docs/cartesianToSpherical.m
Bidski: Is there a Python implementation of these functions? Considering the majority of our code base is C++/Python it would be nicer if we could stick to those languages as much as possible

shared/utility/math/coordinates.hpp
```diff
@@ -33,6 +33,10 @@ namespace math {
      * (r,phi,theta) represent radial distance, bearing (counter-clockwise from x-axis in xy-plane) and elevation
      * (measured from the xy plane) (in radians)
      * @author Alex Biddulph
+     *
+     * NOTE: Some of the functions below lack mechanisms to deal with large input values that would lead to
+     * infinite return values. This has been noted during testing and it has been determined that during
+     * normal operation, such values won't be possible.
```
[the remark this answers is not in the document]
Bidski: I would actually duplicate this comment for each offending function. Make it explicit with maybe a pre-/post-condition style comment for each function

Bidski (changes requested):

shared/tests/utility/math/docs/cartesianToSpherical.m
KipHamiltons: I have found a python implementation in `astropy` - I think it uses another convention though. That means the results it gives would need to be permuted before we used them, which is fine, but certainly worth noting if that's what you think we should go for.

shared/utility/math/coordinates.hpp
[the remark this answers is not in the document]
Crafty15: Removed these comments completely. It was early days so i was explaining things to myself. Probably TMI i think.

shared/tests/utility/math/docs/README.txt
Crafty15: Removed TMI comments

shared/tests/utility/math/TestCoordinates.cpp
Crafty15: Good call. We implemented these changes promptly.....

shared/utility/math/coordinates.hpp
```diff
@@ -121,4 +153,4 @@ namespace math {
 }  // namespace utility
 
 
-#endif  // UTILITY_MATH_COORDINATES_HPP
+#endif  // UTILITY_MATH_COORDINATES_H
```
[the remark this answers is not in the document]
KipHamiltons: We still don't want to change this :P

shared/tests/utility/math/TestCoordinates.cpp
```diff
@@ -0,0 +1,125 @@
...
+                    INFO("Expected epsilon x value: " << epsilon_x.toString());
+
+                    INFO("Calculated y value: " << result.y());
+                    INFO("Expected y value: " << cart_to_spher_expected.at(i).y());
+                    INFO("Expected epsilon y value: " << epsilon_y.toString());
+
+                    INFO("Calculated z value: " << result.z());
+                    INFO("Expected z value: " << cart_to_spher_expected.at(i).z());
+                    INFO("Expected epsilon z value: " << epsilon_z.toString());
+                    INFO("Failed test value at index: " << i);
```
KipHamiltons, line 78: Given that these can be printed on success, it's probably worth changing these `INFO`'s which reference failure as if it did fail.

shared/tests/utility/math/TestCoordinates.cpp
```diff
@@ -0,0 +1,125 @@
...
+                    INFO("Expected epsilon x value: " << epsilon_x.toString());
+
+                    INFO("Calculated y value: " << result.y());
+                    INFO("Expected y value: " << spher_to_cart_expected.at(i).y());
+                    INFO("Expected epsilon y value: " << epsilon_y.toString());
+
+                    INFO("Calculated z value: " << result.z());
+                    INFO("Expected z value: " << spher_to_cart_expected.at(i).z());
+                    INFO("Expected epsilon z value: " << epsilon_z.toString());
+                    INFO("Failed test value at index: " << i);
```
KipHamiltons, line 116: Same here

shared/tests/utility/math/TestCoordinates.cpp
```diff
@@ -0,0 +1,125 @@
...
+            resolve_expression<Eigen::Vector3d, 200>(test_values["cartesianInput"]);
+        // cart to spher results
+        static const std::array<Eigen::Vector3d, 200> cart_to_spher_expected =
+            resolve_expression<Eigen::Vector3d, 200>(test_values["cartToSpher_results"]);
+        for (size_t i = 0; i < cart_coords.size(); i++) {
+            WHEN("Cartesian coordinates are converted to spherical coordinates") {
+                static const Eigen::Vector3d result = cartesianToSpherical(cart_coords.at(i));
+                THEN("The results of the calculation should be approximately equal to the expected values") {
+                    // NOTE:We use approx because floating points dont like being compared
+                    // at the epsilon value of the expected results
```
KipHamiltons, line 62: I don't quite understand this comment. What do you mean by "at the epsilon value of the expected results"? I piece together you're referring to the return value of the `Approx(<thing>).epsilon(ERROR_THRESHOLD)`, but calling it the `epsilon value of the expected results` confuses me. Naming these variables `epsilon_<letter>` confuses me too, since catch's name for the `epsilon` function itself seems like a misnomer (on catch's behalf) to me.

KipHamiltons (commented):
This is really clean! Nice job. 

Sidenote: We talked a while ago about adding a readme or comment somewhere nearby to test data which detailed how inputs were generated and how the ground truth was generated/checked, so that future people could verify, regenerate, or build upon it. Do you think we could include something like that here?

shared/tests/utility/math/TestCoordinates.cpp
Crafty15: Yeah i don't understand it either. Looks like i deleted a line in the middle or something.

shared/utility/math/coordinates.hpp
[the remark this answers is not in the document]
Crafty15: I can't even remember why. Just to clarify, HPP, right?

shared/tests/data/CoordinateTests.yaml
```diff
@@ -0,0 +1,843 @@
...
+  - [-5, 0, 0]
+  - [-1.4853e-307, 0, 0]
+  - [0, 0, 0]
+  - [1.4853e-307, 0, 0]
+  - [1, 0, 0]
+  - [-5, 1.5707963267948966, 0]
+  - [-5, 0, 1.5707963267948966]
+  - [-5, 1.5707963267948966, 1.5707963267948966]
+  - [-1.4853e-307, 1.5707963267948966, 0]
+  # - [-1.4853e-307, 0, 1.5707963267948966]
```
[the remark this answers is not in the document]
Crafty15: > This is really clean! Nice job.
> 
> Sidenote: We talked a while ago about adding a readme or comment somewhere nearby to test data which detailed how inputs were generated and how the ground truth was generated/checked, so that future people could verify, regenerate, or build upon it. Do you think we could include something like that here?

Yeah for sure. I'll whip something up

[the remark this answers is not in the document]
Crafty15: > 

Yeah something didn't like them. I think it was the yaml cpp. I've completely removed them.

shared/utility/math/coordinates.hpp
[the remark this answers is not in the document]
KipHamiltons: Yep. It's the same as the filename (basically ;P )

KipHamiltons:
Once you add in how you generated the test data (and fix that macroguard ;p ), this LGTM :grin:

ysims:
@Crafty15 did you wanna fix this up and convert to non-draft PR? Looks like it's pretty much done.

Crafty15:
> @Crafty15 did you wanna fix this up and convert to non-draft PR? Looks like it's pretty much done.

Will do

shared/tests/data/README.md
```diff
@@ -0,0 +1,5 @@
+# Generating test data
+
+The input data used in these tests was generated using the random number generator functions in the [NumPy](https://github.com/numpy/numpy) python library.
+
+The expected results were calculated using the coordinate conversion functions found in the [astropy](https://github.com/astropy/astropy) python library.
```
ysims, line 5: Should specify which tests/config files these apply to.

shared/tests/utility/math/TestCoordinates.cpp
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * The NUbots Codebase is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright 2021 NUbots <nubots@nubots.net>
```
ysims, line 17:
```suggestion
 * Copyright 2022 NUbots <nubots@nubots.net>
```

shared/tests/utility/math/TestCoordinates.cpp
```diff
@@ -0,0 +1,134 @@
...
+static constexpr double ERROR_THRESHOLD = 1e-6;
+
+/** Test Data */
+static const YAML::Node test_values = YAML::LoadFile("tests/CoordinateTests.yaml");
+
+// cartesian to spherical
+SCENARIO("Cartesian coordinates can be converted to spherical coordinates", "[utility][math][coordinates]") {
+    GIVEN("A set of cartesian coordinates and a set of expected values") {
+        // cart input
+        static const std::array<Eigen::Vector3d, 200> cart_coords =
```
ysims, line 52: Might be good to specify the number of points in the yaml file instead of hardcoding here? or use an `std::vector` if possible?

shared/tests/utility/math/TestCoordinates.cpp
```diff
@@ -0,0 +1,134 @@
...
+            }
+        }
+    }
+}
+
+// spherical to cartesian
+SCENARIO("Spherical coordinates can be converted to cartesian coordinates", "[utility][math][coordinates]") {
+    GIVEN("A set of spherical coordinates and a set of expected values") {
+        // spherical input
+        static const std::array<Eigen::Vector3d, 210> spher_coords =
```
ysims, line 96: and here

shared/tests/utility/math/TestCoordinates.cpp
KipHamiltons: Just for my curiosity - what's the advantage of that?

ysims: new set of data?

KipHamiltons: IIRC, the blame is on me for using an array here. I guess I figured that if you know the size at compile time and it doesn't change, it should be an array, but (at the risk of speaking for Liam here) I would be totally fine with vector here  if you think that's better (I don't think you can use a std::array with a dynamic size/a size only found at runtime).

Crafty15: Doesn't bother me either way. Might be cleaner with a vector

shared/tests/utility/math/TestCoordinates.cpp
Crafty15: lol, i forgot to scroll down

TrentHouliston (approved):

ysims (approved):
Seems good

KipHamiltons (approved):

Bidski (changes requested):
You should add an extra test that randomly generates cartesian coordinates then tries to round trip them (`cart -> spherical -> cart -> spherical`) and then check that the two `cart`s are equal and the two `spherical` are equal. This will let us know that the functions are inverses of each other

PR #370, opened 2021-02-17
Fix mesh not getting object3d properties set on creation

TrentHouliston:
It wasn't getting set on creation, and since geometry objects can get updated without remaking the mesh it was never set

### DIFF
--- a/nusight2/src/client/components/three/builders.ts
+++ b/nusight2/src/client/components/three/builders.ts
@@ -143,7 +143,11 @@ type MeshOpts = Object3DOpts & {
 }
 
 export const mesh = createUpdatableComputed(
-  (opts: MeshOpts) => new Mesh(opts.geometry, opts.material),
+  (opts: MeshOpts) => {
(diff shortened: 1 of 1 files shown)


nusight2/src/client/components/three/builders.ts
```diff
@@ -143,7 +143,11 @@ type MeshOpts = Object3DOpts & {
 }
 
 export const mesh = createUpdatableComputed(
-  (opts: MeshOpts) => new Mesh(opts.geometry, opts.material),
+  (opts: MeshOpts) => {
+    const mesh = new Mesh(opts.geometry, opts.material)
+    updateObject3D(mesh, opts)
```
BrendanAnnable, line 148: This already gets called https://github.com/NUbots/NUbots/blob/f9cd04276825069e40c8842ea7ae6ec9a661620a/nusight2/src/client/components/three/create_updatable_computed.ts#L25

TrentHouliston:
Turns out I fixed the problem elsewhere and misattributed the bug :O

PR #372, opened 2021-02-20
Pins version of OpenCL to fix linker errors

KipHamiltons:
Fixes linking errors when building `ocl-icd` a bit later.

Fix verfied by @JosephusPaye  - I am having certificate issues for github related domains when building packages after that one atm, so I haven't been able to build the full image myself.


TrentHouliston (approved):
Nice to see they finally started making versioned releases of this library

PR #373, opened 2021-02-20
Changes path to prefer our tools

KipHamiltons:
Prepends our tools directories to the path, so that our tools are used first if there is a namespace conflict

### DIFF
--- a/nuclear/b.py
+++ b/nuclear/b.py
@@ -73,8 +73,8 @@ except:
 if __name__ == "__main__":
 
     # Prepend nuclear and user tools to the path, so we prefer our packages
-    sys.path.insert(0, user_tools_path)
     sys.path.insert(0, nuclear_tools_path)
+    sys.path.insert(0, user_tools_path)
 
     # Root parser information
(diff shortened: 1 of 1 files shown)


wongjoel (dismissed):
Looks good to me

nuclear/b.py
```diff
@@ -72,9 +72,9 @@
 
 if __name__ == "__main__":
 
-    # Add our builtin tools to the path and user tools
-    sys.path.append(user_tools_path)
-    sys.path.append(nuclear_tools_path)
+    # Prepend nuclear and user tools to the path, so we prefer our packages
+    sys.path.insert(0, user_tools_path)
+    sys.path.insert(0, nuclear_tools_path)
```
TrentHouliston, line 77: This is backward, we should prefer our own tools. But currently, this will prefer the built-in nuclear ones. For example this will mess up the `./b module generate` tool by not using the built in NUbots one

TrentHouliston (changes requested):

TrentHouliston (approved):
:rocket:

PR #374, opened 2021-02-22
Make the ./b script keep rerunning itself so long as it thinks it can install a package

TrentHouliston:
Currently, it tells you that it installed and to try again. But we can automate that process and try again for you!

### DIFF
--- a/nuclear/b.py
+++ b/nuclear/b.py
@@ -114,7 +114,7 @@ if __name__ == "__main__":
 
                             install_dependency(package)
 
-                            # Try rerunning ourself now the library exists
+                            # Try re-running the current command now that the library exists
                             sys.exit(subprocess.call([sys.executable, *sys.argv]))
 
                         if hasattr(module, "register") and hasattr(module, "run"):
--- a/nuclear/b.py
+++ b/nuclear/b.py
@@ -6,6 +6,8 @@ import re
 import subprocess
 import sys
 
+from dependencies import find_dependency, install_dependency
+
 # Don't make .pyc files
 sys.dont_write_bytecode = True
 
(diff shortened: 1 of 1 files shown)


TrentHouliston:
I now understand that this PR isn't fitting because #355 hasn't yet been merged

nuclear/b.py
```diff
@@ -113,9 +114,8 @@
 
                             install_dependency(package)
 
-                            print()
-                            print("missing dependency installed, re-run command to continue")
-                            sys.exit(1)
+                            # Try rerunning ourself now the library exists
```
JosephusPaye, line 117:
```suggestion
                            # Try re-running the current command now that the library exists
```

KipHamiltons:
@Y0sh1dk pointed out to me that the python docs [recommend using subprocess.run() wherever possible](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module), instead of `.call()` and `.Popen()`. Is there a particular reason we use `.call()` ubiquitously?

TrentHouliston:
Because most of this code was written before `subprocess.run` even existed. Additionally, the return type of call is the return type of the process that you execute. `run` on the other hand returns this wonderful fancy CompletedProcess instance which would require us to store it in a variable and then access the return value from it rather than just using the return value of the function.

That's the main reason it's being used over run in this case, so we can just exit with the return code

JosephusPaye (approved):
LGTM.

From Python 3.5, `subprocess.call()` [is equivalent](https://docs.python.org/3/library/subprocess.html#subprocess.call) to `subprocess.run(...).returncode`:

> Run the command described by _args_. Wait for command to complete, then return the `returncode` attribute.

The difference (`subprocess.call()` lacks the `check` and `input` parameters) is not relevant in this case.

PR #375, opened 2021-02-22
Remove the name from the container

TrentHouliston:
Don't name the container, it means you can't run multiple ./b things at once (e.g. if you're in a ./b shell in one terminal you can't ./b build in another etc)

### DIFF
--- a/tools/dockerise/__init__.py
+++ b/tools/dockerise/__init__.py
@@ -236,8 +236,6 @@ def run_on_docker(func):
                             "--tty",
                             "--workdir",
                             "/home/{}/{}/{}".format(user, directory, code_to_cwd),
-                            "--name",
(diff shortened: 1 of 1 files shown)


KipHamiltons (approved):
LGTM

Y0sh1dk:
Would it be worth naming the container with the name of the tool that is ran (eg `nubots_shell` or `nubots_build`) so that we can tell container resource usage when running `docker stats`?

KipHamiltons:
That sounds like a good idea to me!

TrentHouliston:
Nah that's not a good idea. Then you'd only be able to run a single shell etc which is still limiting. This way it just gets a random name but if you do `docker ps` you'll still see the actual command that was run anyway

PR #376, opened 2021-02-22
Upgrade run_on_docker so it can take different images

TrentHouliston:
This allows the run_on_docker annotation to take an image parameter so specific ./b tools will run on specific images.

This might be useful for example for ensuring that a tool only uses the generic image if you are executing code that was compiled for a target system.

Also if you decide to start doing things such as training neural networks using the ./b script, you can use the TensorFlow docker image to train with, also allowing you to lock versions of TensorFlow for specific training functions


Bidski (approved):
`./b target generic` with #378 seems to work fine. Would need to build image, push to docker hub, clean docker caches/images, then rebuild/pull to ensure that everything works properly.

A more thorough test of all functionality is still needed to test all other changes

PR #377, opened 2021-02-22
Upgrade the formatting tools to the latest versions

TrentHouliston:
Upgrades the formatting for python and cmake and applies it

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,3 +3,4 @@ line-length = 120
 
 [tool.isort]
 line_length=120
+known_first_party=['b', 'utility']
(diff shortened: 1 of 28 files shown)


pyproject.toml
```diff
@@ -3,3 +3,4 @@ line-length = 120
 
 [tool.isort]
 line_length=120
+known_first_party=['b', 'utility']
```
ElderNoSpace, line 6: Should we also include `dockerise`?

ElderNoSpace (changes requested):

pyproject.toml
TrentHouliston: I didn't bother because it no longer lives there as of #376

ElderNoSpace (approved):
LGTM, after #375

PR #326, opened 2020-07-15
Tell docker which images to use as a cache when building

Bidski:

PR #378, opened 2021-02-23
Swap the ./b script to use buildx for building docker images

TrentHouliston:
Swaps the ./b script to use buildx for building docker images.
This means that we can do a much better job of using caches so that people don't have to rebuild the images all the time.

Depends on #376

Will also supercede at least part of #326 if we also swap to #353 that PR will be made totally redundant


Bidski (dismissed):
This seems to work (tested before this was "re-based" on #376 ) 

See https://github.com/NUbots/NUbots/pull/376#pullrequestreview-598099503 for more details of the testing that was performed

cmake/Modules/ToolchainLibraryFinder.cmake
```diff
@@ -90,8 +90,8 @@ function(ToolchainLibraryFinder)
     if(PACKAGE_VERSION_BINARY_ARGUMENTS AND PACKAGE_BINARY)
```
Bidski: Why does this file even exist?

nuclear/cmake/Modules/ToolchainLibraryFinder.cmake
```diff
@@ -90,8 +90,8 @@ function(ToolchainLibraryFinder)
     if(PACKAGE_VERSION_BINARY_ARGUMENTS AND PACKAGE_BINARY)
       exec_program(
         ${${PACKAGE_NAME}_BINARY} ARGS
-        ${PACKAGE_VERSION_BINARY_ARGUMENTS} OUTPUT_VARIABLE
```
Bidski: How is this relevant to this PR?

Bidski (changes requested):
Seems to be quite a few unrelated changes in this PR

nuclear/cmake/Modules/ToolchainLibraryFinder.cmake
TrentHouliston: Why are you commenting on merge commits? This is the merge in from master

cmake/Modules/ToolchainLibraryFinder.cmake
TrentHouliston: Why are you commenting on merge commits? This is the merge in from master

tools/utility/dockerise/platform.py
```diff
@@ -74,33 +77,31 @@ def build(repository, platform):
...
 
-    # Pull the latest version
-    err = pty.spawn(["docker", "pull", remote_tag])
-    if err != 0:
-        cprint("Docker pull returned exit code {}".format(err), "red", attrs=["bold"])
-        exit(err)
+    # Create a buildx instance for building this image
+    builder_name = "{}_{}".format(defaults.image, platform)
+    if subprocess.call(["docker", "buildx", "inspect", builder_name], stderr=DEVNULL, stdout=DEVNULL) != 0:
+        subprocess.call(["docker", "buildx", "create", "--name", builder_name], stderr=DEVNULL, stdout=DEVNULL)
```
Bidski, line 83: You should probably replace these `.call()`s with `.run().returncode`s, per https://github.com/NUbots/NUbots/pull/374#issuecomment-783741219

Bidski: I should probably should have picked that up in the other PR, but oh well

Bidski (changes requested):

tools/utility/dockerise/platform.py
TrentHouliston: You're making the call more complicated for no gain. It's not like `call` is deprecated

Bidski (approved):
LGTM

PR #379, opened 2021-02-25
Update run_on_docker import

Y0sh1dk:
Fix `run_on_docker` import to reflect changes in #376

### DIFF
--- a/tools/tests.py
+++ b/tools/tests.py
@@ -5,9 +5,8 @@ import os
 import subprocess
 import time
 
-from dockerise import run_on_docker
-
 import b
+from utility.dockerise import run_on_docker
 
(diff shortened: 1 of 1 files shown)


TrentHouliston (approved):

TrentHouliston (changes requested):
Actually, this might be a good time to take advantage of the new features of run_on_docker.
Since you should only really be running the unit tests on the `generic` image, changing your `@run_on_docker` annotations to `@run_on_docker(image="nubots:generic")` might be a good idea

Y0sh1dk:
Not sure if specifying image is necessary for the `register(command)` command, however I added it there too.

TrentHouliston (approved):

PR #380, opened 2021-02-26
Removes `wget` from first arch install script

KipHamiltons:
wget has been removed from newer arch iso's, so we have to use curl for this instance. The NUbook arch flash guide will have to change too.


ElderNoSpace (approved):
LGTM

PR #381, opened 2021-03-01
Fix error for detecting internal images properly

TrentHouliston:
Internal images are no longer working since `repository` was changed to `image` to allow for caching


JosephusPaye (approved):
LGTM, assuming it works

tools/utility/dockerise/defaults.py
```diff
@@ -7,6 +7,6 @@
 directory = "NUbots"
 
 
-def is_internal_image(image):
-    split = image.split(":")
-    return len(split) == 2 and split[0] == repository and (split[1] in platform.list() or split[1] == "selected")
+def is_internal_image(tag):
+    split = tag.split(":")
+    return len(split) == 2 and split[0] == image and (split[1] in platform.list() or split[1] == "selected")
```
Bidski, line 12: Surely `repository` is more semantically correct here?

TrentHouliston: Nah, in docker you have things split up into 3
`repository/image:thing`
We're checking here that `image` is `nubots` (local image)

Bidski: Fair enough, I dont see any issues then

PR #383, opened 2021-03-02
Make setting the log level from a file neater.

CMurtagh-LGTM:
Make setting the log level from the config file a one liner by specialising yaml-cpp.


TrentHouliston (changes requested):
Given how common this is I think I would make it automatically included in the configuration extension. Now that I think of it we should probably do that with yaml expression too (another PR). Also once this PR is merged you should make a PR where you spread the word of log level to all the modules

ElderNoSpace:
> Given how common this is I think I would make it automatically included in the configuration extension. Now that I think of it we should probably do that with yaml expression too (another PR). Also once this PR is merged you should make a PR where you spread the word of log level to all the modules

So do you mean that we should check for the "log_level" member and if it exists, automatically set the log level with the yaml-cpp extension?

module/input/ImageDecompressor/src/ImageDecompressor.cpp
```diff
@@ -9,6 +9,8 @@
 #include "message/input/Image.hpp"
 #include "message/output/CompressedImage.hpp"
 
+#include "utility/support/yaml_LogLevel.hpp"
```
TrentHouliston, line 12: this filename is an abomination. it either needs to be snake_case or something else.

shared/utility/support/yaml_LogLevel.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include <fmt/format.h>
+#include <map>
+#include <nuclear>
+#include <string>
+#include <yaml-cpp/yaml.h>
+
+namespace utility::support {
+
+struct LogLevel {
```
TrentHouliston, line 31: Just go directly to NUClear::LogLevel, this intermediate class serves no purpose but to increase complexity

shared/utility/support/yaml_LogLevel.hpp
```diff
@@ -0,0 +1,86 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright 2013 NUbots <nubots@nubots.net>
+ */
+
+#ifndef UTILITY_SUPPORT_yaml_LogLevel_HPP
+#define UTILITY_SUPPORT_yaml_LogLevel_HPP
```
TrentHouliston, line 21: Macroguards should be all caps. (and yes I know some slipt through elsewhere when they shouldn't)

shared/utility/support/yaml_LogLevel.hpp
```diff
@@ -0,0 +1,86 @@
...
+}  // namespace utility::support
+
+namespace YAML {
+
+template <>
+struct convert<utility::support::LogLevel> {
+    static Node encode(const utility::support::LogLevel& rhs) {
+        Node node;
+
+        // Treat as a int???
```
TrentHouliston, line 73: What is this for?
Comments should answer questions not create more!

tools/module/generate.py
```diff
@@ -129,6 +129,8 @@ def generate_cpp(parts):
 
         #include "extension/Configuration.hpp"
 
+        #include "utility/support/yaml_LogLevel.hpp"
```
TrentHouliston, line 132: I think this should be included within `extension/Configuration.hpp` so that you don't need to include something else.
This is also what I was talking about with yaml_expression. That should probably be included in Configuration.hpp since it's such a common usage. (but that's another PR)

tools/module/generate.py
```diff
@@ -137,16 +139,7 @@ def generate_cpp(parts):
...
-                // clang-format off
-                auto lvl = cfg["log_level"].as<std::string>();
-                if (lvl == "TRACE") {{ this->log_level = NUClear::TRACE; }}
-                else if (lvl == "DEBUG") {{ this->log_level = NUClear::DEBUG; }}
-                else if (lvl == "INFO") {{ this->log_level = NUClear::INFO; }}
-                else if (lvl == "WARN") {{ this->log_level = NUClear::WARN; }}
-                else if (lvl == "ERROR") {{ this->log_level = NUClear::ERROR; }}
-                else if (lvl == "FATAL") {{ this->log_level = NUClear::FATAL; }}
-                // clang-format on
+                this->log_level = cfg["log_level"].as<utility::support::LogLevel>();
```
TrentHouliston, line 142: If you setup the converter to do NUClear::LogLevel instances directly you don't need the .as<>

shared/utility/support/yaml_LogLevel.hpp
ElderNoSpace: Should have been a `TODO`, then I forgot about it.

tools/module/generate.py
```diff
@@ -129,6 +129,7 @@ def generate_cpp(parts):
 
         #include "extension/Configuration.hpp"
 
+
```
TrentHouliston, line 132:
```suggestion
```

TrentHouliston (changes requested):
Delete this new line, and one last thought but should this file be kept next to `Configuration.hpp` (not sure if it should be)

shared/utility/support/yaml_log_level.hpp
```diff
@@ -0,0 +1,69 @@
...
+    static bool decode(const Node& node, ::NUClear::LogLevel& rhs) {
+        try {
+            // clang-format off
+            auto lvl = node.as<std::string>();
+            if (lvl == "TRACE") { rhs = NUClear::TRACE; }
+            else if (lvl == "DEBUG") { rhs = NUClear::DEBUG; }
+            else if (lvl == "INFO") { rhs = NUClear::INFO; }
+            else if (lvl == "WARN") { rhs = NUClear::WARN; }
+            else if (lvl == "ERROR") { rhs = NUClear::ERROR; }
+            else if (lvl == "FATAL") { rhs = NUClear::FATAL; }
```
KipHamiltons, line 57: What's the intended behaviour if the string from the file doesn't match any of these options?

CMurtagh-LGTM: Time to look in the NUClear source.

CMurtagh-LGTM: Its `DEBUG`

KipHamiltons: Would it be appropriate to make that an `else` condition (or add a comment indicating as such), so that it's clear what  happens in such a case

shared/utility/support/yaml_log_level.hpp
```diff
@@ -0,0 +1,71 @@
...
+            case NUClear::INFO: node = "INFO"; break;
+            case NUClear::WARN: node = "WARN"; break;
+            case NUClear::ERROR: node = "ERROR"; break;
+            case NUClear::FATAL: node = "FATAL"; break;
+            default: node = "DEBUG"; break;
+        }
+
+        return node;
+    }
+
```
KipHamiltons, line 48: This is a simple function to read the source for, but I think we should still be adding @brief etc. to stuff like this, even if it's only for the day of auto generated doxygen-tation

KipHamiltons (approved):
Lgtm, for what that's worth

KipHamiltons (approved):
LGTM

KipHamiltons (dismissed):

KipHamiltons (approved):

PR #386, opened 2021-03-15
Removes OldWalkEngine

KipHamiltons:
I was going to remove other unused walk engines, but didn't know where to look for them.
Further purges of other stuff are coming soon :D


ysims:
You can probably remove motion/walk/FootStep
It's not really useful and a better version is in the staticwalk2 branch.

Bidski:
> You can probably remove motion/walk/FootStep
> It's not really useful and a better version is in the staticwalk2 branch.

QuinticWalk doesn't use `module/motion/walk/FootStep`, so just get rid of it.

Bidski (changes requested):
Are there any other roles that need changing?

Are there any messages/utilities that are now redundant/unused after removing these modules?

KipHamiltons:
I found a few unused Foot-related messages. I have added a commit which deletes them in this PR.
Otherwise, I couldn't find anything else which was previously used only by the OldWalkEngine or the FootStep which was removed.

Bidski (approved):
LGTM I guess

PR #387, opened 2021-03-16
buildkite use new tests tool

Y0sh1dk:
- Change buildkite pipeline to new new tests tool (tests will now run in parallel)
 
 - Update tests tool to still work when it cannot create a directoy to place log file (such as on buildkite agents, see [here](https://buildkite.com/nubots/nubots/builds/1136#85eae382-5631-42d6-9278-0cfe9e53393a)


tools/tests.py
```diff
@@ -101,22 +108,19 @@ def run(sub_command, num_jobs=0, test=None, given_ctest_args=[], **kwargs):
...
-            subprocess.run(
-                ["/usr/bin/ctest", "--output-log", logPath, "--parallel", str(num_jobs), *ctest_args]
-            ).returncode
-        )
+        else:
+            if dirExists:
+                filename = "all-" + filename
+                logPath = os.path.join(tests_dir, filename)
+                ctest_args.extend(["--output-log", logPath])
+            exit(subprocess.run(["/usr/bin/ctest", "--parallel", str(num_jobs), *ctest_args]).returncode)
```
KipHamiltons, line 123: I think this would be a little cleaner if you just built up the command as its own list, then pass that complete list to the run command. So, for example like this:
```py
tests_command = ["/usr/bin/ctest", "--parallel", str(num_jobs)]
if dirExists:
    if test:
        filename = test + "-" + filename
    else:
        filename = "all-" + filename
    logPath = os.path.join(tests_dir, filename)
    tests_command.extend(["--output-log", logpath])
tests_command.extend([ctest_args])

exit(subprocess.run(tests_command).returncode) 
```

Maybe something like that? What do you think?

Y0sh1dk: Ah yep thats much nicer, I've done something similar in the latest commit [here](https://github.com/NUbots/NUbots/pull/387/files#diff-4a9b0a14dd6637678f0bd2ce01b9337e8fc5d117b759439e54f80daef8f1c579)

tools/tests.py
```diff
@@ -82,41 +82,42 @@ def run(sub_command, num_jobs=0, test=None, given_ctest_args=[], **kwargs):
...
+            ctest_command.extend(given_ctest_args)
 
         # If a test was given to run
         if test:
+            ctest_command.extend(["-R", test])
             filename = test + "-" + filename
+        else:
+            filename = "all-" + filename
+
+        # If directory was created suscessfully
```
KipHamiltons, line 115:
```suggestion
        # If directory was created successfully
```

KipHamiltons (approved):

PR #388, opened 2021-03-16
Removes NUcapLocalisation module

KipHamiltons:
This module is not in any roles. If you add it to a role, it fails to compile since the messages it relies on have been refactored under it.
It hasn't seen any actual maintenance besides formatting changes and very slight refactors in the history of this repo.
I am open to being closing this PR if y'all think I should, but it seems to me like this is dead code.


TrentHouliston (approved):
It would be fairly easy to replace if you need to. All it is is a localisation system that is based on the Motion Capture system.

Should be all of about 10 lines of code provided the motion capture stuff is working properly

TrentHouliston:
Pretty sure `OdometryLocalisation` as odometry is implemented now would be just repeatedly emitting an identity matrix for Hwf

PR #389, opened 2021-03-17
Modernise subprocess usage

Y0sh1dk:
Since python 3.5 `subprocess.run()` is recommended over `subprocess.call()` to invoke a subprocess. [Subprocess docs](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module)

All usage of `subprocess.call()` has been replaced with `subprocess.run()`. In instances where the return code of the process is needed, it has been replaced with `subprocess.run().returncode`


Bidski (approved):
Looks OK to me

KipHamiltons:
@TrentHouliston any objections? you didn't seem too keen [last time this was brought up](https://github.com/NUbots/NUbots/pull/374#issuecomment-783743484), but this seems pretty neat to me 🤷 

(right PR this time 😂 😂  🤣 )

TrentHouliston:
I think .call makes more sense in situations where .run would make the code more complicated. But honestly, it doesn't matter either way

PR #390, opened 2021-03-17
Pacman cache for buildkite agents

Y0sh1dk:
Bit of a spicy one,
 - Docs, config and script for setting up Pacman package cache on buildkite agents
 - Update Dockerfile to tell pacman to try to pull packages from the package cache running on the host first before pulling directly from package server.

To test:
 - Follow `doc/PacmanCache/README.md`, except for step 2 run this instead
```
wget https://raw.githubusercontent.com/NUbots/NUbots/deKoeyer/pacman-cache/doc/PacmanCache/install.sh
```
 - Build a container with updated dockerfile
 - View cache hits/misses with `docker logs flexo`


KipHamiltons:
I would preferably like to see a NUbook PR with more details before merging this in pls.

doc/PacmanCache/install.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+# Download docker-compose.yml for package cache (using nroi/flexo)
+wget https://raw.githubusercontent.com/NUbots/NUbots/deKoeyer/pacman-cache/doc/PacmanCache/docker-compose.yml
```
Bidski, line 4: This link will need to be changed before a merge could happen

docker/Dockerfile
```diff
@@ -11,7 +11,7 @@
 ### the date in the FROM command
 ### Date format is ISO year/month/day
 FROM archlinux:base-20210110.0.13332
-RUN echo "Server=https://archive.archlinux.org/repos/2021/01/10/\$repo/os/\$arch" | tee /etc/pacman.d/mirrorlist
+RUN echo -e "Server=http://172.17.0.1:7878/\$repo/os/\$arch\nServer=https://archive.archlinux.org/repos/2021/01/10/\$repo/os/\$arch" | tee /etc/pacman.d/mirrorlist
```
Bidski, line 14: Is this IP guaranteed to always be this?

Bidski (changes requested):
As per @KipHamiltons we will need a nubook PR to pair with this PR

docker/Dockerfile
Y0sh1dk: Yeah it is, during our build, the buildx container defaults to the `bridge` network driver, from which you can always ping the host on `172.17.0.1`

Alternatively, we could add `--add-host host.docker.internal:host-gateway` to [here](https://github.com/NUbots/NUbots/blob/master/tools/utility/dockerise/platform.py#L87), and then instead of `172.17.0.1` it could be `host.docker.internal`, thoughts?

Y0sh1dk:
Relevent NUbook PR @KipHamiltons @Bidski 

https://github.com/NUbots/NUbook/pull/101

docker/Dockerfile
Bidski: The only alternative I can think of is to set an environment variable to the output of this bash command
```
export DOCKER_GATEWAY=$(ip -o addr show | grep inet\  | grep "docker" | cut -d\  -f 7 | cut -d'/' -f1)
```

But this also assumes that the interface will always have `docker` in its name. But, if the `--add-host` command works then I would go with that option.

Y0sh1dk: I couldn't get `--add-host host.docker.internal:host-gateway` to work with `buildx`, it didnt seem to recognise `host-gateway`, it does work with `build` though...

I ended up using docker inspect to get the docker host ip, and then just passed that too `--add-host`

https://github.com/NUbots/NUbots/pull/390/files#diff-e8c882de7db134796dc7fe8e647cae033d8faa7d34027e24b451ecf25d7487c0R86

Bidski (approved):
Looks good. I assume this has been tested?

doc/PacmanCache/README.md
```diff
@@ -0,0 +1,29 @@
...
+Sets up a package cache for caching [pacman](https://wiki.archlinux.org/index.php/pacman) packages locally.
+
+## Prerequisites
+ - [wget](https://www.gnu.org/software/wget/) or [curl](https://curl.se/)
+ - [docker-compose](https://docs.docker.com/compose/)
+
+## Install
+
+1. Create a directory to store cache config
+```
```
KipHamiltons, line 12:
```suggestion
```sh
```

KipHamiltons:
Yea, we tested by making a dummy dockerfile to see if the packages it installed were cached, and they certainly appeared to be. The cache container said in its logs they were cache hits, and they seemed to install fast enough, so that was good enough for me lol

PR #391, opened 2021-03-18
Removes FixedWalk and WalkOptimiser

KipHamiltons:
(Depends on #386)

FixedWalk was unused besides by WalkOptimiser. WalkOptimiser has not been used in years, and was superseded by the Gazebo optimisation project. This PR simply removes those two.


TrentHouliston (approved):
Reminds me of the Thanos snap.
Good times

PR #393, opened 2021-03-18
Make NUsight2 use the protocol buffers as defined by the NUbots codebase

TrentHouliston:
Next step of making the code into a monorepo, makes NUsight2 use the protocol buffers defined by NUbots codebase rather than it's own copy


Bidski (approved):
Looks OK to me

JosephusPaye:
Does it work?

BrendanAnnable (approved):
Now that I fixed everything... LGTM 😄

PR #394, opened 2021-03-21
Add communication to connect to the Webots Robocup controller.

CMurtagh-LGTM:
See [the webots' controller](https://github.com/cyberbotics/webots/pull/2888) and [the draft api](https://cdn.robocup.org/hl/wp/2021/04/v-hsc_simulator_api_draft2.pdf).

Also checkout [our controller code](https://github.com/NUbots/Webots).


module/platform/webots/src/webots.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+        int fd = tcpip_connect(cfg["port"].as<int>(), cfg["server_address"].as<std::string>());
+
+        send_connect(fd, cfg["teamId"].as<int>(), cfg["robotId"].as<int>());
+
+        on<IO>(fd).then([this]() {
+            // Receiveing
+
+            uint64_t N;
+            // Somehow work out how big the message was
```
Bidski, line 74: Do this the way i did in the webots repo. Have a uint64 which is N, then have the serialised message following

module/platform/webots/src/webots.cpp
```diff
@@ -0,0 +1,124 @@
...
+            }
+
+            emit(sensor_data);
+        });
+
+        on<Every<1, std::chrono::seconds>>().then([this]() {
+            // Sending
+            std::vector<char> data = NUClear::util::serialise::Serialise<ActingMessage>.serialise(to_send);
+            uint64_t N             = data.size();
+            send(fd, data.data(), N, 0);
```
Bidski, line 99: Send N before you send the data

Bidski:
We will need to confirm with the TC about their protocol. Protobuf messages arent designed to be sent end-on-end without break which is why I added the message length first. However, there are alternatives to this scheme

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+                    server->h_length);
+    }
+    else {
+        close(fd);
+        log<NUClear::ERROR>(fmt::format("Cannot resolve server name: {}", server_name));
+        return -1;
+    }
+
+    // Connect to the ip address
+    if (int error = connect(fd, (struct sockaddr*) &address, sizeof(address))) {
```
KipHamiltons, line 77: I want to note that I saw that everyone seems to cast their `sockaddr_in` like this when calling `bind(..)`, but I couldn't see people doing it for `connect(..)`, because they all seemed to do things fundamentally differently, usually using `getaddrinfo(..)` instead of `gethostbyname(..)`

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+int Webots::tcpip_connect(const std::string& server_name, const int& port) {
+    // Create the socket
+    int fd = socket(AF_INET, SOCK_STREAM, 0);
+
+    // Store the ip address information that we will connect to
+    sockaddr_in address;
+    memset(&address, 0, sizeof(sockaddr_in));
+    address.sin_family = AF_INET;
+    address.sin_port   = htons(port);
+    hostent* server    = gethostbyname(server_name.c_str());
```
KipHamiltons, line 62: the [linux docs](https://man7.org/linux/man-pages/man3/gethostbyname.3.html#DESCRIPTION) say you should use `getaddrinfo(..)` instead of `gethostbyname(..)`. I'll leave that for you to decide if that's worth acting on.

KipHamiltons (commented):
[Duplicating my comments from my PR to this branch]

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+        // Store each ServoTarget to send in the next lot
+        for (auto& command : commands.targets) {
+            MotorPosition position_msg = to_send.add_motor_position();
+            position_msg.name          = command.id;
+            position_msg.position      = command.position;
+
+            // TODO(cmurtagh) work out if gain is velocity or force
+            MotorVelocity velocity_msg = to_send.add_motor_velocity();
+            velocity_msg.name          = command.id;
+            velocity_msg.velocity      = command.gain;
```
tayloryoung6396, line 205: Velocity shouldn't be gain or force. We don't set a velocity for the dynamixel servos. Gain is our PID gain

module/platform/Webots/data/config/webots.yaml
```diff
@@ -0,0 +1,10 @@
+# Controls the minimum log level that NUClear log will display
+log_level: DEBUG
+
+# Connection details
+server_address:
+port:
+
+# Info to tell webots who we are
+team_id:
+robot_id:
```
Bidski, line 10: Get this information from [`module::support::configuration::GlobalConfig`](https://github.com/NUbots/NUbots/blob/cmurtagh/webots_communication/module/support/configuration/GlobalConfig/data/config/GlobalConfig.yaml)

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+#include "message/platform/darwin/DarwinSensors.hpp"
+#include "message/platform/webots/ConnectRequest.hpp"
+#include "message/platform/webots/messages.hpp"
+
+// Include headers needed for TCP connection
+#include <netdb.h>      /* definition of gethostbyname */
+#include <netinet/in.h> /* definition of struct sockaddr_in */
+#include <sys/socket.h>
+#include <sys/time.h>
+#include <unistd.h> /* definition of close */
```
Bidski, line 37:
```suggestion
extern "C" {
#include <netdb.h>      /* definition of gethostbyname */
#include <netinet/in.h> /* definition of struct sockaddr_in */
#include <sys/socket.h>
#include <sys/time.h>
#include <unistd.h> /* definition of close */
}
```

module/platform/Webots/src/Webots.cpp
Bidski: Yea, we should avoid using deprecated functions.

module/platform/Webots/data/config/webots.yaml
TrentHouliston: Or more specifically, from the message it emits

module/platform/Webots/src/Webots.cpp
TrentHouliston: We totally do set the velocity for the dynamixel servos. That's how we do our scripts (and potentially everything else)

module/platform/Webots/src/Webots.cpp
TrentHouliston: `sockaddr*` can be considered a "base class" as much as c has such a concept. Beyond that there are variants of `sockaddr` for each of the different network types.

For example `sockaddr_in` is the IPv4 version of the struct. and `sockaddr_in6` would be the IPv6 version. There are also other versions for other socket protocols that can be used (although those two are the most common you'll see)

All the functions require you to cast your type to a `sockaddr*` message to use, and to make sure that the size you pass in is actually the size of the type itself (`sizeof(sockaddr)` might not equal `sizeof(sockaddr_in)`. If you ever need to store a list of addresses that contain a mix of those there is a special type `sockaddr_storage` which will always have a large enough size to store any of the possible `sockaddr*` variants.

In this case though, you should use a reinterpret cast, not a c style cast

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+                log<NUClear::ERROR>("Error: Failed to read message from TCP connection");
+                return;
+            }
+
+            // Deserialise the message into a neutron
+            SensorMeasurements msg = NUClear::util::serialise::Serialise<SensorMeasurements>::deserialise(data);
+
+            // Read each field of msg, translate it to our protobuf and emit the data
+            auto sensor_data = std::make_unique<DarwinSensors>();
+            sensor_data.FromSeconds(msg.time);
```
Bidski, line 130: According to the [API specs document](https://cdn.robocup.org/hl/wp/2021/03/v-hsc_simulator_api.pdf), the sensors measurement has no timestamp?

Bidski: And even if it did, `FromSeconds` is only valid on a protobuf timestamp message. You are using a `Neutron` here so `timestamp` is a `std::chrono::time_point`, so this should work I think
```suggestion
            sensor_data->timestamp = msg.time;
```

Bidski: Althought, perhaps we just want
```suggestion
            sensor_data->timestamp = NUClear::clock::now();
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+                // string name
+                // Vector3 value
+            }
+
+            emit(sensor_data);
+
+            for (auto& camera : msg.camera) {
+                // Convert the incoming image so we can emit it to the PowerPlant.
+                auto compressed_image = std::make_unique<CompressedImage>();
+                compressed_image->timestamp.FromSeconds(msg.time);
```
Bidski, line 167: Same story here as above

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+                compressed_image->lens.centre.y() = (float) camera.centerY;
+                // Radial coefficients
+                // tangential coefficients
+                emit(compressed_image);
+            }
+        });
+
+        on<Every<1, std::chrono::seconds>>().then([this, fd]() {
+            // Sending
+            std::vector<char> data = NUClear::util::serialise::Serialise<ActingMessage>::serialise(to_send);
```
Bidski, line 184:
```suggestion
            std::vector<char> data = NUClear::util::serialise::Serialise<ActuatorRequests>::serialise(to_send);
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+            MotorVelocity velocity_msg = to_send.add_motor_velocity();
+            velocity_msg.name          = command.id;
+            velocity_msg.velocity      = command.gain;
+
+            MotorTorque torque_msg = to_send.add_motor_torque();
+            torque_msg.name        = command.id;
+            torque_msg.torque      = command.torque;
+
+            // MotorPID ? Do we need to send this?
+        }
```
Bidski, line 212: `MotorPosition`, `MotorVelocity`, and `MotorTorque` are all `Neutron`s here. You want something like this
```
MotorTorque torque_msg;
torque_msg.name        = command.id;
torque_msg.torque      = command.torque;
to_send.motor_torque.push_back(torque_msg);
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+            uint64_t N             = data.size();
+            send(fd, &N, sizeof(N), 0);
+            send(fd, data.data(), N, 0);
+        });
+    });
+
+    // Create the message that we are going to send.
+    on<Trigger<ServoTargets>>().then([this](const ServoTargets& commands) {
+        // Maybe keep the `ServoTarget`s we have not sent yet, instead of just overriding them.
+        to_send = ActuatorRequests();
```
Bidski, line 194: You probably want to go through each element and clear the vectors.
```suggestion
        to_send.motor_position.clear();
        to_send.motor_velocity.clear();
        to_send.motor_torque.clear();
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+            torque_msg.torque      = command.torque;
+
+            // MotorPID ? Do we need to send this?
+        }
+    });
+}
+
+void Webots::send_connect(int& fd, int& team_id, int& robot_id) {
+    // TODO(cameron) workout what to do if failes
+    ConnectRequest connect_request = ConnectRequest;
```
Bidski, line 218:
```suggestion
    ConnectRequest connect_request;
```

Bidski (changes requested):

module/platform/Webots/src/Webots.cpp
KipHamiltons: Thanks for the update! that's quite interesting. 
One thing I did find was that `sizeof(sockaddr) == sizeof(sockaddr_in)`, which is convenient for what we're doing

TrentHouliston: `sockaddr` will be the same size as sockaddr_in mostly because of people using the wrong one all the time and sockaddr_in being the most common :P. Since using sizeof(sockaddr) is so common instead of sizeof(sockaddr_in) for people they just made them the same size by padding out the `sockaddr` message with reserved bytes :P

module/platform/Webots/src/Webots.cpp
CMurtagh-LGTM:
```suggestion
                compressed_image->timestamp = msg.time;
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,245 @@
...
+                auto compressed_image = std::make_unique<CompressedImage>();
+                compressed_image->timestamp = NUClear::clock::now();
+                compressed_image->name            = camera.name;
+                compressed_image->dimensions.x()  = camera.width;
+                compressed_image->dimensions.y()  = camera.height;
+                compressed_image->format          = camera.quality;  // This is probably wrong, we havent documented :(
+                compressed_image->data            = camera.data;
+                compressed_image->lens.fov        = (float) camera.horizontalFieldOfView;
+                compressed_image->lens.centre.x() = (float) camera.centerX;
+                compressed_image->lens.centre.y() = (float) camera.centerY;
```
KipHamiltons, line 186:
```suggestion
                compressed_image->lens.fov        = static_cast<float>(camera.horizontalFieldOfView);
                compressed_image->lens.centre.x() = static_cast<float>(camera.centerX);
                compressed_image->lens.centre.y() = static_cast<float>(camera.centerY);
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,245 @@
...
+            velocity_msg.name          = command.id;
+            velocity_msg.velocity      = command.gain;
+			to_send.motor_velocity.push_back(velocity_msg);
+
+            MotorTorque torque_msg;
+            torque_msg.name        = command.id;
+            torque_msg.torque      = command.torque;
+			to_send.motor_torque.push_back(torque_msg);
+
+            // MotorPID ? Do we need to send this?
```
KipHamiltons, line 227: I think MotorPID refers to PID gains. We usually only set `p` (proportional) gain (presumably the first value in the Vector3). We also usually just use one value for it, but that value changes depending on where in the codebase you're looking.
For reference, the quintic walk uses (p) gain value of "13" for the legs.

module/platform/Webots/src/Webots.hpp
```diff
@@ -0,0 +1,56 @@
...
+	
+public:
+    /// @brief Called by the powerplant to build and setup the webots reactor.
+    explicit Webots(std::unique_ptr<NUClear::Environment> environment);
+
+    /// @brief Establish a TCP connection to the specified server/port
+    /// @param server_name The name or IP address to connect to. If it's an IP, it should be in "X.X.X.X" form
+    /// @param port The port number to connect to
+    /// @return If the connection was successful, a file descriptor. Else, -1 is returned
+    int tcpip_connect(const std::string& server_name, const char& port);
```
KipHamiltons, line 51: I have been wondering if this should be made private. Probably? idk

module/platform/Webots/src/Webots.hpp
```diff
@@ -0,0 +1,56 @@
...
+private:
+    /// @brief Send the inital message that tells webots who we are. We should do this when we first connect or when we
+    /// want to reconnect.
+	/// @param The socket we want to send our details to.
+    void send_connect(int& fd);
+
+    /// @brief A single message that stores all the data that webots needs from us.
+    message::platform::webots::ActuatorRequests to_send;
+
+	/// @ brief The information we got from Global Config
```
KipHamiltons, line 40:
```suggestion
    /// @brief The information we got from Global Config
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,245 @@
...
+        }
+    });
+
+	on<Trigger<GlobalConfig>>().then([this](const GlobalConfig& config){
+		player_details.playerId = config.playerId; // TODO(cameron) Do we need to copy this?
+		player_details.teamId = config.teamId;
+	});
+}
+
+void Webots::send_connect(int& fd) {
```
KipHamiltons, line 237: I don't really like this function name

KipHamiltons, line 237:
```suggestion
void Webots::send_connect(const int& fd) {
```

module/platform/Webots/src/Webots.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include "message/platform/webots/ConnectRequest.hpp"
+
+namespace module::platform {
+
+class Webots : public NUClear::Reactor {
+private:
+    /// @brief Send the inital message that tells webots who we are. We should do this when we first connect or when we
+    /// want to reconnect.
+	/// @param The socket we want to send our details to.
+    void send_connect(int& fd);
```
KipHamiltons, line 35:
```suggestion
    void send_connect(const int& fd);
```

module/platform/Webots/src/Webots.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include "message/platform/webots/messages.hpp"
+#include "message/platform/webots/ConnectRequest.hpp"
+
+namespace module::platform {
+
+class Webots : public NUClear::Reactor {
+private:
+    /// @brief Send the inital message that tells webots who we are. We should do this when we first connect or when we
+    /// want to reconnect.
+	/// @param The socket we want to send our details to.
```
KipHamiltons, line 34: You're meant to put the parameter name when doing this javadoc style `@param` stuff. [It looks like you're also "meant" to have an `@param` for every single parameter of every method, (see "required tags" on this link)](https://www.oracle.com/au/technical-resources/articles/java/javadoc-tool.html)
```suggestion
    /// @param fd The socket we want to send our details to.
```

roles/webots.role
```diff
@@ -0,0 +1,9 @@
+nuclear_role(
+  # FileWatcher, ConsoleLogHandler and Signal Catcher Must Go First.
+  extension::FileWatcher
+  support::logging::ConsoleLogHandler
+  # Support and Configuration
+  support::configuration::GlobalConfig
+  # Simulator
```
KipHamiltons, line 7: It's not really a "Simulator" is it?

shared/message/platform/webots/messages.proto
```diff
@@ -0,0 +1,141 @@
...
+    string name  = 1;
+    double value = 2;  // [N]
+}
+
+message Force3DMeasurement {
+    string  name  = 1;
+    Vector3 value = 2;  // [N], x-axis, y-axis, z-axis
+}
+
+message Force6DMeasurement {  // not yet implemented
```
KipHamiltons, line 62: What do you mean "not yet implemented"? Are we waiting on us, or the TC or something?

shared/message/platform/webots/messages.proto
```diff
@@ -0,0 +1,141 @@
...
+}
+
+message SensorTimeStep {  // apply to a Accelerometer, Camera, Gyro, PositionSensor or TouchSensor
+    string name     = 1;
+    double timeStep = 2;  // time between two measurements expressed in [s], disabled if 0
+}
+
+message CameraQuality {
+    string name    = 1;
+    int32  quality = 2;  // JPEG quality
```
KipHamiltons, line 124: wdym quality? Is this the same as the other `quality` field in `CameraMeasurement`? please elaborate

shared/message/platform/webots/messages.proto
```diff
@@ -0,0 +1,141 @@
...
+ * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright 2021 NUbots <nubots@nubots.net>
+ */
+
+syntax = "proto3";
+
+package message.platform.webots;
+
+// Copy of cyberbotics/webots projects/samples/contests/robocup/controllers/player/messages.proto
```
KipHamiltons, line 24: Not sure if the comments I have left for this file can be addressed easily if we don't have more info than just the copy/paste job. 
Also, this filepath doesn't exist any more, unless I am confused and you didn't mean that this was on the `cyberbotics/webots` repo on GitHub. Not sure what they have done with this file at all.

roles/webots.role
CMurtagh-LGTM: ¯\_(ツ)_/¯

module/platform/Webots/src/Webots.cpp
CMurtagh-LGTM: Do you have a better idea?

shared/message/platform/webots/messages.proto
CMurtagh-LGTM: On [the webots' controller](https://github.com/cyberbotics/webots/pull/2888)

module/platform/Webots/src/Webots.cpp
KipHamiltons: `connection_request` perhaps? `send_player_details`?

roles/webots.role
Bidski: Webots is indeed a simulator. We are effectively thinking of `platform::webots` the same way we think of `platform::darwin::HardwareSimulator`

shared/message/platform/webots/messages.proto
```diff
@@ -66,27 +67,33 @@ message Force6DMeasurement {  // not yet implemented
...
 message SensorMeasurements {
-    double time = 1;  // time stamp at which the measurements were performed expressed in [s]
-    repeated PositionSensorMeasurement position_sensor = 2;
-    repeated AccelerometerMeasurement  accelerometer   = 3;
-    repeated GyroMeasurement           gyro            = 4;
-    repeated BumperMeasurement         bumper          = 5;
-    repeated Force3DMeasurement        force_3d        = 6;
-    repeated Force6DMeasurement        force_6d        = 7;
-    repeated CameraMeasurement         camera          = 8;
+    uint32           time     = 1;  // time stamp at which the measurements were performed expressed in [ms]
```
Bidski, line 87: `[ms]` with reference to some fixed point in time?? Simulation start time? Time that we connected to the server? Epoch time?

Bidski (changes requested):

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,257 @@
...
+    if ((error = getaddrinfo(server_name.c_str(), port.c_str(), &hints, &address)) != 0) {
+        log<NUClear::ERROR>(fmt::format("Cannot resolve server name: {}. Error {}. Error code {}",
+                                        server_name,
+                                        gai_strerror(error),
+                                        error));
+        return -1;
+    }
+
+    // Loop through the linked list of potential options for connecting. In order of best to worst.
+    for (addrinfo* addr_ptr; addr_ptr != NULL; addr_ptr = addr_ptr->ai_next) {
```
WongAaronSW, line 81:
```suggestion
    for (addrinfo* addr_ptr = address; addr_ptr != NULL; addr_ptr = addr_ptr->ai_next) {
```
addr_ptr was instantiated, but not assigned, need to assign it to the beginning of the list.

roles/webots.role
KipHamiltons: fair enough.

module/platform/Webots/src/Webots.cpp
```diff
@@ -98,26 +114,26 @@ Webots::Webots(std::unique_ptr<NUClear::Environment> environment) : Reactor(std:
...
         // clang-format on
 
-        int fd = tcpip_connect(cfg["server_address"].as<std::string>(), cfg["port"].as<int>());
+        int fd = tcpip_connect(cfg["server_address"].as<std::string>(), cfg["port"].as<std::string>());
 
         // Tell webots who we are
-        int team_id  = cfg["team_id"].as<int>();
-        int robot_id = cfg["robot_id"].as<int>();
-        send_connect(fd, team_id, robot_id);
+        send_connect(fd);
```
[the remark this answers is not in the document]
CMurtagh-LGTM: Would this be a race condition? It doesn't seem certain that either would come first.

module/platform/Webots/README.md
```diff
@@ -0,0 +1,14 @@
+webots
+======
+
+## Description
+
+
+## Usage
+
+
+## Emits
```
KipHamiltons, line 10: Emits compressed image and sensor data. The rest can be added as we go

module/platform/Webots/README.md
```diff
@@ -0,0 +1,14 @@
+webots
+======
+
+## Description
+
+
+## Usage
+
```
KipHamiltons, line 8: Fill in usage and description for sure

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,258 @@
...
+    // Store the ip address information that we will connect to
+    addrinfo* address;
+
+    int error;
+    if ((error = getaddrinfo(server_name.c_str(), port.c_str(), &hints, &address)) != 0) {
+        log<NUClear::ERROR>(fmt::format("Cannot resolve server name: {}. Error {}. Error code {}",
+                                        server_name,
+                                        gai_strerror(error),
+                                        error));
+        return -1;
```
KipHamiltons, line 79: Do we ever `#define BAD_FD -1`?

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,258 @@
...
+                return;
+            }
+
+            // Deserialise the message into a neutron
+            SensorMeasurements msg = NUClear::util::serialise::Serialise<SensorMeasurements>::deserialise(data);
+            
+            translate_and_emit_sensor(msg);
+        });
+
+        on<Every<1, std::chrono::seconds>>().then([this, fd]() {
```
KipHamiltons, line 149: is this fast enough?

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,258 @@
...
+            // Sending
+            std::vector<char> data = NUClear::util::serialise::Serialise<ActuatorRequests>::serialise(to_send);
+            uint32_t N             = htonl(data.size());
+            send(fd, &N, sizeof(N), 0);
+            send(fd, data.data(), N, 0);
```
KipHamiltons, line 154: Should we keep this consistent?
```suggestion
            uint32_t Nn             = htonl(data.size());
            send(fd, &Nn, sizeof(Nn), 0);
            send(fd, data.data(), Nn, 0);
```

shared/message/platform/webots/messages.proto
CMurtagh-LGTM: That'll be Cybertronics.

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,258 @@
...
+        for (auto& command : commands.targets) {
+            MotorPosition position_msg;
+            position_msg.name     = command.id;
+            position_msg.position = command.position;
+            to_send.motor_positions.push_back(position_msg);
+
+            // TODO(cmurtagh) work out if gain is velocity or force
+            MotorVelocity velocity_msg;
+            velocity_msg.name     = command.id;
+            velocity_msg.velocity = command.gain;
```
KipHamiltons, line 175: velocity != gain

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,257 @@
...
+            torque_msg.torque = command.torque;
+            to_send.motor_torques.push_back(torque_msg);
+
+            // MotorPID ? Do we need to send this?
+        }
+    });
+}
+
+void Webots::send_player_details(const int& fd, const GlobalConfig& player_details) {
+    // TODO(cameron) resend if failes
```
KipHamiltons, line 189: Action this TODO

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,257 @@
...
+    std::vector<char> data = NUClear::util::serialise::Serialise<GlobalConfig>::serialise(player_details);
+    uint32_t N             = htonl(data.size());
+    send(fd, &N, sizeof(N), 0);
+    send(fd, data.data(), N, 0);
```
KipHamiltons, line 193: Should we keep this consistent?
```suggestion
    uint32_t Nn             = htonl(data.size());
    send(fd, &Nn, sizeof(Nn), 0);
    send(fd, data.data(), Nn, 0);
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,257 @@
...
+}
+
+void Webots::translate_and_emit_sensor(const SensorMeasurements& sensor_measurements) {
+            // Read each field of msg, translate it to our protobuf and emit the data
+            auto sensor_data = std::make_unique<DarwinSensors>();
+
+
+            sensor_data->timestamp = sensor_measurements.time;  // Not sure if we want this or the timestamp on the received message.
+
+            // Vecotor3 is a neutron of 3 doubles
```
KipHamiltons, line 203:
```suggestion
            // Vector3 is a neutron of 3 doubles
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+            for (auto& position : sensor_measurements.position_sensors) {
+                // string name
+                // double valvue
```
KipHamiltons, line 207:
```suggestion
                // double value
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,257 @@
...
+            }
+
+            for (auto& force_6d : sensor_measurements.force6ds) {
+                // string name
+                // Vector3 value
+            }
+
+            emit(sensor_data);
+
+            for (auto& camera : sensor_measurements.cameras) {
```
KipHamiltons, line 237: Can we make all of these loop variables like this one `const`?
```suggestion
            for (const auto& camera : sensor_measurements.cameras) {
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,257 @@
...
+                compressed_image->name           = camera.name;
+                compressed_image->dimensions.x() = camera.width;
+                compressed_image->dimensions.y() = camera.height;
+                compressed_image->format         = fourcc("JPEG");
+                compressed_image->data           = camera.image;
+                emit(compressed_image);
+            }
+
+            // Parse the messages from Webots and log them. Maybe check for certain things.
+            for (auto& message : sensor_measurements.messages) {
```
KipHamiltons, line 249:
```suggestion
            for (const auto& message : sensor_measurements.messages) {
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,257 @@
...
+                auto compressed_image            = std::make_unique<CompressedImage>();
+                compressed_image->name           = camera.name;
+                compressed_image->dimensions.x() = camera.width;
+                compressed_image->dimensions.y() = camera.height;
+                compressed_image->format         = fourcc("JPEG");
+                compressed_image->data           = camera.image;
+                emit(compressed_image);
+            }
+
+            // Parse the messages from Webots and log them. Maybe check for certain things.
```
KipHamiltons, line 248: This comment doesn't quite match what the switch is doing, since it's only logging errors and warnings. The comment implies to me that all messages are being logged.

KipHamiltons (commented):
More review action

module/platform/Webots/src/Webots.cpp
TrentHouliston: Velocity is neither gain nor force

module/platform/Webots/src/Webots.cpp
CMurtagh-LGTM: Probably not

module/platform/Webots/src/Webots.cpp
CMurtagh-LGTM: https://www.youtube.com/watch?v=nJKyRvn3k7E

module/platform/Webots/src/Webots.cpp
CMurtagh-LGTM: As long as we have a cont iterator

shared/message/platform/webots/messages.proto
CMurtagh-LGTM: Simulation time since the start of the connection with the Webots controller.

KipHamiltons:
Besides fixing formatting, does this branch need anything else @CMurtagh-LGTM ?

module/platform/Webots/src/Webots.cpp
CMurtagh-LGTM: Doesn't look like this is even used :p

CMurtagh-LGTM:
> Besides fixing formatting, does this branch need anything else @CMurtagh-LGTM ?
### Super necessary
- Do the weird clock thing.
- @miikyla 's dev day code
- A ton of testing

### Can get stuff done if not yet complete
- Finish off the robustness
- Halt when denied connection
- Add `close(fd)` to the `on<Shutdown>()`

module/platform/Webots/src/Webots.cpp
CMurtagh-LGTM: `constexpr int BAD_FD = -1;` would be a more c++ way of doing it.

shared/message/platform/webots/messages.proto
CMurtagh-LGTM: We don't really use it anyway.

shared/message/platform/webots/ConnectRequest.proto
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright 2021 NUbots <nubots@nubots.net>
+ */
+
+syntax = "proto3";
+
+package message.platform.webots;
+
+// The message to send when we connect or want to reconnect to the webots robocup
+// https://cdn.robocup.org/hl/wp/2021/03/v-hsc_simulator_api.pdf
+
```
KipHamiltons, line 26: To keep it consistent with my document proto efforts :pp
```suggestion
/** The message to send when we connect or want to reconnect to the webots robocup simulation
 * https://cdn.robocup.org/hl/wp/2021/03/v-hsc_simulator_api.pdf
 */
```

module/platform/Webots/src/Webots.cpp
KipHamiltons: might be overkill either way :shrug:

module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,377 @@
...
+                // Use configuration here from file webots.yaml
+
+                // clang-format off
+            auto lvl = local_config["log_level"].as<std::string>();
+            if (lvl == "TRACE") { this->log_level = NUClear::TRACE; }
+            else if (lvl == "DEBUG") { this->log_level = NUClear::DEBUG; }
+            else if (lvl == "INFO") { this->log_level = NUClear::INFO; }
+            else if (lvl == "WARN") { this->log_level = NUClear::WARN; }
+            else if (lvl == "ERROR") { this->log_level = NUClear::ERROR; }
+            else if (lvl == "FATAL") { this->log_level = NUClear::FATAL; }
```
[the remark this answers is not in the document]
CMurtagh-LGTM: See #383 for future plans.

module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,377 @@
...
+                // clang-format on
+
+                on<Watchdog<Webots, 5, std::chrono::seconds>>().then([this] {
+                    // We haven't received any messages lately
```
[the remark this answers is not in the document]
CMurtagh-LGTM: TODO

module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,377 @@
...
+                log<NUClear::ERROR>(
+                    fmt::format("Error in sending actuator requests message size, {}", strerror(errno)));
+            }
+            if (send(fd, data.data(), Nn, 0) == Nn) {
+                log<NUClear::ERROR>("Error in sending actuator requests message, {}", strerror(errno));
+            }
+        });
+
+        error_io = on<IO>(fd, IO::CLOSE | IO::ERROR).then([this, fd](const IO::Event& event) {
+            // Something went wrong
```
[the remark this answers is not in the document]
CMurtagh-LGTM: TODO

module/platform/webots/Webots/src/Webots.cpp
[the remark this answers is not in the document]
KipHamiltons: That PR doesn't address the question of what happens if the string is not one of those options tho :p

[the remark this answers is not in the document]
CMurtagh-LGTM: Comment on that PR then

module/platform/webots/Webots/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Start the Robocup contoller and run NUbots.
+
+## Emits
+platform::darwin::DarwinSensors
+output::CompressedImage
+
+## Dependencies
+Configuration
+GlobalConfiguration -- probably not
```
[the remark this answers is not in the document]
CMurtagh-LGTM: I think we may actually need it now.

CMurtagh-LGTM:
We're up to trying to integrate it with https://github.com/NUbots/NUWebots/pull/31 and https://github.com/RoboCup-Humanoid-TC/webots

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,656 @@
...
+                velocity = 0;
+            }
+            velocity_msg.name     = translate_id_servo(target.id);
+            velocity_msg.velocity = velocity;
+            to_send_next.motor_velocities.push_back(velocity_msg);
+
+            MotorTorque torque_msg;
+            torque_msg.name   = translate_id_servo(target.id);
+            torque_msg.torque = target.torque;
+            to_send_next.motor_torques.push_back(torque_msg);
```
Bidski, line 212: We need to query how the robocup controller is going to use this message. If this is part of force/torque control mode then we dont need to set this message at all. If it is used to set the maximum available torque then we do need to set it (but I dont think this is the case)

https://cyberbotics.com/doc/reference/motor#force-and-torque-control

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,656 @@
...
+            // TODO(Cameron) Try to reconnect after a delay.
+            // setup_connection(server_address, port);
+            log<NUClear::FATAL>("Quitting due to failed connection attempt");
+            powerplant.shutdown();
+            return;
+        }
+
+        // Initaliase the string with ???????
+        std::string initial_message = std::string(7, '?');
+        const int n                 = recv(fd, initial_message.data(), sizeof(initial_message), 0);
```
Bidski, line 322: Make `recv` wait until the entire message has been received (I know its only 7 bytes, but forcing the wait might not be a bad thing)
```suggestion
        const int n                 = recv(fd, initial_message.data(), sizeof(initial_message), MSG_WAITALL);
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,656 @@
...
+            uint64_t message = uint64_t(recv(fd, data.data(), Nh, MSG_WAITALL));
+            if (message != Nh) {
+                log<NUClear::ERROR>("Failed to read message from TCP connection");
+                return;
+            }
+
+            log<NUClear::TRACE>("Received sensor measurements");
+
+            // Deserialise the message into a neutron
+            SensorMeasurements msg = NUClear::util::serialise::Serialise<SensorMeasurements>::deserialise(data);
```
Bidski, line 381: If `Nh == 0` we should set `msg` to a default constructed `SensorMeasurements` message (if I am not mistaken, a protobuf message with all fields set to their defaults has a 0 byte serialised length)

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,656 @@
...
+            utility::clock::custom_rtf =
+                static_cast<double>(sim_delta + prev_sim_delta) / static_cast<double>(real_delta + prev_real_delta);
+
+            // Update our current times
+            current_sim_time  = msg.time;
+            current_real_time = msg.real_time;
+        });
+
+        send_io =
+            on<IO, With<ServoTargets>, With<DarwinSensors>>(fd, IO::READ)
```
Bidski, line 410: `IO::READ`?? not `IO::WRITE`? Maybe this would be better as a 90Hz (or whatever) update loop like we have in in `HardwareIO`?

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,656 @@
...
+        uint32_t Nn = htonl(data.size());
+
+        // Send the message size first
+        if (send(fd, &Nn, sizeof(Nn), 0) != sizeof(Nn)) {
+            log<NUClear::ERROR>(fmt::format("Error in sending ActuatorRequests' message size,  {}", strerror(errno)));
+        }
+        // then send the data
+        if (send(fd, data.data(), data.size(), 0) != (signed) data.size()) {
+            log<NUClear::ERROR>(fmt::format("Error sending ActuatorRequests message, {}", strerror(errno)));
+        }
```
Bidski, line 458: These seem like fatal conditions?

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,656 @@
...
+        // target.position = 2;
+        // target.gain     = 10;
+        // target.torque   = 10;
+        // auto targets    = std::make_unique<ServoTargets>();
+        // targets->targets.push_back(target);
+        // emit(targets);
+    }
+
+    void Webots::translate_and_emit_sensor(const SensorMeasurements& sensor_measurements) {
+        if (false) {  // Change to true to print the received sensor_measurements, will remove after this is done
```
Bidski, line 474: Maybe just change all of these to `log<NUClear::TRACE>()` statements? Optionally, also wrap them in a `#ifndef NDEBUG ... #endif` block?

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,656 @@
...
+        for (const auto& force_3d : sensor_measurements.force3ds) {
+            // string name
+            // Vector3 value
+        }
+
+        for (const auto& force_6d : sensor_measurements.force6ds) {
+            // string name
+            // Vector3 value
+        }
+        */
```
Bidski, line 630: Delete commented code? Pretty sure we will never have these? Unless FSRs manifest as bumpers?

module/platform/Webots/src/Webots.hpp
```diff
@@ -0,0 +1,81 @@
...
+    class Webots : public NUClear::Reactor {
+    private:
+        /// @brief Store a handle to the io reaction so we can unbind it
+        ReactionHandle read_io;
+        /// @brief Store a handle to the io reaction so we can unbind it
+        ReactionHandle send_io;
+        /// @brief Store a handle to the io error reaction so we can unbind it
+        ReactionHandle error_io;
+        /// @brief Store a handle to the shutdown reaction so we can unbind it
+        ReactionHandle shutdown_handle;
```
Bidski, line 39: Better comments on these please

Bidski (changes requested):

module/platform/Webots/src/Webots.cpp
KipHamiltons: Pretty sure FSRs are bumpers.The rest can go

module/platform/Webots/src/Webots.cpp
tayloryoung6396: 90Hz is our current rate, but we will need to look at network bandwidth (probably more for images than sensors)

module/platform/Webots/src/Webots.cpp
tayloryoung6396: the ifdef is handles by the if(false) statement.. and i dont care for changing it :P

KipHamiltons (commented):
Small questions. Once these are answered I'll be approving

shared/clock/clock.cpp
```diff
@@ -16,7 +17,7 @@ namespace NUClear {
 
         // now is multiplied by the real time factor to sync
         // NUClear with the simulation time
-        auto now = std::chrono::steady_clock::now();
+        auto now = NUClear::base_clock::now();
```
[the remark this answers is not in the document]
Bidski: Not sure, but my guess would be that `NUClear::base_clock` is not always the same thing as `std::chrono::steady_clock`? Which probably means there are more places that this change needs to be applied

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,625 @@
...
+            auto lvl = config["log_level"].as<std::string>();
+            if      (lvl == "TRACE") { this->log_level = NUClear::TRACE; }
+            else if (lvl == "DEBUG") { this->log_level = NUClear::DEBUG; }
+            else if (lvl == "INFO")  { this->log_level = NUClear::INFO; }
+            else if (lvl == "WARN")  { this->log_level = NUClear::WARN; }
+            else if (lvl == "ERROR") { this->log_level = NUClear::ERROR; }
+            else if (lvl == "FATAL") { this->log_level = NUClear::FATAL; }
+            // clang-format on
+
+            on<Watchdog<Webots, 5, std::chrono::seconds>>().then([this, config] {
```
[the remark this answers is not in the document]
Bidski: That should be fine

KipHamiltons (dismissed):
L G T M

KipHamiltons:
and when I say that, I mean that I'll make a PR tomorrow fixing typos and making inconsequential changes after it's merged tomorrow :yum:

tayloryoung6396 (commented):
LGTM

Bidski (approved):
Sure

PR #407, opened 2021-03-27
Add systemd units files and instructions for docker cleaning

Bidski:


KipHamiltons (commented):
I think we should add stuff to set this up to the NUbook guide to set up a buildkite agent, and to the script which does that automatically

doc/BuildKiteDockerClean/docker_clean.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
KipHamiltons, line 1: this also presupposes bash

KipHamiltons (changes requested):

KipHamiltons:
This is _kind of_ redundant now we have GHA.. Not sure if it will be useful if and when we use self-hosted runners though. Keen to hear thoughts @Bidski @Y0sh1dk

CMurtagh-LGTM:
We have moved to github actions, I'm assuming this in now irrelevant.

PR #408, opened 2021-03-28
Progresses communication efforts

KipHamiltons:
It still doesn't compile, but it's getting close-ish I think

We need to convert to make `gethostbyname`'s param a `char*` because C. 

Also need to decide what to do about the `log` calls. they're not valid as they are outside of a reactor or powerplant, so I'll have to talk to u Cameron about how you wanted to do those.


module/platform/Webots/src/Webots.cpp
```diff
@@ -1,21 +1,53 @@
...
+// Include headers needed for TCP connection
+#ifdef _WIN32
+    #include <winsock.h>
+#else
+    #include <netdb.h>      /* definition of gethostbyname */
+    #include <netinet/in.h> /* definition of struct sockaddr_in */
+    #include <sys/socket.h>
+    #include <sys/time.h>
+    #include <unistd.h> /* definition of close */
+#endif
```
CMurtagh-LGTM, line 41: Do we need to include Windows headers as we always run on the docker image?

module/platform/Webots/src/Webots.cpp
```diff
@@ -1,21 +1,53 @@
...
 
 namespace module::platform {
 
 using extension::Configuration;
 using message::motion::ServoTargets;
+using message::output::CompressedImage;
 using message::platform::darwin::DarwinSensors;
-using msessage::output::CompressedImage;
+
+using message::platform::webots::ActuatorRequests;
```
CMurtagh-LGTM, line 50:
```suggestion
using message::platform::webots::ActuatorRequests;
using message::platform::webots::ConnectRequest;
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -1,21 +1,53 @@
...
 // TODO(cmurtagh) work out namespace of protobuf files from webots
```
CMurtagh-LGTM, line 51:
```suggestion
```
I suppose we've worked this out.

CMurtagh-LGTM (changes requested):

module/platform/Webots/src/Webots.cpp
KipHamiltons: I don't think so?

CMurtagh-LGTM: From the slack discussion, "Probably not".

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+                    server->h_length);
+    }
+    else {
+        close(fd);
+        log<NUClear::ERROR>(fmt::format("Cannot resolve server name: {}", server_name));
+        return -1;
+    }
+
+    // Connect to the ip address
+    if (int error = connect(fd, (struct sockaddr*) &address, sizeof(address))) {
```
KipHamiltons, line 77: I want to note that I saw that everyone seems to cast their `sockaddr_in` like this when calling `bind(..)`, but I couldn't see people doing it for `connect(..)`, because they all seemed to do things fundamentally differently, usually using `getaddrinfo(..)` instead of `gethostbyname(..)`

module/platform/Webots/src/Webots.cpp
```diff
@@ -0,0 +1,228 @@
...
+int Webots::tcpip_connect(const std::string& server_name, const int& port) {
+    // Create the socket
+    int fd = socket(AF_INET, SOCK_STREAM, 0);
+
+    // Store the ip address information that we will connect to
+    sockaddr_in address;
+    memset(&address, 0, sizeof(sockaddr_in));
+    address.sin_family = AF_INET;
+    address.sin_port   = htons(port);
+    hostent* server    = gethostbyname(server_name.c_str());
```
KipHamiltons, line 62: the [linux docs](https://man7.org/linux/man-pages/man3/gethostbyname.3.html#DESCRIPTION) say you should use `getaddrinfo(..)` instead of `gethostbyname(..)`. I'll leave that for you to decide if that's worth acting on.

KipHamiltons (commented):
Still a couple of compiler errors, but we are actually getting close to building.

CMurtagh-LGTM (approved):
LGTM other than the protobuf vs neutron stuff, that can be a separate pr.

PR #409, opened 2021-03-28
Removes KickAtGoal

KipHamiltons:
This module isn't in any roles, and doesn't appear to have ever been in the `robocup` role. It doesn't compile if it is put in a role, since it uses functions and methods from `utility/time/time.hpp`, which doesn't exist.


ysims:
I assume this is just an old version of kickplanner + walk path planner? @Bidski @TrentHouliston

PR #410, opened 2021-04-11
Removes old field description config files

KipHamiltons:
I don't know why these still exist.
The current one will need updating before robocup too.


JosephusPaye (approved):
LGTM

PR #411, opened 2021-04-11
Removes look message

KipHamiltons:
This message is unused, undocumented, and no one in the team knows how to use it.


ysims (approved):
If nothing uses it I don't see why we should keep it. Behaviour needs a big overhaul anyway.

PR #412, opened 2021-04-11
Removes soccer simulator

KipHamiltons:
"Soccer simulator stinks" - @Bidski, probably


ysims (approved):
I'm good for this to happen. Tbh any simulator stuff can be removed since webots will replace it. The only exception I think would be HardwareSimulator since it's still nice to be able to run code without any external tools or robots.

PR #413, opened 2021-04-11
Update the image Compressor and Decompressor modules

TrentHouliston:
Applies some minor bug fixes found by clang-tidy as well as some general cleanup
Note this also makes ImageDecompressor compile when it didn't before


module/input/ImageDecompressor/src/ImageDecompressor.cpp
```diff
@@ -9,130 +9,128 @@
...
+                it = decompressors.insert(std::make_pair(image.id, std::make_shared<DecompressorContext>())).first;
+                it->second->width  = image.dimensions[0];
+                it->second->height = image.dimensions[1];
+                it->second->format = image.format;
+
+                for (auto& f : config.factories) {
+                    for (int i = 0; i < f.second; ++i) {
+                        auto a = std::make_unique<std::atomic<bool>>();
+                        it->second->decompressors.emplace_back(DecompressorContext::Decompressor{
+                            std::move(a),
```
CMurtagh-LGTM, line 66: It seems a bit redundant to create a lvalue then immediately move it.

module/input/ImageDecompressor/src/ImageDecompressor.cpp
```diff
@@ -9,130 +9,128 @@
...
-                                            100 * double(decompressed) / double(decompressed + dropped)));
-            decompressed = 0;
-            dropped      = 0;
-        });
-    }
-
-}  // namespace input
-}  // namespace module
+        }
+        // We failed to compress this image
```
CMurtagh-LGTM, line 121:
```suggestion
        // We failed to decompress this image
```

module/output/ImageCompressor/src/ImageCompressor.cpp
```diff
@@ -12,175 +12,168 @@
...
+                it = compressors.insert(std::make_pair(image.id, std::make_shared<CompressorContext>())).first;
+                it->second->width  = image.dimensions[0];
+                it->second->height = image.dimensions[1];
+                it->second->format = image.format;
+
+                for (auto& f : config.factories) {
+                    for (int i = 0; i < f.second; ++i) {
+                        auto a = std::make_unique<std::atomic<bool>>();
+                        it->second->compressors.emplace_back(CompressorContext::Compressor{
+                            std::move(a),
```
CMurtagh-LGTM, line 108: Same as before.

module/output/ImageCompressor/src/compressor/CompressorFactory.hpp
```diff
@@ -10,7 +10,7 @@ namespace module::output::compressor {
 
 class CompressorFactory {
 public:
-    virtual std::shared_ptr<Compressor> make_compressor(const uint32_t width,
+    virtual std::shared_ptr<Compressor> make_compressor(uint32_t width,
                                                         const uint32_t& height,
                                                         const uint32_t& format) = 0;
```
CMurtagh-LGTM, line 15: It seems odd that width is passed by value.

CMurtagh-LGTM (changes requested):
Maybe someone who knows typescript can have a peek at those changes.

module/input/ImageDecompressor/src/ImageDecompressor.cpp
TrentHouliston: Wasn't previously this way, but clang static analysis flags this as a memory leak (incorrectly) if you don't do it this way.

module/output/ImageCompressor/src/ImageCompressor.cpp
TrentHouliston: Same as before, clang static analysis thinks it's a memory leak if you don't

module/output/ImageCompressor/src/compressor/CompressorFactory.hpp
TrentHouliston: That does seem odd..., but the answer would be to make it a const reference, not remove the const

CMurtagh-LGTM: I've had a look and it seems that width is immediately passed to a constructor that takes width as a const reference.

CMurtagh-LGTM (approved):
The c++ side LGTM, but I've no idea on ts.

nusight2/src/client/components/vision/network.ts
```diff
@@ -128,8 +120,8 @@ export class VisionNetwork {
         goal.side === message.vision.Goal.Side.LEFT
           ? 'left'
           : goal.side === message.vision.Goal.Side.RIGHT
-          ? 'right'
-          : 'unknown',
+            ? 'right'
```
JosephusPaye, line 123: Why did the formatting here change? Indented nested ternaries are from an older version of Prettier.

PR #414, opened 2021-04-12
Add in message reflection capabilities to Neutrons

TrentHouliston:
Also fix up some minor errors that were in the Neutron.cmake file

Merge in #413 first as it fixes the id fields for image/compressed image


Bidski (approved):
Looks OK, I think

PR #417, opened 2021-04-12
Make the recordings symlink be generated by utility not DataPlayback

TrentHouliston:
Currently if you make a role that tries to do something with the recordings directory in docker, it will only find it if you happend to build the DataPlayback module in a role.

This ensures it's always created by making it a dependency of utility rather than that module


shared/utility/libraries.cmake
```diff
@@ -21,4 +21,13 @@ if(CMAKE_BUILD_TYPE STREQUAL "Debug")
   target_link_libraries(nuclear_utility PUBLIC libbacktrace::libbacktrace ${CMAKE_DL_LIBS})
 endif()
 
+# Generate the decoders for each of the message types we know
+add_custom_command(
+  OUTPUT "${source_file}"
```
KipHamiltons, line 26: I am probably missing something, but `${source_file}` is undefined here, so I don't know what `add_custom_command` is meant to output

shared/utility/libraries.cmake
```diff
@@ -21,4 +21,13 @@ if(CMAKE_BUILD_TYPE STREQUAL "Debug")
...
+  COMMAND ${PYTHON_EXECUTABLE} ARGS "${CMAKE_CURRENT_SOURCE_DIR}/src/generate_dataplayback.py"
+          "${PROJECT_BINARY_DIR}/nuclear/message/python" "${source_file}"
```
KipHamiltons, line 28: that also makes this line strange

KipHamiltons (commented):
Bit confused about this one

shared/utility/libraries.cmake
TrentHouliston: I clearly copied the wrong add_custom_command.... this is what happens when you do too many PRs at the same time

KipHamiltons (approved):
Presuming this works, LGTM

PR #419, opened 2021-04-12
Upgrades the DataLogging module to write an index as it writes the log

TrentHouliston:
Depends on #414 as it uses the message reflection code


module/support/logging/DataLogging/src/DataLogging.cpp
```diff
@@ -1,138 +1,200 @@
...
-#include <fmt/format.h>
+#include <filesystem>
 #include <iomanip>
 #include <sstream>
+#include <zstr.hpp>
 
 #include "extension/Configuration.hpp"
 
-#include "utility/file/fileutil.hpp"
+#include "utility/nbs/Index.hpp"
```
CMurtagh-LGTM, line 10: NUbots does not have this :P.

CMurtagh-LGTM (changes requested):

module/support/logging/DataLogging/src/DataLogging.cpp
TrentHouliston: Seems like an oversight...

CMurtagh-LGTM (approved):
¯\_(ツ)_/¯

PR #420, opened 2021-04-12
Adds clang-tidy to the codebase

TrentHouliston:
This adds clang-tidy to the codebase, also it adjusts a clang-format rule regarding indentation as the new clang-tidy rules will enforce c++17 style namespace declarations which mess up the nice indentation of our code.

We should probably not enforce these clang-tidy rules until we have resolved all the issues they find


CMakeLists.txt
```diff
@@ -70,12 +76,26 @@ if(USE_ASAN)
   link_libraries(asan)
 endif()
 
+# Enable testing if we are building tests
+if(BUILD_TESTS)
```
KipHamiltons, line 80: This if already exists in here

CMakeLists.txt
```diff
@@ -70,12 +76,26 @@ if(USE_ASAN)
...
 set(CMAKE_EXPORT_COMPILE_COMMANDS
     ON
     CACHE STRING "Enable/Disable output of compile commands during generation." FORCE
 )
 
+# Enable all warnings that we can
+add_compile_options(-Wall -Wpedantic -Wextra)
+
+# This is needed everywhere so that we can catch segmentation faults
+add_compile_options(-fnon-call-exceptions)
```
KipHamiltons, line 94: You've removed `-Wno-int-in-bool-context`. Is that warning gone now?

CMakeLists.txt
```diff
@@ -38,29 +42,31 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
-
-# Make sure our binaries can find our libraries both in docker and on the robot
-list(APPEND CMAKE_INSTALL_RPATH /home/nubots/toolchain/)
-list(APPEND CMAKE_INSTALL_RPATH /usr/local/lib)
-
-# Disable armadillo bounds checking in release and min size release builds
-list(APPEND CMAKE_CXX_FLAGS_RELEASE "-DARMA_NO_DEBUG")
-list(APPEND CMAKE_CXX_FLAGS_MINSIZEREL "-DARMA_NO_DEBUG")
```
KipHamiltons, line 56: why are you deleting these lines?

CMakeLists.txt
```diff
@@ -38,29 +42,31 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
-# Default not to run the clang-tidy checks
+# Default not to run the clang-tidy checks, default to whatever our CI_BUILD is
 option(ENABLE_CLANG_TIDY "Enable building with clang-tidy checks." OFF)
-if(ENABLE_CLANG_TIDY)
-  set(CMAKE_CXX_CLANG_TIDY clang-tidy)
-  set(CMAKE_C_CLANG_TIDY clang-tidy)
-endif()
+if(ENABLE_CLANG_TIDY OR CI_BUILD)
+  find_package(PythonInterp 3 REQUIRED)
+  set(CMAKE_CXX_CLANG_TIDY "${PYTHON_EXECUTABLE}" "${PROJECT_SOURCE_DIR}/cmake/scripts/clang-tidy.py"
```
KipHamiltons, line 49: where is `PYTHON_EXECUTABLE` defined? am I missing something?

KipHamiltons (commented):
a couple of questions

CMakeLists.txt
TrentHouliston: It gets defined by `find_package(PythonInterp 3 REQUIRED)`. That find script finds python and sets that variable

CMakeLists.txt
TrentHouliston: Because they were solved by the modern cmake stuff and just were missed. We are setting public attributes on things we're building that require cxx_17 rather than setting them globally like this.

CMakeLists.txt
TrentHouliston: This warning was fixed by the modern cmake stuff. The reason it showed up was that the Eigen library we used had that warning and it wasn't properly being included as SYSTEM. Which meant our compiler treated it as "our" codes issue not one in an included library. Now that it's properly set we can have that rule again, and if we break it in our code we should just not use an int in a bool context

CMakeLists.txt
TrentHouliston: Indeed, I have turned off tests being extra enabled

CMakeLists.txt
```diff
@@ -38,29 +42,31 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
+  set(CMAKE_C_CLANG_TIDY ${CMAKE_CXX_CLANG_TIDY})
+
+  # Create a target that will apply clang-tidy fixes to the codebase
+  add_custom_target(
+    apply-clang-tidy
+    COMMAND clang-apply-replacements --format --style=file --style-config="${PROJECT_SOURCE_DIR}"
+            --remove-change-desc-files "${PROJECT_BINARY_DIR}/clang-tidy-fixes"
+    COMMENT "Applying fixes from clang-tidy to the codebase."
+  )
+endif(ENABLE_CLANG_TIDY OR CI_BUILD)
```
KipHamiltons, line 61: We don't use endif conditions any more
```suggestion
endif()
```

CMakeLists.txt
```diff
@@ -38,29 +42,31 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
+
+# If we are doing a CI build then we want to enable -Werror when compiling warnings are bad. We will also make it fail
+# if clang-tidy has an error
+if(CI_BUILD)
+  add_compile_options(-Werror)
+  set(CMAKE_CXX_CLANG_TIDY ${CMAKE_CXX_CLANG_TIDY} -warnings-as-errors=*)
+  set(CMAKE_C_CLANG_TIDY ${CMAKE_C_CLANG_TIDY} -warnings-as-errors=*)
+endif(CI_BUILD)
```
KipHamiltons, line 69:
```suggestion
endif()
```

CMakeLists.txt
```diff
@@ -38,29 +42,31 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
+    COMMAND clang-apply-replacements --format --style=file --style-config="${PROJECT_SOURCE_DIR}"
+            --remove-change-desc-files "${PROJECT_BINARY_DIR}/clang-tidy-fixes"
+    COMMENT "Applying fixes from clang-tidy to the codebase."
+  )
+endif(ENABLE_CLANG_TIDY OR CI_BUILD)
+
+# If we are doing a CI build then we want to enable -Werror when compiling warnings are bad. We will also make it fail
+# if clang-tidy has an error
+if(CI_BUILD)
+  add_compile_options(-Werror)
```
KipHamiltons, line 66: We have compiler warnings which exist already - are we going to get rid of those? surely we'll have to, to use this

CMakeLists.txt
```diff
@@ -76,6 +82,15 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS
...
 )
 
+# Enable all warnings that we can
+add_compile_options(-Wall -Wpedantic -Wextra)
+
+# This is needed everywhere so that we can catch segmentation faults
+add_compile_options(-fnon-call-exceptions)
+
+# Make NUClear use the system clock, as we are syncing it with ptp
+add_compile_definitions(NUCLEAR_CLOCK_TYPE=std::chrono::steady_clock)
```
KipHamiltons, line 92: I presume this plays nicely with and without a custom clock?

cmake/scripts/clang-tidy.py
```diff
@@ -0,0 +1,22 @@
...
+
+# Hash the remaining arguments as the output file name
+output_file = "{}.yaml".format(hashlib.sha256(" ".join(sys.argv[2:]).encode("utf-8")).hexdigest())
+
+# Work out our arguments to call the clang-tidy binary with
+args = [clang_tidy_binary, "--export-fixes={}".format(os.path.join(output_folder, output_file)), *sys.argv[3:]]
+result = subprocess.run(args)
+
+# Pass the return code back to the caller
+exit(result.returncode)
```
KipHamiltons, line 22: you can `exit(subprocess.run(args).returncode` if you like...

CMakeLists.txt
TrentHouliston: Well on CI build you'll need to get rid of all compiler warnings as well as all clang-tidy warnings. That's why I'm not turning on this mode in the CI build currently

CMakeLists.txt
TrentHouliston: With the latest version of NUClear it does, I should check you guys are using that

cmake/scripts/clang-tidy.py
TrentHouliston, line 22:
```suggestion

# Run clang-tidy and pass the return code back to the caller
exit(subprocess.run(args).returncode)
```

KipHamiltons (dismissed):
LGTM

KipHamiltons (approved):

PR #421, opened 2021-04-14
Comments for changing mirrors

Y0sh1dk:
Adds comments to files where mirrors need to be changed so we can link to file instead of to line.

Related to this [PR](https://github.com/NUbots/NUbook/pull/101)


Bidski (approved):
I think this should work

PR #422, opened 2021-04-14
Implementing sensor translations to and from webots

miikyla:


module/platform/Webots/src/Webots.cpp
```diff
@@ -309,4 +329,32 @@ void Webots::translate_and_emit_sensor(const SensorMeasurements& sensor_measurem
...
+        case "right_knee_pitch": return servos.rKnee;
+        case "left_knee_pitch": return servos.lKnee;
+        case "right_ankle_pitch": return servos.rAnklePitch;
+        case "left_ankle_pitch": return servos.lAnklePitch;
+        case "right_ankle_roll": return servos.rAnkleRoll;
+        case "left_ankle_roll": return servos.lAnkleRoll;
+        case "neck_yaw": return servos.headPan;
+        case "head_pitch": return servos.headTilt;
+
+        default: throw std::runtime_error("Out of bounds on servo name");
```
KipHamiltons, line 356: Since we are switching on a string, is it correct to say "out of bounds"? 
```suggestion
        default: throw std::runtime_error("Invalid servo name");
```
Perhaps?

module/platform/Webots/src/Webots.cpp
```diff
@@ -260,18 +270,27 @@ void Webots::translate_and_emit_sensor(const SensorMeasurements& sensor_measurem
...
         // double value
+        translate_servo_id(position.name, sensor_data.servo).presentPosition = position.value;
     }
 
     for (const auto& accelerometer : sensor_measurements.accelerometers) {
         // string name
         // Vector3 value
+        sensor_data.accelerometer.x = static_cast<float>(accelerometers.value.X);
+        sensor_data.accelerometer.y = static_cast<float>(accelerometers.value.Y);
+        sensor_data.accelerometer.z = static_cast<float>(accelerometers.value.Z);
```
KipHamiltons, line 281: A couple of issues with this I think. I don't think `accelerometers` is defined (the right side variable which owns the values). I think that was supposed to be `accelerometer` - the thing the for loop gives you.

The other thing is I am pretty sure that `sensor_data.accelerometers` (note the `s`) is an array, where you actually want to be assigning to the specific id associated with that servo. This will probably get messy.

module/platform/Webots/src/Webots.cpp
```diff
@@ -260,18 +270,27 @@ void Webots::translate_and_emit_sensor(const SensorMeasurements& sensor_measurem
...
     }
 
     for (const auto& gyro : sensor_measurements.gyros) {
         // string name
         // Vector3 value
+        sensor_data.gyroscope.x = static_cast<float>(gyros.value.X);
+        sensor_data.gyroscope.y = static_cast<float>(gyros.value.Y);
+        sensor_data.gyroscope.z = static_cast<float>(gyros.value.Z);
```
KipHamiltons, line 289: similar story here. I don't think `gyros` is defined, and we need to be assigning to the correct index, for the given `gyro`. As it stands both of these loops would (if `sensor_data.gyroscope` was a thing) just overwrite the same values for each loop iteration

module/platform/Webots/src/Webots.cpp
```diff
@@ -149,7 +152,14 @@ Webots::Webots(std::unique_ptr<NUClear::Environment> environment) : Reactor(std:
             torque_msg.torque = command.torque;
             to_send.motor_torques.push_back(torque_msg);
 
-            // MotorPID ? Do we need to send this?
+            // MotorPID, only sending P gain, set I and D to zero
```
KipHamiltons, line 155:
```suggestion
            // MotorPID, only sending P gain. Set I and D to zero
```

KipHamiltons (commented):
I think this will need another look. Also, it's quite difficult to know if we're on the right track when the base branch doesn't build for other reasons.

Side note: to fix the formatting, just run `./b format`, then commit those changes.

module/platform/Webots/src/Webots.cpp
```diff
@@ -309,4 +329,32 @@ void Webots::translate_and_emit_sensor(const SensorMeasurements& sensor_measurem
     }
 }
 
+// Converts the NUgus.proto servo name to the equivalent DarwinSensor.proto name
+DarwinSensors::Servo& translate_servo_id(const std::string& name, const DarwinSensors::Servos& servos) {
+    switch(name){
+        case "right_shoulder_pitch_sensor": return servos.rShoulderPitch;
+        case "left_shoulder_pitch": return servos.lShoulderPitch;
```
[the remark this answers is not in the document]
JosephusPaye: Left is what we've called it in the NUgus.proto file from webots, right is our existing stuff. We should probably change the webots proto to match our DarwinSensors.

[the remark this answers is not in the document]
JosephusPaye: Seems like we need the `_sensor` suffix, the corresponding motor is given the non-suffixed name:

```
device [
  RotationalMotor {
    name "right_shoulder_roll"
    maxVelocity 8.16814
    minPosition -3.14159265359
    maxPosition 3.14159265359
    maxTorque 7.3
  }
  PositionSensor {
    name "right_shoulder_roll_sensor"
  }
]
```
I'll at least make sure they all have the suffix.

module/platform/Webots/src/Webots.cpp
JosephusPaye: Updated the error message. Also, we can't actually switch on strings.

module/platform/Webots/src/Webots.cpp
```diff
@@ -251,27 +301,28 @@ void Webots::translate_and_emit_sensor(const SensorMeasurements& sensor_measurem
...
     for (const auto& position : sensor_measurements.position_sensors) {
-        // string name
-        // double value
+        translate_servo_id(position.name, sensor_data->servo).presentPosition = position.value;
     }
 
     for (const auto& accelerometer : sensor_measurements.accelerometers) {
-        // string name
-        // Vector3 value
+        sensor_data->accelerometer.x = static_cast<float>(accelerometer.value.X);
```
KipHamiltons, line 313: Why are we iterating and thus reading through a list of accelerometers when we are writing to one accelerometer? as it stands, this will just write `sensor_data` with whatever accelerometer is last on the `sensor_measurements` list of accelerometers. This applies for the gyros too.

KipHamiltons: unless I have confused myself into a corner lol

JosephusPaye: Missed this. It looks like the Webots `SensorMeasurements.accelerometers` is the list of all accelerometers on the robot? In our case it'll always be of length 1.

We could get rid of the loop and assume the first or last in the list is the one we want.

KipHamiltons: I suppose we should test and just see what works?

JosephusPaye: Yep, waiting for Cameron's comms to test this. Seems pretty close.

PR #423, opened 2021-04-18
Removes unused line(s) message

KipHamiltons:
We don't use this. If we want to pass a message with a `Line`, we can use `Line` from `Geometry.proto`. If we want to pass multiple lines in one message, that is dead easy to recreate.


ysims (approved):
LGTM

PR #429, opened 2021-04-22
Refactors motionmodel to abstract away the monolithic vector representation

KipHamiltons:
The monolithic vector representation of all of the vectors in the `StateVec` was quite difficult to deal with and read (for me at least...), so I did something about it.
This PR is focused around replacing that `StateVec` representation with a `StateVec` struct which stores each vector separately, and only combines them into a monolithic vector when required.
I have left several TODOs, which represent questions I want to answer before merging:

- [x] In `time(..)` on `MotionModel`, when we are adding our rotational velocity to our orientation, I think there should surely be a cleaner way to do it, but I have no idea how. Help would be appreciated pls
- **I have made it cleaner, but it has spawned a follow up question:**
- [x] In that same line in `time(..)`, we update the rotation by **adding** the old rotation quaternion to the discrete change in the rotation. This seems dirty and wrong, because adding unit quaternions is from what I gather totally wrong(???). I have updated it so that it is multiplication - is that correct?
- [x] In one of the `predict(..)` methods, we add our bias - shouldn't we be subtracting bias? I suppose you could (and we might) just negate your bias which would make this correct, but I thought bias was generally something you wanted to remove, instead of add, conceptually speaking.
- **It is correct adding the bias**, because it's predicting what the measurement will be based on the current state, which will be the measurement plus the bias.
- [x] This isn't so much of a question, as a thought - we should probably put some effort into making sure that we always use normalised quaternions. When we swapped from arma to eigen, we stopped using our unitquaternion wrapper which did that for us, and I think it's worth verifying that we have maintained the guarantees that wrapper gave us.
- **I am going to make an issue, such that this can be addressed later, possibly in a follow up PR**
- [x] Should a type conversion operator or method return a ref? They basically create a new object of that type and return it, so should they return a ref, just because refs==good, or some other reason?
- **It should return a copy**, since it's creating a new object, so that we enable copy elision.

Interestingly, these changes made the compiler notice a warning which previously it didn't notice on code I didn't change? lol

Further work would probably add a `StateVec` struct to the other `*Model` classes so that we can abstract away the underlying vector representation completely on `UKF`.


module/platform/darwin/SensorFilter/src/MotionModel.hpp
```diff
@@ -89,64 +143,72 @@ namespace module {
...
+                    Eigen::Quaternion<Scalar> Rwt(newState.Rwt);
 
                     // Apply our rotational velocity to our orientation
                     // https://fgiesen.wordpress.com/2012/08/24/quaternion-differentiation/
                     // Quaternions are stored internally as (x, y, z, w)
                     const Scalar t_2 = deltaT * Scalar(0.5);
-                    newState.template segment<4>(QX) =
-                        Rwt.coeffs()
-                        + t_2 * (Eigen::Quaternion<Scalar>(0.0, state[WX], state[WY], state[WZ]) * Rwt).coeffs();
+                    // TODO(KipHamiltons) this could probably be cleaner...
```
Bidski, line 172: Break it up into multiple statements

Bidski (commented):
Seems ok from what I can see. Splitting up the "complicated" expression in the `time()` function to multiple statements should help make that more readable. 

Making efforts to ensure the quaternion always remains normalised is not a bad idea

module/platform/darwin/SensorFilter/src/MotionModel.hpp
KipHamiltons: Roger that. In doing so, I have realised that the current implementation in master adds unit quaternions there - isn't that a no-no??? I have replaced it with a multiplication - is that correct?

KipHamiltons: Now I think about it, I am not sure that the scalar multiplication by 0.5 is proper too, because wouldn't that mess with the unit norm?? Quarternions are a headache lol

TrentHouliston: This math is based on some random StackOverflow somewhere that told us how to add a set of rotation rates to a quaternion. It then had a bunch of caveats as to where it would work

KipHamiltons: So above that code there was a link trying to explain part of it ([link](https://fgiesen.wordpress.com/2012/08/24/quaternion-differentiation/)), but that is more of a derivation. I found [this](https://math.stackexchange.com/a/2099673) stack overflow answer (which links that page) which justifies the math which existed there, but it has the caveat:
"[the formula we use's result] would be very inaccurate for any reasonable dt; quaternions are not closed under addition unless the added value is truly differential (this is the nature of being a manifold)."

This begs the question - are the time steps we are using small enough that we can confidently say that our time steps are small enough to be "truly differential"? I would _guess_ that they _probably_ aren't, which then begs the question - is it possible to do odometry without quaternions? they seem very hard to do properly...

KipHamiltons:
I have migrated to a different - easier to follow - method of updating our rotation, based on [the method here](https://stackoverflow.com/a/41226401/11108223). This seems much nicer to me.

Bidski:
> I have migrated to a different - easier to follow - method of updating our rotation, based on [the method here](https://stackoverflow.com/a/41226401/11108223). This seems much nicer to me.

The answer that you linked and the lecture slides that your linked answer links give differing formulae: `Q = update * Q` and `Q = Q * update` are not the same formula as quaternion multiplication is not commutative.

Furthermore, everything I find when looking at how to update a quaternion orientation with an angular velocity measurement leads me back to the formula we were using before `Q += 0.5 * t * omega * Q`

https://gamedev.stackexchange.com/questions/108920/applying-angular-velocity-to-quaternion 
which links to https://fgiesen.wordpress.com/2012/08/24/quaternion-differentiation/ (the link we originally had in our code) and references the book [Game Physics Engine Development by Ian Millington](http://www.r-5.org/files/books/computers/algo-list/realtime-3d/Ian_Millington-Game_Physics_Engine_Development-EN.pdf) which describes this formula on page 160

[This link](https://stackoverflow.com/questions/46908345/integrate-angular-velocity-as-quaternion-rotation) also talks about this

KipHamiltons:
I saw that the answer had a different formula to the lecture slides, but isn't the answer correct? you're composing your existing rotation with the rotation update - which I would have thought would be left multiplying your update?

The reason I am worried about that method is that [this answer](https://math.stackexchange.com/questions/1896379/how-to-use-the-quaternion-derivative/2099673#2099673) states in no unclear terms that the formula we were using is only accurate if the `dt` is truly differential. I am sceptical that the frequency of 90 Hz is "truly differential".

If we're confident enough in the existing method to dismiss that answer, that's all good. I certainly wasn't that confident, but it seems like basically everyone does it like this, so I am fine with continuing to use it if you like

Bidski:
Im honestly not hugely confident with quaternions in general. Im happy to keep the method we had before simply because it seems to be the way that everyone does it (not that this is a validbreason to do anything, but this does seem to be "the way" *click click click*).

I think we will find that the two methods probably are the same and that the composition is just being formed differently, but someone smarter than me would need to show/prove this.

KipHamiltons:
I have gone back to the old method of doing it, linking that gamedev stackexchange thread as well as the wordpress derivation which was already linked

Bidski (approved):
Seems OK

PR #431, opened 2021-04-24
Turn on clang-tidy in the CI build

TrentHouliston:
This will enforce all the clang-tidy rules we have (so this won't merge yet)
Keep merging fixes into master until this PR has a green tick!

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -52,7 +52,7 @@ steps:
 
   # Build generic target and test
   - label: "Build generic and Test"
-    command: "./b configure -- -DBUILD_TESTS=ON && ./b build && ./b tests run"
+    command: "./b configure -- -DBUILD_TESTS=ON -DCI_BUILD=ON && ./b build && ./b tests run"
     plugins:
       - docker-compose#v3.0.3:
           config: .buildkite/docker-compose.yml


KipHamiltons:
The size of this is absolutely insane. I reckon we should break this into multiple PRs, only enabling the CI build at the last one.
This one is already huge enough...

KipHamiltons:
Also, note to any reviewer - it's probably easiest to review this commit by commit.
The nested namespace change makes things very messy, so if it were me, I would just take that commit on faith because I can certify that after that commit, it built, and I didn't change anything except namespaces and the end namespace comments.

KipHamiltons:
Oh shit. I just understood what the intention of this PR was....
well, now that I have effectively hijacked it, I suppose we should turn this PR into one of the ones which merges changes :sweat_smile:

TrentHouliston:
Nothing a force push can't solve!

tayloryoung6396:
@TrentHouliston @KipHamiltons What is the state of this PR? As I understand it's no longer needed

KipHamiltons:
@tayloryoung6396 It is needed, but it has to be updated. What this PR is going to do is add clang-tidy to the CI checks. However, we have to fix the clang-tidy warnings which are in the codebase atm for that to be useful. So this is waiting for #734 (and the ones which come after it) to fix those warnings. Then, we'll update this PR so that it adds clang-tidy to the github actions workflows, instead of buildkite.
Long story short, keep this PR, and approve my clang-tidy fix PRs as they come in :sweat_smile:

CMurtagh-LGTM:
Is this already a feature?
https://github.com/NUbots/NUbots/blob/main/.github/workflows/nubots.yaml#L145

PR #432, opened 2021-04-24
Swaps from nested namespaces

KipHamiltons:
All this PR does is change to use one namespace scope in each C++ file with nested namespaces, then reformats the files.


TrentHouliston:
For those reviewing this, I highly recommend you turn on the option to ignore white space changes!

shared/utility/math/geometry/UnitQuaternion.hpp
```diff
@@ -24,128 +24,125 @@
...
+            inline arma::subview_col<double> imaginary() {
+                return rows(1, 3);
+            }
+
+            UnitQuaternion slerp(const UnitQuaternion& p, const double& t);
+            static inline UnitQuaternion Identity() {
+                return (arma::vec4({1.0, 0.0, 0.0, 0.0}));
+            }
+        };
+    }  // namespace geometry
```
TrentHouliston, line 146: This shouldn't be here

TrentHouliston (changes requested):
Found a misplaced comment!

shared/utility/math/geometry/UnitQuaternion.hpp
KipHamiltons: It actually should though
the way this was written, it has two namespaces within that one. We _could_ fully qualify each of them, but that seems less nice than the current situation.
The author did something similar in `Rotation3D`, as well as `Configuration.hpp`, `FileWatch.hpp` and `Script.hpp` :p

TrentHouliston (approved):
My bad I didn't see the opening

PR #433, opened 2021-04-24
Apply some clang-tidy fixes

KipHamiltons:
Wait for #432 to merge. 

This PR includes:
- a large batch of sane auto-fixes which I verified
- removal of a string initialisation rule I didn't like
- adding `NOLINT` comments above our global variables
- Removing a stack of narrowing conversions
- A bunch of assorted fixes for `uart.cpp`


module/behaviour/strategy/PS3Walk/src/Joystick.hpp
```diff
@@ -12,62 +12,62 @@
...
 
-#include <string>
+    #include <string>
 
-#define JS_EVENT_BUTTON 0x01  // button pressed/released
-#define JS_EVENT_AXIS   0x02  // joystick moved
-#define JS_EVENT_INIT   0x80  // initial state of device
+    #define JS_EVENT_BUTTON 0x01  // button pressed/released
+    #define JS_EVENT_AXIS   0x02  // joystick moved
+    #define JS_EVENT_INIT   0x80  // initial state of device
```
CMurtagh-LGTM, line 21:
```suggestion
enum struct JS_EVENT : unsigned char {
    BUTTON = 0x01, AXIS = 0x02, INIT = 0x80
};
```
Probably a more c++ way of doing things. Also put in the namespace.

module/behaviour/strategy/PS3Walk/src/Joystick.hpp
```diff
@@ -12,62 +12,62 @@
...
      * For axes, this will range between -32768 and 32767.
      */
-    short value;
+    short value{0};
 
     /**
      * The event type.
      */
-    unsigned char type;
+    unsigned char type{0};
```
CMurtagh-LGTM, line 45: Give `std::byte` a look, I think in this file it will give more robust code.

shared/utility/motion/splines/Footstep.cpp
```diff
@@ -85,15 +85,15 @@ namespace utility::motion::splines {
         stepFromSupport(tmpDiff);
     }
 
-    Eigen::Vector3f Footstep::poseAdd(const Eigen::Vector3f& pose, const Eigen::Vector3f& diff) const {
+    Eigen::Vector3f Footstep::poseAdd(const Eigen::Vector3f& pose, const Eigen::Vector3f& diff) {
```
CMurtagh-LGTM, line 88: Removing const?

shared/utility/motion/splines/Footstep.cpp
```diff
@@ -85,15 +85,15 @@ namespace utility::motion::splines {
...
         const float cos_z = std::cos(pose.z());
         const float sin_z = std::sin(pose.z());
         return Eigen::Vector3f(pose.x() + diff.x() * cos_z - diff.y() * sin_z,
                                pose.y() + diff.x() * sin_z + diff.y() * cos_z,
                                utility::math::angle::normalizeAngle(pose.z() + diff.z()));
     }
 
-    Eigen::Vector3f Footstep::diffInv(const Eigen::Vector3f& diff) const {
+    Eigen::Vector3f Footstep::diffInv(const Eigen::Vector3f& diff) {
```
CMurtagh-LGTM, line 96: Removing const?

CMurtagh-LGTM (changes requested):
Very good.

shared/utility/motion/splines/Footstep.cpp
KipHamiltons: This was made static. Const doesn't make sense for static method

shared/utility/motion/splines/Footstep.cpp
KipHamiltons: Also made static

module/behaviour/strategy/PS3Walk/src/Joystick.hpp
CMurtagh-LGTM: Probs another pr

module/behaviour/strategy/PS3Walk/src/Joystick.hpp
CMurtagh-LGTM: probs another pr

CMurtagh-LGTM (approved):
LGTM

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.cpp
```diff
@@ -50,7 +49,7 @@ namespace module::behaviour::strategy {
         initscr();
 
         // Initialise colours
-        if (has_colors() == TRUE) {
+        if (static_cast<int>(has_colors()) == TRUE) {
```
TrentHouliston, line 52: This seems incorrect. I think clang-tidy has correctly deduced that TRUE is defined somewhere as `#define TRUE 1` and is, therefore, an int. And `has_colors()` is returning a bool so it has correctly decided that comparing a bool to an integer is a bad thing. Where it went wrong was rather than changing `TRUE` to `true` so it's bool to bool comparison, it changed `has_colors` to an int

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.hpp
```diff
@@ -73,7 +73,7 @@ namespace module::behaviour::strategy {
         void look_up();
         void look_down();
         void walk_toggle();
-        void quit();
+        static void quit();
```
TrentHouliston, line 76: Generally, when I see clang-tidy suggest this, I move the function into the cpp file as a raw function. It clearly didn't need to be on the class

module/behaviour/strategy/PS3Walk/src/Joystick.hpp
TrentHouliston: Agreed, there is a more c++ way of doing this that should probably be used

module/behaviour/strategy/PS3Walk/src/Joystick.hpp
```diff
@@ -12,62 +12,62 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>
 
 #ifndef __JOYSTICK_HH__
-#define __JOYSTICK_HH__
+    #define JOYSTICK_HH_
```
TrentHouliston, line 15: This macroguard should be fixed

module/extension/FileWatcher/src/FileWatcher.cpp
```diff
@@ -38,7 +38,7 @@ namespace module::extension {
         std::lock_guard<std::mutex> lock(reactor.paths_mutex);
 
         // Work out what path we are watching
-        std::array<char, 512> pathbuff;
+        std::array<char, 512> pathbuff{};
         pathbuff.fill('\0');
```
TrentHouliston, line 42: Now that we did pathbuff{} the fill is uneccasary

module/support/SignalCatcher/src/SignalCatcher.cpp
```diff
@@ -24,12 +24,14 @@
...
     // Set our initial shutdown request state
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
     volatile bool userRequestedShutdown = false;
 
     // Initialize our powerplant variable
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
     NUClear::PowerPlant* POWER_PLANT = nullptr;
 
-    void sigint(int) {
+    void sigint(int /*unused*/) {
```
TrentHouliston, line 34: instead of `/*unused*/` in these, we should put /*signal*/ to show what the variable actually would be named if we used it

module/support/SignalCatcher/src/SignalCatcher.cpp
```diff
@@ -62,7 +64,7 @@ namespace module::support {
 
         // Store our powerplant in the static variable
         POWER_PLANT = &powerplant;
-        struct sigaction action;
+        struct sigaction action {};
```
TrentHouliston, line 67: remove the struct from out front of this since we are in c++

shared/extension/FileWatch.hpp
```diff
@@ -76,19 +76,17 @@ namespace NUClear::dsl {
             }
 
             template <typename DSL>
-            static inline ::extension::FileWatch get(threading::Reaction&) {
+            static inline ::extension::FileWatch get(threading::Reaction& /*unused*/) {
```
TrentHouliston, line 79: unused here is `/*reaction*/`

shared/utility/input/LimbID.cpp
```diff
@@ -25,20 +25,27 @@ namespace utility::input {
...
             value = Value::RIGHT_ARM;
-        else if (str == "HEAD")
+        }
+        else if (str == "HEAD") {
             value = Value::HEAD;
-        else
+        }
+        else {
             throw std::runtime_error("String did not match any enum for LimbID");
+        }
```
TrentHouliston, line 48: If you want for this you can turn off clang-format and make if statements inline. Makes it much easier to read

shared/utility/input/ServoID.cpp
```diff
@@ -23,48 +23,69 @@
...
             value = Value::HEAD_YAW;
-        else if (str == "HEAD_PITCH")
+        }
+        else if (str == "HEAD_PITCH") {
             value = Value::HEAD_PITCH;
-        else
+        }
+        else {
             throw std::runtime_error("String " + str + " did not match any enum for ServoID");
+        }
```
TrentHouliston, line 88: Same here for turning off clang-format and having all these be one liners

shared/utility/io/uart.cpp
```diff
@@ -37,7 +37,7 @@ namespace utility::io {
         }
     }
 
-    void uart::close() {
+    void uart::close() const {
```
TrentHouliston, line 40: It's incorrect that this should be able to be a const function. Within this function, we should instead be setting `fd = -1` when we close.

shared/utility/io/uart.cpp
```diff
@@ -163,11 +163,11 @@ namespace utility::io {
         return !(fcntl(fd, F_GETFL) < 0 && errno == EBADF);
     }
 
-    ssize_t uart::read(void* buf, size_t count) {
+    ssize_t uart::read(void* buf, size_t count) const {
         return ::read(fd, buf, count);
     }
 
-    ssize_t uart::write(const void* buf, size_t count) {
+    ssize_t uart::write(const void* buf, size_t count) const {
```
TrentHouliston, line 170: I feel like these also shouldn't be const but i'm not sure how to enforce that

shared/utility/io/uart.hpp
```diff
@@ -43,7 +43,14 @@ namespace utility::io {
...
          */
         uart(const uart& uart) = delete;
-        uart(uart&&)           = default;
+        uart& operator=(const uart& uart) = delete;
+
+        /**
+         * @brief We can move these because it won't close the device twice
+         */
+        uart(uart&&)  = default;
+        uart& operator=(uart&& uart) = default;
```
TrentHouliston, line 52: I'm not sure we can move them... would the other instance no longer have something to close?
We might need explicit move constructors and operators

shared/utility/math/angle.hpp
```diff
@@ -43,11 +43,13 @@ namespace utility::math::angle {
...
 
-        if (angle <= -M_PI)
+        if (angle <= -M_PI) {
             angle += M_PI * 2;
+        }
 
-        if (angle > M_PI)
+        if (angle > M_PI) {
             angle -= 2 * M_PI;
+        }
```
TrentHouliston, line 52: For some reason it makes me upset that one is pi * 2 and the other is 2 * pi

shared/utility/math/angle.hpp
```diff
@@ -165,7 +167,7 @@ namespace utility::math::angle {
 
         // Find alpha such that $\sin(\alpha) = a\_$ and $\cos(\alpha) = b\_$, which is possible because $a\_^2
         // + b\_^2 = 1$
-        float alpha = atan2(a_, b_);
+        float alpha = atan2f(a_, b_);
```
TrentHouliston, line 170: this should have been changed to `std::atan2`, not to `atan2f`
The std versions are overloaded such that they will return the type that's put into them.

shared/utility/motion/splines/Footstep.cpp
TrentHouliston, line 88: I would leave it up to personal preference in this case, but in general, I prefer to just have regular functions in the cpp file if there is a function that is static and it doesn't use any state from the class. Makes it clearer that it doesn't require any class state to function

shared/utility/nusight/NUhelpers.hpp
```diff
@@ -54,7 +55,7 @@ namespace utility::nusight {
                                           && std::is_same<decltype(test_end<T>(0)), yes>::value;
         };
 
-        inline void buildGraph(DataPoint&) {}
+        inline void buildGraph(DataPoint& /*unused*/) {}
```
TrentHouliston, line 58: `/*dataPoint*/`

shared/utility/support/evil/pure_evil.hpp
```diff
@@ -30,18 +31,19 @@ namespace utility::support::evil {
 
     struct StackFrame {
 
-        StackFrame() : pc(), file(), lineno(), function() {}
+        StackFrame() {}
 
         StackFrame(uintptr_t pc, std::string file, int lineno, std::string function)
-            : pc(pc), file(file), lineno(lineno), function(function) {}
+            : pc(pc), file(std::move(std::move(file))), lineno(lineno), function(std::move(std::move(function))) {}
```
TrentHouliston, line 37: The double move here is 100% not needed. One move is fine

shared/utility/vision/Vision.cpp
```diff
@@ -186,42 +186,42 @@ namespace utility::vision {
...
+        BayerPixelType type    = row   ? col ? BayerPixelType::R : BayerPixelType::GR
+                                 : col ? BayerPixelType::GB
+                                       : BayerPixelType::B;
 
         return getBayerPixel(patch, type);
     }
 
     Pixel getGrey16Pixel(uint x, uint y, int width, int /*height*/, const std::vector<uint8_t>& data) {
-        int origin = (y * width + x) * 2;
+        const unsigned int origin = (y * width + x) * 2;
```
TrentHouliston, line 199: All these shoudl 100% be kept as regular ints not unsigned ints. Doing math for pixel coordinates in unsigned math is asking for trouble. Every so often an expression will dip below 0 and back up, but with unsigned it explodes and has memory issues.

In general, try to always use signed integers unless there is an explicit reason to use unsigned integers. For the most common case where it gets mad at you for looping on a .size() of something, use `int(x.size())` instead. Until we get c++20 when they add `x.ssize()`

Even the original authors of c++ consider it one of their blunders that .size() returns an unsigned integer

shared/utility/vision/Vision.hpp
```diff
@@ -54,24 +54,33 @@ namespace utility::vision {
...
                 value = Value::MAGENTA;
-            else if (str == "WHITE_GREEN")
+            }
+            else if (str == "WHITE_GREEN") {
                 value = Value::WHITE_GREEN;
-            else
+            }
+            else {
                 throw std::runtime_error("String " + str + " did not match any enum for Colour");
+            }
```
TrentHouliston, line 83: You can turn off clang-format here and collapse these into one liners for readability if you want

TrentHouliston (commented):
Curse you automerge!!!!!

PR #434, opened 2021-04-24
Remove some of the least helpful/actionable/crustiest TODOs

KipHamiltons:
All of the todos will eventually have to be dealt with for clang-tidy, but these ones are a good start. These were so obvious to remove to me that it's not worth tagging them with a contributor or issue.


JosephusPaye:
Things like TODO: document can go for sure. But some of these are signaling something wrong with the current implementation or indicating a possible area of improvement. That's useful information that could help whoever will work on the piece of code next. Is there a way we can keep the information without framing it as a "to-do"?

KipHamiltons:
Were there any specifically here you thought were worth keeping? Each of these are either extremely old, extremely unhelpful, or reference things that don't apply any more, which is why I selected them as not worth assigning or issuing up.

module/motion/IKKick/src/IKKick.cpp
```diff
@@ -150,9 +150,9 @@ namespace module::motion {
                 Eigen::Vector3d directionSupportFoot = torsoPose.rotation() * directionTorso;
 
                 Eigen::Vector3d ballPosition = targetSupportFoot;
-                ballPosition.z()             = 0.05;  // TODO: figure out why ball height is unreliable
```
Bidski, line 153: Change this to "TODO: get ball height from config" (I think we want FieldDescription for this??)

module/platform/darwin/HardwareSimulator/data/config/DarwinHardwareSimulator.yaml
```diff
@@ -12,4 +12,4 @@ noise:
     y: 0.01
     z: 0.01
 
-bodyTilt: 0 #25 * pi / 180 # set this to be the same as the WalkEngine.yaml:stance:bodyTilt TODO: read this from WalkEngine
+bodyTilt: 0 #25 * pi / 180
```
Bidski, line 15: Although this may need to be updated to follow the QuinticWalk
```suggestion
bodyTilt: 0 #25 * pi / 180 # set this to be the same as the WalkEngine.yaml:stance:bodyTilt
```

module/behaviour/skills/DirectWalkController/src/DirectWalkController.cpp
```diff
@@ -85,11 +85,5 @@ namespace module::behaviour::skills {
                 emit(std::make_unique<ActionPriorities>(ActionPriorities{subsumptionId, {0, 0}}));
             }
         });
-
-        on<Trigger<WalkStopped>>().then([this] {
-            // TODO : Right now, this causes the walk engine to become unaware of positions and negatively
-            // impact servo positions... emit(std::make_unique<ActionPriorities>(ActionPriorities {
-            // subsumptionId, { 0, 0 }}));
-        });
```
Bidski, line 93: I feel like this should remain with a more pointed TODO message. Unless there are no actions that need to be taken by the walk controllers (this should be compared with the other walk controllers) when the walk stops?

Bidski (changes requested):

cmake/Modules/FindOpenBLAS.cmake
```diff
@@ -7,5 +7,3 @@ ToolchainLibraryFinder(
   VERSION_FILE openblas_config.h
   VERSION_REGEX "OPENBLAS_VERSION \" ?OpenBLAS (([0-9]+\\.?)+) ?\""
 )
-
-# TODO also find LAPACK if needed
```
Bidski, line 11: I feel like we should make a check to see if LAPACK has been compiled into OpenBLAS or not (this is not guaranteed, it is optional for OpenBLAS, but would almost certainly indicate an error in the building of the docker image if it isnt there). The only way I can think of to do this is to use [try_compile](https://cmake.org/cmake/help/latest/command/try_compile.html)

module/behaviour/skills/DirectWalkController/src/DirectWalkController.cpp
KipHamiltons: That TODO refers directly to the oldwalkengine (it was written 4 years ago), and there aren't any other triggers in the codebase for `WalkStopped`

KipHamiltons: That's wrong. other controllers do use it. I had files in the `files to include` in my search :facepalm:

KipHamiltons: [`SimpleWalkPathPlanner`](https://github.com/NUbots/NUbots/blob/master/module/behaviour/planning/SimpleWalkPathPlanner/src/SimpleWalkPathPlanner.cpp#L123) has a trigger for it where it emits a new `ActionPriorities` which just includes the `subsumptionId`

cmake/Modules/FindOpenBLAS.cmake
KipHamiltons: When you say use try_compile, I don't quite understand. I have read that docs page, but how would that guarantee it builds with LAPACK? Does it return false if it can't link against one of the options that you said to link against?

Bidski: You would pick a function that is only provided by LAPACK and write a small program that tries to call that function. If the program fails to compile and/or link then LAPACK is not provided by OpenBLAS.

Might even be able to use [CheckFunctionExists](https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html) 
```cmake
# If you compiled OpenBLAS with LAPACK in it then it should have the
# sgetrf_single function in it.  So if we find that function in
# OpenBLAS then just use OpenBLAS's LAPACK. 
CHECK_FUNCTION_EXISTS(sgetrf_single OPENBLAS_HAS_LAPACK)
```

Bidski (approved):
Seems OK. Dont know if we want to add in the LAPACK stuff to this PR or not. If we don't we should leave that TODO in

PR #439, opened 2021-04-26
Fix first clang-tidy general things PR issues

KipHamiltons:
@TrentHouliston, most of your review was implementable, but there were a few parts where I wanted further clarification or couldn't apply them:

- [x] in `SignalCatcher.cpp`, I think we need the `struct` keyword, because `sigaction` is also the name of a function in that namespace (because of course it is...), and the compiler tries to interpret it as a function without that keyword. There is probably a way around that, but I couldn't find one. 
** Confirmed**.
- [x] in `uart.cpp`, there are a couple of functions which are `const` that you pointed out, but yea, I am not sure why they wouldn't be able to be const. like, you're writing or reading from a buffer which isn't a class variable. I don't see why they shouldn't be `const`?
**const has been removed from these functions**. 
- [x] regarding the `move` constructor and operator in `uart.cpp`, shouldn't they be fine? wouldn't the new owner of the object be given the responsibility of closing the `fd`, with the old owner having nothing there to close? I don't understand the finer parts of move semantics, but this seems fine to me.
**I'll `= delete` the move constructor and operator for now.** They seem a bit spicy.

~Other than those, I have applied your suggestions without issue.~
Those have been addressed


module/behaviour/strategy/PS3Walk/src/Joystick.hpp
```diff
@@ -11,14 +11,14 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>
 
-#ifndef __JOYSTICK_HH__
-    #define JOYSTICK_HH_
+#ifndef __JOYSTICK_HPP__
+#define __JOYSTICK_HPP__
```
TrentHouliston, line 15: It's mad at you because of the `__THING__` Things that start with a double underscore are reserved for use by the compiler. That's why clang-tidy removed it in the first place, but it only removed it for the `#define` which made the indentation go all screwy because it was no longer a valid macroguard
```suggestion
#ifndef MODULE_BEHAVIOUR_STRATEGY_PS3WALK_JOYSTICK_HPP
#define MODULE_BEHAVIOUR_STRATEGY_PS3WALK_JOYSTICK_HPP
```

module/behaviour/strategy/PS3Walk/src/Joystick.hpp
```diff
@@ -123,4 +123,4 @@ class Joystick {
     bool sample(JoystickEvent* event) const;
 };
 
-#endif
+#endif  // __JOYSTICK_HPP__
```
TrentHouliston, line 126:
```suggestion
#endif  // MODULE_BEHAVIOUR_STRATEGY_PS3WALK_JOYSTICK_HPP
```

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.hpp
```diff
@@ -87,6 +86,10 @@ namespace module::behaviour::strategy {
         /// @brief Called by the powerplant to build and setup the KeyboardWalk reactor.
         explicit KeyboardWalk(std::unique_ptr<NUClear::Environment> environment);
     };
+
+    // quit does not need any of the state from the KeyboardWalk class
+    void quit();
+
```
TrentHouliston, line 92: Not needed unless it's called from another file.
Better to avoid adding things to your public interface unless you need to
```suggestion
```

shared/utility/math/angle.hpp
```diff
@@ -41,14 +41,14 @@ namespace utility::math::angle {
...
-        T angle = std::fmod(value, static_cast<T>(2.0 * M_PI));
+        T angle = std::fmod(value, static_cast<T>(M_2_PI));
 
         if (angle <= -M_PI) {
-            angle += M_PI * 2;
+            angle += M_2_PI;
         }
 
         if (angle > M_PI) {
-            angle -= 2 * M_PI;
```
TrentHouliston, line 51: M_2_PI is `2/pi` not `2*pi` So these changes are no good!

shared/utility/motion/splines/Footstep.hpp
```diff
@@ -114,17 +114,17 @@ namespace utility::motion::splines {
...
 
-        /**
-         * Compute and return the delta from (zero + diff) to (zero) in (zero + diff) frame.
-         */
-        [[nodiscard]] static Eigen::Vector3f diffInv(const Eigen::Vector3f& diff);
-    };
+    /**
+     * Compute and return the delta from (zero + diff) to (zero) in (zero + diff) frame.
+     */
+    [[nodiscard]] Eigen::Vector3f diffInv(const Eigen::Vector3f& diff);
```
TrentHouliston, line 127: If these aren't used outside of this `Footstep.cpp`, remove them from the header

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
```diff
@@ -51,14 +51,14 @@ namespace module::behaviour::strategy {
                         case AXIS_LEFT_JOYSTICK_HORIZONTAL:
                             // y is left relative to robot
                             // strafe[1] = -event.value;
-                            rotationalSpeed = float(-event.value);
+                            rotationalSpeed = static_cast<float>(-event.value);
```
TrentHouliston, line 54: What rule is forcing you to change to static casts here?
I don't think it should be required

TrentHouliston (changes requested):

module/behaviour/strategy/PS3Walk/src/Joystick.hpp
KipHamiltons: Ohhh, that makes a lot of sense. Thanks

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.cpp
```diff
@@ -360,7 +360,7 @@ namespace module::behaviour::strategy {
         update_window(command_window, LogColours::TRACE_COLOURS, "", message, false);
     }
 
-    void KeyboardWalk::quit() {
+    void quit() {
         endwin();
         std::raise(SIGTERM);  // Change back to SIGINT if required by NUbots messaging system//
```
CMurtagh-LGTM, line 365: Is `powerplant.shutdown();` a nicer way to do this?

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.hpp
KipHamiltons: If that's the issue, perhaps it should be `private static`?

shared/utility/math/angle.hpp
KipHamiltons: ahh yes. classic me

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
KipHamiltons: It's not required. I just like static casts

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.cpp
KipHamiltons: I suppose? @TrentHouliston

TrentHouliston: Still need to do that endwin stuff etc but maybe

module/behaviour/strategy/PS3Walk/src/PS3Walk.cpp
TrentHouliston: Fair enough :P I like functional casts but I am aware it's a stylistic choice so idm

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.hpp
TrentHouliston: Still would have to be in the header file

KipHamiltons: Right. I am suggesting moving it back into the header, and making it a private, static function declared in the header

TrentHouliston: Personal preference, my preference is to keep the headers with as little in them as possible. Less reasons to trigger a recompilation of files

KipHamiltons: Fair enough. This should be good to rereview then

TrentHouliston (dismissed):

TrentHouliston (approved):

PR #441, opened 2021-04-26
More comms suggestions

KipHamiltons:
This doesn't build, because NUClear's clock (which the `ServoTarget`s use) has a different duration to this PRs custom clock, and there's no operator to subtract between the two duration types.

I think this should be along the right track though, and we're going to have to build the custom clock into the other stuff too eventually (perhaps next...)

### DIFF
--- a/module/platform/webots/Webots/src/Webots.cpp
+++ b/module/platform/webots/Webots/src/Webots.cpp
@@ -248,6 +248,8 @@ namespace module::platform::webots {
         // Tell webots who we are
         // send_player_details(fd, global_config); Not needed? Set by the port we connect to?
 
+        // Now that we are connected, we can set up our reaction handles with this file descriptor
+
         // Receiving
         read_io = on<IO>(fd, IO::READ).then([this, fd]() {
             // Get the size of the message
@@ -257,7 +259,8 @@ namespace module::platform::webots {
                 return;
             }
 
-            uint32_t Nh = ntohl(Nn);  // Convert from network endian to host endian
+            // Convert from network endian to host endian
+            const uint32_t Nh = ntohl(Nn);
 
             // Get the message
             std::vector<char> data(Nh, 0);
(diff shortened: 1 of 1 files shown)


module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -172,39 +176,54 @@ namespace module::platform::webots {
...
+                    to_send.motor_positions.push_back(position_msg);
+
+                    MotorVelocity velocity_msg;
+                    // We need to calculate the servo velocity to add to the velocity message
+                    // (method stolen from HardwareIO.cpp)
+                    // velocity = distance / time
+                    const float distance =
+                        utility::math::angle::difference(target.position,
+                                                         getDarwinServo(target.id, sensors).presentPosition);
+                    Clock::duration time = target.time - Clock::now();
```
CMurtagh-LGTM, line 201: These two have different epochs, as things stand at the moment.
They are also different time_point specialisations it should not compile.

All in all the webots::clock needs more work.

module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -65,42 +67,44 @@ namespace module::platform::webots {
     using message::support::GlobalConfig;
 
     using utility::input::ServoID;
+    using utility::platform::darwin::getDarwinServo;
     using utility::vision::fourcc;
 
+    using module::platform::webots::Clock;
```
CMurtagh-LGTM, line 73: Not needed as they are in the same namespace.

module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -242,13 +260,16 @@ namespace module::platform::webots {
             log<NUClear::FATAL>("Connection was closed.");
         }
 
+        // TODO(KipHamiltons): Does this need to be replaced with our custom clock's now()?
         connect_time = NUClear::clock::now();
```
CMurtagh-LGTM, line 264: I should clarify, this is currently the real connect time.

```suggestion
        // Set the real time of the connection initiation
        connect_time = NUClear::clock::now();
        // Reset the simulation connection time
        Clock::current_tick = 0;
```

CMurtagh-LGTM (changes requested):

CMurtagh-LGTM:
Depending if anyone answers https://hl.forum.robocup.org/t/realtime-in-webots-controller/504, we may move to a clock like gazebo.

KipHamiltons:
Have you tried bugging people on the robocup discord for an answer? could be worth doing

CMurtagh-LGTM (approved):
LGTM

module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -214,11 +231,10 @@ namespace module::platform::webots {
...
-        int fd = tcpip_connect(server_address, port);
-
+        const int fd = tcpip_connect(server_address, port);
 
         char initial_message[8];
-        int n = recv(fd, initial_message, sizeof(initial_message), 0);
+        const int n = recv(fd, initial_message, sizeof(initial_message), 0);
 
         if (n > 0) {
             if (strncmp(initial_message, "Welcome", sizeof(initial_message)) == 0) {
```
JosephusPaye, line 240: We should probably just convert initial message to a proper string above and just use `==` to compare. Did this [here in Mikyla's PR](https://github.com/NUbots/NUbots/pull/422/commits/6d40546fad6da098a386e9bc2a9a4df133e735d1#diff-fc0b2b2d580a7b5d606168708dc45d1b335285a24d5c80f1f4ef7846672cc8b4L176-R223) but it got overwritten in the merge.

JosephusPaye:
> Have you tried bugging people on the robocup discord for an answer? could be worth doing

Yeah, a Discord post in the [#webots-technical-questions](https://discord.com/channels/462951096374132736/810971909046009876) channel will probably get a quicker response.

PR #446, opened 2021-04-29
Fix NUsight2 CI

KipHamiltons:
~Making this draft PR so that I can see what happens when it goes to buildkite.~
I consolidated the nusight pipeline into the nubots one. Lmk if I should make any changes of course


JosephusPaye (approved):
LGTM

PR #447, opened 2021-04-30
Adds NUsight's dependencies to the dockerfile

KipHamiltons:
To run yarn and node stuff inside the container (as is done in `./b format`, for example), we need the node stuff

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -35,7 +35,7 @@ RUN install-package \
     ninja \
     cmake \
     meson \
-    nodejs \
+    nodejs-lts-fermium \
     npm \
     yarn \
     git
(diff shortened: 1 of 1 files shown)


Y0sh1dk:
This fixes the issue with `./b format` not formatting NUsight files. However with the depenedencies, `./b format` changes a few NUsight files, will this cause issues with the CI pipeline as it currently passes without these changes?
image

docker/Dockerfile
```diff
@@ -35,6 +35,9 @@ RUN install-package \
     ninja \
     cmake \
     meson \
+    nodejs \
```
JosephusPaye, line 38: Use the LTS ([fermium is v14](https://wiki.archlinux.org/index.php/Node.js_#Installation), the current LTS):

```suggestion
    nodejs-lts-fermium \
```

JosephusPaye (changes requested):

JosephusPaye (approved):
LGTM

PR #449, opened 2021-04-30
Changes `./b target` default behaviour to pull from dockerhub

KipHamiltons:
Pulling rules, building drools


tools/target.py
```diff
@@ -15,18 +15,35 @@ def register(command):
         "target", nargs="?", choices=platform.list(), help="the platform to select as the default platform"
     )
 
+    command.add_argument(
+        "-l",
+        "--local-build",
+        dest="build_local",
+        action="store_true",
+        default=False,
+        help="build the image locally, instead of pulling and tagging it",
```
KipHamiltons, line 24: `./b format` adds the comma at the end of this line. Weird, but sure lol...

tools/target.py
```diff
@@ -15,18 +15,35 @@ def register(command):
         "target", nargs="?", choices=platform.list(), help="the platform to select as the default platform"
     )
 
+    command.add_argument(
```
Y0sh1dk, line 18: Probably a good idea to document this in NUbook

tools/utility/dockerise/platform.py
```diff
@@ -119,8 +120,25 @@ def build(repository, platform):
...
+
+
+def pull(repository, platform):
+    # Define our tag strings
+    remote_tag = "{0}/{0}:{1}".format(repository, platform)
+    local_tag = "{0}:{1}".format(repository, platform)
+    selected_tag = "{}:selected".format(repository)
+
+    print("Pulling remote image", remote_tag)
+    err = subprocess.run(["docker", "pull", remote_tag], stderr=DEVNULL, stdout=DEVNULL).returncode
```
Y0sh1dk, line 137: Might be a good idea to print the output of `docker pull` to console so the user can see that something is actually happening, especially for those with slow internet (like me :stuck_out_tongue:). Possibly with `stdout=sys.stdout`.

Y0sh1dk (changes requested):

tools/target.py
TrentHouliston: Of course, trailing commas are the correct way to do multi line lists.
The rationale is that if you were to add or remove an item from the list the diff would only be for that line. If you didn't have the trailing comma then when you add a new element it would be a diff of adding a comma to the previous (unrelated) line as well as adding the new line.

TrentHouliston:
This is already supposed to happen via the cache from in buildx so you get the best of both worlds. This PR is just going to cause confusion in the future IMO

JosephusPaye:
I think we should add this. buildx is not integrated/working yet, and for the majority of the team currently re-building the code each time they do `./b target generic`, this will be a great time saver.

tools/target.py
JosephusPaye: Yep, trailing commas all the way

tools/target.py
```diff
@@ -15,18 +15,35 @@ def register(command):
...
+        "-l",
+        "--local-build",
```
JosephusPaye, line 20: We could probably just call this `--build`, since pulling + tagging is not actually building

JosephusPaye (changes requested):

tools/utility/dockerise/platform.py
JosephusPaye: For sure. Let's print the output (both stdout and stderr) from the subprocess so the user knows what's going on.

TrentHouliston:
You could also automatic work out if we need to build based on if the docker folder is different from the one on master

JosephusPaye (approved):
Tested, and it works. LGTM.

Y0sh1dk (dismissed):
Works well, LGTM.

tools/target.py
```diff
@@ -15,18 +16,55 @@ def register(command):
...
     else:
-        # Ensure the platform image is built
-        platform.build(defaults.image, target)
+        # We check if a build is necessary by diffing the docker folder with master
+        output = subprocess.run(
+            ["git", "diff", "origin/master", os.path.join(b.project_dir, "docker")], capture_output=True
+        ).stdout
+        # If there are changes in tracked files in the docker folder when comparing with master,
+        # we want to build the image and we don't want to pull the image
+        if output != b"":
```
Bidski, line 52:
```suggestion
        ).stdout.decode("utf-8")
        # If there are changes in tracked files in the docker folder when comparing with master,
        # we want to build the image and we don't want to pull the image
        if output != "":
```

Bidski (changes requested):

Bidski (approved):
Sure

PR #451, opened 2021-05-01
Fixes Field.proto message comment

KipHamiltons:
This field was changed during the arma purge, and the comments weren't updated


Bidski (approved):
Sure

PR #452, opened 2021-05-01
Updates CM740 Sensor Conversion

KipHamiltons:
This adds more comments explaining the factors, as well as changing the speed conversion
The speed conversion factor was set for old servo models. This updates it with the values appropriate for the MX64/106, which are fortunately the same as each other

### DIFF
--- a/module/platform/darwin/HardwareIO/src/Convert.hpp
+++ b/module/platform/darwin/HardwareIO/src/Convert.hpp
@@ -35,7 +35,7 @@ namespace module::platform::darwin {
 
         /// The Gyrosocope value from the Darwin is between 0 and 1023, representing a value between -500 and
         /// 500 degrees per second. This means 512 = 0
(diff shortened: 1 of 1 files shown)


module/platform/darwin/HardwareIO/src/Convert.hpp
```diff
@@ -35,6 +35,10 @@ namespace module::platform::darwin {
 
         /// The Gyrosocope value from the Darwin is between 0 and 1023, representing a value between -500 and
         /// 500 degrees per second. This means 512 = 0
+        /// 1888.0 is an empirically measured value for this factor, which was calculated by spinning the robot on an
```
Bidski, line 38:
```suggestion
        /// 1880.0 is an empirically measured value for this factor, which was calculated by spinning the robot on an
```

Bidski (changes requested):

PR #453, opened 2021-05-01
Examine and update SensorFilter

KipHamiltons:
@TheCoolKidsTable, Cameron Steggles and I went through SensorFilter to try to make some sense of it, annotating the .cpp as we went.
This PR looks to add comments/make the code generally more readable, as well as highlight some strange parts, and seek clarification on others. I'll add review comments on the file to highlight the areas which I would like to discuss more


module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -491,43 +514,37 @@ namespace module::platform::darwin {
...
                     sensors->feet[side].Hwf = footlanding_Hwf[side].matrix();
                 }
 
                 // Calculate our time offset from the last read
-                double deltaT = std::max(
+                // too hard to read
+                const double deltaT = std::max(
                     (input.timestamp - (previousSensors ? previousSensors->timestamp : input.timestamp)).count()
                         / double(NUClear::clock::period::den),
                     0.0);
```
KipHamiltons, line 534: this is unreadable IMO

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -491,43 +514,37 @@ namespace module::platform::darwin {
...
-                // Gives us the quaternion representation
+                // Convert the motion filter's state vector to a nicer representation, so we can access its elements
                 const auto o = MotionModel<double>::StateVec(motionFilter.get());
 
-                // Map from world to torso coordinates (Rtw)
+                // We make Hwt first, because `o` is in world space
                 Eigen::Affine3d Hwt;
                 Hwt.linear()      = o.Rwt.toRotationMatrix();
-                Hwt.translation() = o.rTWw;
+                Hwt.translation() = -o.rTWw;
```
KipHamiltons, line 545: pretty sure this is negative to what it should be in master. On master we are storing a vector from world to torso in a transform from torso to world. That seems wrong to me

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -491,43 +514,37 @@ namespace module::platform::darwin {
...
                 sensors->Htw      = Hwt.inverse().matrix();
 
-                // Integrate gyro to get angular positions
-                sensors->angular_position = o.omegaTTt / 90.0;
```
KipHamiltons, line 520: This is baffling, but the result was never used anyway (besides below, where it logged the rotational "position" as `p_x`, `p_y`, and `p_z`???). Also, that field in the message file had a comment saying that it was deprecated, so I removed it.

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -537,26 +554,23 @@ namespace module::platform::darwin {
...
+                    // Also removing the yaw in the same way
                     Eigen::AngleAxisd(-Hwt.rotation().eulerAngles(0, 1, 2).z(), Eigen::Vector3d::UnitZ())
                     * Hwt.rotation());
-                Eigen::Affine3d Hgt(Hwt);
+                Eigen::Affine3d Hgt;
                 Hgt.translation() = Eigen::Vector3d(0, 0, Hwt.translation().z());
-                Hgt.linear()      = yawlessWorldInvR.linear();
-                sensors->Hgc      = Hgt * Htc;  // Rwt * Rth
-
+                Hgt.linear()      = yawlessWorldInvR.toRotationMatrix();
```
KipHamiltons, line 573: We calculate a different `Hgt` to the one we just put in the `sensors` message, then do nothing with it. This seems wrong.

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
TrentHouliston: `Hwt.translation()` should always be `rTWw` by definition. `-rTWw` would make it `rWTw` which is definitely wrong here

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -145,7 +144,7 @@ namespace module::platform::darwin {
             process_noise.omegaTTt_bias      = this->config.motionFilter.noise.process.gyroscopeBias;
             motionFilter.model.process_noise = process_noise;
 
-            // Update our mean configs and if it changed, reset the filter
+            // Set our mean configs
```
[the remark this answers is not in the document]
TrentHouliston: Resetting the filter when you change the config is a good idea for when you're trying to tune it.
Also, make sure that when you're reinitialising the filter you're not just setting it to some default value, set it based on one of the sensors (e.g. the accelerometer) which will make it less likely to explode. @ysims was going to do that at some point but I'm not sure if they ever go around to it

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -303,9 +302,9 @@ namespace module::platform::darwin {
                         break;
                     }
 
-                    // If we have previous sensors and our current sensors have an error
-                    // we then use our previous sensor values with some updates
-                    if (previousSensors && error != DarwinSensors::Error::OK) {
+                    // If we have a previous Sensors message and our current Sensors message for this servo has an
+                    // error, then we use our previous sensor values with some updates
+                    if (error != DarwinSensors::Error::OK && previousSensors) {
```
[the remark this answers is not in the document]
TrentHouliston: Such computational savings! Much wow!

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -338,26 +337,32 @@ namespace module::platform::darwin {
                     }
                 }
 
+                // It's worth making absolutely sure that these are correct for a NUgus...
```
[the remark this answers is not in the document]
TrentHouliston: Likely some combination of NUsight2 chart view and spinning the robot

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -338,26 +337,32 @@ namespace module::platform::darwin {
...
                 // gyro_x to the right
                 // gyro_y to the back
                 // gyro_z down
 
+                // This is a left-handed coordinate system?? that's not just a rotation of the CM740, if it's normally
+                // a right-handed coordinate system
```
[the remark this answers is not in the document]
TrentHouliston: It shouldn't be one, but the chip inside of the CM740 might be a left-handed coordinate system. That thing is weird. It is worth checking though

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -338,26 +337,32 @@ namespace module::platform::darwin {
...
+                        // accelerometer.x is backwards
                         Eigen::Vector3d(-input.accelerometer.x, input.accelerometer.y, input.accelerometer.z);
                 }
 
-                // If we have a previous sensors and our cm740 has errors then reuse our last sensor value
+                // If we have a previous sensors and (our cm740 has errors or we are spinning too quickly), then
+                // reuse our last sensor value
                 if (previousSensors
                     && (input.cm740_error_flags
+                        // if it's rotating this quick, something is probably wrong
```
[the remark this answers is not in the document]
TrentHouliston: That's correct, for a while we had a Darwin with a gyroscope that would occasionally throw massive numbers without an error flag which would make it freak out. This check is a result of that although if your hardware is working as intended this should never happen.

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -366,6 +371,7 @@ namespace module::platform::darwin {
                     sensors->gyroscope = previousSensors->gyroscope;
                 }
                 else {
+                    // This smells wrong. y x -z? at the very least this needs an explanatory comment
                     sensors->gyroscope = Eigen::Vector3d(input.gyroscope.y, input.gyroscope.x, -input.gyroscope.z);
```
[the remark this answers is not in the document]
TrentHouliston: This is where people go to play with the axes of the gyroscope, the git history on this line is probably insane

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -399,27 +407,31 @@ namespace module::platform::darwin {
                 sensors->feet[BodySide::LEFT].down  = true;
 
                 std::array<bool, 2> feet_down = {true};
+
+                // If we're using the load value on the foot to work out if our foot is down, do that
                 if (config.footDown.fromLoad) {
-                    // Use our virtual load sensor class to work out which feet are down
+                    // Use our load sensor to work out which foot is down
```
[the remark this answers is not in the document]
TrentHouliston: And hopefully one day on the real robots

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -449,6 +461,7 @@ namespace module::platform::darwin {
                                      MeasurementType::GYROSCOPE());
 
                 // Calculate accelerometer noise factor
+                // Tom has never seen this. Updating the noise dynamically based on the acceleration seems very spicy
                 Eigen::Matrix3d acc_noise =
                     config.motionFilter.noise.measurement.accelerometer
                     + ((sensors->accelerometer.norm() - std::abs(G)) * (sensors->accelerometer.norm() - std::abs(G)))
```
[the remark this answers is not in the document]
TrentHouliston: Blame this one on me. It's a software engineers solution to a mechatronics problem :P 

Since our filter doesn't actually track the acceleration, it just uses the accelerometer as a down vector, however when it's going all over the place (and isn't 9.8m/s^2) what we do is if we don't have a vector that is 9.8m/s^2 we increase the "noise" so it has less impact on the state of the filter. At one point in the past doing this slightly improved our filter when the robot was walking and the noisy "hits" on the ground came in and messed with our filter.

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -474,11 +492,16 @@ namespace module::platform::darwin {
                         footlanding_Hwf[side]                   = Hwt * Htg;
```
TrentHouliston, line 492: I would probably add in something to explain the whole Htg concept since it's a little bit interesting as to how it is supposed to work.

`Htg` does some magic so that the "foot down" position is where the foot would be if it were flat. Therefore as the foot "flattens" it becomes true. So even if your foot hits the ground on the angle it doesn't store the angled position as the "foot down" position, it stores the position that foot would be if/when it is flat on the ground

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -474,11 +492,16 @@ namespace module::platform::darwin {
...
                         footlanding_Hwf[side].translation().z() = 0.0;
 
+                        // This foot was down at this time step, so next time step's previous time step should
+                        // have the foot down
```
TrentHouliston, line 496:
```suggestion
                        // Store the current foot down state for next time
```

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -474,11 +492,16 @@ namespace module::platform::darwin {
...
                         previous_foot_down[side] = true;
                     }
+                    // Else is down, and didn't hit the ground this time step
                     else if (foot_down && prev_foot_down) {
                         // Use stored Hwf and Htf to calculate Hwt
-                        Eigen::Affine3d footlanding_Hwt = footlanding_Hwf[side] * Htf.inverse();
+                        Eigen::Affine3d Hft = Htf.inverse();
+                        // I think that this (footlanding_Hwf) could have weird stuff stored / be uninitialised
+                        Eigen::Affine3d footlanding_Hwt = footlanding_Hwf[side] * Hft;
```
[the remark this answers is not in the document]
TrentHouliston: Shouldn't it have been set when we did the first loop of `foot_down && !prev_foot_down`

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -491,43 +514,37 @@ namespace module::platform::darwin {
...
+                    // Else the foot is off the ground, so we make sure that for the next time step, we know that
+                    // this time step, the foot was off the ground
                     else if (!foot_down) {
                         previous_foot_down[side] = false;
                     }
 
+                    // Note that the Hwf is set, even if the foot is not down. This means that moving feet in the air
+                    // will have an Hwf associated with them which is the transform from when that foot last hit
+                    // the ground
                     sensors->feet[side].Hwf = footlanding_Hwf[side].matrix();
```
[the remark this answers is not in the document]
TrentHouliston: We record it and send it because it makes debugging easier. The (still not quite complete) odometry view in NUsight2 shows the last foot landing position which makes it easier to see where it was when the foot was on the ground.

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
TrentHouliston: Yeah and it's doing dumb stuff... Probably written by me before I knew the std::chrono library better

TrentHouliston, line 534:
```suggestion
                using namespace std::chrono;
                const double deltaT = std::max(
                    duration_cast<duration<double>>(input.timestamp - (previousSensors ? previousSensors->timestamp : input.timestamp)).count(), 0.0);
```

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
TrentHouliston: This is someone getting mad at the odometry filter not giving Euler angles and making their own. Moral of the story, don't ever use Euler angles :P

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -537,26 +554,23 @@ namespace module::platform::darwin {
...
                  ************************************************/
-                Eigen::Affine3d Rwt(sensors->Htw.inverse());
-                // remove translation components from the transform
-                Rwt.translation() = Eigen::Vector3d::Zero();
-                Eigen::Affine3d Rgt(
-                    Eigen::AngleAxisd(-Rwt.rotation().eulerAngles(0, 1, 2).z(), Eigen::Vector3d::UnitZ()) * Rwt);
-                // sensors->Hgt : Mat size [4x4] (default identity)
-                // createRotationZ : Mat size [3x3]
-                // Rwt : Mat size [3x3]
+                // Extract the inverse of the rotation component of Htw
```
TrentHouliston, line 557: Do we even use kinematics horizon anywhere anymore? I thought it was gone

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
TrentHouliston: This is two things sharing the same name, This Hgt is an very old thing that is hopefully not used anywhere anymore

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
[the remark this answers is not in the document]
tayloryoung6396: https://emanual.robotis.com/docs/en/platform/op2/getting_started/ Near the bottom of this is a diagram for the cm740

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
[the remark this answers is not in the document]
ysims: I did do something like that. It is on [sims/kalman-filter](https://github.com/NUbots/NUbots/tree/sims/kalman-filter). It seemed to be okay in the odometry NUsight view but it kept giving Cholesky decomposition errors, but after changing some of the config values it seemed ok? Probably worth revisiting.

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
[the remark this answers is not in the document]
ysims: The board is upside down and turned 90 degrees. If we want to confirm it, can test it on the robot easily enough.

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
ysims: https://nubook.nubots.net/system/foundations/mathematics#homogeneous-transformations 😄

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
ysims: I think @Bidski did that at last RoboCup

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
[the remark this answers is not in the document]
KipHamiltons: I looked at the links [here](https://emanual.robotis.com/docs/en/platform/op2/getting_started/#sub-controllercm-740) for the accelerometer and gyro separately, and both products appear to have right handed coordinates on their documentation

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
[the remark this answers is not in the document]
KipHamiltons: I would like to confirm it :p

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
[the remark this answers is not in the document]
KipHamiltons: yea. that makes sense.

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
KipHamiltons: yikes. I had this wrong then lol 😬

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
KipHamiltons: I don't know. @ysims

ysims: I don't even know what that is... so no? 😆

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -338,26 +337,32 @@ namespace module::platform::darwin {
                     }
                 }
 
+                // It's worth making absolutely sure that these are correct for a NUgus...
                 // gyro_x to the right
```
tayloryoung6396, line 341: IMO This block of comments is correct, The Gyro is right handed and the accelerometer is left handed

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
tayloryoung6396, line 375: Howevet I believe it should be like this?
sensors->gyroscope = Eigen::Vector3d(-input.gyroscope.y, -input.gyroscope.x, -input.gyroscope.z);

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
TrentHouliston: The way I remember it is if you have `Hab`, it transforms from `b` to `a`, however, what it is constructed from is 4 vectors as the columns, all these values are measured in `a` space. They are the X, Y and Z axis of the `b` space as measured in `a` space, and then the translation vector to `b` space measured in `a` space.

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
KipHamiltons: You've omitted the `den` term - was that deliberate?

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
KipHamiltons: oh, so the rotation matrix is like a change of basis. Of course it is... I never realised that lol :facepalm: 
thanks!

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
KipHamiltons: so I should delete everything from the `kinematics horizon` block comment up to but not including the `emit(std::move(sensors))`?

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -267,15 +266,15 @@ namespace module::platform::darwin {
...
-                // Read through all of our sensors
-                for (uint32_t i = 0; i < 20; ++i) {
-                    auto& original = utility::platform::darwin::getDarwinServo(i, input);
+                // Loop through all the servos, emitting a message::input::Sensors for each one
+                for (uint32_t id = 0; id < 20; ++id) {
+                    auto& original = utility::platform::darwin::getDarwinServo(id, input);
                     auto& error    = original.error_flags;
 
                     // Check for an error on the servo and report it
                     while (error != DarwinSensors::Error::OK) {
```
KipHamiltons, line 275: Could this be an `if`, or am I missing something?

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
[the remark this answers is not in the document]
KipHamiltons: @TheCoolKidsTable and @CampbellRoss had a look at this with me, and it seems that the filter is reset just after this anyway - the initial mean, initial covariance, and process noise are reset to the new configuration values, and the UKF has `set_state` called with those new initial matrices.

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
TrentHouliston: It's literally a change of basis :P

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
TrentHouliston: I omitted the `num` term, and that was technically lazyness. because nano/pico/milli etc all have a 1 as the num term. Technically shouldn't do that

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
TrentHouliston: Yeah, I have no idea why this is a while, it unconditionally breaks at the end of the first loop...

TrentHouliston: https://github.com/NUbots/NUbots/commit/2c98722097ac19311860f2e90ea88617fabeb1d4
Apparently @JakeFountain did it :P

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
KipHamiltons: In the previous calculation it also had a `den` term - I don't think those are 1 :p

KipHamiltons: I have gone with multiplying it by a constexpr double which is `nuclear::..::num / nuclear::...::den`. lmk if that's incorrect

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
KipHamiltons: I have deleted the stuff which wasn't added to the `sensors` message which seemed to serve zero purpose.

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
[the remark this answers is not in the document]
Bidski: The comments do seem to imply that the code is. If you verified the correctness of this code then this needs updating

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
Bidski: This was matt trying to compensate pitch/roll in the walk engine

KipHamiltons:
Superseded by #557

PR #188, opened 2018-10-12
[WIP] Gazebo Optimisation (Brandon)

Jade-Fountain:
Implements both a connection to [Gazebo](github.com/NUbots/Gazebo) plugin and also NSGA2 Multiobjective optimiser for scripts

PR #454, opened 2021-05-02
Move to the gazebo branch's way of dealing with time/clocks

KipHamiltons:
Now that the addition of real time to the webots status messages has been ratified ([see here](https://github.com/RoboCup-Humanoid-TC/webots/blob/release/projects/samples/contests/robocup/controllers/player/messages.proto#L64)), as I understand it, we can basically just deal with clocks the exact same way that the gazebo project did.
This PR updates the message as such, and copies that implementation for everything clock related.


module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -165,64 +165,69 @@ namespace module::platform::webots {
...
+                // velocity = distance / time
+                const float distance =
+                    utility::math::angle::difference(target.position,
+                                                     getDarwinServo(target.id, sensors).present_position);
+
+                NUClear::clock::duration time = target.time - NUClear::clock::now();
+                float velocity;
+                if (time.count() > 0) {
+                    velocity =
+                        distance / static_cast<float>(time.count()) / static_cast<float>(NUClear::clock::period::den);
```
CMurtagh-LGTM, line 202: Do we need to multiply by the numerator, to make sure nothing breaks if the numerator is not one?

module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -301,13 +306,29 @@ namespace module::platform::webots {
...
+            this->sim_delta  = msg.time - this->current_sim_time;
+            this->real_delta = msg.real_time - this->current_real_time;
+
+            // Calculate our custom rtf - the ratio of the past two sim deltas and the past two real time deltas
+            utility::clock::custom_rtf = static_cast<double>(this->sim_delta + prev_sim_delta)
+                                         / static_cast<double>(this->real_delta + prev_real_delta);
+
+            // Update our current times
+            this->current_sim_time  = msg.time;
+            this->current_real_time = msg.real_time;
```
CMurtagh-LGTM, line 325: Do we really need all this `this->`?

module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -165,64 +165,69 @@ namespace module::platform::webots {
...
+                motorpid_msg.PID.Y = 0.0;
+                motorpid_msg.PID.Z = 0.0;
+                to_send_next.motor_pids.push_back(motorpid_msg);
+            }
+
+            // TODO(KipHamiltons) does this _still_ cause a race condition - as in, if it goes to send a message while
+            // this assignment is happening, what happens?
+            // Avoid a race condition by reassigning the whole message at once, so it won't be sent while the
+            // next message to send is being generated
+            to_send_now = std::move(to_send_next);
```
CMurtagh-LGTM, line 229: This `std::move` does nothing.

module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -301,13 +306,29 @@ namespace module::platform::webots {
...
+            // Calculate our custom rtf - the ratio of the past two sim deltas and the past two real time deltas
+            utility::clock::custom_rtf = static_cast<double>(this->sim_delta + prev_sim_delta)
+                                         / static_cast<double>(this->real_delta + prev_real_delta);
+
+            // Update our current times
+            this->current_sim_time  = msg.time;
+            this->current_real_time = msg.real_time;
         });
 
+        // TODO(KipHamiltons) why every 10 ms? That seems a little random?
```
CMurtagh-LGTM, line 328: It is completely arbitrary. It should be once per message received/time step. Maybe send it when they send us a message?

module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -321,29 +342,27 @@ namespace module::platform::webots {
...
         });
 
         shutdown_handle = on<Shutdown>().then([this] {
             // Disconnect the fd gracefully
-            shutdown(fd, SHUT_RDWR);
-            close(fd);
-            fd = -1;
+            shutdown(this->fd, SHUT_RDWR);
+            close(this->fd);
+            this->fd = -1;
```
CMurtagh-LGTM, line 354: No need for `this`

CMurtagh-LGTM (changes requested):

module/platform/webots/Webots/src/Webots.cpp
KipHamiltons: I don't understand what you mean

module/platform/webots/Webots/src/Webots.cpp
KipHamiltons: I think so, because I'm pretty sure the lamdas can't capture class variables. I think that if we want to use them, we have to use `this->var` or call a getter/setter on `this->get/set()` anyway

module/platform/webots/Webots/src/Webots.cpp
KipHamiltons: I suppose that would be correct. That avoids the potential race condition completely as well

module/platform/webots/Webots/src/Webots.cpp
KipHamiltons: there is a need for `this`. the class variable `fd` can't be captured by the lambda. It's either this, or `this->get_fd/set_fd`

module/platform/webots/Webots/src/Webots.cpp
CMurtagh-LGTM: a `std::ratio` has both a numerator and a denominator, to properly apply it you have to times by `num / den`.

module/platform/webots/Webots/src/Webots.cpp
CMurtagh-LGTM:
```c++
// Type your code here, or load an example.
#include <iostream>
class A {
    int x = 1;

public:
    void lambda(){
        [this](){std::cout << x;}();
    }
};

int main(){
    A a;
    a.lambda();
}
```
This compiles and prints 1

module/platform/webots/Webots/src/Webots.cpp
CMurtagh-LGTM: Wait nvm it actually does :P.

module/platform/webots/Webots/src/Webots.cpp
KipHamiltons: Oh, interesting. Both [`HardwareIO`](https://github.com/NUbots/NUbots/blob/master/module/platform/darwin/HardwareIO/src/HardwareIO.cpp#L381) and the [gazebo set up](https://github.com/NUbots/NUbots/pull/188/files#diff-6f602e793f7adecdd41295d32cee5df47594442833c48e2c77c497b2ec251a94R338) do it this way, so I suppose not?

module/platform/webots/Webots/src/Webots.cpp
CMurtagh-LGTM: It does not prevent the race condition.

KipHamiltons: I think it would. at the end of the `on<Trigger<ServoTargets>, With<DarwinSensors>>` where we do the `std::move`, it could emit the `ActuatorRequest` instead. Then, we would change this to be `on<Trigger<ActuatorRequest>>` instead of periodically. Then we wouldn't need a persisent `to_send_x` at all.

module/platform/webots/Webots/src/Webots.cpp
CMurtagh-LGTM: It just seems like code that will break if the numerator is not 1.

module/platform/webots/Webots/src/Webots.cpp
KipHamiltons: I mean, you can try without the `this`. When I tried with the `fd` one you also commented on, it said I couldn't capture it, and I found a stackoverflow thread which said yea you can't do that, so I used `this` here too 🤷

module/platform/webots/Webots/src/Webots.cpp
KipHamiltons: This makes sense to me, and I'll explain my thinking with an example.
Say I went 2 units of distance in one unit of time => `time.count() == 1`.
Then, alternatively I go 2 units of distance in two units of time => `time.count() == 2`. 
This math works out so that if `time.count()` doubles, with the distance covered equal, you go half as fast, as you would expect to.

CMurtagh-LGTM: Lets say that the tick ratio was `2/3` so after 2 seconds 3 ticks will have passed. Now if we multiply the number of ticks by `1/3` we get that `1` seconds have pass, when in really `2` seconds have passed.

KipHamiltons: I think I see your point. What's your suggested solution?

TrentHouliston: Basically multiply by the numerator, we have been lazy because we have always assumed the numerator is 1 :P

module/platform/webots/Webots/src/Webots.cpp
TrentHouliston: You don't need the this if you capture this in the lambda

KipHamiltons: Roger that. No idea why I had that issue before. Cameron's PR to this branch removes the this's #455 and multiplies by the `num`

module/platform/webots/Webots/src/Webots.cpp
TrentHouliston: You don't need the this if you capture the [this] in the lambda

PR #456, opened 2021-05-03
Revert arma purge

KipHamiltons:
Something in the arma purge didn't go to plan :cry: :sob: 
When running `keyboardwalk` or `keyboardwalkfake` on commits after this PR was merged, the following error occurs when `moving = true` is toggled for the first time, crashing the binary:
`keyboardwalkfake: /usr/local/include/eigen3/Eigen/src/Core/util/XprHelper.h:110: Eigen::internal::variable_if_dynamic<T, Value>::variable_if_dynamic(T) [with T = long int; int Value = 1]: Assertion 'v == T(Value)' failed.`

I couldn't find the exact change which did it, but this PR will get things back to usable while I hunt it down.
Tragic stuff

### DIFF
--- a/cmake/Modules/FindArmadillo.cmake
+++ b/cmake/Modules/FindArmadillo.cmake
@@ -1,0 +1,12 @@
+include(ToolchainLibraryFinder)
+
+# OpenBLAS for armadillo
+find_package(OpenBLAS REQUIRED)
+
+ToolchainLibraryFinder(
+  NAME Armadillo
+  HEADER armadillo
+  LIBRARY armadillo
+)
+
+target_link_libraries(Armadillo::Armadillo INTERFACE OpenBLAS::OpenBLAS)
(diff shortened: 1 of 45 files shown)


ysims (approved):
😭

TrentHouliston:
That error to me looks like you're assigning something to the wrong sized field. Go and look for dynamic eigen matrices/vectors and it'll be an operation with one of those.

KipHamiltons:
The problem is caused by swapping `QuinticWalk.cpp`'s call to `calculateLegJoints` to use the Eigen version of that function. On this branch, if I make that change I have the issue again.
So basically, we know that the problem is somewhere in `calculateLegJoints` or a function it calls. I haven't found any dynamic matrices in my first looks there yet, but I'll keep looking

KipHamiltons:
Wow... the fix for this is so whack man... hahahaha I'll submit a PR now. incredible.

PR #457, opened 2021-05-03
Fix Keyboardwalk Eigen Error

KipHamiltons:
Not sure why this can compile if it's going to fail at run time for sure :shrug:


ysims (approved):
This does indeed fix things :smile:

PR #458, opened 2021-05-05
Adds time step info to message

ysims:
and testing things


module/platform/webots/Webots/src/Webots.cpp
```diff
@@ -292,6 +325,8 @@ namespace module::platform::webots {
                         return;
                     }
 
+                    std::cout << "Received sensor measurements" << std::endl;
```
CMurtagh-LGTM, line 328:
```suggestion
                    log<NUClear::DEBUG>("Received sensor measurements");
```

PR #461, opened 2021-05-18
Reset the UKF "properly" on startup, configuration, and cholesky failure

KipHamiltons:
Basically, at the moment 
1. We only reset the filter on configuration, which means that if we have issues (e.g. caused by cholesky decomposition failure), we can't recover.
2. We assume that we're in the zombie position (or something close to it) when we do reset, which means that if we're not, it's likely to be screwed up/fail again (as I understand)

This PR aims to deal with both of those issues. To address the first point, it changes the return type of functions which perform cholesky decompositions, so we can track when they go bad, triggering a reset in such a case. To address the second point, (I think) we'll average 10 states generated from sensor updates, and use that mean state as an initial state to reset to.

TODO:

- [ ] [Work out how to construct a state from a sensors message](https://github.com/NUbots/NUbots/compare/hamiltons/examine_sensorfilter...hamiltons/odometry#diff-42df2408b0c979d0cb1b44cbae10d75db2b727526c02e780d4dafd1b28fcfc2dR121-R127)

Fixing the space which the gyro is in is beyond the scope of this PR. I think this one will be big enough without that change.


Bidski:
Or you can average measurements from the first N seconds (N need not be an integer) and use those averages to seed the initial state of the filter.

We can average IMU readings and the CoM

shared/utility/math/filter/UKF.hpp
```diff
@@ -177,27 +187,33 @@ namespace utility::math::filter {
...
+            sigma_mean                  = mean;
+            auto&& points_chole_success = generate_sigma_points(mean, covariance, covariance_sigma_weight);
+
+            sigma_points = points_chole_success.first;
 
             // Reset our state for more measurements
             covariance_update = covariance_weights.asDiagonal();
             d.setZero();
             centred_sigma_points = sigma_points.colwise() - sigma_mean;
+            return points_chole_success.second;
```
Bidski, line 207: I think we need an early termination/reset here as well. If `generate_sigma_points` failed then `centred_sigma_points is a useless calculation?
```suggestion
            auto&& sigma_points = generate_sigma_points(mean, covariance, covariance_sigma_weight);

            // Reset our state for more measurements
            covariance_update = covariance_weights.asDiagonal();
            d.setZero();
            centred_sigma_points = sigma_points.first.colwise() - sigma_mean;
            return sigma_points.second;
```

shared/utility/math/filter/UKF.hpp
```diff
@@ -177,27 +187,33 @@ namespace utility::math::filter {
...
         }
 
         template <typename... Args>
-        void time(const Scalar& dt, const Args&... params) {
+        [[no_discard]] bool time(const Scalar& dt, const Args&... params) {
             // Generate our sigma points
-            sigma_points = generate_sigma_points(mean, covariance, covariance_sigma_weight);
+            auto&& points_chole_success = generate_sigma_points(mean, covariance, covariance_sigma_weight);
```
Bidski, line 213: `points_chole_success` makes be barf every time I read it :vomiting_face:

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -87,6 +88,44 @@ namespace module::platform::darwin {
...
+        return mean_state;
+    }
+
+    MotionModel<double>::StateVec sensorsToState(DarwinSensors sensors) {
+        MotionModel<double>::StateVec state;
+
+        // TODO: I have no idea how to construct a state from sensors
+        // maybe look at sims/kalman_filter for ideas
+        return state;
+    }
```
Bidski, line 127: I think the best way to do this is to actually take measurements (calling `motionFilter.measure()`). 

Might be better to average the sensor readings to get a better idea on initial system state and use these readings to initialise the filter rather than using the mean values from the config file. Might also be able to use these averages to get an idea on noise values?

Bidski (changes requested):

shared/utility/math/filter/UKF.hpp
KipHamiltons: I find naming a pair hard as hahaha. do you have any suggestions?

shared/utility/math/filter/UKF.hpp
KipHamiltons: Is this suggestion saying that the class variable `sigma_points` should be changed to a `std::pair`?

That being said, this `std::pair` thing isn't clean at all. I am wondering if there's a better way to do it

Bidski: Oh, didnt realise that was a class variable. I just hate that variable name lol

KipHamiltons: @CMurtagh-LGTM has suggested using `std::optional<sigma_points_type>` or throwing an exception if this is expected to happen more often. How often are we expecting to have cholesky decomposition errors - as in, is it common enough to warrant throwing errors?

tagging @TrentHouliston too. come join the c++ party :sweat_smile:

Bidski: Everywhere that `generate_sigma_points` is called, the first thing we do is assign the result to `sigma_points`. You can always just return a `bool` from `generate_sigma_points` and `generate_sigma_points` can set `sigma_points` internally if the decomposition succeeded

KipHamiltons: Roger. I'll have a go with that.

KipHamiltons: Ahh, it already throws lol. My code was poorly thought through hahahha

module/platform/darwin/SensorFilter/src/SensorFilter.cpp
```diff
@@ -264,15 +264,32 @@ namespace module::platform::darwin {
...
+                                                                                covariance.asDiagonal(),
+                                                                                motionFilter.ALPHA_DEFAULT,
+                                                                                motionFilter.KAPPA_DEFAULT,
+                                                                                motionFilter.BETA_DEFAULT);
+
+                    // If a Cholesky decomposition failed, log the error, then wait another MIN_UPDATES and try again
+                    if (cholesky_result != Eigen::Success) {
+                        switch (cholesky_result) {
+                            case Eigen::NumericalIssue:
+                                NUClear::log<NUClear::WARN>(
```
Bidski, line 277: Inside a reactor you don't need to prefix `log` with `NUClear::` (unless the compiler is getting confused with `std::log` ... in which case it might be better to use `this->log`)

Bidski (changes requested):

KipHamiltons:
Superseded by #481

PR #462, opened 2021-05-19
Moves Convert to be a utility so webots can use it

KipHamiltons:


ysims (dismissed):
LGTM

TrentHouliston (changes requested):
Change the namespace now that you moved convert

KipHamiltons:
We are going to do the conversion for webots in the webots module, since the webots side isn't going to exactly mirror the CM740's output

PR #463, opened 2021-05-19
Make SensorFilter platform agnostic and move it the input meta-module

Bidski:


KipHamiltons (approved):
This compiles so I don't think anything should break with a rename-refactor? LGTM

Bidski:
> This compiles so I don't think anything should break with a rename-refactor? LGTM

You are a brave, brave man :stuck_out_tongue:

PR #464, opened 2021-05-21
Change accelerometer axes so that gravity should be +1g not -1g

Bidski:


tayloryoung6396 (commented):
LGTM You should probably test this..

tayloryoung6396 (approved):
Works

PR #350, opened 2021-01-27
Fix the mutex inside FileWatcher to make it reentrant + Fix unit test.

CMurtagh-LGTM:
Made FileWatcher mutex less restrictive, so you can now have a FileWatch inside a FileWatch without the internal mutex waiting for itself to unlock.

PR #467, opened 2021-05-22
Some more clang tidy "fixes"

KipHamiltons:
Most of this is pretty run of the mill stuff. Lmk where I have done stupid stuff lol :sweat_smile: 

~`pure_evil.cpp` won't compile with the `__attribute__((noreturn))`, but I remember you (@TrentHouliston) telling me something about how that required further fixes because it really should have that attribute (and there's a non-clang-tidy warning without it lol). What were those fixes?~

```cpp
// It's this in `master`, but it won't compile with clang
static void (*const rethrow)(void*, void*, void (*)(void*)) __attribute__((noreturn)) = get_real_throw_func();

// This does compile, but generates a warning because the function `void __cxa_throw(..)` does return when it calls this
static void (*const rethrow)(void*, void*, void (*)(void*)) = get_real_throw_func();
```

(that was dealt with by #350)


shared/utility/vision/fourcc.hpp
```diff
@@ -6,11 +6,13 @@
 
 namespace utility::vision {
 
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
     inline constexpr uint32_t fourcc(const char (&c)[5]) {
         return c[3] << 24 | c[2] << 16 | c[1] << 8 | c[0];
     }
 
     inline std::string fourcc(const uint32_t& v) {
+        // NOLINTNEXTLINE(cppcoreguidelines-avoid-c-arrays,modernize-avoid-c-arrays)
```
TrentHouliston, line 15: I think this section can just be replaced with an initialiser list for a string.

return {...};

shared/utility/math/angle.hpp
```diff
@@ -108,25 +109,19 @@ namespace utility::math::angle {
...
-            if (angleSrc > angleDst) {
-                return dist2;
-            }
-            else {
-                return -dist2;
-            }
+        if (angleSrc > angleDst) {
+            return dist2;
         }
+        return -dist2;
```
TrentHouliston, line 124: This looks like it should really have just been a ternary
```cpp
return (angleSrc > angleDst ? -1 : 1) * (dist1 < dist2 ? dist1 : dist2);

File: shared/utility/io/uart.cpp [hunk 3]
```diff
@@ -164,10 +162,13 @@ namespace utility::io {
         return !(fcntl(fd, F_GETFL) < 0 && errno == EBADF);
     }
 
-    ssize_t uart::read(void* buf, size_t count) {
+    ssize_t uart::read(void* buf, size_t count) const {
```
TrentHouliston, line 165: I would say this shouldn't be const either because reading will change the state of this object (next read will return different data)

shared/utility/io/uart.cpp
```diff
@@ -140,6 +137,7 @@ namespace utility::io {
             }
 
             // Set the speed flags to "Custom Speed" (clear the existing speed, and set the custom speed flags)
+            // TODO(KipHamiltons) Pretty sure this is undefined behaviour
```
TrentHouliston, line 140: This is entirely defined and OK behaviour. This is a standard C way of clearing a flag
```suggestion
```

shared/utility/io/uart.cpp
```diff
@@ -10,10 +10,7 @@
 #include <unistd.h>
 
 namespace utility::io {
-
-    uart::uart() : device(""), fd(-1) {}
```
TrentHouliston, line 14: Why did we get rid of the default constructor?

shared/utility/file/fileutil.cpp
```diff
@@ -158,7 +157,7 @@ namespace utility::file {
             // loop through every file within the directory
             for (auto&& file : listDir(directory)) {
                 // specify the correct path within the directory
-                auto path = directory + "/" + file;
+                auto path = directory.append("/").append(file);
```
TrentHouliston, line 160: This is introducing a giant bug.
This will modify the directory object and appending that text to it so next time you'll end up with the previous file plus a new file every time.

Really this should be done using the `std::filesystem` api to avoid this problem.
We don't need listFiles or listDir anymore since std::filesystem does it

shared/utility/file/fileutil.cpp
```diff
@@ -114,7 +114,7 @@ namespace utility::file {
             closedir(dir);
         }
         else {
-            // TODO Throw an error or something
+            throw std::runtime_error("Attempted to list directory which didn't exist.\n");
```
TrentHouliston, line 117: Don't put newlines in exceptions, it's up to the consumer of these exceptions how to format them and putting a newline in there just makes it more difficult
```suggestion
            throw std::runtime_error("Attempted to list directory which didn't exist.");
```

shared/utility/math/angle.hpp
```diff
@@ -98,7 +98,8 @@ namespace utility::math::angle {
         angleSrc = normalizeAngle(angleSrc);
         angleDst = normalizeAngle(angleDst);
 
-        Scalar max, min;
+        Scalar max;
+        Scalar min;
```
TrentHouliston, line 102: Suprised clang-tidy isn't yelling about these not being initialised

module/extension/FileWatcher/src/FileWatcher.cpp
```diff
@@ -234,7 +234,7 @@ namespace module::extension {
                 for (const auto& element : utility::file::listDir(path)) {
                     // Set our thread local event details
                     FileWatch watch;
-                    watch.path   = path + "/" + element;
+                    watch.path   = path.append("/").append(element);
```
TrentHouliston, line 237: This is once again another bug as you're modifying the path element on each loop, use `std::filesystem`

TrentHouliston (changes requested):

shared/utility/support/evil/pure_evil.cpp
```diff
@@ -21,30 +40,31 @@ extern "C" {
...
-// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables,cert-err58-cpp)
-backtrace_state* state = backtrace_create_state(nullptr, int(true), error_callback, nullptr);
+// Ignore the fact this can potentially throw an exception
+// NOLINTNEXTLINE(cert-err58-cpp,cppcoreguidelines-avoid-non-const-global-variables)
+backtrace_state* state = backtrace_create_state(nullptr, 1, error_callback, nullptr);
 
-using cxa_throw_func_type = void (*)(void*, void*, void (*)(void*));
+// We can't use a using statement here as the attribute on the end will make it explode
+// NOLINTNEXTLINE(modernize-use-using,cppcoreguidelines-avoid-non-const-global-variables)
+typedef void (*cxa_throw_func_type)(void*, void*, void (*)(void*)) __attribute__((noreturn));
```
CMurtagh-LGTM, line 49: What's the benefit of having the attribute?

shared/utility/support/evil/pure_evil.hpp
```diff
@@ -25,28 +23,28 @@
...
 namespace utility::support::evil {
 
     struct StackFrame {
 
         StackFrame() = default;
 
-        StackFrame(uintptr_t pc_, std::string file_, int lineno_, std::string function_)
-            : pc(pc_), file(std::move(std::move(file_))), lineno(lineno_), function(std::move(function_)) {}
+        StackFrame(uintptr_t pc, std::string file, int lineno, std::string function)
+            : pc(pc), file(std::move(file)), lineno(lineno), function(std::move(function)) {}
```
CMurtagh-LGTM: @KipHamiltons Didn't you say this was bad?

shared/utility/support/evil/pure_evil.cpp
TrentHouliston: Because it has to match the signature of the cxa throw function which has this attribute

shared/utility/io/uart.cpp
KipHamiltons: ahh shit. I was misreading this as a minus not tilde.
It still warns that it's implementation defined though

shared/utility/io/uart.cpp
KipHamiltons: I set `uart() = default;` in the .hpp

shared/utility/support/evil/pure_evil.hpp
KipHamiltons: Yes, it is bad. Not sure how this got through lol

shared/utility/vision/fourcc.hpp
```diff
@@ -12,13 +12,11 @@ namespace utility::vision {
...
-                            static_cast<char>((v >> 8) & 0xFF),
-                            static_cast<char>((v >> 16) & 0xFF),
-                            static_cast<char>((v >> 24) & 0xFF),
-                            0};
-        return ch;
+        return {static_cast<char>((v >> 0) & 0xFF),
+                static_cast<char>((v >> 8) & 0xFF),
+                static_cast<char>((v >> 16) & 0xFF),
+                static_cast<char>((v >> 24) & 0xFF),
+                0};
```
TrentHouliston, line 19: Not 100% sure if you need this null on the end when you're constructing a string directly, probably should be checked

shared/utility/support/evil/pure_evil.hpp
TrentHouliston: For the record, this doesn't shadow variables since in context the meaning of both is entirely defined. Up to you if you want to add the underscores to the end. I don't generally bother unless I have a specific need to refer to both individually within the constructor body itself

shared/utility/io/uart.cpp
TrentHouliston: It does? Maybe there is something strange we are doing then...

KipHamiltons: Pretty sure it was a narrowing conversion in the implicit cast before the bitwise `and`. I [checked on compiler explorer](https://godbolt.org/z/PrE5eh71v) and both gcc and clang treat narrowing into an int like this as we would expect

KipHamiltons:
@TrentHouliston could I grab a review on this bad boy pls?
Not being able to compile with clang is preventing us from having more sanitizers :(

shared/utility/math/angle.hpp
```diff
@@ -94,39 +94,17 @@ namespace utility::math::angle {
...
-                return dist2;
-            }
-            else {
-                return -dist2;
-            }
-        }
+        const Scalar dist1 = max - min;
+        const Scalar dist2 = (2.0 * M_PI) - max + min;
+
+        return dist1 < dist2 ? (source > dest ? -dist1 : dist1) : (source > dest ? dist2 : -dist2);
```
CMurtagh-LGTM, line 107: Comment please.

shared/utility/nusight/NUhelpers.hpp
```diff
@@ -27,11 +27,9 @@
 
 namespace utility::nusight {
 
-    namespace {
+    using message::support::nusight::DataPoint;
 
-        using message::support::nusight::DataPoint;
-
-        constexpr float TIMEOUT = 2.5;
+    namespace helpers {
```
CMurtagh-LGTM, line 32: `buildGraph` should be inside an anon namespace and `is_iterable` should be moved to a header in the type_traits folder.

shared/utility/vision/Vision.cpp
```diff
@@ -43,7 +43,7 @@ namespace utility::vision {
         ofs.close();
     }
 
-    const auto getSubImage(uint x, uint y, uint width, uint height, const std::vector<uint8_t>& data) {
+    auto getSubImage(uint x, uint y, uint width, uint height, const std::vector<uint8_t>& data) {
```
CMurtagh-LGTM, line 46: Doesn't this remove the constness of the data?

Also, should we pass the ints by reference?

CMurtagh-LGTM (changes requested):

shared/utility/vision/Vision.cpp
TrentHouliston: having const on a return type that is not a reference is meaninglesss

TrentHouliston (dismissed):

shared/utility/vision/Vision.cpp
CMurtagh-LGTM: We return a eigen map, which does have an internal reference to the data.

shared/utility/nusight/NUhelpers.hpp
KipHamiltons: There is a clang-tidy check (which we have enabled) that warns on anonymous namespaces.

KipHamiltons: and I'll leave the type trait change to you, for a further PR :p

shared/utility/vision/Vision.cpp
KipHamiltons: The map is mapping a `const Eigen::Matrix` though, so I think that it should still be const where it needs to be

CMurtagh-LGTM: We're mapping a `std::vector<uint8_t>`?

shared/utility/nusight/NUhelpers.hpp
CMurtagh-LGTM: Which check is this, so I can look into it.

shared/utility/vision/Vision.cpp
KipHamiltons: I am reading this code as it creates an `const Eigen::Matrix` from the data from the vector, then returns a submatrix (`Eigen::Map`) of that `const Eigen::Matrix`
It's not a hash map. It's basically a submatrix view. [eigen map docs](https://eigen.tuxfamily.org/dox/classEigen_1_1Map.html)
```c++
        return Eigen::Map<const Eigen::Matrix<uint8_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>(data.data(),
                                                                                                         height,
                                                                                                         width)
            .block<5, 5>(y - 2, x - 2);
```

Trent's point is a thing too. It doesn't do anything, and I am pretty sure that's what the clang-tidy warning was looking at.

shared/utility/nusight/NUhelpers.hpp
KipHamiltons: https://clang.llvm.org/extra/clang-tidy/checks/google-build-namespaces.html

[relevant guideline](https://google.github.io/styleguide/cppguide.html#Internal_Linkage)

[relevant reason provided by CMU](https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL59-CPP.+Do+not+define+an+unnamed+namespace+in+a+header+file#:~:text=Do%20not%20define%20an%20unnamed%20namespace%20in%20a%20header%20file&text=When%20an%20unnamed%20namespace%20is,used%20within%20that%20translation%20unit.)

shared/utility/vision/Vision.cpp
TrentHouliston: Basically, it does nothing because it would return a const version of whatever you return. Then if you assigned it to a variable it'd just copy it into the new variable (because it has to copy over the const thing). Therefore it does nothing

shared/utility/nusight/NUhelpers.hpp
CMurtagh-LGTM: Yeah your right.

CMurtagh-LGTM: #532

CMurtagh-LGTM (approved):

shared/utility/nusight/NUhelpers.hpp
```diff
@@ -32,9 +32,10 @@ namespace utility::nusight {
     using message::support::nusight::DataPoint;
 
     namespace helpers {
+        using message::support::nusight::DataPoint;
+        using utility::type_traits::is_iterable;
 
-
-        inline void buildGraph(DataPoint& /*dataPoint*/) {}
+        inline void(DataPoint& /*dataPoint*/) {}
```
CMurtagh-LGTM, line 38: No, bad. :frowning_face:

CMurtagh-LGTM (changes requested):

KipHamiltons:
@CMurtagh-LGTM u enabled auto-merge but still have requested changes?? looool

CMurtagh-LGTM:
> @CMurtagh-LGTM u enabled auto-merge but still have requested changes?? looool

Wrong button, too tired.

CMurtagh-LGTM (approved):

PR #468, opened 2021-05-22
Tools for our RoboCup docker image

Y0sh1dk:
Currently builds and tags a image locally as `nugus_sim:robocup`. This image contains all runtime depedancies and our compiled binaries. Should be tested with the docker command directly as that is how it will be run for robocup. Currently `nugus_controller` crashes a couple of seconds after connecting for me so cannot test rigerously. This will be used for the image we need to submit on 23rd May. This could probably be split up into multiple PR's later.

Build image:
`./b webots build` (probably a good idea to run with `-c` flag to clean config)

Run container:
`./b webots run webots` (runs `webots` role with `127.0.0.1:10020` as the address of Webots server)
alternatively,
```
docker container run --rm --network=host -e ROBOCUP_ROBOT_ID=1 -e ROBOCUP_TEAM_COLOR=Blue -e ROBOCUP_SIMULATOR_ADDR=127.0.0.1:10020 nugus_sim:robocup webots
```
Key Points:
 - Needs to be run on `host` network
 - Needs to be provided all 3 `ROBOCUP` env vars, or will not run
 - `ROBOCUP_TEAM_COLOR` is not set anywhere as we dont use it?

TODO (before submission):
- Fix `nugus_controller` on Webots side
- Make a role for a simple walk with get-up

TODO (later):
 - target for robocup hardware (`robocup2021` target currently the same as generic, and not  used)
 - log to `/robocup-logs`
 - `./b install` logic can probably be cleaned up and test I didnt break it for installing to a robot
 - Integrate with current dockerise tools?
 - GPU support?

Pls test before submission :sweat_smile:


docker/nugus_sim/run.py
```diff
@@ -0,0 +1,103 @@
...
+    # ROBOCUP_ROBOT_ID
+    file_name = "GlobalConfig.yaml"
+    with open(file_name) as file:
+        data = yaml.load(file)
+    data["player_id"] = int(config["ROBOCUP_ROBOT_ID"])
+
+    with open(file_name, "w") as file:
+        yaml.dump(data, file)
+
+    file_name = "GameController.yaml"
```
KipHamiltons, line 67: reusing the same variable names threw me for a loop for a second

docker/nugus_sim/run.py
```diff
@@ -0,0 +1,103 @@
...
+    with open(file_name) as file:
+        data = yaml.load(file)
+    data["player_id"] = int(config["ROBOCUP_ROBOT_ID"])
+
+    with open(file_name, "w") as file:
+        yaml.dump(data, file)
+
+    # ROBOCUP_TEAM_COLOR
+    # ??
```
KipHamiltons, line 76: If we don't use the team colour, does that mean we have an implicit mapping from team number to colour somewhere? @Bidski 
If so, we should make sure that that mapping is correct for this robocup.

.gitignore
```diff
@@ -36,7 +36,6 @@ packer/build.pwi
 
 # LaTeX stuff
 *.aux
-
```
KipHamiltons, line 39: Unrelated change?

docker/usr/local/toolchain/generate_robocup2021_toolchain.py
```diff
@@ -0,0 +1,23 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Generate toolchain files for the docker images")
+    parser.add_argument(
+        "--prefix", default=os.path.join("usr", "local"), help="Prefix path to store generated files in"
+    )
+
+    args = parser.parse_args()
+
+    generate_toolchains.generate(args.prefix, "generic", target)
```
KipHamiltons, line 23:
```suggestion
    generate_toolchains.generate(args.prefix, "robocup2021", target)
```

tools/install.py
```diff
@@ -36,11 +36,12 @@ def run(target, user, config, toolchain, **kwargs):
...
-        "{}{}".format(k, num): "10.1.1.{}".format(num)
-        for num in range(1, num_robots + 1)
-        for k, v in zip(("nugus", "n", "i", "igus"), [num] * num_robots)
-    }.get(target, target)
+    if target != "local":
+        target = {
+            "{}{}".format(k, num): "10.1.1.{}".format(num)
+            for num in range(1, num_robots + 1)
+            for k, v in zip(("nugus", "n", "i", "igus"), [num] * num_robots)
+        }.get(target, target)
```
KipHamiltons, line 44: this hurts my brain lol. Not your changes, just this code :joy:

tools/webots.py
```diff
@@ -0,0 +1,109 @@
...
+    )
+
+    command_run.add_argument("role", nargs=1, help="Role to run")
+
+
+def run(sub_command, role=None, **kwargs):
+    if sub_command == "build":
+
+        # # Set target
+        # err = subprocess.run(["./b", "target", "robocup2021"]).returncode
```
KipHamiltons, line 44: Remove commented code if it doesn't have a purpose

tools/webots.py
```diff
@@ -0,0 +1,109 @@
...
+        # if err != 0:
+        #     cprint("returned exit code {}".format(err), "red", attrs=["bold"])
+        #     exit(err)
+
+        # Remove build volume
+        if kwargs["clean"] == True:
+            # Clean config
+            err = subprocess.run(["./b", "configure", "--clean"]).returncode
+            if err != 0:
+                cprint("returned exit code {}".format(err), "red", attrs=["bold"])
```
KipHamiltons, line 54:
```suggestion
                cprint("returned exit code {} during NUbots docker volume clean and default configure.".format(err), "red", attrs=["bold"])
```

tools/webots.py
```diff
@@ -0,0 +1,109 @@
...
+                exit(err)
+
+        # Set build config
+        err = subprocess.run(
+            ["./b", "configure", "--", "-DCMAKE_BUILD_TYPE=Release", "--", "-DROLE_webots=ON"]
```
KipHamiltons, line 59: double `"--"`? I don't follow what the second `"--"` is meant to do

tools/webots.py
```diff
@@ -0,0 +1,109 @@
...
+        ).returncode
+        if err != 0:
+            cprint("returned exit code {}".format(err), "red", attrs=["bold"])
```
KipHamiltons, line 62:
```suggestion
            cprint("returned exit code {} configuring NUbots.".format(err), "red", attrs=["bold"])
```

tools/webots.py
```diff
@@ -0,0 +1,109 @@
...
+            exit(err)
+
+        # Build code
+        err = subprocess.run(["./b", "build"]).returncode
+        if err != 0:
+            cprint("returned exit code {}".format(err), "red", attrs=["bold"])
```
KipHamiltons, line 68:
```suggestion
            cprint("returned exit code {} building NUbots.".format(err), "red", attrs=["bold"])
```

tools/webots.py
```diff
@@ -0,0 +1,109 @@
...
+            exit(err)
+
+        # Copy compiled binaries and runtime dependancies out of build volume
+        err = subprocess.run(["./b", "install", "local", "-t"]).returncode
+        if err != 0:
+            cprint("returned exit code {}".format(err), "red", attrs=["bold"])
```
KipHamiltons, line 74:
```suggestion
            cprint("returned exit code {} during copying of binaries to local directory.".format(err), "red", attrs=["bold"])
```

tools/webots.py
```diff
@@ -0,0 +1,109 @@
...
+            exit(err)
+
+        # Build image!
+        os.chdir(b.project_dir + "/docker")
+        err = subprocess.run(
+            ["docker", "build", "-t", "nugus_sim:robocup", "-f", "./nugus_sim_Dockerfile", "."]
+        ).returncode
+        if err != 0:
+            cprint("returned exit code {}".format(err), "red", attrs=["bold"])
```
KipHamiltons, line 83:
```suggestion
            cprint("returned exit code {} during build stage of RoboCup2021 docker image.".format(err), "red", attrs=["bold"])
```

KipHamiltons (commented):
I think we should make a `config.json` file to test with, so we can test our entrypoint and other stuff
Great progress! It's looking really good to me so far! :grin:

.gitignore
Y0sh1dk: Unrelated indeed :stuck_out_tongue:

tools/install.py
Y0sh1dk: Tell me about it :sweat_smile:

tools/webots.py
Y0sh1dk: Once a target is made for robocup this will just need to be uncommented out to use it so I think It can stay for now. It can be uncommented now, but It will just build a image the exact same as generic so no point

tools/webots.py
Y0sh1dk: Woops, the second `"--"` shouldnt be there

docker/nugus_sim/run.py
```diff
@@ -0,0 +1,103 @@
...
+import subprocess
+import sys
+from typing import Tuple
+
+import ruamel.yaml
+
+BINARIES_DIR = "/home/nubots/NUbots/binaries"
+CONFIG_DIR = BINARIES_DIR + "/config"
+
+ENV_VARS = {
```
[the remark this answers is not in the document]
Y0sh1dk: The reason I did it this way was to get the unset vars easily [here](https://github.com/NUbots/NUbots/pull/468/files#diff-3c31cc1e636ef5b1f14b4ed9baa005211f5d862419a498154a0642fd6237680dR42), I'm not sure how you would use a list instead to make it cleaner?

tools/install.py
```diff
@@ -49,18 +50,25 @@ def run(target, user, config, toolchain, **kwargs):
...
 
     # Target location to install to
-    target_dir = "{0}@{1}:/home/{0}/".format(user, target)
+    if target != "local":
+        target_binary_dir = "{0}@{1}:/home/{0}/".format(user, target)
+        target_toolchain_dir = "{0}@{1}:/home/{0}/".format(user, target)
+    else:
+        target_binary_dir = b.project_dir + "/docker/nugus_sim/binaries"
+        target_toolchain_dir = b.project_dir + "/docker/nugus_sim/toolchain"
+
```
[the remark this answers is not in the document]
Y0sh1dk: It was already `rsync`ing to two different directories (on the robot) before, but because they need to be set to different paths if we are installing locally, it made sense to set them there

docker/nugus_sim/run.py
Bidski: This should be set and used somewhere otherwise we are loading the wrong model

docker/nugus_sim/run.py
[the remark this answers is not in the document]
Bidski: You dont actually need/use the values in this map as they are just duplicates of the keys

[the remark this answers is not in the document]
JosephusPaye: Changed to a list

PR #471, opened 2021-05-22
Hopefully corrects conversions

tayloryoung6396:


module/platform/Webots/src/Webots.cpp
```diff
@@ -563,17 +564,17 @@ namespace module::platform {
         if (sensor_measurements.accelerometers.size() > 0) {
             // .accelerometers is a list of one, since our robots have only one accelerometer
             const auto& accelerometer    = sensor_measurements.accelerometers[0];
-            sensor_data->accelerometer.x = static_cast<float>(accelerometer.value.X);
-            sensor_data->accelerometer.y = static_cast<float>(accelerometer.value.Y);
-            sensor_data->accelerometer.z = static_cast<float>(accelerometer.value.Z);
+            sensor_data->accelerometer.x = Convert::accelerometer(static_cast<float>(accelerometer.value.X));
+            sensor_data->accelerometer.y = Convert::accelerometer(static_cast<float>(accelerometer.value.Y));
+            sensor_data->accelerometer.z = Convert::accelerometer(static_cast<float>(accelerometer.value.Z));
```
Bidski, line 569: Should already be in SI units, just need to offset the readings
```suggestion
            sensor_data->accelerometer.x = static_cast<float>(accelerometer.value.X) - 100.0f;
            sensor_data->accelerometer.y = static_cast<float>(accelerometer.value.Y) - 100.0f;
            sensor_data->accelerometer.z = static_cast<float>(accelerometer.value.Z) - 100.0f;
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -563,17 +564,17 @@ namespace module::platform {
...
 
         if (sensor_measurements.gyros.size() > 0) {
             // .gyros is a list of one, since our robots have only one gyroscope
             const auto& gyro         = sensor_measurements.gyros[0];
-            sensor_data->gyroscope.x = static_cast<float>(gyro.value.X);
-            sensor_data->gyroscope.y = static_cast<float>(gyro.value.Y);
-            sensor_data->gyroscope.z = static_cast<float>(gyro.value.Z);
+            sensor_data->gyroscope.x = Convert::gyroscope(static_cast<float>(gyro.value.X));
+            sensor_data->gyroscope.y = Convert::gyroscope(static_cast<float>(gyro.value.Y));
+            sensor_data->gyroscope.z = Convert::gyroscope(static_cast<float>(gyro.value.Z));
```
Bidski, line 577:
```suggestion
            sensor_data->gyroscope.x = static_cast<float>(gyro.value.X) - 100.0f;
            sensor_data->gyroscope.y = static_cast<float>(gyro.value.Y) - 100.0f;
            sensor_data->gyroscope.z = static_cast<float>(gyro.value.Z) - 100.0f;
```

Bidski (changes requested):
I dont think we need the Convert files given the simplicity of the convert functions

module/platform/Webots/src/Webots.cpp
```diff
@@ -562,18 +562,22 @@ namespace module::platform {
 
         if (sensor_measurements.accelerometers.size() > 0) {
```
KipHamiltons, line 563: If this is false, doesn't that mean that the accelerometers are default values (because protobuf) and should be set as such?

Bidski (approved):
Assuming the axes are correct, LGTM

PR #472, opened 2021-05-23
Change from using vectors of ServoTarget to using ServoTargets

ysims:
`std::vector<ServoTarget>` to `ServoTargets`. 
`std::vector<ServoCommand>` to `ServoCommands`.


Bidski (approved):
This looks OK to me

PR #473, opened 2021-05-23
Fix DataLogging so that it compiles

ysims:
DataLogging does not compile on `master`. This fixes it so it compiles.


Bidski:
This change has nothing to do with DataLogging?

ysims:
> This change has nothing to do with DataLogging?

From what I can tell, this is to do with DataLogging using a template that checks for `timestamp` in the proto and treats it as a `google.protobuf.Timestamp`. `MotionCapture.proto` did not have this kind of setup so it stopped DataLogging compiling.

shared/message/input/MotionCapture.proto
```diff
@@ -22,6 +22,7 @@ syntax = "proto3";
 package message.input;
 
 import "Vector.proto";
+import "google/protobuf/timestamp.proto";
```
TrentHouliston, line 25:
```suggestion
```

TrentHouliston (changes requested):
This will break the NatNet module, might be better to replace timestamp with a different name instead.
In NatNet.cpp line 178/181 change mocap->timestamp to something else

ysims:
This gets it to a point where it's not going to effect NatNet, but DataLogging compiles. NatNet does not compile anyway, but I think that should be in a different PR

shared/message/input/MotionCapture.proto
```diff
@@ -101,7 +101,7 @@ message MotionCapture {
     /// https://en.wikipedia.org/wiki/SMPTE_timecode
     uint32 timecode     = 3;
     uint32 timecode_sub = 4;
-    double timestamp    = 5;
+    double time         = 5;
```
Bidski, line 104: `time` probably isnt the best name either to he honest. Maybe `time_step`? Or `message_time`? @TrentHouliston any thoughts?

Bidski (changes requested):

shared/message/input/MotionCapture.proto
ysims: What does it actually represent?

module/input/NatNet/src/NatNet.cpp
```diff
@@ -173,12 +173,12 @@ namespace module::input {
         mocap->timecode    = ReadData<uint32_t>::read(ptr, version);
         mocap->timecodeSub = ReadData<uint32_t>::read(ptr, version);
 
-        // In version 2.9 timestamp went from a float to a double
+        // In version 2.9 time went from a float to a double
         if (version >= 0x02090000) {
```
ysims, line 177: What even is this if-else??

Bidski: It checks the version and reads in different data?

shared/message/input/MotionCapture.proto
Bidski: I think it is the time since the software (motive) was started. So it is a timestamp

module/input/NatNet/src/NatNet.cpp
ysims: Actually no this makes sense now, I thought it was doing something else.

shared/message/input/MotionCapture.proto
ysims: What about `external_timestamp`? Or if this stuff is very specific to motive, then `motive_timestamp`? Or `received_timestamp`?

Bidski: I think the idea would be to try and sync up this timestamp with the system time. So maybe `motive_timestamp` or `natnet_timestamp` since this is the NatNet module.

Bidski (approved):
Sure. I guess this will do

PR #475, opened 2021-05-23
Fix webots communication

ysims:
This fixes communication so that sending and receiving works as intended.


JosephusPaye:
Testing now

module/platform/Webots/src/Webots.cpp
```diff
@@ -254,6 +254,11 @@ namespace module::platform {
                 if (duration.count() > 0) {
                     speed = diff / (double(duration.count()) / double(NUClear::clock::period::den));
                 }
+                else {
+                    // We have a speed of 0, which means 'move as fast as you can'
+                    // 5.236 == 50 rpm which is similar to the max speed of the servos
+                    speed = 5.236;
```
KipHamiltons, line 260: this seems strange. 
Basically if the clock hasn't progressed, we set the servo to move as quickly as possible??

ysims: If the servo was meant to get to its target now or before now, then we make it move as quick as possible.

KipHamiltons: That makes a lot more sense. Could you please update the comment as such? That comment as is made no sense to me without that context

module/platform/Webots/src/Webots.cpp
```diff
@@ -245,15 +245,24 @@ namespace module::platform {
...
+                // Get the difference between the current time and the time the servo should reach its target
                 NUClear::clock::duration duration = target.time - NUClear::clock::now();
 
                 double speed = 0.0;
+                // If we have a positive duration, find the velocity
                 if (duration.count() > 0) {
                     speed = diff / (double(duration.count()) / double(NUClear::clock::period::den));
                 }
+                else {
+                    // The duraction is negative, so the servo should have reached its position before now
```
ysims, line 261:
```suggestion
                    // The duration is negative, so the servo should have reached its position before now
```

module/platform/Webots/src/Webots.cpp
ysims: How is this?

module/platform/Webots/src/Webots.cpp
```diff
@@ -245,17 +245,21 @@ namespace module::platform {
...
                 NUClear::clock::duration duration = target.time - NUClear::clock::now();
 
                 double speed = 0.0;
+                // If we have a positive duration, find the velocity
                 if (duration.count() > 0) {
                     speed = diff / (double(duration.count()) / double(NUClear::clock::period::den));
                 }
                 else {
-                    // We have a speed of 0, which means 'move as fast as you can'
+                    // The duraction is negative, so the servo should have reached its position before now
```
KipHamiltons:
```suggestion
                    // The duration is negative, so the servo should have reached its position before now
```

module/platform/Webots/src/Webots.cpp
KipHamiltons: Perfect

JosephusPaye:
Summary of findings from my test:

- Doesn't seem to work when we're not TRACE logging. It connects to webots and shows the error below. Meanwhile webots seems continue sending valid SensorMeasurement messages. This happened consistently, with both `webots` and `webots_keyboard` roles.

   ```
   Webots INFO: Connected to 127.0.0.1:10020
   Webots (╯°□°）╯︵ ┻━┻: Failed to read message size from TCP connection
   ```

- With TRACE logging, and with the second controller arg (the timestep) in `kid.wbt` set to 8 (instead of 32), I was able to connect and share messages. With `webots_keyboard`, the robot moves and attempts to walk but falls over. When it starts moving, it looks like webots stops sending the expected SensorMeasurments, and starts sending ones with empty data:

	image

module/platform/Webots/src/Webots.cpp
```diff
@@ -191,7 +191,7 @@ namespace module::platform {
 
         // Loop through the linked list of potential options for connecting. In order of best to worst.
         for (addrinfo* addr_ptr = address; addr_ptr != NULL; addr_ptr = addr_ptr->ai_next) {
-            int fd = socket(addr_ptr->ai_family, addr_ptr->ai_socktype, addr_ptr->ai_protocol);
+            fd = socket(addr_ptr->ai_family, addr_ptr->ai_socktype, addr_ptr->ai_protocol);
```
CMurtagh-LGTM, line 194: This functions was not meant to directly modify the state of the reactor.
```suggestion
            int fd_temp = socket(addr_ptr->ai_family, addr_ptr->ai_socktype, addr_ptr->ai_protocol);
```

CMurtagh-LGTM (changes requested):

module/platform/Webots/src/Webots.cpp
ysims: I assume the other `fd`'s need to be changed to `fd_temp` too? All of them in the rest of this function?

module/platform/Webots/src/Webots.cpp
```diff
@@ -293,168 +299,197 @@ namespace module::platform {
...
+
+                        // TODO do something here with this payload and length to extract the protocol buffer and emit
+                        // it
+                        log<NUClear::TRACE>("Received sensor measurements");
+                        translate_and_emit_sensor(
+                            NUClear::util::serialise::Serialise<SensorMeasurements>::deserialise(payload, length));
+
+                        // Delete the packet we just read ready to read the next one
+                        buffer.erase(buffer.begin(), std::next(buffer.begin(), sizeof(length) + length));
+                    }
```
CMurtagh-LGTM, line 422: This does not make any sense.

CMurtagh-LGTM: After reading it for 10 minutes it does but we still need more comments.

module/platform/Webots/src/Webots.cpp
```diff
@@ -293,168 +299,197 @@ namespace module::platform {
...
+                        char* payload = reinterpret_cast<char*>(buffer.data()) + sizeof(length);
+
+                        // TODO do something here with this payload and length to extract the protocol buffer and emit
+                        // it
+                        log<NUClear::TRACE>("Received sensor measurements");
+                        translate_and_emit_sensor(
+                            NUClear::util::serialise::Serialise<SensorMeasurements>::deserialise(payload, length));
+
+                        // Delete the packet we just read ready to read the next one
+                        buffer.erase(buffer.begin(), std::next(buffer.begin(), sizeof(length) + length));
```
CMurtagh-LGTM, line 421: How does this affect size and capacity?

module/platform/Webots/src/Webots.cpp
```diff
@@ -292,172 +301,201 @@ namespace module::platform {
...
-            }
-            // If we don't have any data then this should correspond to a message with all fields set to defaults
-            else {
-                translate_and_emit_sensor(SensorMeasurements(SensorMeasurements::protobuf_type()));
-            }
+            // Receiving
+            read_io = on<IO>(fd, IO::READ | IO::CLOSE | IO::ERROR).then("Read Stream", [this](const IO::Event& event) {
+                if ((event.events & IO::READ) != 0) {
+                    // Service the watchdog
+                    emit<Scope::WATCHDOG>(ServiceWatchdog<Webots>());
```
TrentHouliston, line 333: I would recommend you service the watchdog only once you process a packet, so that if you start getting garbage you reset

nuclear/message/include/message/MessageBase.hpp
```diff
@@ -25,19 +25,25 @@ namespace NUClear::util::serialise {
             return output;
         }
 
-        static inline T deserialise(const std::vector<char>& in) {
+        static inline T deserialise(const char* in, const size_t& length) {
```
KipHamiltons, line 28:
```suggestion
        [[nodiscard]] static inline T deserialise(const char* in, const size_t& length) {
```

nuclear/message/include/message/MessageBase.hpp
```diff
@@ -25,19 +25,25 @@ namespace NUClear::util::serialise {
...
             }
 
             return out;
         }
 
+        static inline T deserialise(const std::vector<char>& in) {
+
+            // Deserialize it
+            return deserialise(in.data(), in.size());
+        }
```
KipHamiltons, line 45:
```suggestion
        [[nodiscard]] static inline T deserialise(const std::vector<char>& in) {
            return deserialise(in.data(), in.size());
        }
```

CMurtagh-LGTM (commented):
Looks fine, other than all the comments that are still unresolved.

module/platform/Webots/src/Webots.cpp
```diff
@@ -292,172 +301,201 @@ namespace module::platform {
...
-            // Connection failed
-            // TODO(Cameron) Try to reconnect after a delay.
-            // setup_connection(server_address, port);
-            log<NUClear::FATAL>("Quitting due to failed connection attempt");
-            powerplant.shutdown();
-            return;
-        }
+    void Webots::setup_connection() {
+        // This will return false if a reconnection is not currently in progress
+        if (!active_reconnect.exchange(true)) {
```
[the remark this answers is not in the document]
CMurtagh-LGTM: I think it is fine.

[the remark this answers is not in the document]
ysims: Quickly looking at what this means, I think it is fine?
@Bidski opinion?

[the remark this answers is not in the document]
TrentHouliston: You can't use load and store, it would become non atomic

module/platform/Webots/src/Webots.cpp
```diff
@@ -292,172 +304,206 @@ namespace module::platform {
...
+                        unsigned long available = 0;
+                        if (::ioctl(fd, FIONREAD, &available) < 0) {
+                            log<NUClear::ERROR>(fmt::format("Error querying for available data, {}", strerror(errno)));
+                            return;
+                        }
+                        const size_t old_size = buffer.size();
+                        buffer.resize(old_size + available);
+
+                        // Read data into our buffer and resize it to the new data we read
+                        auto bytes_read = ::read(fd, buffer.data() + old_size, available);
```
ysims, line 389: I'm getting a warning about this being an unused variable. Does this need to be stored in a variable? And what's with the `::` here?

TrentHouliston: The bytes_read variable should be used here? if it's not that's a problem

The `::` here is likely because it was stolen from some code I wrote for NUClear or similar. In that code I do this to make sure I'm getting the read function in POSIX, since I think there were other read functions floating around. The `::` at the start defines the root namespace, technically there is a `::` at the start of every single

TrentHouliston, line 387: Bytes read needs to be used here!
If you read less than requested with available you'll end up with junk in your buffer!

TrentHouliston (changes requested):
`bytes_read` needs to be used!

module/platform/Webots/src/Webots.cpp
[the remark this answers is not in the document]
ysims: Sounds like we're all good then.

module/platform/Webots/src/Webots.cpp
ysims: @Bidski @TrentHouliston I'm hoping one of you can help here

module/platform/Webots/src/Webots.cpp
```diff
@@ -408,6 +405,9 @@ namespace module::platform {
                             buffer.erase(buffer.begin(), std::next(buffer.begin(), sizeof(length) + length));
                         }
                     }
+
+                    // Service the watchdog
+                    emit<Scope::WATCHDOG>(ServiceWatchdog<Webots>());
```
TrentHouliston, line 411: This is also still in the wrong place, should be on line 403ish

TrentHouliston (changes requested):

module/platform/Webots/src/Webots.cpp
ysims: If you understand it would you be able to put in the needed comments? Or is there someone else who can? @Bidski?

module/platform/Webots/src/Webots.cpp
TrentHouliston: Erase will reduce the size, but I believe it will not reduce the capacity

module/platform/Webots/src/Webots.cpp
Bidski: something like this then
```suggestion
                        auto bytes_read = ::read(fd, buffer.data() + old_size, available);
                        if (bytes_read < available) {
                            buffer.resize(old_size + bytes_read);
                        }
```

CMurtagh-LGTM:
```suggestion
                        auto bytes_read = ::read(fd, buffer.data() + old_size, available);
                        // Shrink the buffer to the size that was actually read.
                        buffer.resize(old_size + bytes_read);
```

module/platform/Webots/src/Webots.cpp
Bidski: This was fixed here :( https://github.com/NUbots/NUbots/pull/475/commits/3a807992b4ef901582def9316930d79540786ac8

module/platform/Webots/src/Webots.cpp
```diff
@@ -222,34 +234,40 @@ namespace module::platform {
...
-                if (duration.count() > 0) {
-                    speed = diff / std::chrono::duration_cast<std::chrono::duration<double>>(duration).count();
-                }
+
+                // If we have a positive duration, find the velocity.
+                // Otherwise, if the duration is negative or 0, the servo should have reached its position before now
+                // Because of this, we move the servo as fast as we can to reach the position.
+                // 5.236 == 50 rpm which is similar to the max speed of the servos
+                double speed = duration.count() > 0
+                                   ? diff / std::chrono::duration_cast<std::chrono::duration<double>>(duration).count()
```
Bidski, line 269: Pretty sure the `duration_cast` isn't needed here
```suggestion
                                   ? diff / std::chrono::duration<double>(duration).count()
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -292,172 +310,206 @@ namespace module::platform {
...
-                        actuator_requests.motor_velocities.emplace_back(MotorVelocity(servo.name, servo.moving_speed));
+                    // Size of the message, in network endian
+                    const uint32_t Nn = htonl(data.size());
 
-                        // TODO: Add FSRs message
+                    // Send the message size first
+                    if (send(fd, &Nn, sizeof(Nn), 0) != sizeof(Nn)) {
+                        log<NUClear::ERROR>(
+                            fmt::format("Error in sending ActuatorRequests' message size,  {}", strerror(errno)));
+                    }
```
Bidski, line 477: If `connection_active` is false then we have not yet received the welcome message from the robocup controller, so we should not be sending actuator requests yet. (include the next send statement in this if statement as well)
```suggestion
                    if (!connection_active) {
                        // Send the message size first
                        if (send(fd, &Nn, sizeof(Nn), 0) != sizeof(Nn)) {
                            log<NUClear::ERROR>(
                                fmt::format("Error in sending ActuatorRequests' message size,  {}", strerror(errno)));
                        }
                    }
```

Bidski (changes requested):

module/platform/Webots/src/Webots.cpp
```diff
@@ -470,17 +469,21 @@ namespace module::platform {
                     // Size of the message, in network endian
                     const uint32_t Nn = htonl(data.size());
 
-                    // Send the message size first
-                    if (send(fd, &Nn, sizeof(Nn), 0) != sizeof(Nn)) {
-                        log<NUClear::ERROR>(
-                            fmt::format("Error in sending ActuatorRequests' message size,  {}", strerror(errno)));
-                    }
+                    if (connection_active) {
```
Bidski, line 472:
```suggestion
                    // Only send actuator requests if we are connected to the controller
                    if (connection_active) {
```

Bidski (changes requested):

module/platform/Webots/src/Webots.cpp
ysims: Is this all good now?

Bidski (approved):
LGTM

module/platform/Webots/src/Webots.cpp
```diff
@@ -292,172 +308,212 @@ namespace module::platform {
...
+                             length          = read_length(buffer)) {
+                            // Decode the protocol buffer and emit it as a message
+                            char* payload = reinterpret_cast<char*>(buffer.data()) + sizeof(length);
+
+                            translate_and_emit_sensor(
+                                NUClear::util::serialise::Serialise<SensorMeasurements>::deserialise(payload, length));
+
+                            // Delete the packet we just read ready to read the next one
+                            buffer.erase(buffer.begin(), std::next(buffer.begin(), sizeof(length) + length));
+                        }
```
CMurtagh-LGTM, line 417: This still needs better comments.

CMurtagh-LGTM (approved):

PR #477, opened 2021-05-26
Allow extra args to be passed to cmake and ninja

Bidski:


KipHamiltons:
I have seen this before somewhere 🤔😜

TrentHouliston:
Does this mean https://github.com/NUbots/NUbots/pull/477/files#diff-728a6a243e6aa631a402856e897c7c1e5585ad206e98674fb18fcfee2d4ef224R16 isn't as important?

Bidski:
Is this done somewhere else?

KipHamiltons:
We definitely don't need this for `./b configure`. Something does that for us (perhaps the wrap pty stuff?) - we can pass CMake flags the same as in NUWebots.
This is definitely worth having for `./b build` though. Atm ninja doesn't interpret the flags properly if they're after `--`, and if they're before `--`, then argparse throws a fit.

Bidski:
WrapPty certainly doesnt do anything of the sort

KipHamiltons:
Well I don't know what does it for NUbots then, but it's a thing. You can try it. `./b configure -- -DENABLE_CLANG_TIDY=ON` will work in this repo, for example

KipHamiltons:
but `./b build -- --version` won't pass it to ninja without this change

Bidski:
I dont know how its working on configure then. Unless cmake is just ignoring `--`

Bidski:
Does that flag actaully get set in the cmake cache or is it ignored?

KipHamiltons:
I don't know what sets it. It does get set in the cmakecache though. just tested it

Cmake doesn't like the `--`.
```
$ cmake .. -- -GNinja
CMake Error: Unknown argument --
CMake Error: Run 'cmake --help' for all supported options.
```

Bidski:
What does `cmake -GNinja -- ..` do?

KipHamiltons:
```
$ cmake -GNinja -- ..
CMake Error: The source directory "/home/nubots/build/--" does not exist.
Specify --help for usage, or press the help button on the CMake GUI.
```

Bidski:
Interesting, no idea what is handling this then.

Bidski:
@TrentHouliston what is your comment referencing?

KipHamiltons (approved):
This code isn't needed in `./b configure` at the moment, but there's no harm in having it there I suppose and we don't know if it'll stay like that, so :shrug:

PR #479, opened 2021-05-26
Move compiler warnings settings to target level for modules

KipHamiltons:


cmake/CompilerWarnings.cmake
```diff
@@ -3,9 +3,9 @@
 # * https://github.com/lefticus/cppbestpractices/blob/master/02-Use_the_Tools_Available.md
 # * https://github.com/lefticus/cpp_starter_project/blob/master/cmake/CompilerWarnings.cmake
 
-function(set_project_warnings)
+function(set_project_warnings target_name)
```
Bidski, line 6: Might be better to move this into `nuclear/cmake` now

Bidski (changes requested):

nuclear/cmake/CompilerWarnings.cmake
```diff
@@ -3,9 +3,9 @@
 # * https://github.com/lefticus/cppbestpractices/blob/master/02-Use_the_Tools_Available.md
 # * https://github.com/lefticus/cpp_starter_project/blob/master/cmake/CompilerWarnings.cmake
 
-function(set_project_warnings)
+function(set_project_warnings target_name)
```
Bidski, line 6:
```suggestion
function(set_target_warnings target_name)
```

nuclear/cmake/CompilerWarnings.cmake
```diff
@@ -20,30 +20,25 @@ function(set_project_warnings)
...
-    set(PROJECT_WARNINGS ${GCC_WARNINGS})
+  if(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    set(PROJECT_WARNINGS ${TARGET_WARNINGS})
   else()
+    # User is using some other unsupported compiler, such as MSVC
     message(AUTHOR_WARNING "No compiler warnings set for '${CMAKE_CXX_COMPILER_ID}' compiler.")
   endif()
 
-  add_compile_options(${PROJECT_WARNINGS})
+  target_compile_options(${target_name} INTERFACE ${PROJECT_WARNINGS})
```
Bidski, line 42: Should these be private?

nuclear/cmake/CompilerWarnings.cmake
```diff
@@ -20,30 +20,25 @@ function(set_project_warnings)
...
-      -Wlogical-op # warn about logical operations being used where bitwise were probably wanted
-      -Wuseless-cast # warn if you perform a cast to the same type
-  )
-
-  if(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang" OR ENABLE_CLANG_TIDY)
-    set(PROJECT_WARNINGS ${CLANG_WARNINGS})
-  elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    set(PROJECT_WARNINGS ${GCC_WARNINGS})
+  if(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    set(PROJECT_WARNINGS ${TARGET_WARNINGS})
```
Bidski, line 36: Is this even needed?

nuclear/module/NUClearModule.cmake
```diff
@@ -191,6 +191,16 @@ function(NUCLEAR_MODULE)
   # Put it in an IDE group for the module's directory
   set_target_properties(${module_target_name} PROPERTIES FOLDER ${module_path})
 
+  # ####################################################################################################################
+  # Warnings #
+  # ####################################################################################################################
+
+  # * We need to tell cmake where the function for the compiler warnings is (src/cmake)
+  list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/nuclear/cmake")
```
Bidski, line 199: Move this line [here](https://github.com/NUbots/NUbots/blob/10337a29ed89d30b44563b83bee487141de0c9a3/nuclear/CMakeLists.txt#L2)

nuclear/module/NUClearModule.cmake
```diff
@@ -191,6 +191,16 @@ function(NUCLEAR_MODULE)
...
+  include(CompilerWarnings)
+  # * Then we set the warnings with it for the module
+  set_project_warnings(${module_target_name})
```
Bidski, line 202: Or `set_module_warnings`? @TrentHouliston any suggestions on naming scheme?
```suggestion
  set_target_warnings(${module_target_name})
```

Bidski (changes requested):

nuclear/cmake/CompilerWarnings.cmake
KipHamiltons: Sure

nuclear/cmake/CompilerWarnings.cmake
KipHamiltons: Nah. good point

CMakeLists.txt
```diff
@@ -109,6 +103,9 @@ set(NUCLEAR_ROLE_BANNER_FILE
     CACHE PATH "The path the banner to print at the start of each role execution" FORCE
 )
 
+# Treat all normal compiler warnings as errors
+option(WARNINGS_AS_ERRORS "Fail the build when we have compiler or clang-tidy warnings" ON)
+
```
TrentHouliston, line 108: Only do this in CI if you don't want to have angry developers

tools/utility/dockerise/platform.py
```diff
@@ -116,11 +117,19 @@ def build(repository, platform):
...
+    remote_tag = "{0}/{0}:{1}".format(repository, platform)
+    local_tag = "{0}:{1}".format(repository, platform)
+    selected_tag = "{}:selected".format(repository)
+
+    print("Pulling remote image", remote_tag)
+    err = subprocess.run(["docker", "pull", remote_tag]).returncode
+    if err != 0:
+        raise RuntimeError("docker image pull returned a non-zero exit code")
+
+    print("Tagging ", remote_tag, "as", local_tag)
```
TrentHouliston, line 132: Well this seems like poor cohesion for a function named pull :P

tools/target.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import os
```
TrentHouliston, line 3: Shouldn't all the changes in this file be in another PR?

KipHamiltons: You're right. They have been merged to master. I merged master into my branch off his branch, without merging it into his branch. Rookie error on my part lel

tools/utility/dockerise/platform.py
KipHamiltons: Pretty sure I had it named `pull_and_tag` and changed it for some reason I can't remember

nuclear/module/NUClearModule.cmake
TrentHouliston: Or maybe like set default warnings?

Bidski (changes requested):
Need to add utility, message, and extension too

nuclear/extension/NUClearExtension.cmake
```diff
@@ -30,5 +30,9 @@ if(NUCLEAR_LINK_TYPE STREQUAL "SHARED")
   set_target_properties(nuclear_extension PROPERTIES LIBRARY_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin/lib")
 endif()
 
+# Set warnings
+include(CompilerWarnings)
+set_target_warnings(nuclear_extension)
+
```
TrentHouliston, line 36: I feel like these changes would be better to put in the libraries.cmake file in each of these rather than putting them in NUClear Roles directly

Bidski: What benefit does that serve?

TrentHouliston: Means it won't break the 4Tel code when I copy stuff like this back across

Bidski: Would you only be copying extensions across or nuclear roles?

Bidski: Or do you mean all of these changes?

KipHamiltons: If you do mean all of these, does that mean you want a `libraries.cmake` in `shared/message`?

KipHamiltons:
@Bidski can I grab a review pls xx

nusight2/typings/bindings/index.d.ts
```diff
@@ -1,7 +1,11 @@
 declare module 'bindings' {
-  const bindings: (name: string | {
-    bindings: string,
-    module_root: string,
-  }) => any
+  const bindings: (
```
Bidski, line 2: What has changed to our formatter??? Are we using the correct formatter config for nusight files?

Bidski (changes requested):

nusight2/typings/bindings/index.d.ts
KipHamiltons: god damnit. I merged in master again to a PR onto another PR...
These formatting changes are on master but not your branch.

The formatter fixes these if we actually run it properly, but doesn't notice if they're not fixed? idk. it's weird. Basically I found these things after I rebuilt target generic with nodejs inside docker. These are old news :shrug: @JosephusPaye

KipHamiltons:
Alright. Unrelated changes have been stricken from this PR lol

Bidski (approved):
Is @TrentHouliston happy?

nuclear/module/NUClearModule.cmake
```diff
@@ -191,6 +191,14 @@ function(NUCLEAR_MODULE)
   # Put it in an IDE group for the module's directory
   set_target_properties(${module_target_name} PROPERTIES FOLDER ${module_path})
 
+  # ####################################################################################################################
+  # Warnings #
+  # ####################################################################################################################
+
+  include(CompilerWarnings)
+  # * Set the warnings with this function for the module
+  set_target_warnings(${module_target_name})
```
Bidski, line 200: Also apply to `test_module_target_name` target to apply the same scrutiny to tests. Also don't forget to apply to the tests for utility as well. Hopefully catch code isnt too nasty lol

Bidski (changes requested):

KipHamiltons:
I have added the warnings for tests and fixed the existing warnings that created :+1:

KipHamiltons:
@TrentHouliston does this PR match what you had in mind?

module/platform/darwin/HardwareIO/tests/ConversionTest.cpp
```diff
@@ -146,7 +147,7 @@ TEST_CASE("Testing the hardware position conversions to radians", "[hardware][co
                                                            {3073, M_PI_2},
                                                            {4095, M_PI}};
 
-        for (size_t i = 0; i < 20; ++i) {
+        for (uint8_t i = 0; i < 20; ++i) {
```
TrentHouliston, line 150: Why uint8_t? just make it an int.
uint8_t doesn't make anything faster, it still has to go in a register anyway

nuclear/cmake/CompilerWarnings.cmake
```diff
@@ -20,30 +20,18 @@ function(set_project_warnings)
...
       -Wmisleading-indentation # warn if indentation implies blocks where blocks do not exist
-      -Wduplicated-cond # warn if if / else chain has duplicated conditions
-      -Wduplicated-branches # warn if if / else branches have duplicated code
-      -Wlogical-op # warn about logical operations being used where bitwise were probably wanted
-      -Wuseless-cast # warn if you perform a cast to the same type
+      # * We want the warnings below if clang adds support for them
+      # * Until then, these make clang-tidy mad, and we would have to have an extra CI pipeline step to cover them
+      # * -Wlogical-op # warn about logical operations being used where bitwise were probably wanted
+      # * -Wuseless-cast # warn if you perform a cast to the same type
   )
```
TrentHouliston, line 29: Clang-tidy already finds these on it's own, so I don't think we need our compilers to also find them so you can just delete these

nuclear/cmake/CompilerWarnings.cmake
```diff
@@ -3,9 +3,9 @@
 # * https://github.com/lefticus/cppbestpractices/blob/master/02-Use_the_Tools_Available.md
 # * https://github.com/lefticus/cpp_starter_project/blob/master/cmake/CompilerWarnings.cmake
 
-function(set_project_warnings)
+function(set_target_warnings target_name)
```
TrentHouliston, line 6: If you're going to move these into NUClear proper with some defaults then there needs to be an API to change them. NUClear roles is used in more than one project.
Have some configuration variable like the one that is used to set the banner/etc and then you can move these into the nuclear directory.

shared/extension/libraries.cmake
```diff
@@ -3,3 +3,7 @@ find_package(yaml-cpp REQUIRED)
 target_link_libraries(nuclear_extension PUBLIC yaml-cpp)
 
 target_compile_features(nuclear_extension PUBLIC cxx_std_17)
+
+# Set warnings
+include(CompilerWarnings)
+set_target_warnings(nuclear_extension)
```
TrentHouliston, line 9: I have decided you can move these into NUClear proper so long as you add a cache variable thing so that these warnings are configurable from outside of NUClear

shared/tests/utility/math/filter/TestFilters.cpp
```diff
@@ -165,12 +165,12 @@ static const std::array<MeasurementType, 101> measurements = {
     MeasurementType(-1.08344978673423),   MeasurementType(-1.69966623692842),  MeasurementType(-0.777801698154316),
     MeasurementType(-1.57714784288996),   MeasurementType(-0.971672418368469)};
 
-static constexpr int number_of_particles = 1000;
-static constexpr double deltaT           = 0.05;
+static constexpr int num_particles = 1000;
```
TrentHouliston, line 168: What's up with this change?

shared/utility/libraries.cmake
```diff
@@ -27,3 +27,7 @@ add_custom_command(
 target_sources(nuclear_utility PRIVATE "${CMAKE_BINARY_DIR}/recordings")
 
 target_compile_features(nuclear_utility PUBLIC cxx_std_17)
+
+# Set warnings
+include(CompilerWarnings)
+set_target_warnings(nuclear_utility)
```
TrentHouliston, line 33: These can go in NUClear proper so long as you have a way to override it

TrentHouliston (changes requested):

module/platform/darwin/HardwareIO/tests/ConversionTest.cpp
KipHamiltons: It's not to make it faster, it's because `Convert::servoPosition` takes a `uint8_t`. The variable was throwing a narrowing conversion warning IIRC, and since it's a fixed size loop with fewer than 256 iterations, I figured the cleanest solution was just to make the loop variable the same type. The other option would be to explicitly cast it to that type when that function is called, and this just seemed nicer

KipHamiltons:
@TrentHouliston two things:
1. How do you want it to be configurable? I have a couple ideas: add a `${NUCLEAR_COMPILER_WARNINGS}` variable which is what we use when we `include(${NUCLEAR_COMPILER_WARNINGS})` the function, or alternatively perhaps a `NUCLEAR_COMPILER_WARNINGS_DIR` variable which is added by `nuclear/CMakeLists.txt`?
2. If you're keen to add this to nuclear proper, woudl you like me to add back the MSVC support? The repo I got this off had a list of MSVC warnings which I removed because we'll never use MSVC on this repo.

shared/tests/utility/math/filter/TestFilters.cpp
KipHamiltons: shadowed global declaration from `ParticleFitler.hpp`

module/platform/darwin/HardwareIO/tests/ConversionTest.cpp
Bidski: unsigned arithmetic is actually slower as the compiler has to generate extra code to check for overflows (not sure if it turns that off if it knows the bounds or not). But I suppose that given this is a test we dont really care about this performance hit here

module/platform/darwin/HardwareIO/tests/ConversionTest.cpp
```diff
@@ -132,7 +133,7 @@ TEST_CASE("Testing the hardware position conversions to radians", "[hardware][co
 
     for (size_t i = 0; i < 20; ++i) {
```
Bidski, line 134: You should make this a `uint8_t` since you are doing it everywhere else

shared/tests/utility/math/filter/TestFilters.cpp
Bidski: why is there a global variable in `ParticleFilter.hpp`?

shared/tests/utility/motion/TestSplines.cpp
```diff
@@ -198,12 +199,12 @@ TEST_CASE("Test Polynom", "[utility][motion][splines][Polynom]") {
...
         std::vector<long int> coefs = {P[i][5], P[i][4], P[i][3], P[i][2], P[i][1], P[i][0]};
-        Eigen::Vector4i result_true(P[i][7], P[i][8], P[i][9], P[i][10]);
-        int x = P[i][6];
+        Eigen::Matrix<long int, 4, 1> result_true(P[i][7], P[i][8], P[i][9], P[i][10]);
+        long int x = P[i][6];
 
         // Create the polynomial from test coefficients and get the computed differential values for the given x
         utility::motion::splines::Polynom poly(coefs);
-        Eigen::Matrix<int long, 4, 1> result_calc(poly.pos(x), poly.vel(x), poly.acc(x), poly.jerk(x));
+        Eigen::Matrix<long int, 4, 1> result_calc(poly.pos(x), poly.vel(x), poly.acc(x), poly.jerk(x));
```
Bidski, line 207: Should these not be `size_t`?

Bidski (changes requested):

shared/tests/utility/math/filter/TestFilters.cpp
KipHamiltons: There's not. There's a global variable here. The (non-global) variable in particlefilter.hpp shadows this global variable.

module/platform/darwin/HardwareIO/tests/ConversionTest.cpp
KipHamiltons: Interesting point that it's slower generally. I would be interested in reading more about that.
In this case though, I am quite confident that since there is a fixed, small number of iterations, it would unroll the loop anyway and do away with the loop variable entirely

shared/tests/utility/motion/TestSplines.cpp
KipHamiltons: I don't think we want to be doing unsigned arithmetic here. These aren't indices

TrentHouliston: You really don't want to be doing unsigned arithmetic anywhere ever indices or not, it's so bug-prone and gives you literally nothing except a higher max value in exchange. And if you need to deal with numbers bigger than an int64... that sounds insane

shared/tests/utility/motion/TestSplines.cpp
```diff
@@ -280,7 +281,7 @@ TEST_CASE("Test Smooth Spline", "[utility][motion][splines][SmoothSpline]") {
             std::vector<double> splineResult = spline.part(i).polynom.getCoefs();
 
             // Check contraints for first endpoint and then the second endpoint in the polynomial
-            for (int k = 0; k < 2; k++) {
+            for (size_t k = 0; k < 2; k++) {
```
TrentHouliston, line 284: Cast things to int rather than making things size_t, don't let unsigned infect everything that's how you get bugs.
If somewhere for instance you wanted to do k-1 suddenly it's a massive number.
So often you have people go `if (k - 1 < 0)` and that statement is literally never true with unsigned math.

Rather than making everything unsigned to avoid warnings downstream, make the downstream stuff signed

TrentHouliston (changes requested):
Two things,
One, you should probably make those warnings variables cached
Two, stop all the conversions of things to unsigned integers! They are the root cause of so many bugs in code and they provide you literally nothing in exchange

nuclear/extension/NUClearExtension.cmake
```diff
@@ -30,5 +30,11 @@ if(NUCLEAR_LINK_TYPE STREQUAL "SHARED")
   set_target_properties(nuclear_extension PROPERTIES LIBRARY_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin/lib")
 endif()
 
+# Add warnings for the extensions. Use the default ROLE warnings setting if it's not configured manually
+if(NOT NUCLEAR_ROLE_WARNINGS)
```
Bidski, line 34:
```suggestion
if(NOT DEFINED NUCLEAR_ROLE_WARNINGS)
```

nuclear/message/Neutron.cmake
```diff
@@ -176,9 +176,11 @@ target_link_libraries(nuclear_message PUBLIC Eigen3::Eigen)
 target_include_directories(nuclear_message PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
 target_include_directories(nuclear_message PUBLIC ${nt_out})
 
-# Add warnings for the C++ message files
-include(CompilerWarnings)
-set_target_warnings(nuclear_message)
+# Add warnings for the C++ message files. Use the default ROLE warnings setting if it's not configured manually
+if(NOT NUCLEAR_MESSAGE_WARNINGS)
```
Bidski, line 180:
```suggestion
if(NOT DEFINED NUCLEAR_MESSAGE_WARNINGS)
```

nuclear/module/NUClearModule.cmake
```diff
@@ -4,6 +4,11 @@ include(CMakeParseArguments)
 find_package(Threads REQUIRED)
 find_package(NUClear REQUIRED)
 
+# Set up the warnings variable for the modules. Use the default ROLE warnings setting if it's not configured manually
+if(NOT NUCLEAR_MODULE_WARNINGS)
```
Bidski, line 8:
```suggestion
if(NOT DEFINED NUCLEAR_MODULE_WARNINGS)
```

nuclear/utility/NUClearUtility.cmake
```diff
@@ -31,5 +31,11 @@ if(NUCLEAR_LINK_TYPE STREQUAL "SHARED")
   set_target_properties(nuclear_utility PROPERTIES LIBRARY_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin/lib")
 endif()
 
+# Add warnings for the utilities. Use the default ROLE warnings setting if it's not configured manually
+if(NOT NUCLEAR_UTILITY_WARNINGS)
```
Bidski, line 35:
```suggestion
if(NOT DEFINED NUCLEAR_UTILITY_WARNINGS)
```

shared/tests/CMakeLists.txt
```diff
@@ -7,8 +7,11 @@ if(test_src)
 
   set_target_properties(TestShared PROPERTIES FOLDER "shared/tests")
 
-  # Add warnings for the tests
-  set_target_warnings(TestShared)
+  # Add warnings. Use the default ROLE warnings setting if it's not configured manually
+  if(NOT NUCLEAR_UTILITY_WARNINGS)
```
Bidski, line 11:
```suggestion
  if(NOT DEFINED NUCLEAR_UTILITY_WARNINGS)
```

Bidski (changes requested):

shared/tests/utility/motion/TestSplines.cpp
```diff
@@ -19,110 +19,111 @@
...
-                                        {24, 6, 134596},
-                                        {15, 1, 15},
-                                        {39, 21, 62359143990},
-                                        {44, 20, 1761039350070},
-                                        {21, 21, 1},
-                                        {14, 11, 364},
-                                        {43, 0, 1},
-                                        {1, 0, 1},
-                                        {39, 21, 62359143990}};
+std::vector<std::vector<size_t>> C = {{3, 0, 1},
```
KipHamiltons, line 30: Note that these ones testing the combination should be `size_t`, because the `Combination` functions take `size_t` arguments. Fixing that is beyond the scope of this PR

module/platform/darwin/HardwareIO/tests/ConversionTest.cpp
```diff
@@ -146,7 +147,7 @@ TEST_CASE("Testing the hardware position conversions to radians", "[hardware][co
                                                            {3073, M_PI_2},
                                                            {4095, M_PI}};
 
-        for (size_t i = 0; i < 20; ++i) {
+        for (ssize_t i = 0; i < 20; ++i) {
```
TrentHouliston, line 150: I would just use int here, yes `ssize_t` is technically not a bad choice, but having choices makes it harder on your brain. If you can use int, use int :P

module/platform/darwin/HardwareIO/tests/ConversionTest.cpp
```diff
@@ -173,42 +174,42 @@ TEST_CASE("Testing the hardware position conversions to radians", "[hardware][co
     {
         INFO("Testing the inverse position conversions");
 
-        for (size_t i = 0; i < 20; ++i) {
+        for (ssize_t i = 0; i < 20; ++i) {
 
             INFO("Testing inverse motor " << i);
 
-            for (auto& test : inverseTests[i]) {
+            for (auto& test : inverseTests[size_t(i)]) {
```
TrentHouliston, line 181: Is this cast really needed? If so then something needs to be done to fix it so it's not.

module/platform/darwin/HardwareIO/tests/ConversionTest.cpp
```diff
@@ -173,42 +174,42 @@ TEST_CASE("Testing the hardware position conversions to radians", "[hardware][co
     {
         INFO("Testing the inverse position conversions");
 
-        for (size_t i = 0; i < 20; ++i) {
+        for (ssize_t i = 0; i < 20; ++i) {
```
TrentHouliston, line 177: I'd really just use int here, no need for ssize_t when an int will do the trick. It's just one more thing for people to need to understand to read this code

module/platform/darwin/HardwareIO/tests/ConversionTest.cpp
```diff
@@ -224,16 +225,16 @@ TEST_CASE("Testing the hardware speed conversions to radians/second", "[hardware
     {
         INFO("Testing the forward position conversions");
 
-        const std::pair<uint16_t, float> tests[] = {{0, 0.0}, {1023, 1.0}, {1024, 0.0}, {2047, -1.0}};
+        const std::pair<uint16_t, float> tests[] = {{0, 0.0f}, {1023, 1.0f}, {1024, 0.0f}, {2047, -1.0f}};
 
-        for (size_t i = 0; i < 20; ++i) {
+        for (ssize_t i = 0; i < 20; ++i) {
```
TrentHouliston, line 230: Repeat previous comment for all these `ssize_t`s

nuclear/CMakeLists.txt
```diff
@@ -1,6 +1,3 @@
-# We use additional modules for the NUClear roles system
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules/")
```
TrentHouliston, line 2: Pretty sure this should stay here. NUClear roles needs to be able to use the things in NUClear roles.
And no the one in the root cmakelists doesn't count, it's not NUclear roles yet `IWYU`

nuclear/extension/NUClearExtension.cmake
```diff
@@ -30,5 +30,14 @@ if(NUCLEAR_LINK_TYPE STREQUAL "SHARED")
   set_target_properties(nuclear_extension PROPERTIES LIBRARY_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin/lib")
 endif()
 
+# Add warnings for the extensions. Use the default ROLE warnings setting if it's not configured manually
+if(NOT DEFINED NUCLEAR_ROLE_WARNINGS)
```
TrentHouliston, line 34: I'm not 100% sold on the whole NUCLEAR_ROLE name, because there are things called roles in here that could have warnings. Maybe change it to NUCLEAR_ROLES?

nuclear/roles/NUClearRole.cmake
```diff
@@ -42,6 +42,9 @@ function(NUCLEAR_ROLE)
   # modules are used in the role
   set_target_properties(${role} PROPERTIES NUCLEAR_MODULES "${role_modules}")
 
+  # Add warnings for the roles
+  target_compile_options(${role} PRIVATE ${NUCLEAR_ROLE_WARNINGS})
```
TrentHouliston, line 46: Yeah this one here is where it becomes ambiguous with the variable, are you referring to NUCLEAR_ROLE as in this role, or NUCLEAR_ROLES as in the system?

shared/tests/utility/math/filter/TestFilters.cpp
TrentHouliston: If this is a constant, shouldn't it be `SCREAMING_SNAKE_CASE`? which should also solve the problem

TrentHouliston (changes requested):

module/platform/darwin/HardwareIO/tests/ConversionTest.cpp
```diff
@@ -131,9 +131,9 @@ TEST_CASE("Testing the hardware position conversions to radians", "[hardware][co
                                                0.0,      0.0,      0.0,       0.0,      0.0,       0.0,       0.0,
                                                0.0,      0.0,      0.0,       0.0,      0.0,       0.0};
 
-    for (uint8_t i = 0; i < 20; ++i) {
-        Convert::SERVO_DIRECTION[i] = direction[i];
-        Convert::SERVO_OFFSET[i]    = float(offset[i]);
+    for (ssize_t i = 0; i < 20; ++i) {
+        Convert::SERVO_DIRECTION[i] = direction[size_t(i)];
```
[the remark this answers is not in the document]
KipHamiltons: There is a signedness warning, so if you would like the loop variable to be signed, then yes.

nuclear/extension/NUClearExtension.cmake
KipHamiltons: I changed the global one to NUCLEAR_ROLES :+1:

nuclear/roles/NUClearRole.cmake
KipHamiltons: Amended :+1:

KipHamiltons:
@Bidski FYI I removed `Wsign-conversion`, because it warns on [this snippet (courtesy trent)](https://godbolt.org/z/P96a6ch1j), which is silly

TrentHouliston (approved):
LGTM

KipHamiltons:
This dismissal was approved by Alex

PR #482, opened 2021-05-28
[WIP] Add python modules

KipHamiltons:
Building on #146. All of what I have done so far is based on or stolen completely from there.
This is still early stages - looking for feedback on what I have done so far pls


KipHamiltons:
At the moment it configures and starts to build, but fails because in a `.py.cpp` file, we have a conflicting declaration for a sub-message of one of the messages

TrentHouliston (commented):
Seems I started a review ages ago and never finished. So here is what's sitting here that I typed lol

nuclear/message/generate_python_messages.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python3
+import os
+import pkgutil
+import shutil
+import sys
+
+import google.protobuf.message
+
+# TODO(Kip): This file is broken. See `messages.txt` in the build directory. It's full of duplicates. It's created
+# inside this file. We need to fix the custom command which calls this, and we need to fix this.
```
Bidski, line 10: It's not broken. `meesages.txt` is just a log file, it looks the way it is meant to look

PR #488, opened 2021-05-29
Fixes NatNet so that it compiles.

ysims:
NatNet didn't compile. This wasn't noticed because it's not in any roles. 
Looks like `MotionCapture.proto` was changed to snake_case and no one realised it affected NatNet.
Changed `frameNumber` to `frame_number` in `MotionCapture.proto` since it seemed to be the only thing not snake_case in there.


Bidski (changes requested):
Had a quick look. Seems ok (maybe get a second opinion, this is hard to read on my phone). Also, you should make sure there is role that builds this module

CMurtagh-LGTM (approved):
LGTM

KipHamiltons:
@Bidski

Bidski (approved):
Sure

Issue #398, opened 2021-03-24
We should handle unused modules better


KipHamiltons:
Unused modules can become annoying if they're not in roles. Not being in roles means they aren't ever compiled, so things change under them making them more difficult to use if we ever want to unearth them.
Possible solutions include:
- A script to find unused modules. This could be a standalone solution, but that would require it to be ran manually, and deliberated on manually etc. This could also be used as part of other solutions. PRs (#48 #290) have been made to do this.
- A role which has all modules in it. this role could be optional to build, perhaps with a CMAKE option (analagous to -DBUILD_TESTS=ON), but we would get the buildkite agents to run it to make sure that nothing is broken unknowingly by things that aren't in "real" roles.  This role should be ignored by git, and would ideally be generated somehow.
- A b script module integrating those two ideas. It would generate the role, and optionally list the unused modules. the buildkite agents would use it to generate the role for their checks. This is the cleanest solution of those listed.

Something that these solutions don't cover though are modules which aren't connected to other modules. They might be in roles, but they might not emit or consume any messages which are used by the other modules in the roles they're in. This isn't a big issue, but it's something that could theoretically be covered by work on top of a graph of message interactions, such as that provided by #343.

KipHamiltons:
The first bullet point has been addressed (#290 has been merged).
The third bullet point should be the next action for this issue.

KipHamiltons:
As of right now, the following modules are unused, as found by `./b unused`:

| Module | Status |
| ---------- | -------- |
| motion::IKKick | Apparently going to be rewritten (# ) |
| support::VirtualCamera | Useful |
| behaviour::skills::GoalSaver | Dead |
| localisation::OdometryLocalisation | Dead |
| output::eSpeak | Dead |
| support::ReactionTimer | Dead |
| behaviour::tools::SpeakTest | Dead |
| behaviour::skills::WalkEngineStand | I don't know |
| behaviour::skills::WalkPathFollower | I don't know |

I am going to submit PRs to remove the ones I claim are dead.
@ysims @TheCoolKidsTable @Bidski: `WalkEngineStand`/`WalkPathFollower` - are they going to be used again ever?

ysims:
`WalkEngineStand`: I'm not sure how it'd perform with the QuinticWalk. We should probably use it.
[...]

PR #489, opened 2021-05-29
A role with all modules! [not to be merged]

ysims:
This is not intended to be merged, but intended to be a way for us to see, for now, if there are broken modules. Might be something that newer members could work through, since a lot of these are likely to be easier fixes that have just been missed because they're never compiled. 

If you work on fixing anything here, please make a new PR with the fix!

Relates to #398. A better solution should be created. 

Current issue: eSpeak has a library that is not in the docker image (PortAudio). 
Solution: Delete eSpeak or add the library?

### DIFF
--- a/roles/all_modules.role
+++ b/roles/all_modules.role
@@ -1,0 +1,88 @@
+nuclear_role(
+    # We don't care about order because this is just for compiling purposes
+
+    # Behaviour
+    behaviour::Controller
+
+    behaviour::planning::KickPlanner
+    behaviour::planning::SimpleWalkPathPlanner
+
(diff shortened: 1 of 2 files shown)


KipHamiltons:
I think the neatest solution for this long term (suggested by @Bidski a while back) is have a `./b` tool which calls `./b unused`, and pipes the result - the list of unused modules - into a role. The role will be generated as a buildkite pipeline step, so that changes which break unused modules would have their builds fail. It could also be generated by any of us when we needed to check it.

ysims:
Superseded by #633

PR #491, opened 2021-06-01
Add utility to get hostname

JosephusPaye:
This adds a utility to get the system hostname, which uses the env variable `ROBOT_HOSTNAME` if set, otherwise uses `gethostname()`. Also updates everywhere we use `gethostname()` to use the utility.


TrentHouliston (changes requested):
Need to fix up Nuclear Network module as well

JosephusPaye:
> Need to fix up Nuclear Network module as well

Where is the hostname used in there?

shared/utility/support/hostname.cpp
```diff
@@ -0,0 +1,17 @@
+#include "hostname.hpp"
+
+namespace utility::support {
+
+    std::string getHostname() {
+        char* hostname = std::getenv("ROBOT_HOSTNAME");
+
+        if (hostname == nullptr) {
+            char buffer[255];
+            gethostname(buffer, 255);
```
Bidski, line 10: Pedanticness I guess
```suggestion
            ::gethostname(buffer, 255);
```

shared/utility/support/hostname.hpp
```diff
@@ -0,0 +1,13 @@
+#ifndef UTILITY_SUPPORT_HOSTNAME_HPP
+#define UTILITY_SUPPORT_HOSTNAME_HPP
+
+#include <string>
+#include <unistd.h>
```
Bidski, line 5:
```suggestion
extern "C" {
#include <unistd.h>
}
```

shared/utility/support/hostname.hpp
```diff
@@ -0,0 +1,13 @@
...
+
+namespace utility::support {
+
+    std::string getHostname();
```
Bidski, line 9: I feel like you could just move the entire function into here
```suggestion
    inline std::string getHostname() {
        char* hostname = std::getenv("ROBOT_HOSTNAME");

        if (hostname == nullptr) {
            char buffer[255];
            ::gethostname(buffer, 255);    
            hostname = buffer;
        }

        return std::string(hostname);            
    }
```

Bidski (changes requested):

Bidski:
> > Need to fix up Nuclear Network module as well
> 
> Where is the hostname used in there?

Apparently here
https://github.com/NUbots/NUbots/blob/master/module/network/NUClearNet/src/NUClearNet.cpp#L30

shared/utility/support/hostname.hpp
JosephusPaye: Yeah, was wondering that too

module/network/NUClearNet/src/NUClearNet.cpp
```diff
@@ -27,7 +29,8 @@ namespace module::network {
 
         on<Configuration>("NUClearNet.yaml").then([this](const Configuration& config) {
             auto netConfig              = std::make_unique<NUClear::message::NetworkConfiguration>();
-            netConfig->name             = config["name"].as<std::string>();
+            std::string name            = config["name"].as<std::string>();
+            netConfig->name             = name.length() == 0 ? utility::support::getHostname() : name;
```
Bidski, line 33:
```suggestion
            netConfig->name             = name.empty() ? utility::support::getHostname() : name;
```

Bidski (changes requested):

Bidski (approved):
Sure, if it works

JosephusPaye:
Tested and it works!

TrentHouliston (approved):

PR #492, opened 2021-06-01
Add robocup run script

JosephusPaye:
Splitting up @Y0sh1dk's work on #468 with minor improvements.

This one is just the Python script that reads the env variables, sets the config values, and runs the requested role.


docker/nugus_sim/run.py
```diff
@@ -0,0 +1,101 @@
...
+    built_roles = []
+
+    # Built roles
+    for filename in os.listdir(BINARIES_DIR):
+        if os.path.isfile(os.path.join(BINARIES_DIR, filename)):
+            built_roles.append(filename)
+
+    if role not in built_roles:
+        print("The role '" + role + "' does not exist!")
+        sys.exit(1)
```
Bidski:
```suggestion
    if not os.path.exists(os.path.join(BINARIES_DIR, role)):
        print("The role '" + role + "' does not exist!")
        sys.exit(1)
```

docker/nugus_sim/run.py
```diff
@@ -0,0 +1,101 @@
...
+import subprocess
+import sys
+from typing import Tuple
+
+import ruamel.yaml
+
+BINARIES_DIR = "/home/nubots/NUbots/binaries"
+CONFIG_DIR = BINARIES_DIR + "/config"
+
+ENV_VARS = ["ROBOCUP_ROBOT_ID", "ROBOCUP_TEAM_COLOR", "ROBOCUP_SIMULATOR_ADDR"]
```
Bidski, line 12: Might as well make it immutable
```suggestion
ENV_VARS = ("ROBOCUP_ROBOT_ID", "ROBOCUP_TEAM_COLOR", "ROBOCUP_SIMULATOR_ADDR")
```

docker/nugus_sim/run.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def set_env_vars(config: dict) -> None:
+    yaml = ruamel.yaml.YAML()
+    os.chdir(CONFIG_DIR)
+
+    # ROBOCUP_ROBOT_ID
+    global_config_fname = "GlobalConfig.yaml"
+    with open(global_config_fname) as file:
+        global_config_data = yaml.load(file)
```
Bidski: `PyYAML` provides a `safe_load` function, if we need to use `ruamel.yaml` for whatever reason we should make sure we use a similar function
```suggestion
    with open(os.path.join(CONFIG_DIR, "GlobalConfig.yaml") as file:
        global_config_data = yaml.safe_load(file)
```

docker/nugus_sim/run.py
```diff
@@ -0,0 +1,101 @@
...
+        yaml.dump(webots_config_data, file)
+
+
+def run_role(inRole: str) -> None:
+    print(inRole)
+    # Change into binaries directory
+    os.chdir(BINARIES_DIR)
+
+    # Run Binary
+    exit(subprocess.run("./" + inRole).returncode)
```
Bidski:
```suggestion
    exit(subprocess.run(f"./{inRole}").returncode)
```

docker/nugus_sim/run.py
```diff
@@ -0,0 +1,101 @@
...
+        webots_config_data = yaml.load(file)
+    webots_config_data["server_address"] = str(webots_addr)
+    webots_config_data["port"] = int(webots_port)
+
+    with open(webots_config_fname, "w") as file:
+        yaml.dump(webots_config_data, file)
+
+
+def run_role(inRole: str) -> None:
+    print(inRole)
```
Bidski: to be deleted?

docker/nugus_sim/run.py
```diff
@@ -0,0 +1,110 @@
...
+    if len(sys.argv) < 2:
+        print("Please specify a role to run!")
+        sys.exit(1)
+
+    role = sys.argv[1]
+    is_goalie = False
+
+    # Check for the goalie flag
+    if len(sys.argv) > 2:
+        is_goalie = sys.argv[2] == "--goalie"
```
Bidski, line 26: Use `argparse` for handling command line arguments

Bidski (changes requested):

docker/nugus_sim/run.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
+
+import os
+import subprocess
+import sys
+from typing import Tuple
+
+import ruamel.yaml
```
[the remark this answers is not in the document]
JosephusPaye: I don't think there's any specific reason other than it's used in some of the b scripts.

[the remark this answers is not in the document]
Bidski: Hmmm, only in `calibrate_cameras.py`. I would prefer we only used one of either `PyYAML` or `ruamel.yaml`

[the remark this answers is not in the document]
JosephusPaye: Seems like `ruamel.yaml` is used only in [calibrate_cameras.py](https://github.com/NUbots/NUbots/blob/5149f1c0cd1e4ff902d749269d1aa90f52adc075/tools/nbs/calibrate_cameras.py#L14)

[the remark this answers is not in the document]
Bidski: `PyYAML` is used in `tools/nbs/footdown.py` as well as some supporting scripts in `doc`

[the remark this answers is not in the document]
TrentHouliston: `ruamel.yaml` is better than regular yaml if you're editing yaml files. It keeps the same position etc of the elements and keeps comments. If you edit a yaml file using the regular yaml, it'll blow away all comments and formatting you have in the file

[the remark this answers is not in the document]
Bidski: Fair enough, I dont mind if we switch to ruamel for everything, just no point having 2 libraries for the same thing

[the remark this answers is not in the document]
JosephusPaye: Aight, I'll keep `ruamel.yaml` here.

docker/nugus_sim/run.py
```diff
@@ -0,0 +1,98 @@
...
+
+import argparse
+import os
+import subprocess
+import sys
+
+import ruamel.yaml
+
+BINARIES_DIR = "/home/nubots/NUbots/binaries"
+CONFIG_DIR = BINARIES_DIR + "/config"
```
Bidski, line 11: Put these guys in the `argparse` thing, you can make these values the defaults if you want

docker/nugus_sim/run.py
```diff
@@ -0,0 +1,98 @@
...
+
+ENV_VARS = ("ROBOCUP_ROBOT_ID", "ROBOCUP_TEAM_COLOR", "ROBOCUP_SIMULATOR_ADDR")
+
+
+def read_args() -> dict:
+    parser = argparse.ArgumentParser(description="Run a role for webots RoboCup")
+
+    parser.add_argument("role", help="The role to run")
+    parser.add_argument(
+        "--goalie", action=argparse.BooleanOptionalAction, default=False, help="Run the role with goalie behaviour"
```
Bidski, line 21: I have never seen this before, I usually just go with `store_true`
```suggestion
        "--goalie", action="store_true", default=False, help="Run the role with goalie behaviour"
```

docker/nugus_sim/run.py
```diff
@@ -0,0 +1,98 @@
...
+    yaml = ruamel.yaml.YAML(typ="safe")
+
+    # Change into the config directory
+    os.chdir(CONFIG_DIR)
+
+    # Set `player_id` in GlobalConfig.yaml from ROBOCUP_ROBOT_ID
+    with open("GlobalConfig.yaml", "rw") as file:
+        global_config = yaml.load(file)
+        global_config["player_id"] = int(env_vars["ROBOCUP_ROBOT_ID"])
+        yaml.dump(global_config, file)
```
Bidski, line 61: Will this overwrite the file or append to it?

Bidski (changes requested):

docker/nugus_sim/run.py
JosephusPaye: Fixed here, to override: https://github.com/NUbots/NUbots/pull/492/commits/fbbe63f0cf0fb85c8ff5e7bfbd8ea09ce78fdea8. "rw" was not a valid mode.

docker/nugus_sim/run.py
```diff
@@ -7,32 +7,40 @@
 
 import ruamel.yaml
 
-BINARIES_DIR = "/home/nubots/NUbots/binaries"
-CONFIG_DIR = BINARIES_DIR + "/config"
+DEFAULT_BINARIES_DIR = "/home/nubots/NUbots/binaries"
+DEFAULT_CONFIG_DIR = DEFAULT_BINARIES_DIR + "/config"
```
Bidski, line 11: Sorry, I should have picked up on this earlier, but we should be consistent with path usages and should always use `os.path.join` with constructing paths
```suggestion
DEFAULT_BINARIES_DIR = os.path.abspath(os.path.join(os.sep, "home", "nubots", "NUbots", "binaries"))
DEFAULT_CONFIG_DIR = os.path.join(DEFAULT_BINARIES_DIR, "config")
```

docker/nugus_sim/run.py
```diff
@@ -7,32 +7,40 @@
...
+    "ROBOCUP_TEAM_PLAYER3_IP",
+    "ROBOCUP_TEAM_PLAYER4_IP",
+)
 
 
 def read_args() -> dict:
     parser = argparse.ArgumentParser(description="Run a role for webots RoboCup")
 
     parser.add_argument("role", help="The role to run")
+    parser.add_argument("--goalie", action="store_true", default=False, help="Run the role with goalie behaviour")
```
Bidski, line 26: Remove this if we are going to be changing the configs in a different/conflicting way

Bidski (changes requested):

Bidski (approved):
LGTM

PR #496, opened 2021-06-02
Add docker image for Robocup simulation

JosephusPaye:
Continues from @Y0sh1dk's work on #468. This adds the docker image and a `./b` script to build it.

Depends on #497 for the updated `./b install`.


docker/nugus_sim.Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+# Get python to look in /usr/local for packages
+RUN echo $(python -c "import site; print(site.getsitepackages()[0].replace('/usr', '/usr/local'))") \
+    > $(python -c "import site; print(site.getsitepackages()[0])")/local.pth
+COPY /etc/xdg/pip/pip.conf /etc/xdg/pip/pip.conf
+
+# Install python libraries
+RUN pip install ruamel.yaml
+
+# Copy our code in yo
```
Bidski, line 22: yo, buddy

docker/nugus_sim.Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+# Copy run script
+COPY ./nugus_sim/run.py .
+
+# Make sure /usr/local is checked for libraries and binaries
+COPY /etc/ld.so.conf.d/usrlocal.conf /etc/ld.so.conf.d/usrlocal.conf
+RUN ldconfig
+
+# Make a symlink from /usr/local/lib to /usr/local/lib64 so library install location is irrelevant
+RUN cd /usr/local && ln -sf lib lib64
```
Bidski, line 36: I would run ldconfig after making the symlink

docker/nugus_sim/run.py
```diff
@@ -23,13 +23,21 @@
...
 
 def read_config(file_path):
-    with open(file_path, "r") as file:
-        return yaml.load(file)
+    try:
+        with open(file_path, "r") as file:
+            return yaml.load(file)
+    except FileNotFoundError:
+        print(f"config file not found: {file_path}")
+        sys.exit(1)
```
Bidski, line 31: How about returning an empty dictionary (maybe also ensuring directories exist)? Just in case the config doesn't already exist, this would allow you to create it

tools/webots.py
```diff
@@ -0,0 +1,196 @@
...
+
+    build_subcommand = subparsers.add_parser("build", help="Build the docker image for use with Webots")
+    build_subcommand.add_argument("roles", nargs="+", help="The roles to build for the image")
+
+    push_subcommand = subparsers.add_parser(
+        "push",
+        help=textwrap.dedent(
+            """
+            Push the simulation docker image to the TC registry. Run the aws CLI and login with the TC credentials before running this command.
+            See "Uploading Docker Images" in the Robocup "API Specifications for Virtual Soccer Competition" document for details.
```
Bidski, line 28: Is there are way to check that we are logged in properly? If you can detect this it would be nice to get a clean exit if login hasnt happened

tools/webots.py
```diff
@@ -0,0 +1,196 @@
...
+    roles_dir = os.path.join(b.project_dir, "roles")
+    roles_glob = os.path.join(roles_dir, "*.role")
+
+    available_roles = [Path(role_path).stem for role_path in glob.glob(roles_glob)]
+
+    # Ensure that all the roles requested are available
+    for role in roles_to_build:
+        if role not in available_roles:
+            print(f"role '{role}' not found")
+            sys.exit(1)
```
Bidski, line 47:
```suggestion
    all_roles_available = True
    for role in roles_to_build:
        if not os.path.exists(os.path.join(roles_dir, f"{role}.role")):
            print(f"role '{role}' not found")
            all_roles_available = True
    if not all_roles_available:
        sys.exit(1) 
```

tools/webots.py
```diff
@@ -0,0 +1,196 @@
...
+
+    role_flags = [f"-DROLE_{role}=ON" for role in available_roles if role in roles_to_build] + [
+        f"-DROLE_{role}=OFF" for role in available_roles if role not in roles_to_build
+    ]
```
Bidski, line 51: I feel like there is a better way to do this, but cant think of it
```suggestion
    available_roles = [Path(role_path).stem for role_path in glob.glob(roles_glob)]
    role_flags = [f"-DROLE_{role}=ON" for role in available_roles if role in roles_to_build] + [
        f"-DROLE_{role}=OFF" for role in available_roles if role not in roles_to_build
    ]
```

tools/webots.py
```diff
@@ -0,0 +1,196 @@
...
+        "robocup-vhsc-nubots:robocup2021",
+        role,
+    ]
+
+    subprocess.run(docker_run_command)
+
+
+def exec_push():
+    image_name = "robocup-vhsc-nubots"
+    image_tag = "robocup2021"
```
Bidski, line 159: This is duplicated on lines 108-109

Bidski (dismissed):

docker/nugus_sim.Dockerfile
JosephusPaye: [yo](https://en.wikipedia.org/wiki/Yo_(app))

docker/nugus_sim/run.py
JosephusPaye: I reckon we should fail early with the error if we're attempting to load a config that's not in the image. Creating the config could lead to unexpected behaviour if we later read it in the code and try to access a key that doesn't exist.

tools/webots.py
JosephusPaye: Docker will show an authentication error if we attempt to push without first logging in.

KipHamiltons (commented):
Looking basically done :heart_eyes: . All of these are consistency suggestions and things we need to remember for later.

Y0sh1dk (approved):
LGTM, tested and is working. Once a specific toolchain is made some things will need to change, but that can be done later.

KipHamiltons (approved):
LGTM. We can add toolchain stuff in the next day or so

PR #497, opened 2021-06-02
Extend `./b install` to install to local directory

JosephusPaye:
Extracted from #468 and improved. Used as follows:

```
./b install ./path/to/local/dir --local
```

### DIFF
--- a/tools/install.py
+++ b/tools/install.py
@@ -4,7 +4,6 @@ import glob
 import os
 import re
 import subprocess
-from pathlib import Path
 
 from termcolor import cprint
 
@@ -56,8 +55,8 @@ def run(target, local, user, config, toolchain, **kwargs):
         target_toolchain_dir = os.path.abspath(os.path.join(target, "toolchain"))
 
         # Ensure the directories exist
-        Path(target_binaries_dir).mkdir(parents=True, exist_ok=True)
-        Path(target_toolchain_dir).mkdir(parents=True, exist_ok=True)
+        os.makedirs(target_binaries_dir, exist_ok=True)
+        os.makedirs(target_toolchain_dir, exist_ok=True)
     else:
         target_binaries_dir = "{0}@{1}:/home/{0}/".format(user, target)
         target_toolchain_dir = "{0}@{1}:/usr/".format(user, target)
--- a/tools/install.py
+++ b/tools/install.py
@@ -4,6 +4,7 @@ import glob
 import os
 import re
 import subprocess
+from pathlib import Path
 
 from termcolor import cprint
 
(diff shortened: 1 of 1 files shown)


tools/install.py
```diff
@@ -15,32 +16,33 @@
 def register(command):
     command.help = "Install the system onto the target system"
 
-    # Configuration
-    command.add_argument("target", help="the target host to install the packages to")
+    command.add_argument("target", help="The target host or directory to install the packages to")
 
-    command.add_argument("--user", "-u", help="the user to install onto the target with")
+    command.add_argument("--local", "-l", dest="local", action="store_true", help="Install to a local directory")
```
KipHamiltons, line 21:
```suggestion
    command.add_argument("--local", "-l", dest="local", action="store_true", default=False, help="Install to a local directory")
```

tools/install.py
```diff
@@ -15,32 +16,33 @@
...
     command.add_argument(
         "--config",
         "-c",
         choices=["new", "n", "update", "u", "overwrite", "o", "ignore", "i"],
         default="new",
-        help="method to use for configuration files when installing",
+        help="The method to use for configuration files when installing",
     )
 
     command.add_argument("-t", "--toolchain", dest="toolchain", action="store_true")
```
KipHamiltons, line 33: should probably add a default to this too
```suggestion
    command.add_argument("-t", "--toolchain", dest="toolchain", default=False, action="store_true")
```

KipHamiltons (commented):
LGTM otherwise

tools/install.py
JosephusPaye: `False` [is the default](https://docs.python.org/3/library/argparse.html#action).

KipHamiltons:
Is there a way to pass the local directory as the argument for `--local`?
as in, use it like `./b install --local ./local/dir` instead of `./b install ./local/dir --local`. That seems much more idiomatic to me

JosephusPaye:
That could be another way to do it.

I went with `./b install /path/to/folder --local` because the folder _is_ the target, and `--local` is just indicating what type of target it is.

I'm not tied to this approach though, we could go either way.

tools/install.py
```diff
@@ -4,6 +4,7 @@
 import os
 import re
 import subprocess
+from pathlib import Path
```
Bidski, line 7:
```suggestion
```

tools/install.py
```diff
@@ -49,18 +51,28 @@ def run(target, user, config, toolchain, **kwargs):
...
 
     # Target location to install to
-    target_dir = "{0}@{1}:/home/{0}/".format(user, target)
+    if local:
+        target_binaries_dir = os.path.abspath(os.path.join(target, "binaries"))
+        target_toolchain_dir = os.path.abspath(os.path.join(target, "toolchain"))
+
+        # Ensure the directories exist
+        Path(target_binaries_dir).mkdir(parents=True, exist_ok=True)
+        Path(target_toolchain_dir).mkdir(parents=True, exist_ok=True)
```
Bidski, line 60:
```suggestion
        os.makedirs(target_binaries_dir, exist_ok=True)
        os.makedirs(target_toolchain_dir, exist_ok=True)
```

Bidski (changes requested):

Bidski (approved):
LGTM

JosephusPaye:
Merging now, we can adjust the flags as @KipHamiltons suggested later if we need to.

PR #498, opened 2021-06-02
Add a role for RoboCup 2021 with appropriate logging

JosephusPaye:


docker/nugus_sim/run.py
```diff
@@ -100,6 +101,25 @@ def update_config_files(args: dict) -> None:
...
+        # Ensure the log directory exists
+        Path(robot_log_dir).mkdir(parents=True, exist_ok=True)
+
+        # Configure FileLogHandler to write to the RoboCup log directory
+        file_log_handler_config = read_config("FileLogHandler.yaml")
+        file_log_handler_config["log_file"] = f"{robot_log_dir}/log"
+        write_config("FileLogHandler.yaml", file_log_handler_config)
+
+        # Configure DataLogging to write to the RoboCup log directory
+        data_logging_config = read_config("DataLogging.yaml")
```
JosephusPaye, line 119: Should this be reading and writing `robocup2021/DataLogging.yaml` (i.e. the role-specific config)?

roles/robocup2021.role
```diff
@@ -0,0 +1,45 @@
+nuclear_role(
```
JosephusPaye, line 1: Let me know if I've missed anything that should or shouldn't be in here.

JosephusPaye:
Looks like DataLogging is failing to compile due to the issue fixed in #473.

Y0sh1dk:
I think @ysims suggested earlier that the role be named something like `webotsrobocup` instead of something specific to this year, particularly if we end up using webots again next year.

roles/robocup2021.role
```diff
@@ -0,0 +1,45 @@
...
+  # Support
+  support::configuration::GlobalConfig
+  network::NUClearNet
+  # support::NUsight
+  support::configuration::SoccerConfig
+  # input
+  input::GameController
+  # Simulator connection
+  platform::Webots
+  input::SensorFilter
```
ysims, line 18: For some reason webots needs to be after sensorfilter, or it wont work. Was probably the issue in the lab today.

docker/nugus_sim/run.py
Bidski: It can, but doesn't need to be

docker/nugus_sim/run.py
```diff
@@ -4,6 +4,7 @@
 import os
 import subprocess
 import sys
+from pathlib import Path
```
Bidski, line 7:
```suggestion
```

docker/nugus_sim/run.py
```diff
@@ -100,6 +101,25 @@ def update_config_files(args: dict) -> None:
...
     write_config("webots.yaml", webots_config)
 
+    # Configure logging to /robocup-logs if it exists
+    if os.path.exists("/robocup-logs"):
+        # Compute the log directory for this robot, based on ROBOCUP_ROBOT_ID
+        hostname = f"webots{env_vars['ROBOCUP_ROBOT_ID']}"
+        robot_log_dir = f"/robocup-logs/{hostname}"
+
+        # Ensure the log directory exists
+        Path(robot_log_dir).mkdir(parents=True, exist_ok=True)
```
Bidski, line 111:
```suggestion
        os.makedirs(robot_log_dir, exist_ok=True)
```

docker/nugus_sim/run.py
```diff
@@ -100,6 +101,25 @@ def update_config_files(args: dict) -> None:
     webots_config["port"] = int(port)
     write_config("webots.yaml", webots_config)
 
+    # Configure logging to /robocup-logs if it exists
+    if os.path.exists("/robocup-logs"):
```
Bidski, line 105: Why check if it exists? Is this meant to be a mounted volume?

module/support/logging/DataLogging/data/config/robocup2021/DataLogging.yaml
```diff
@@ -0,0 +1,10 @@
+output:
+  directory: log
+  split_size: 5 * GiB
+
+# You can put any messages that is emitted in the system here and it will be logged
+# The below ones are just the most common and here as an example
+messages:
+  message.output.CompressedImage: true
```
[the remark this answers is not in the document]
Bidski: - Vision balls and goals (so we can check to see that we are getting detections)
- GreenHorizon would be nice, but is also a large message (visual mesh too, but also a large message)
- Localisation messages

roles/webotsrobocup.role
```diff
@@ -0,0 +1,45 @@
...
+  # FileWatcher, FileLogHandler and Signal Catcher Must Go First. KinematicsConfiguration usually goes after these and
+  # without it many roles do not run
+  extension::FileWatcher
+  support::SignalCatcher
+  support::logging::FileLogHandler
+  # This must come first as it emits config which many roles depend on (e.g. SensorFilter, WalkEngine)
+  motion::KinematicsConfiguration
+  # Support
+  support::configuration::GlobalConfig
+  network::NUClearNet
```
Bidski, line 11: Not sure we need NUClearNet? Unless we plan on trying to organise inter-robot communication? @TrentHouliston

roles/webotsrobocup.role
```diff
@@ -0,0 +1,45 @@
...
+  # Behaviour
+  behaviour::Controller
+  behaviour::skills::Stand
+  behaviour::skills::FallingRelax
+  behaviour::skills::Getup
+  behaviour::skills::KickScript
+  behaviour::skills::HeadBehaviourSoccer
+  behaviour::planning::KickPlanner
+  behaviour::planning::SimpleWalkPathPlanner # Creates WalkPaths. behaviour::planning::BezierWalkPathPlanner  # Creates
+                                             # WalkPaths.
```
Bidski, line 36:
```suggestion
  behaviour::planning::SimpleWalkPathPlanner # Creates WalkPaths.
```

roles/webotsrobocup.role
```diff
@@ -0,0 +1,45 @@
...
+  behaviour::strategy::SoccerStrategy
+  behaviour::skills::Nod
```
Bidski, line 38: Probably not needed

Bidski (changes requested):

docker/nugus_sim/run.py
JosephusPaye: Yep, it will be a mounted volume which may or may not be there

docker/nugus_sim/run.py
JosephusPaye: OK

module/support/logging/DataLogging/data/config/webotsrobocup/DataLogging.yaml
```diff
@@ -0,0 +1,17 @@
+output:
+  directory: log
+  split_size: 5 * GiB
+
+# You can put any messages that is emitted in the system here and it will be logged
+messages:
+  message.output.CompressedImage: true
```
JosephusPaye, line 7: As I understand we'll be getting `Image`s from webots, not `CompressedImage`s. So we probably need to setup the image compressor to do the conversion, if it isn't already.

module/support/logging/DataLogging/data/config/robocup2021/DataLogging.yaml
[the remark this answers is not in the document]
JosephusPaye: Added all of those. After we get the docker image working with webots we should run the robocup role and see how much log space we use with those large messages enabled.

module/support/logging/DataLogging/data/config/webotsrobocup/DataLogging.yaml
ysims: I think we just need to have ImageCompressor in the role and it'll do the compression when it gets an image? Then it'll log that image.

roles/webotsrobocup.role
```diff
@@ -0,0 +1,43 @@
...
+  vision::GreenHorizonDetector
+  vision::GoalDetector
+  vision::BallDetector
+  # Localisation
+  localisation::BallLocalisation
+  localisation::RobotParticleLocalisation
+  # Behaviour
+  behaviour::Controller
+  behaviour::skills::Stand
+  behaviour::skills::FallingRelax
```
ysims, line 30: We shouldn't have falling relax - the robot wont break itself anyway, and relaxing the joints currently just sends NaN to webots which is not happy at getting NaNs!

module/support/logging/DataLogging/data/config/webotsrobocup/DataLogging.yaml
JosephusPaye: Yep, seems so.

roles/webotsrobocup.role
JosephusPaye: Commenting out for now.

Bidski (approved):
Sure

PR #499, opened 2021-06-03
Create motion scripts to use in Webots

ysims:
This creates a folder for scripts for `webots#` robots. The scripts are tuned to work for the NUgus in Webots.

- `StandUpFront.yaml`
- `StandUpBack.yaml`
- `KickLeft.yaml`
- `KickRight.yaml`

It also updates from using the roll over method for getting up to just using either front or back get up.

When this is run normally (outside of RoboCup), the robot is `docker`, so those files are also updated. (Do we want to somehow make it use the `webots` folder when using webots?)

### DIFF
--- a/module/behaviour/skills/Getup/src/Getup.cpp
+++ b/module/behaviour/skills/Getup/src/Getup.cpp
@@ -104,7 +104,7 @@ namespace module::behaviour::skills {
             if (isFront) {
                 emit(std::make_unique<ExecuteScriptByName>(
                     id,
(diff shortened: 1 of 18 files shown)


wongjoel:
Tested `StandUpFront.yaml` with Taylor and it appears to fail due to arms colliding with the hips

tayloryoung6396 (approved):
LGTM

wongjoel (commented):
LGTM

PR #500, opened 2021-06-04
Only want to use one camera in Webots

ysims:
We don't want to use both cameras because then we go over the bandwidth limit for RoboCup.


tayloryoung6396 (approved):
lgtm

PR #501, opened 2021-06-04
Set a max bound on the velocity we send to webots

ysims:
At the moment we don't set a max bound, which results in webots being sent ridiculous velocity requests. This gives the simulator warnings. We don't want a load of warnings in the simulator console (particularly not in official games!). This PR fixes this so we have a bound.

### DIFF
--- a/module/platform/Webots/src/Webots.cpp
+++ b/module/platform/Webots/src/Webots.cpp
@@ -224,6 +224,8 @@ namespace module::platform {
             min_camera_time_step = config["min_camera_time_step"].as<int>();
             min_sensor_time_step = config["min_sensor_time_step"].as<int>();
 
+            max_velocity = config["max_velocity"].as<double>();
+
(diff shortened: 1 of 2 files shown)


wongjoel (approved):
lgtm

module/platform/Webots/src/Webots.cpp
```diff
@@ -257,8 +259,9 @@ namespace module::platform {
                 // Otherwise, if the duration is negative or 0, the servo should have reached its position before now
                 // Because of this, we move the servo as fast as we can to reach the position.
                 // 5.236 == 50 rpm which is similar to the max speed of the servos
```
wongjoel, line 261: This comment needs to be updated

wongjoel (changes requested):

wongjoel (approved):
lgtm again :)

PR #502, opened 2021-06-04
Update the Webots config so that it works

ysims:
`max_velocity` was missing from the other PR :( so things broke
Updates the server address too, since I don't think this changes? And the port just so it's easier to run.


wongjoel (approved):
lgtm

PR #503, opened 2021-06-04
Update the GameController message

ysims:
There were two new fields added, `numberOfWarnings` and `goalKeeper`. This adds those messages to the GameController message we are expecting.


tayloryoung6396 (commented):
It looks good to me, will need some changes in GameController.cpp if we ever want to use it @Bidski

module/input/GameController/src/GameControllerData.hpp
```diff
@@ -125,8 +127,10 @@ namespace module::input::gamecontroller {
     inline std::ostream& operator<<(std::ostream& os, const Robot& robot) {
         os << "\t\tPenalty state......: " << uint(robot.penaltyState) << std::endl
            << "\t\tPenalised time left: " << uint(robot.penalisedTimeLeft) << std::endl
+           << "\t\tNumber of warnings: " << uint(robot.numberOfWarnings) << std::endl
            << "\t\tYellow Card Count..: " << uint(robot.yellowCardCount) << std::endl
-           << "\t\tRed Card Count.....: " << uint(robot.redCardCount) << std::endl;
+           << "\t\tRed Card Count.....: " << uint(robot.redCardCount) << std::endl
+           << "\t\tGoalkeeper: " << uint(robot.goalKeeper) << std::endl;
```
tayloryoung6396, line 133: this should probs be bool(robot.goalKeeper) or something

tayloryoung6396 (changes requested):

module/input/GameController/src/GameControllerData.hpp
ysims: fixed the same way the other bool is done

tayloryoung6396 (approved):
Lgtm

PR #504, opened 2021-06-05
Force visual mesh to use cpu

tayloryoung6396:


tayloryoung6396 (commented):
@TrentHouliston This is how @Bidski told me to fix the no docker gpu issue, does work, there's probably a better way

TrentHouliston:
> @TrentHouliston This is how @Bidski told me to fix the no docker gpu issue, does work, there's probably a better way

This is the way

TrentHouliston:
Actually, wait, this problem was entirely solved by merging the visual mesh 2 branch

PR #508, opened 2021-06-05
Fix communication in the Webots module to successfully connect to the official RoboCup controller

ysims:
The welcome message was read in as 7 bytes, but 8 bytes were associated with the message. One byte left over messed with the rest of communication. This fixed that.
It also updates the config so that we are using the right team id (as from GameController) and the right host IP address.

### DIFF
--- a/module/platform/Webots/src/Webots.cpp
+++ b/module/platform/Webots/src/Webots.cpp
@@ -312,6 +312,7 @@ namespace module::platform {
             read_io.unbind();
             send_io.unbind();
             error_io.unbind();
+            buffer.clear();
 
             if (fd != -1) {
                 // Disconnect the fd gracefully
@@ -335,16 +336,16 @@ namespace module::platform {
                     emit<Scope::WATCHDOG>(ServiceWatchdog<Webots>());
                     // If we have not seen the welcome message yet, look for it
                     if (!connection_active) {
-                        // Initaliase the string with ???????
-                        std::string initial_message = std::string(7, '?');
-                        const int n                 = ::read(fd, initial_message.data(), sizeof(initial_message));
+                        // Initaliase the string with 0s
+                        std::array<char, 8> initial_message = {'\0', '\0', '\0', '\0', '\0', '\0', '\0', '\0'};
+                        const int n = ::read(fd, initial_message.data(), initial_message.size());
 
                         if (n >= 0) {
-                            if (initial_message == "Welcome") {
+                            if (std::strncmp(initial_message.data(), "Welcome", initial_message.size()) == 0) {
                                 // good
                                 log<NUClear::INFO>(fmt::format("Connected to {}:{}", server_address, server_port));
                             }
-                            else if (initial_message == "Refused") {
+                            else if (std::strncmp(initial_message.data(), "Refused", initial_message.size()) == 0) {
                                 log<NUClear::FATAL>(
                                     fmt::format("Connection to {}:{} refused: your IP is not white listed.",
                                                 server_address,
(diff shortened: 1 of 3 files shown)


module/platform/Webots/src/Webots.cpp
```diff
@@ -335,16 +336,16 @@ namespace module::platform {
...
-                        // Initaliase the string with ???????
-                        std::string initial_message = std::string(7, '?');
-                        const int n                 = ::read(fd, initial_message.data(), sizeof(initial_message));
+                        // Initaliase the string with 0s
+                        std::array<char, 8> initial_message{};
+                        const int n = ::read(fd, initial_message.data(), initial_message.size());
 
                         if (n >= 0) {
-                            if (initial_message == "Welcome") {
+                            if (std::strncmp(initial_message.data(), "Welcome", initial_message.size()) == 0) {
```
TrentHouliston, line 344: Does this work instead?
```suggestion
                            if (std::string("Welcome") == initial_message.data()) {
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -335,16 +336,16 @@ namespace module::platform {
...
                                 // good
                                 log<NUClear::INFO>(fmt::format("Connected to {}:{}", server_address, server_port));
                             }
-                            else if (initial_message == "Refused") {
+                            else if (std::strncmp(initial_message.data(), "Refused", initial_message.size()) == 0) {
```
TrentHouliston, line 348:
```suggestion
                            else if (std::string("Refused") == initial_message.data()) {
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -404,11 +405,16 @@ namespace module::platform {
                             // Decode the protocol buffer and emit it as a message
                             char* payload = reinterpret_cast<char*>(buffer.data()) + sizeof(length);
 
-                            translate_and_emit_sensor(
-                                NUClear::util::serialise::Serialise<SensorMeasurements>::deserialise(payload, length));
-
-                            // Delete the packet we just read ready to read the next one
-                            buffer.erase(buffer.begin(), std::next(buffer.begin(), sizeof(length) + length));
+                            try {
```
TrentHouliston, line 408: If you're going to do a try/catch like this i'd be careful you could end up in a bad state forever.

Maybe only service the watchdog down here on success

TrentHouliston: Actually wait, this try/catch is really bad, you're in a loop here and you only erase from the buffer if you successfully deserialise the protobuf.
So if you fail, you'll go back around this loop, over and over again

TrentHouliston (changes requested):

module/platform/Webots/src/Webots.cpp
Bidski: Yea, this was a hack ibadded to stop the program crashing and to grab some extra info. Should get rid of the try/catch

module/platform/Webots/src/Webots.cpp
TrentHouliston: If you like you can probably swap the sides of the ==

module/platform/Webots/src/Webots.cpp
```diff
@@ -335,16 +336,16 @@ namespace module::platform {
                     emit<Scope::WATCHDOG>(ServiceWatchdog<Webots>());
```
TrentHouliston, line 336: Can you move this emit to after the translate and emit sensor line?

module/platform/Webots/src/Webots.cpp
```diff
@@ -404,9 +403,10 @@ namespace module::platform {
                             // Decode the protocol buffer and emit it as a message
                             char* payload = reinterpret_cast<char*>(buffer.data()) + sizeof(length);
 
+                            // Service the watchdog
+                            emit<Scope::WATCHDOG>(ServiceWatchdog<Webots>());
```
TrentHouliston, line 407: After you deserialise, so if it keeps failing it will reset the connection (one line lower)

module/platform/Webots/src/Webots.cpp
```diff
@@ -331,20 +332,18 @@ namespace module::platform {
...
-                    // Service the watchdog
-                    emit<Scope::WATCHDOG>(ServiceWatchdog<Webots>());
                     // If we have not seen the welcome message yet, look for it
                     if (!connection_active) {
-                        // Initaliase the string with ???????
-                        std::string initial_message = std::string(7, '?');
-                        const int n                 = ::read(fd, initial_message.data(), sizeof(initial_message));
+                        // Initaliase the string with 0s
+                        std::array<char, 8> initial_message{};
+                        const int n = ::read(fd, initial_message.data(), initial_message.size());
```
TrentHouliston, line 339: You know... the more I think about this, the more terrified I am that you'll get a random crash here sometime because the welcome message ends up without a null terminator and the comparison later goes on forever and crashes... Adding an extra null will keep us safe

```suggestion
                        // Initaliase the string with 0s
                        // make sure we have an extra character just in case we read something that isn't a null terminator
                        std::array<char, 9> initial_message{};
                        const int n = ::read(fd, initial_message.data(), initial_message.size() - 1);
```

TrentHouliston (approved):

TrentHouliston (approved):

module/platform/Webots/src/Webots.cpp
JosephusPaye: Unfortunate typo: "Initaliase"

PR #510, opened 2021-06-05
Updates the get ups for Webots

ysims:
Turns out we weren't tuning on the exact field used in the RoboCup world (see NUbots/NUWebots#77). This updates the get ups so they work on the right field.


tayloryoung6396 (approved):
lgtm

PR #511, opened 2021-06-05
Fix the get up so that we don't keep trying to get up

ysims:
Before, we were disabling the reaction so it never got new sensor readings until it finished the script/s. Because of the averaging of the last 20 sensors, it was still using readings from when it first fell down. This caused a loop of trying to get up even when it was stood up again. 
This PR stops the reaction being disabled. Existing checks already make sure the robot doesn't get up again when trying to get up. 
Tested in Webots with the official RoboCup controller, and works.


tayloryoung6396 (commented):
This should continuously try to getup whilst getting up?? Seems like a bad idea, hence disabling fallen check.

ysims:
There's a check already for if we're already getting up.

tayloryoung6396 (approved):
Sure

PR #513, opened 2021-06-08
Changes return types of walk engine functions which mutated many references

KipHamiltons:
~`computeCartesianPosition[AtTime]` takes a boolean parameter `isLeftsupportFoot` (passed by value), which is then passed to `TrajectoriesSupportFootState`, which modifies the value as a reference. The call to that function does nothing because the bool is passed by value into the previous function call (this is the only use of `TrajectoriesSupportFootState`).~

~This changes it to be passed by reference, which is certainly the intention~

Update:
I have made the trajectoryutils functions return a tuple instead of mutating its parameters. This is much nicer

### DIFF
--- a/module/motion/QuinticWalk/src/WalkEngine.cpp
+++ b/module/motion/QuinticWalk/src/WalkEngine.cpp
@@ -31,7 +31,7 @@ namespace module::motion {
         , trunk_axis_acc_at_last()
         , trajs() {
         // Make sure to call the reset method after having the parameters
-        TrajectoriesInit(trajs);
+        trajectoriesInit(trajs);
     }
 
     bool QuinticWalkEngine::updateState(const float dt, const Eigen::Vector3f& orders) {
@@ -300,7 +300,7 @@ namespace module::motion {
         }
 
         // Reset the trajectories
-        TrajectoriesInit(trajs);
+        trajectoriesInit(trajs);
 
         // full period (float step) is needed for trunk splines
         float period = 2.0f * half_period;
@@ -538,7 +538,7 @@ namespace module::motion {
         foot_step.stepFromOrders(orders);
 
         // Reset the trajectories
-        TrajectoriesInit(trajs);
+        trajectoriesInit(trajs);
 
         // Time length of float and single support phase during the half cycle
         float doubleSupportLength = params.double_support_ratio * half_period;
@@ -687,9 +687,9 @@ namespace module::motion {
 
     QuinticWalkEngine::PositionSupportTuple QuinticWalkEngine::computeCartesianPositionAtTime(const float time) const {
         // Evaluate target cartesian state from trajectories
-        const auto [trunkPos, trunkAxis, footPos, footAxis] = TrajectoriesTrunkFootPos(time, trajs);
+        const auto [trunkPos, trunkAxis, footPos, footAxis] = trajectoriesTrunkFootPos(time, trajs);
         // Discard isDoubleSupport because we don't use it
-        const auto [_, isLeftSupportFoot] = TrajectoriesSupportFootState(time, trajs);
+        const auto [_, isLeftSupportFoot] = trajectoriesSupportFootState(time, trajs);
         return {trunkPos, trunkAxis, footPos, footAxis, isLeftSupportFoot};
     }
 
--- a/module/motion/QuinticWalk/src/WalkEngine.cpp
+++ b/module/motion/QuinticWalk/src/WalkEngine.cpp
@@ -679,29 +679,21 @@ namespace module::motion {
         trunk_axis_acc_at_last.setZero();
     }
 
-    void QuinticWalkEngine::computeCartesianPosition(Eigen::Vector3f& trunkPos,
-                                                     Eigen::Vector3f& trunkAxis,
-                                                     Eigen::Vector3f& footPos,
-                                                     Eigen::Vector3f& footAxis,
-                                                     bool& isLeftsupportFoot) {
+    using PositionSupportTuple = std::tuple<Eigen::Vector3f, Eigen::Vector3f, Eigen::Vector3f, Eigen::Vector3f, bool>;
+
+    PositionSupportTuple QuinticWalkEngine::computeCartesianPosition() const {
         // Compute trajectories time
-        float time = getTrajsTime();
-
-        computeCartesianPositionAtTime(trunkPos, trunkAxis, footPos, footAxis, isLeftsupportFoot, time);
+        const float time = getTrajsTime();
+        return computeCartesianPositionAtTime(time);
     }
 
-
-    void QuinticWalkEngine::computeCartesianPositionAtTime(Eigen::Vector3f& trunkPos,
-                                                           Eigen::Vector3f& trunkAxis,
-                                                           Eigen::Vector3f& footPos,
-                                                           Eigen::Vector3f& footAxis,
-                                                           bool& isLeftsupportFoot,
-                                                           float time) {
-        // We don't use the double support capability of the trajectory utils
-        // We need to initialise this (l-value) to pass in as a reference to the function though
-        bool isDoubleSupport = false;
+    PositionSupportTuple QuinticWalkEngine::computeCartesianPositionAtTime(const float time) const {
         // Evaluate target cartesian state from trajectories
-        TrajectoriesTrunkFootPos(time, trajs, trunkPos, trunkAxis, footPos, footAxis);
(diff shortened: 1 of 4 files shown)


TrentHouliston:
This isn't a very good way to pass this sort of information around in the first place.
Is there a way this can be done without requiring something to be passed by reference?
(via a return or otherwise?)

KipHamiltons:
`TrajectoriesSupportFootState` is designed to set its two boolean references, so we would have to split it into two separate tracks, but I don't see why we couldn't
(assuming we don't want to use `pair` or something, which seems like a terrible idea lol)

Bidski:
> `TrajectoriesSupportFootState` is designed to set its two boolean references, so we would have to split it into two separate tracks, but I don't see why we couldn't
> (assuming we don't want to use `pair` or something, which seems like a terrible idea lol)

I would honestly make it return a pair, I dont see what is horrible about that

TrentHouliston:
Especially with the awesome power of structured bindings in c++17!

KipHamiltons:
oh dang, I didn't realised structured binding was in C++ :heart_eyes: that's one of my favourite things in python

ysims (commented):
Make sure this is tested before merging.

KipHamiltons:
I removed `DefaultCheckState` as well, because it was never used

module/motion/QuinticWalk/src/WalkEngine.cpp
```diff
@@ -679,27 +679,21 @@ namespace module::motion {
...
     }
 
-    void QuinticWalkEngine::computeCartesianPosition(Eigen::Vector3f& trunkPos,
-                                                     Eigen::Vector3f& trunkAxis,
-                                                     Eigen::Vector3f& footPos,
-                                                     Eigen::Vector3f& footAxis,
-                                                     bool isLeftsupportFoot) {
-        // Compute trajectories time
-        float time = getTrajsTime();
+    using PositionSupportTuple = std::tuple<Eigen::Vector3f, Eigen::Vector3f, Eigen::Vector3f, Eigen::Vector3f, bool>;
```
CMurtagh-LGTM, line 682:
```suggestion
```

module/motion/QuinticWalk/src/WalkEngine.cpp
```diff
@@ -679,27 +679,21 @@ namespace module::motion {
...
 
-        computeCartesianPositionAtTime(trunkPos, trunkAxis, footPos, footAxis, isLeftsupportFoot, time);
+    PositionSupportTuple QuinticWalkEngine::computeCartesianPosition() const {
```
CMurtagh-LGTM, line 684:
```suggestion
    QuinticWalkEngine::PositionSupportTuple QuinticWalkEngine::computeCartesianPosition() const {
```

module/motion/QuinticWalk/src/WalkEngine.cpp
```diff
@@ -679,27 +679,21 @@ namespace module::motion {
...
     }
 
-
-    void QuinticWalkEngine::computeCartesianPositionAtTime(Eigen::Vector3f& trunkPos,
-                                                           Eigen::Vector3f& trunkAxis,
-                                                           Eigen::Vector3f& footPos,
-                                                           Eigen::Vector3f& footAxis,
-                                                           bool isLeftsupportFoot,
-                                                           float time) {
+    PositionSupportTuple QuinticWalkEngine::computeCartesianPositionAtTime(const float time) const {
```
CMurtagh-LGTM, line 690:
```suggestion
    QuinticWalkEngine::PositionSupportTuple QuinticWalkEngine::computeCartesianPositionAtTime(const float time) const {
```

CMurtagh-LGTM (changes requested):

module/motion/QuinticWalk/src/WalkEngine.cpp
TrentHouliston: Are these needed to make it compile?

CMurtagh-LGTM: If we remove the duplicate using statement it shouldn't compile.

CMurtagh-LGTM: This suggestion is in relation to the previous one.

ysims:
> Make sure this is tested before merging.

I've just checked now and it looks fine, doesn't look like there is any difference...

shared/utility/motion/splines/TrajectoryUtils.hpp
```diff
@@ -189,111 +194,68 @@ namespace utility::motion::splines {
...
-        }
-        if (trunkPos.z() - footPos.z() < 0.20f) {
-            cost += 1000.0f - 1000.0f * (trunkPos.z() - footPos.z() - 0.20f);
-        }
-        if (footAxis.norm() >= M_PI_2) {
-            cost += 1000.0f + 1000.0f * (footAxis.norm() - M_PI_2);
-        }
-
-        return cost;
+    [[nodiscard]] inline std::pair<bool, bool> TrajectoriesSupportFootState(float t, const Trajectories& traj) {
```
TrentHouliston, line 255: Why is this function PascalCase?

shared/utility/motion/splines/TrajectoryUtils.hpp
```diff
@@ -189,111 +194,68 @@ namespace utility::motion::splines {
      * Compute from given spline container trajectory Cartesian trunk and foot position/velocity/acceleration
      * and assign it to given vector
      */
-    inline void TrajectoriesTrunkFootPos(float t,
-                                         const Trajectories& traj,
-                                         Eigen::Vector3f& trunkPos,
-                                         Eigen::Vector3f& trunkAxis,
-                                         Eigen::Vector3f& footPos,
-                                         Eigen::Vector3f& footAxis) {
+    [[nodiscard]] inline Vector3fQuadruple TrajectoriesTrunkFootPos(const float t, const Trajectories& traj) {
```
CMurtagh-LGTM, line 197:
```suggestion
    [[nodiscard]] inline Vector3fQuadruple trajectoriesTrunkFootPos(const float t, const Trajectories& traj) {
```

shared/utility/motion/splines/TrajectoryUtils.hpp
```diff
@@ -189,111 +194,68 @@ namespace utility::motion::splines {
...
+        return {trunkPosVel, trunkAxisVel, footPosVel, footAxisVel};
     }
 
-    inline void TrajectoriesTrunkFootAcc(float t,
-                                         const Trajectories& traj,
-                                         Eigen::Vector3f& trunkPosAcc,
-                                         Eigen::Vector3f& trunkAxisAcc,
-                                         Eigen::Vector3f& footPosAcc,
-                                         Eigen::Vector3f& footAxisAcc) {
+    [[nodiscard]] inline Vector3fQuadruple TrajectoriesTrunkFootAcc(const float t, const Trajectories& traj) {
```
CMurtagh-LGTM, line 231:
```suggestion
    [[nodiscard]] inline Vector3fQuadruple trajectoriesTrunkFootAcc(const float t, const Trajectories& traj) {
```

shared/utility/motion/splines/TrajectoryUtils.hpp
```diff
@@ -189,111 +194,68 @@ namespace utility::motion::splines {
...
+        return {trunkPos, trunkAxis, footPos, footAxis};
     }
 
-    inline void TrajectoriesTrunkFootVel(float t,
-                                         const Trajectories& traj,
-                                         Eigen::Vector3f& trunkPosVel,
-                                         Eigen::Vector3f& trunkAxisVel,
-                                         Eigen::Vector3f& footPosVel,
-                                         Eigen::Vector3f& footAxisVel) {
+    [[nodiscard]] inline Vector3fQuadruple TrajectoriesTrunkFootVel(const float t, const Trajectories& traj) {
```
CMurtagh-LGTM, line 214:
```suggestion
    [[nodiscard]] inline Vector3fQuadruple trajectoriesTrunkFootVel(const float t, const Trajectories& traj) {
```

shared/utility/motion/splines/TrajectoryUtils.hpp
CMurtagh-LGTM:
```suggestion
    [[nodiscard]] inline std::pair<bool, bool> trajectoriesSupportFootState(float t, const Trajectories& traj) {
```

CMurtagh-LGTM (changes requested):

CMurtagh-LGTM (approved):
LGTM

PR #515, opened 2021-06-09
Fixes the race condition when reconnecting in the Webots module

ysims:
When we communicate with Webots, we create a buffer that holds the data we read. When we have enough data to create a sensor measurements protobuf message, we create that message and then erase those bytes from the buffer. Just before erasing the data, the connection would sometimes disconnect and clear the buffer. This resulted in trying to erase data from a buffer with nothing in it. 

This PR adds the `Sync` DSL word to the `read_io` and `Watchdog` reactors to prevent this race condition.

### DIFF
--- a/module/platform/Webots/src/Webots.cpp
+++ b/module/platform/Webots/src/Webots.cpp
@@ -333,7 +333,7 @@ namespace module::platform {
                         if ((event.events & IO::READ) != 0) {
                             // If we have not seen the welcome message yet, look for it
                             if (!connection_active) {
-                                // Initaliase the string with 0s
+                                // Initialise the string with 0s
                                 // make sure we have an extra character just in case we read something that isn't a null
                                 // terminator
                                 std::array<char, 9> initial_message{};
@@ -401,7 +401,7 @@ namespace module::platform {
                                 buffer.resize(old_size + available);
 
                                 // Read data into our buffer and resize it to the new data we read
-                                auto bytes_read = ::read(fd, buffer.data() + old_size, available);
+                                const auto bytes_read = ::read(fd, buffer.data() + old_size, available);
                                 // Shrink the buffer to the size that was actually read.
                                 buffer.resize(old_size + bytes_read);
 
--- a/module/platform/Webots/src/Webots.cpp
+++ b/module/platform/Webots/src/Webots.cpp
@@ -376,17 +376,6 @@ namespace module::platform {
                                 utility::clock::last_update = NUClear::base_clock::now();
 
                                 connection_active = true;
-
-                                // Clear socket && buffer
-                                unsigned long available = 0;
-                                if (::ioctl(fd, FIONREAD, &available) < 0) {
-                                    log<NUClear::ERROR>(
-                                        fmt::format("Error querying for available data, {}", strerror(errno)));
-                                    return;
-                                }
-                                const size_t old_size = buffer.size();
-                                ::read(fd, buffer.data() + old_size, available);
-                                buffer.clear();
                             }
                             else {
                                 // Work out how many bytes are available to read in the buffer and ensure we have enough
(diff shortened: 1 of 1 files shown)


module/platform/Webots/src/Webots.cpp
```diff
@@ -327,107 +327,122 @@ namespace module::platform {
...
+                                connection_active = true;
+
+                                // Clear socket && buffer
+                                unsigned long available = 0;
+                                if (::ioctl(fd, FIONREAD, &available) < 0) {
+                                    log<NUClear::ERROR>(
+                                        fmt::format("Error querying for available data, {}", strerror(errno)));
+                                    return;
+                                }
+                                const size_t old_size = buffer.size();
```
CMurtagh-LGTM, line 387: Should we resize the buffer?

CMurtagh-LGTM (changes requested):

module/platform/Webots/src/Webots.cpp
ysims: If we don't how will we fit all the new data in? And the size is important later on. Also not relevant to this PR.

ysims: Oh wait this is a different line. I don't see why, we're clearing it anyway?

ysims: Tbh I'm not sure this is really needed anyway.

module/platform/Webots/src/Webots.cpp
```diff
@@ -327,107 +327,122 @@ namespace module::platform {
...
+                                buffer.resize(old_size + available);
+
+                                // Read data into our buffer and resize it to the new data we read
+                                auto bytes_read = ::read(fd, buffer.data() + old_size, available);
```
KipHamiltons, line 404:
```suggestion
                                const auto bytes_read = ::read(fd, buffer.data() + old_size, available);
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -327,107 +327,122 @@ namespace module::platform {
...
-                                // Halt and don't retry as reconnection is pointless.
-                                close(fd);
-                                powerplant.shutdown();
+            read_io =
+                on<IO, Sync<Webots>>(fd, IO::READ | IO::CLOSE | IO::ERROR)
+                    .then("Read Stream", [this](const IO::Event& event) {
+                        if ((event.events & IO::READ) != 0) {
+                            // If we have not seen the welcome message yet, look for it
+                            if (!connection_active) {
+                                // Initaliase the string with 0s
```
KipHamiltons, line 336:
```suggestion
                                // Initialise the string with 0s
```

module/platform/Webots/src/Webots.cpp
TrentHouliston: What does this section do?

ysims: It's clearing the socket and buffer when reconnecting but I don't think it's needed - testing now.

ysims: I've got rid of it, it didn't seem to be needed.

CMurtagh-LGTM (approved):
LGTM

module/platform/Webots/src/Webots.cpp
```diff
@@ -229,7 +229,7 @@ namespace module::platform {
             server_address = config["server_address"].as<std::string>();
             server_port    = config["port"].as<std::string>();
 
-            on<Watchdog<Webots, 5, std::chrono::seconds>>().then([this, config] {
+            on<Watchdog<Webots, 30, std::chrono::seconds>, Sync<Webots>>().then([this, config] {
```
KipHamiltons, line 232: Was changing to 30s timeout meant to be part of this PR?

PR #517, opened 2021-06-14
Add clippy extension

Bidski:


LachlanCourt (approved):
This seems important. Go for your life

TrentHouliston (approved):
I wonder how long before someone just clicks install all recommended

PR #518, opened 2021-06-15
Here's some velocity things

tayloryoung6396:
Someone should test this, also maybe check that this gives you sim speed factor

const double ratio = static_cast<double>(sim_delta + prev_sim_delta) / static_cast<double>(real_delta + prev_real_delta);

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,7 @@ add_subdirectory(nuclear)
 # Setting to use the custom clock with an adaptable rate
 option(USE_CUSTOM_CLOCK "Replace NUClear's clock with a custom implementation." ON)
 if(USE_CUSTOM_CLOCK)
+  add_compile_definitions(NUCLEAR_CUSTOM_CLOCK)
   add_subdirectory(shared/clock)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,7 +110,6 @@ add_subdirectory(nuclear)
 # Setting to use the custom clock with an adaptable rate
 option(USE_CUSTOM_CLOCK "Replace NUClear's clock with a custom implementation." ON)
 if(USE_CUSTOM_CLOCK)
-  add_compile_definitions(NUCLEAR_CUSTOM_CLOCK)
   add_subdirectory(shared/clock)
 endif()
 
--- a/module/platform/Webots/src/Webots.cpp
+++ b/module/platform/Webots/src/Webots.cpp
@@ -40,9 +40,9 @@
 
 // Include headers needed for TCP connection
 extern "C" {
-#include <netdb.h> /* definition of gethostbyname */
+#include <netdb.h>      /* definition of gethostbyname */
 #include <netinet/in.h> /* definition of struct sockaddr_in */
-#include <sys/ioctl.h> /* definition of ioctl and FIONREAD */
+#include <sys/ioctl.h>  /* definition of ioctl and FIONREAD */
 #include <sys/socket.h>
 #include <sys/time.h>
 #include <unistd.h> /* definition of close */
--- a/module/platform/Webots/src/Webots.cpp
+++ b/module/platform/Webots/src/Webots.cpp
@@ -257,9 +257,7 @@ namespace module::platform {
                 // The fastest speed is determined by the config, which comes from the max servo velocity from
                 // NUgus.proto in Webots
                 double speed = duration.count() > 0
-                                   ? diff
-                                         / (std::chrono::duration<double, NUClear::clock::period>(duration).count()
-                                            / double(NUClear::clock::period::den))
+                                   ? diff / (double(duration.count()) / double(NUClear::clock::period::den))
                                    : max_velocity;
                 speed        = std::min(max_velocity, speed);
                 // Update our internal state if anything has changed for this servo
--- a/module/platform/Webots/src/Webots.cpp
+++ b/module/platform/Webots/src/Webots.cpp
@@ -39,9 +39,9 @@
 
 // Include headers needed for TCP connection
 extern "C" {
-#include <netdb.h>      /* definition of gethostbyname */
+#include <netdb.h> /* definition of gethostbyname */
 #include <netinet/in.h> /* definition of struct sockaddr_in */
-#include <sys/ioctl.h>  /* definition of ioctl and FIONREAD */
+#include <sys/ioctl.h> /* definition of ioctl and FIONREAD */
 #include <sys/socket.h>
 #include <sys/time.h>
 #include <unistd.h> /* definition of close */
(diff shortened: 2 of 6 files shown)


ysims:
I don't see any difference.

CMakeLists.txt
```diff
@@ -110,6 +110,7 @@ add_subdirectory(nuclear)
 # Setting to use the custom clock with an adaptable rate
 option(USE_CUSTOM_CLOCK "Replace NUClear's clock with a custom implementation." ON)
 if(USE_CUSTOM_CLOCK)
+  add_compile_definitions(NUCLEAR_CUSTOM_CLOCK)
```
Bidski, line 113: Try without this

module/platform/Webots/src/Webots.cpp
```diff
@@ -256,8 +256,17 @@ namespace module::platform {
...
-                    duration.count() > 0 ? diff / std::chrono::duration<double>(duration).count() : max_velocity;
+                double speed = duration.count() > 0
+                                   ? diff / (double(duration.count()) / double(NUClear::clock::period::den))
+                                   : max_velocity;
+
+                log<NUClear::DEBUG>(
+                    fmt::format("Servo velocities {} -> {}, {}",
+                                target.id,
+                                speed,
+                                diff / (double(duration.count()) / double(NUClear::clock::period::den))));
```
Bidski, line 267: Purge the logs!

Bidski (changes requested):

CMakeLists.txt
TrentHouliston: Now that you solved the clock problem you shouldn't need this. We want to use modern cmake patterns which this is not.

module/platform/Webots/src/Webots.cpp
```diff
@@ -256,8 +256,10 @@ namespace module::platform {
                 // Because of this, we move the servo as fast as we can to reach the position.
                 // The fastest speed is determined by the config, which comes from the max servo velocity from
                 // NUgus.proto in Webots
-                double speed =
-                    duration.count() > 0 ? diff / std::chrono::duration<double>(duration).count() : max_velocity;
+                double speed = duration.count() > 0
+                                   ? diff / (double(duration.count()) / double(NUClear::clock::period::den))
+                                   : max_velocity;
+
                 speed = std::min(max_velocity, speed);
```
TrentHouliston, line 263: Do we need this log line anymore? Is it useful? Otherwise delete it

module/platform/Webots/src/Webots.cpp
```diff
@@ -256,8 +256,10 @@ namespace module::platform {
                 // Because of this, we move the servo as fast as we can to reach the position.
                 // The fastest speed is determined by the config, which comes from the max servo velocity from
                 // NUgus.proto in Webots
-                double speed =
-                    duration.count() > 0 ? diff / std::chrono::duration<double>(duration).count() : max_velocity;
+                double speed = duration.count() > 0
```
TrentHouliston, line 259: This shouldn't have changed so maybe keep the old one?

shared/clock/clock.cpp
```diff
@@ -17,12 +18,16 @@ namespace NUClear {
...
-                                  (now - utility::clock::last_update) * utility::clock::custom_rtf));
-        return utility::clock::last_update;
+        auto now   = NUClear::base_clock::now();
+        auto delta = std::chrono::duration_cast<std::chrono::steady_clock::duration>((now - utility::clock::last_update)
+                                                                                     * utility::clock::custom_rtf);
+        utility::clock::state       = clock::time_point(utility::clock::state + delta);
+        utility::clock::last_update = now;
+        return utility::clock::state;
+
+        // return clock::time_point(utility::clock::start
```
TrentHouliston, line 28: Commented code

shared/clock/clock.hpp
```diff
@@ -5,7 +5,7 @@
 
 namespace utility::clock {
     extern double custom_rtf;  // real time factor
-    extern std::chrono::steady_clock::time_point last_update;
+    extern std::chrono::steady_clock::time_point start;
```
TrentHouliston, line 8: This probably doesn't need to be in the header. Just in the cpp file is enough

TrentHouliston (changes requested):

shared/clock/clock.cpp
```diff
@@ -17,12 +18,12 @@ namespace NUClear {
...
-            clock::time_point(utility::clock::last_update
-                              + std::chrono::duration_cast<std::chrono::steady_clock::duration>(
-                                  (now - utility::clock::last_update) * utility::clock::custom_rtf));
-        return utility::clock::last_update;
+        auto now   = NUClear::base_clock::now();
+        auto delta = std::chrono::duration_cast<std::chrono::steady_clock::duration>((now - utility::clock::last_update)
+                                                                                     * utility::clock::custom_rtf);
+        utility::clock::state       = clock::time_point(utility::clock::state + delta);
+        utility::clock::last_update = now;
+        return utility::clock::state;
```
wongjoel, line 26: After doing a manual trace, I can confirm that the old formula was broken and the new formula works.

module/platform/Webots/src/Webots.cpp
```diff
@@ -256,8 +259,17 @@ namespace module::platform {
                 // Because of this, we move the servo as fast as we can to reach the position.
                 // The fastest speed is determined by the config, which comes from the max servo velocity from
                 // NUgus.proto in Webots
-                double speed =
-                    duration.count() > 0 ? diff / std::chrono::duration<double>(duration).count() : max_velocity;
+                double max_velocity = 0.0;
+                if (target.id >= ServoID::R_HIP_YAW && target.id <= ServoID::L_ANKLE_ROLL) {
+                    max_velocity == max_velocity_mx106;
```
ysims, line 264:
```suggestion
                    max_velocity = max_velocity_mx106;
```

module/platform/Webots/src/Webots.cpp
```diff
@@ -256,8 +259,17 @@ namespace module::platform {
...
+                }
+                else {
+                    max_velocity == max_velocity_mx64;
```
ysims, line 267:
```suggestion
                    max_velocity = max_velocity_mx64;
```

ysims (dismissed):

ysims:
I have tested this in the RoboCup controller and it looks fine. Walk looks different too. Will need retuning after this is merged.

module/platform/Webots/src/Webots.cpp
Bidski: @TrentHouliston which cast do you like better? Check previous comment

Bidski (changes requested):
Something seems strange with the includes in Webots.cpp

Bidski (approved):
LGTM

PR #520, opened 2021-06-16
Improve upon our scripts after time fixes

ysims:
Both get ups and both kicks.


wongjoel:
Tested
- Stand up front works, the robot ends leaning forwards a bit, but seems stable
- Stand up back works, the robot ends leaning forwards a bit, but seems stable. Also seems a bit slow, but otherwise fine
- Left kick doesn't move a ball placed in front of the left leg, and the robot topples over at the end of the script
- Right kick moves a ball placed in front of the right leg

JosephusPaye:
From my tests:

- Left kick moves the ball if it's close enough to the left foot. Robot falls backwards after.
- Right kick is nice and stable, moves the ball fine. Robot remains standing after.
- StandUpBack works, and stands with a slight forward lean.
- StandUpFront works, with similar lean.

I think these are good enough to merge. Perhaps left kick can be further tuned in a separate PR.

JosephusPaye (approved):
Tested and they work.

ysims:
Nice. I think further improvements can be done by someone else.

PR #523, opened 2021-06-16
Upgrade configuration `.as<>` so it can use yaml-cpp overloads

TrentHouliston:
Yaml-cpp has some overloads that let you set a default for .as calls.
This forwards arguments using perfect forwarding to yaml-cpp so it can use them

### DIFF
--- a/shared/extension/Configuration.hpp
+++ b/shared/extension/Configuration.hpp
@@ -176,7 +176,7 @@ namespace extension {
         }
 
         template <typename T, typename... Args>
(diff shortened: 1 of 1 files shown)


Bidski (commented):
Seems too simple. Make it more complicated please

Bidski:
Still not complicated enough. You can do better @TrentHouliston

JosephusPaye (approved):
Builds for me, and works as expected.

PR #524, opened 2021-06-16
Add support for filtering GameController packets to a given IP address

JosephusPaye:
Set `udp_filter_address` in `GameController` to the Game Controller IP address and the GameController module will ignore any packets not from that address.


module/input/GameController/src/GameController.cpp
```diff
@@ -86,6 +88,16 @@ namespace module::input {
...
                                   .then([this](const UDP::Packet& p, const GameState& gameState) {
+                                      std::string remoteAddr = ipAddressIntToString(p.remote.address);
+
+                                      if (!udp_filter_address.empty() && remoteAddr != udp_filter_address) {
+                                          log<NUClear::INFO>("Ignoring UDP packet from",
+                                                             remoteAddr,
+                                                             "as it doesn't match configured filter address",
+                                                             udp_filter_address);
+                                          return;
+                                      }
```
tayloryoung6396, line 99: It is worth doing something like this, don't really want to know it's filtering every packet.. kinda annoying spam
```
static bool hasPrinted = false;
if (!udp_filter_address.empty() && remoteAddr != udp_filter_address) {

    if (!hasPrinted) {
        log<NUClear::INFO>("Ignoring UDP packet from",
                            remoteAddr,
                            "as it doesn't match configured filter address",
                            udp_filter_address);
        hasPrinted = true;
    }
    return;
}```

**Review by tayloryoung6396 — CHANGES_REQUESTED** [review]
LGTM, I've tested it and it works, but i don't like the spam

File: module/input/GameController/src/GameController.cpp (hunk 1 above)
Comment (JosephusPaye) [reply module/input/GameController/src/GameController.cpp:99]: It seems like we'll need more than a single boolean to keep track of different IPs.

Comment (tayloryoung6396) [reply module/input/GameController/src/GameController.cpp:99]: Sure.. just annoying having it spam. Maybe make a list of previously filtered ips and check against that

**Review by tayloryoung6396 — COMMENTED** [review]
@JosephusPaye This should do what you want now, it works for me

**Review by tayloryoung6396 — COMMENTED** [review]
This shouldn't work? isn't the return required...?

File: module/input/GameController/src/GameController.cpp [hunk 2]
```diff
@@ -86,6 +88,22 @@ namespace module::input {
...
+                                          if (std::find(ignored_ip_addresses.begin(),
+                                                        ignored_ip_addresses.end(),
+                                                        remoteAddr)
+                                              == ignored_ip_addresses.end()) {
+                                              ignored_ip_addresses.insert(remoteAddr);
+                                              log<NUClear::INFO>("Ignoring UDP packet from",
+                                                                 remoteAddr,
+                                                                 "as it doesn't match configured filter address",
+                                                                 udp_filter_address);
+                                          }
```
JosephusPaye, line 104:
```suggestion
                                          }

                                          return;
```

tayloryoung6396 (approved):
LGTM

PR #525, opened 2021-06-17
Remove unused param warning

tayloryoung6396:


shared/utility/motion/InverseKinematics.cpp
```diff
@@ -337,7 +337,7 @@ namespace utility::motion::kinematics {
         joints.insert(joints.end(), joints2.begin(), joints2.end());
         return joints;
     }
-    std::vector<std::pair<ServoID, double>> calculateCameraLookJoints(const KinematicsModel& model,
+    std::vector<std::pair<ServoID, double>> calculateCameraLookJoints(const KinematicsModel&,  // model,
```
KipHamiltons, line 340:
```suggestion
    std::vector<std::pair<ServoID, double>> calculateCameraLookJoints(const KinematicsModel&  /*model*/,
```

KipHamiltons (commented):
Last time I looked at removing this warning, IIRC @Bidski said that this function really should be using this parameter, and thus the warning was a valid reminder that it should be dealt with. :shrug: 
EDIT: I did not remember correctly :sweat_smile: see his comment

Bidski:
> Last time I looked at removing this warning, IIRC @Bidski said that this function really should be using this parameter, and thus the warning was a valid reminder that it should be dealt with. :shrug: 

What i said was that it probably was used at one point, but obviously its not needed any more.

Rather than commenting it, you should just remove it

ysims:
> > Last time I looked at removing this warning, IIRC @Bidski said that this function really should be using this parameter, and thus the warning was a valid reminder that it should be dealt with. 🤷
> 
> What i said was that it probably was used at one point, but obviously its not needed any more.
> 
> Rather than commenting it, you should just remove it

Sorry team, I created this warning last year when finishing off the camera calibration PR Trent did. IK used to do the whole thing to the cameras and took the last transform from the kinematics model, but now vision does the last part from the neck pitch servo to the cameras and takes the transform from the Hcp matrix in the camera config. This might have affected head behaviour though?

KipHamiltons:
why did you close this @tayloryoung6396 ? why not just remove the parameter

tayloryoung6396:
Because 

> > > Last time I looked at removing this warning, IIRC @Bidski said that this function really should be using this parameter, and thus the warning was a valid reminder that it should be dealt with. shrug
> > 
> > 
> > What i said was that it probably was used at one point, but obviously its not needed any more.
> > Rather than commenting it, you should just remove it
> 
> Sorry team, I created this warning last year when finishing off the camera calibration PR Trent did. Kinematics used to do the whole thing to the cameras and took the last transform from the kinematics model, but now vision does the last part from the neck pitch servo to the cameras and takes the transform from the Hcp matrix in the camera config. This might have affected head behaviour though?

Bidski:
Still needs fixing?

tayloryoung6396 (commented):
LGTM

KipHamiltons (approved):

PR #527, opened 2021-06-17
Fixes left kick

weepeefa:


wongjoel:
I just tested this in `kid.wbt`: for me, the robot kicks the ground with its left foot, then falls over due to kicking the ground. Because the robot kicks the ground where the foot was, it doesn't end up moving a ball placed right in front of the foot. I assume it was different when you were tuning it?

Bidski (dismissed):
I assume this actually works?

KipHamiltons:
Worked on my machine. Definite improvement

KipHamiltons (approved):
LGTM. Tested on another computer. Very stable, decent power. Excellent work

PR #528, opened 2021-06-17
Fix eigen quaternion coeffs problems

KipHamiltons:
Eigen::Quaternion `.coeffs()` returns the coefficients in a different order to the order that the constructor expects.
This PR looks to find and fix all the places where I screwed that up.


module/motion/QuinticWalk/src/QuinticWalk.cpp
```diff
@@ -219,22 +219,20 @@ namespace module::motion {
...
-            const float halfYaw   = yaw * 0.5f;
-            const float halfPitch = pitch * 0.5f;
-            const float halfRoll  = roll * 0.5f;
-            const float cosYaw    = std::cos(halfYaw);
-            const float sinYaw    = std::sin(halfYaw);
-            const float cosPitch  = std::cos(halfPitch);
-            const float sinPitch  = std::sin(halfPitch);
-            const float cosRoll   = std::cos(halfRoll);
-            const float sinRoll   = std::sin(halfRoll);
-            return Eigen::Quaternionf(cosRoll * cosPitch * cosYaw + sinRoll * sinPitch * sinYaw,  // formerly yzx
```
TrentHouliston, line 231: I'm not actually sure I like this change, it makes things more complicated...
Maybe we just need to document what's going on whenever we use the various constructors.
Or at the very least just wrap it so we still return directly but have the vector4f in the quaternion constructor

module/input/SensorFilter/src/MotionModel.hpp
```diff
@@ -158,22 +158,15 @@ namespace module::input {
...
-            const Eigen::Quaternion<Scalar> dq_dt = Eigen::Quaternion<Scalar>(0.0,
-                                                                              newState.omegaTTt.x() * 0.5,
-                                                                              newState.omegaTTt.y() * 0.5,
-                                                                              newState.omegaTTt.z() * 0.5)
-                                                    * Rwt;
+            const Eigen::Matrix<Scalar, 4, 1> half_omegaTTt_coeffs(newState.omegaTTt.x() * 0.5,
+                                                                   newState.omegaTTt.y() * 0.5,
+                                                                   newState.omegaTTt.z() * 0.5,
+                                                                   0.0);
+            const Eigen::Quaternion<Scalar> dq_dt = Eigen::Quaternion<Scalar>(half_omegaTTt_coeffs) * Rwt;
```
TrentHouliston, line 165: I would just wrap these straight into the constructor, e.g. `Quat(Vec(a,b,c,d));`

TrentHouliston (changes requested):

KipHamiltons:
I agree. Both of those are definitely nicer wrapped in a temporary vector. I also noticed another simplification I could make, which I added in a follow up commit.

module/motion/QuinticWalk/src/QuinticWalk.cpp
Bidski: Goddamn this is horrible :vomiting_face:

Bidski (dismissed):
I dont see anything obviously wrong

KipHamiltons:
@TrentHouliston can I grab a review pls :pleading_face:

module/input/SensorFilter/src/MotionModel.hpp
```diff
@@ -206,7 +207,7 @@ namespace module::input {
 
         Eigen::Matrix<Scalar, 4, 1> predict(const Eigen::Matrix<Scalar, size, 1>& state,
                                             const MeasurementType::FLAT_FOOT_ORIENTATION&) {
-            return StateVec(state).Rwt.coeffs();
+            return Eigen::Matrix<Scalar, 4, 1>(state.template segment<4>(StateVec::QX));
```
TrentHouliston, line 210: Can't you just do this?
```suggestion
            return state.template segment<4>(StateVec::QX);
```

TrentHouliston (changes requested):

module/input/SensorFilter/src/MotionModel.hpp
KipHamiltons: Yep. Good eye.

Bidski (approved):
I guess?

Bidski (dismissed):
LGTM

module/input/SensorFilter/src/MotionModel.hpp
```diff
@@ -158,15 +158,16 @@ namespace module::input {
...
+                Eigen::Quaternion<Scalar>(Eigen::Matrix<Scalar, 4, 1>(newState.omegaTTt.x() * 0.5,
+                                                                      newState.omegaTTt.y() * 0.5,
+                                                                      newState.omegaTTt.z() * 0.5,
+                                                                      0.0))
+                * Rwt;
             // The change in the rotation is the derivative times the differential (which is the time-step)
-            const Eigen::Quaternion<Scalar> change = Eigen::Quaternion<Scalar>(deltaT * dq_dt.coeffs());
-            // We can add the change to the original, as long as our time step is small enough
-            newState.Rwt = Eigen::Quaternion<Scalar>(Rwt.coeffs() + change.coeffs()).normalized();
+            const auto change = deltaT * dq_dt.coeffs();
```
TrentHouliston, line 168: Do not use auto when you are working with Eigen math operations.
https://eigen.tuxfamily.org/dox/TopicPitfalls.html
You fell for one of the classic blunders

TrentHouliston (changes requested):
Fix the bad thing, rest seems ok

module/input/SensorFilter/src/MotionModel.hpp
tayloryoung6396:
``` suggestion 
const Eigen::Matrix<double, 4, 1> change(deltaT * dq_dt.coeffs());
```

KipHamiltons: Imma have to repost the meme :joy:

module/input/SensorFilter/src/MotionModel.hpp
```diff
@@ -158,15 +158,16 @@ namespace module::input {
...
+                Eigen::Quaternion<Scalar>(Eigen::Matrix<Scalar, 4, 1>(newState.omegaTTt.x() * 0.5,
+                                                                      newState.omegaTTt.y() * 0.5,
+                                                                      newState.omegaTTt.z() * 0.5,
+                                                                      0.0))
+                * Rwt;
             // The change in the rotation is the derivative times the differential (which is the time-step)
-            const Eigen::Quaternion<Scalar> change = Eigen::Quaternion<Scalar>(deltaT * dq_dt.coeffs());
-            // We can add the change to the original, as long as our time step is small enough
-            newState.Rwt = Eigen::Quaternion<Scalar>(Rwt.coeffs() + change.coeffs()).normalized();
+const Eigen::Matrix<double, 4, 1> change(deltaT * dq_dt.coeffs());
```
KipHamiltons, line 168:
```suggestion
const Eigen::Matrix<Scalar, 4, 1> change(deltaT * dq_dt.coeffs());
```

module/input/SensorFilter/src/MotionModel.hpp
KipHamiltons: Well that page is quite interesting. Especially the part about passing by value. that makes things a bit spicy

TrentHouliston (approved):
Sure

PR #530, opened 2021-06-18
Upgrades the codebase to C++ 20

KipHamiltons:
Not exactly sure if this requires any special testing. I ran keyboardwalkfake and it didn't crash :shrug: 

I also noticed that the standard wasn't explicitly set for modules, so this change explicitly sets them to C++20 as well. Tests too

### DIFF
--- a/nuclear/message/Neutron.cmake
+++ b/nuclear/message/Neutron.cmake
@@ -154,7 +154,7 @@ target_include_directories(nuclear_message_protobuf SYSTEM INTERFACE ${pb_out})
 target_link_libraries(nuclear_message_protobuf protobuf::libprotobuf)
 
 add_library(nuclear_message ${NUCLEAR_LINK_TYPE} ${neutron_src})
-target_compile_features(nuclear_message PUBLIC cxx_std_17)
+target_compile_features(nuclear_message PUBLIC cxx_std_20)
 target_link_libraries(nuclear_message PUBLIC nuclear_message_protobuf)
 target_link_libraries(nuclear_message PUBLIC Eigen3::Eigen)
 target_include_directories(nuclear_message PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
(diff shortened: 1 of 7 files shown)


CMurtagh-LGTM:
We should try and update llvm before we go to c++20.

nuclear/message/Neutron.cmake
```diff
@@ -154,7 +154,7 @@ target_include_directories(nuclear_message_protobuf SYSTEM INTERFACE ${pb_out})
 target_link_libraries(nuclear_message_protobuf protobuf::libprotobuf)
 
 add_library(nuclear_message ${NUCLEAR_LINK_TYPE} ${neutron_src})
-target_compile_features(nuclear_message PUBLIC cxx_std_17)
+target_compile_features(nuclear_message PUBLIC cxx_std_20)
```
TrentHouliston, line 157: Don't add compile features to things that don't need it. Neutron doesn't use c++20 so you shouldn't add it here. Add it once you want to use a c++20 feature

nuclear/module/NUClearModule.cmake
```diff
@@ -191,6 +191,9 @@ function(NUCLEAR_MODULE)
   # Put it in an IDE group for the module's directory
   set_target_properties(${module_target_name} PROPERTIES FOLDER ${module_path})
 
+  # Set the C++ standard
+  target_compile_features(${module_target_name} PUBLIC cxx_std_20)
```
TrentHouliston, line 195: I wouldn't add this here either, not until you're using c++20 almost everywhere. Until then you can add it on a per-module basis, or rely on it coming in through utilities that use it

TrentHouliston (changes requested):
Not sure this is a good idea just yet, normally I'd wait a bit before using the new c++ standard as it can often cause more bugs than it's worth and for the most part we don't need anything it provides yet.

nuclear/message/Neutron.cmake
KipHamiltons: > Don't add compile features to things that don't need it

True. I didn't think about it like that.

nuclear/module/NUClearModule.cmake
KipHamiltons: So I was trying to compile the codebase with clang so we can get clang sanitizers, but clang complains because apparently I introduced some C++20isms in #516 with `constexpr` usage. GCC doesn't seem to care, but clang gets quite upset. 
I'll set it on a per-module basis for those then :+1:

KipHamiltons:
Wrong approach, and premature

PR #532, opened 2021-06-19
Moved iterable type trait

CMurtagh-LGTM:


KipHamiltons:
If this is fully replaced by range concepts in c++20, why don't we add C++20 for that module that uses this type trait?

Bidski (commented):
Is this needed for something?

KipHamiltons:
@Bidski see [these comments](https://github.com/NUbots/NUbots/pull/467#discussion_r653194435)
Just a quick thing Cameron noted while reviewing #467

CMurtagh-LGTM:
> Is this needed for something?

#467 removes the anon namespace, and I don't like that this type trait just sitting in the header file.

Bidski (approved):
Sure, I guess

PR #533, opened 2021-06-19
Add a module for testing the sensor filter easily

TrentHouliston:
This module will read an nbs file and run the filter as fast as possible and then use the Sensors output object.

This will make it easier to test the sensor filter in situ quickly by repeatedly running and comparing the results.


module/tools/FilterTester/src/FilterTester.cpp
```diff
@@ -0,0 +1,63 @@
+#include "FilterTester.hpp"
+
+#include <filesystem>
+
+#include "extension/Configuration.hpp"
+
+#include "message/input/Sensors.hpp"
+#include "message/platform/RawSensors.hpp"
+
+#include "utility/nbs/Decoder.hpp"
```
CMurtagh-LGTM, line 10: NUbots does not have this.

TrentHouliston: I suppose I should add this to NUbots

module/tools/FilterTester/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+
+## Usage
+
+
+## Emits
+
+
+## Dependencies
+
```
CMurtagh-LGTM, line 14:
```suggestion
FilterTester
============

## Description
Plays back sensor measurements from an nbs file one at a time.

## Usage
Command line arguments are interpreted as nbs files to play the raw sensors from.
`./b run <role_name> <file_name>...`

## Emits
`RawSensors`

## Dependencies

```

module/tools/FilterTester/src/FilterTester.cpp
```diff
@@ -0,0 +1,63 @@
...
+        : Reactor(std::move(environment)), config{} {
+
+        on<Configuration>("FilterTester.yaml").then([this](const Configuration& cfg) {
+            // Use configuration here from file FilterTester.yaml
+            this->log_level = cfg["log_level"].as<NUClear::LogLevel>();
+        });
+
+        // This is the information that will come back from the sensor filter
+        on<Trigger<Sensors>, Sync<FilterTester>>().then([this](const Sensors& sensors) {
+            // TODO do something with the sensors coming out of the module
```
CMurtagh-LGTM, line 30: Do you have any specific plans on what to do with these sensors? Rather than just having "something".

CMurtagh-LGTM (changes requested):

module/tools/FilterTester/src/FilterTester.cpp
TrentHouliston: I was going to leave that to you guys!

Bidski (changes requested):

Tom0Brien:
It would be nice to bring this up to date. I would like something along these lines that works with an optimization routine to tune filter parameters. Essentially, you'd record an NBS file with raw data and ground truth, and use something like this module with an optimization running along side it.

PR #535, opened 2021-06-19
Add tools for working with nbs files in c++

TrentHouliston:
Adds tools for working with nbs files in c++.
Contains a decoder that can read through an nbs file and an encoder for writing them


JosephusPaye:
While we're at it, it would be nice to get the fix for the broken NBS index generation. Currently index files created during DataLogging recording are invalid, and I think this was fixed on the 4tel end.

TrentHouliston:
Will do in another PR once this one gets merged

Bidski (changes requested):
DataLogger will need to be updated to use this Encoder? Same with DataPlayback?

Bidski (approved):
Fine

PR #536, opened 2021-06-19
Update datalogging to use the nbs tools

TrentHouliston:
This also fixes the generation of incorrect indexes

Merge #535 first


module/support/logging/DataLogging/src/DataLogging.cpp
```diff
@@ -46,9 +45,12 @@ namespace module::support::logging {
...
             // If the file isn't open, or writing this message will exceed our max size, make a new file
-            if (!output_file.is_open()
-                || (bytes_written + size + 3 + sizeof(timestamp_us) + sizeof(data.hash)) >= config.output.split_size) {
-                output_file.close();
+            if (!encoder || !encoder->is_open()
+                || (encoder->get_bytes_written() + size + 3 + sizeof(timestamp_us) + sizeof(data.hash))
+                       >= config.output.split_size) {
+                if (encoder) {
+                    encoder->close();
+                }
```
Bidski, line 53: Seems redundant
```suggestion
            if (!encoder || !encoder->is_open()
                || (encoder->get_bytes_written() + size + 3 + sizeof(timestamp_us) + sizeof(data.hash))
                       >= config.output.split_size) {
                    encoder->close();
```

module/support/logging/DataLogging/src/DataLogging.cpp
```diff
@@ -72,61 +74,22 @@ namespace module::support::logging {
...
-            index_file->write(reinterpret_cast<const char*>(&data.hash), sizeof(data.hash));
-            index_file->write(reinterpret_cast<const char*>(&data.id), sizeof(data.id));
-            index_file->write(reinterpret_cast<const char*>(&data.message_timestamp), sizeof(data.message_timestamp));
-            index_file->write(reinterpret_cast<const char*>(&bytes_written), sizeof(bytes_written));
-            index_file->write(reinterpret_cast<const char*>(&full_size), sizeof(full_size));
-            index_file->flush();
-
-            // Update the number of bytes we have written to the nbs file
-            bytes_written += size + 3 + sizeof(timestamp_us) + sizeof(data.hash);
+            encoder->write(data.timestamp, data.message_timestamp, data.hash, data.id, data.data);
```
Bidski, line 88:
```suggestion
            if (encoder) {
                encoder->write(data.timestamp, data.message_timestamp, data.hash, data.id, data.data);
            }
```

Bidski (changes requested):
My previous comment should have been a request changes

module/support/logging/DataLogging/src/DataLogging.cpp
CMurtagh-LGTM: As `encoder` is a `unique_ptr` `!encoder` will be true if `encoder` points to null, this means that `encoder->close()` will deference a null pointer in this case.

CMurtagh-LGTM: `encoder` is a pointer because it has a `ofstream` from zstr. I've opened a [pr](https://github.com/mateidavid/zstr/issues/28) a while ago.

module/support/logging/DataLogging/src/DataLogging.cpp
CMurtagh-LGTM: Here `encoder` should be non-null, as in the scope above we construct it.

Bidski (approved):

PR #544, opened 2021-06-20
Upgrade the NetworkForwarder so it can forward to multiple targets

TrentHouliston:


module/output/NetworkForwarder/src/NetworkForwarder.cpp
```diff
@@ -23,46 +23,53 @@ namespace module::output {
             else if (lvl == "FATAL") { this->log_level = NUClear::FATAL; }
             // clang-format on
 
-            config.target = cfg["target"].as<std::string>();
-
             // Update which types we will be forwarding
-            for (const auto& setting : cfg["messages"].config) {
-                // Get the name of the type
-                auto name = setting.first.as<std::string>();
+            for (const auto& target_config : cfg["targets"].config) {
```
Bidski, line 27: shouldnt need `.config` any more
https://github.com/NUbots/NUbots/blob/master/shared/extension/Configuration.hpp#L183-L195
```suggestion
            for (const auto& target_config : cfg["targets"]) {
```

Bidski: Although, I think we can do that better

Bidski (changes requested):

Bidski (approved):
LGTM

PR #553, opened 2021-06-21
Adds sync to reactions which caused use after free

KipHamiltons:
This adds `Sync<ServoTargets>` to two reactions which had a race condition. We found it from an ASAN crash report. 
@tayloryoung6396 has reported that this hasn't crashed after running for a while. We're presuming this works based on that :yum: 

All of the other lines edited are formatting


module/platform/Webots/src/Webots.cpp
```diff
@@ -312,8 +312,8 @@ namespace module::platform {
         });
 
         // This trigger updates our current servo state
-        on<Trigger<ServoTargets>, With<RawSensors>>().then([this](const ServoTargets& targets,
-                                                                  const RawSensors& sensors) {
+        on<Trigger<ServoTargets>, With<RawSensors>, Sync<ServoTargets>>().then([this](const ServoTargets& targets,
```
CMurtagh-LGTM, line 315: We shouldn't sync on `ServoTargets`, we should sync on a type local to the webots module.

CMurtagh-LGTM (changes requested):

module/platform/Webots/src/Webots.cpp
tayloryoung6396: This was an @Bidski suggestion implemented.. not sure there's anything else you would want to sync on..?

Bidski: This may be a point. We shouldn't sync on `Webots` as this would sync us with non-related reactions. Maybe `ServoState`? If a good type can't be found then a mutex would be a simpler option

Bidski (changes requested):

module/platform/Webots/src/Webots.cpp
CMurtagh-LGTM: We can always make an empty struct to sync on.

KipHamiltons: I'll add a mutex to the servostate instance the webots reactor owns then.

KipHamiltons: instead of a `Sync`

CMurtagh-LGTM (approved):
LGTM

KipHamiltons:
Syncing on the problem struct made the most sense to @CMurtagh-LGTM and I, so that's what we went with.
Also, removed the `Single` specifier, because that reaction is now on a `Sync` group

CMurtagh-LGTM:
Should have we removed the `Single`? `Single` causes new reactions to be ditched if we are already running, whilst `Sync` puts them on a priority queue. Having both would keep the original function.

tayloryoung6396 (commented):
@TrentHouliston To single or not to single?

Bidski (approved):
LGTM

PR #555, opened 2021-06-21
Make robocup container uncrashable

Y0sh1dk:
Restart binary if it crashes instead of stopping the container.


KipHamiltons (approved):
Assuming this works, LGTM. Excellent idea.

PR #559, opened 2021-06-21
Refactor behaviour into an explicit state machine

ysims:
Needs testing. Has some issues still.


tayloryoung6396 (commented):
This hurts.

ysims:
New and better stuff is in a new branch.

PR #562, opened 2021-06-22
Fix config iterators

Bidski:


KipHamiltons (approved):
holy shit, I really should not have approved this hahahahahaha

shared/extension/Configuration.hpp
```diff
@@ -180,15 +180,15 @@ namespace extension {
             return config.as<T>(std::forward<Args>(args)...);
         }
 
-        // Allow interating through configuration
+        // Allow iterating through configuration
         YAML::iterator begin() {
```
KipHamiltons, line 184: These should all be `[[nodiscard]]`

KipHamiltons (changes requested):

KipHamiltons (approved):
`Configuration.hpp` should be gone through to add more nodiscards at some point.

PR #565, opened 2021-06-22
Make logs a bit nicer

CMurtagh-LGTM:
Put all the logs in a single folder that we .gitignore and make the FileLogHandler put a noticeable split between runs.


.gitignore
```diff
@@ -77,6 +77,7 @@ tests_output/
 # Logs
 /home/nubots/NUbots/log/*
```
KipHamiltons, line 78: I don't think this one does anything

KipHamiltons (approved):

PR #567, opened 2021-06-22
Soccer strategy clean-up

KipHamiltons:
This code is a mess. This does some small things, and really is the bare minimum.

Big thing - a function created to get distance returned a bool, and that has been fixed.
I also removed a proto file that only this module used, as an enum.
The rest of it is classic _high impact software engineering_ that I love so much :heart:


module/behaviour/strategy/SoccerStrategy/src/SoccerStrategy.cpp
```diff
@@ -398,51 +370,41 @@ namespace module::behaviour::strategy {
...
-                    soccerObjectPriority->search_type = SearchType::GOAL_SEARCH;
-
-                    break;
-                }
-                case FieldTarget::Target::BALL: {
-                    soccerObjectPriority->ball        = 1;
-                    soccerObjectPriority->search_type = SearchType::LOST;
-                    break;
-                }
-                default: throw std::runtime_error("Soccer strategy attempted to find a bad object");
```
abihall, line 433: Should we still have this default?

abihall (commented):
Looks good apart from my one question

KipHamiltons:
This PR addresses the symptoms and not the cause, which is Soccer Strategy needing a complete rewrite. Closing

PR #569, opened 2021-06-23
Fix particle filter hypothesis

moosahassanx:
Test branch that test all of the other branches combined https://github.com/NUbots/NUbots/tree/young/test-branches

### DIFF
--- a/shared/utility/math/filter/ParticleFilter.hpp
+++ b/shared/utility/math/filter/ParticleFilter.hpp
@@ -115,7 +115,7 @@ namespace utility::math::filter {
             // If this is not an even division, the last N particles will just be randomly initialised
             const int particles_per_state = n_particles / mean_size;
 
-            for (size_t state = 0; state < mean.size(); ++state) {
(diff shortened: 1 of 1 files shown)


shared/utility/math/filter/ParticleFilter.hpp
```diff
@@ -112,9 +113,9 @@ namespace utility::math::filter {
 
             // We want to evenly distribute the particles across all hypotheses
             // If this is not an even division, the last N particles will just be randomly initialised
-            const int particles_per_state = n_particles / mean.size();
+            const int particles_per_state = n_particles / mean_size;
 
-            for (int state = 0; state < int(mean.size()); ++state) {
+            for (size_t state = 0; state < mean.size(); ++state) {
```
TrentHouliston, line 118: It should be int not size_t.

Use unsigned types when you want unsigned math. Not just because you feel a value should only ever be positive.

unless you want 0-1 to be a massive positive number, you probably don't want unsigned math

TrentHouliston (changes requested):

PR #578, opened 2021-06-23
Fix some .gitignores

Y0sh1dk:
Some of our ignores should only be relative to the project root, and only match against directories.

the `bin/` ignore causes issues adding files to `docker/usr/bin`


CMurtagh-LGTM (approved):
LGTM

JosephusPaye (approved):
Lᴏᴏᴋs Gᴏᴏᴅ Tᴏ Mᴇ

PR #579, opened 2021-06-23
Installs ncurses using pacman, instead of building from source

KipHamiltons:
I originally added a build from source for ncurses because it's listed in the arch repositories as a dep for IGC.
Its not really a package impacting performance though, and it's being slightly more annoying than trivial during our current cross compilation efforts.
I shouldn't have added it in the first place (and you questioned me doing it at the time @TrentHouliston), and I am keen to remove it now.


CMurtagh-LGTM (approved):
But what about the MICRO-OPTIMISATIONS.

PR #582, opened 2021-06-23
Removes repeated measurement from goals measure

KipHamiltons:


CMurtagh-LGTM (approved):
lgtm

TrentHouliston (changes requested):
Why are we removing repeated measurements from goalposts? just because we're not using them right now doesn't mean they're important!

You are able to measure the distance to the goal posts by both projecting to the centre like you are now, but also by measuring the goal post width. Therefore two measurements.

Someone just needs to reimplement the goal post width using the SVD stuff at some point

Bidski (changes requested):
I agree with @TrentHouliston this PR serves no purpose. It removes code that causes no harm, but which we will have to add again later on

KipHamiltons:
I was actioning a TODO on one of the localisation PRs, but if you guys think it's a bad idea, I am quite fine with closing.

PR #583, opened 2021-06-24
AWS target for robocup

Y0sh1dk:
`./b webots build webotsrobocup --target g4dnxlarge`

### DIFF
--- a/docker/usr/local/package/openblas/g4dnxlarge.sh
+++ b/docker/usr/local/package/openblas/g4dnxlarge.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+install-make-from-source $1 \
+    CROSS=1 \
+    BINARY=64 \
+    SMP=1 \
+    NUM_THREADS=8 \
+    TARGET=HASWELL \
+    USE_THREAD=1 \
+    NO_SHARED=0 \
+    NO_STATIC=0
--- a/docker/usr/local/toolchain/generate_g4dnxlarge_toolchain.py
+++ b/docker/usr/local/toolchain/generate_g4dnxlarge_toolchain.py
@@ -1,0 +1,32 @@
+#! /usr/bin/env python3
+
+import argparse
+import os
+
+import generate_toolchains
+
+target = {
+    "flags": [
+        "-march=skylake",
+        "-mtune=skylake",
+        "-mabm",
+        "-mno-sgx",
+        "--param l1-cache-size=32",
+        "--param l1-cache-line-size=64",
+        "--param l2-cache-size=36608",
+        "-fPIC",
+    ],
+    "release_flags": ["-O3", "-DNDEBUG"],
+    "asm_flags": ["-DELF", "-D__x86_64__", "-DPIC"],
+    "asm_object": "elf64",
+}
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Generate toolchain files for the docker images")
+    parser.add_argument(
+        "--prefix", default=os.path.join("usr", "local"), help="Prefix path to store generated files in"
+    )
+
+    args = parser.parse_args()
+
+    generate_toolchains.generate(args.prefix, "g4dnxlarge", target)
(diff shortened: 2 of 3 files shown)


docker/usr/local/toolchain/generate_g4dnxlarge_toolchain.py
```diff
@@ -0,0 +1,32 @@
+#! /usr/bin/env python3
```
KipHamiltons, line 1: I would love a comment at the top of this explaining what this target is for. Future years will thank us

tools/webots.py
```diff
@@ -178,9 +198,9 @@ def exec_push():
         sys.exit(exit_code)
 
 
-def run(sub_command, roles=None, role=None, **kwargs):
+def run(sub_command, roles=None, role=None, target="generic", **kwargs):
```
KipHamiltons, line 201: IMO the default should be this. AMD and old intel users can use the flag lel
```suggestion
def run(sub_command, roles=None, role=None, target="g4dnxlarge", **kwargs):
```

KipHamiltons: @JosephusPaye what do you reckon

Y0sh1dk: I think It should default to "generic" because we will likely only be running on aws for this robocup

KipHamiltons: we're also only planning to use this webots build tool for this robocup?

Y0sh1dk: I believe the webots tool will still be useful after robocup.

docker/usr/local/package/openblas/g4dnxlarge.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
```
abihall, line 2: Please add a comment here, I have no idea what i'm looking at

tools/webots.py
CMurtagh-LGTM: Set the default to `g4dnxlarge`, if people want to build generic they can do a normal build.

As this module stand it makes sense to have this target, if we want to change it in the furture, we can change the default then.

tools/webots.py
```diff
@@ -78,6 +92,12 @@ def exec_build(roles):
         cprint(f"unable to build code, exit code {exit_code}", "red", attrs=["bold"])
         sys.exit(exit_code)
 
+    print("Setting target 'generic'...")
```
abihall, line 95: Why do you set the target to the passed target and then set the target to generic after clean, configure and build? A comment would be nice here if possible. Comments a nice here for people that don't understand... like me

tools/webots.py
```diff
@@ -178,9 +198,9 @@ def exec_push():
...
 
 
-def run(sub_command, roles=None, role=None, **kwargs):
+def run(sub_command, roles=None, role=None, target="generic", **kwargs):
     if sub_command == "build":
-        exec_build(roles)
+        exec_build(target, roles)
     elif sub_command == "push":
         exec_push()
     elif sub_command == "run":  # For testing
```
abihall, line 206: What is this comment? Not very clear what you are trying to do here

abihall (commented):
A few comment requests, I cannot understand your code without some comments on what the code is achieving.

tools/webots.py
Y0sh1dk: The image built is intended to be ran on AWS for RoboCup, the `./b webots run` is usually used just for testing the built image.

tools/webots.py
Y0sh1dk: 99% of the time we are compiling for target `generic`. So its probably best to set it back to it after building a docker image for the given target.

tools/webots.py
Y0sh1dk: Finneee

tools/webots.py
KipHamiltons:
```suggestion
    print("Setting back to target 'generic'...")
```

KipHamiltons (approved):

KipHamiltons:
LGTM - Let's get this merged

PR #584, opened 2021-06-24
Fixes FileExists exception caused by the log file not being a directory

KipHamiltons:
if the file `$NUBOTS_HOME/log` existed, and it wasn't a directory, an exception is thrown, and the script exits. This change catches the exception, and moves that existing log into the new log directory

### DIFF
--- a/tools/run.py
+++ b/tools/run.py
@@ -49,7 +49,14 @@ def run(args, use_gdb, use_valgrind, **kwargs):
     env = os.environ
 
     # Make sure the log directory exists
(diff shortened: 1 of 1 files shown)


miikyla (commented):
Looks good

TrentHouliston:
How was this happening in the first place?

CMurtagh-LGTM (approved):
LGTM

KipHamiltons:
@TrentHouliston in `FileLogHandler.yaml` the default log file is `$NUBOTS_HOME/log`, so anyone who ran a role with file logging before #565 then ran stuff afterwards runs into this error

PR #585, opened 2021-06-24
Goal detector throwout for colinear goals, changed to merging clusters to solve the problem instead

tayloryoung6396:
This logic seems back to front for the throwout, can someone look through it


module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -260,6 +262,100 @@ namespace module::vision {
...
+                                rGOo_b /= distance_b;
+
+                                log<NUClear::DEBUG>(fmt::format("Goal post b at distance {}", distance_b));
+
+                                // Calculate the angular distance between the two posts
+                                const float post_to_post_theta = rGOo_a.dot(rGOo_b);
+
+                                const float d = distance_a < distance_b ? distance_a : distance_b;
+                                const float radius =
+                                    (2.0 * d)
```
KipHamiltons, line 306: I have no idea how this calculation works lol

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -260,6 +262,100 @@ namespace module::vision {
...
+                                        // If we delete the outer loop iterator, no need to continue the inner loop for
+                                        // this outer loop
+                                        increment_a = true;
+                                        break;
+                                    }
+                                    else {
+                                        // Remove the goal post
+                                        goal_post_b = goals->goals.erase(goal_post_b);
+                                        log<NUClear::DEBUG>(fmt::format(
+                                            "Throwout post at distance {}, post in front diatance {} because {} < {}",
```
KipHamiltons, line 335:
```suggestion
                                            "Throwout post at distance {}, post in front distance {} because {} < {}",
```

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -260,6 +262,100 @@ namespace module::vision {
                         }
                     }
 
+                    log<NUClear::DEBUG>("Culling colinear goals");
+
+                    for (auto it1 = goals->goals.begin(); it1 != goals->goals.end(); it1 = std::next(it1)) {
+                        log<NUClear::DEBUG>(fmt::format("Post Disances {}", it1->post.distance));
```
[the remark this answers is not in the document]
tayloryoung6396: [WIP]

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -260,6 +262,100 @@ namespace module::vision {
...
+                    }
+                    // If a sided goal is "behind" another sided goal it cannot be a goal
+                    // loop through all goals (sided), if any other goal is behind it (ie ~same phi theta && goal1_dist
+                    // < goal2_dist) throw it away! (or for now, make it an unknow goal or something)
+                    const Eigen::Affine3f Hcw(horizon.Hcw.cast<float>());
+                    for (auto goal_post_a = goals->goals.begin(); goal_post_a != goals->goals.end();) {
```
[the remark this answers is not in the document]
tayloryoung6396: ?

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -260,6 +262,100 @@ namespace module::vision {
...
+                            if (goal_post_a != goal_post_b) {
+                                log<NUClear::DEBUG>(fmt::format("Goal post a at distance {}", distance_a));
+                                // Vector from camera to the bottom center of the post, projected to the observation
+                                Eigen::Matrix<float, 2, 1> rGOo_b =
+                                    (goal_post_b->post.bottom.head<2>() * goal_post_b->post.distance);
+
+                                // Length of that vector
+                                const float distance_b = rGOo_b.norm();
+
+                                rGOo_b /= distance_b;
```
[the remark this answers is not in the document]
tayloryoung6396: Yes from what I can see @Bidski

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -260,6 +262,100 @@ namespace module::vision {
...
+
+                                log<NUClear::DEBUG>(fmt::format("Goal post b at distance {}", distance_b));
+
+                                // Calculate the angular distance between the two posts
+                                const float post_to_post_theta = rGOo_a.dot(rGOo_b);
```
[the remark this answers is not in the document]
tayloryoung6396: @KipHamiltons As the name of the vector would suggest `rGOo` is Observation plane to Goal
@Bidski Any suggestions for a better name, this math doesn't lend itself to good names

module/vision/GoalDetector/src/GoalDetector.cpp
tayloryoung6396: This is the inverse of a calculation above

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -260,6 +262,100 @@ namespace module::vision {
...
+                    for (auto goal_post_a = goals->goals.begin(); goal_post_a != goals->goals.end();) {
+                        bool increment_a = false;
+                        // Vector from camera to the bottom center of the post, projected to the observation
+                        Eigen::Matrix<float, 2, 1> rGOo_a =
+                            (goal_post_a->post.bottom.head<2>() * goal_post_a->post.distance);
+
+                        // Length of that vector
+                        const float distance_a = rGOo_a.norm();
+
+                        rGOo_a /= distance_a;
```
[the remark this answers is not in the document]
Bidski: It is completely obvious to me that this is guaranteed not to be zero :stuck_out_tongue:

module/vision/GoalDetector/src/GoalDetector.cpp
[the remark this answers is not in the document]
Bidski: `angular_distance`, it might be scaled by `cos` but its still an angular distance :stuck_out_tongue:

miikyla (commented):
Comments look good

miikyla (commented):
Looks good

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -119,6 +119,41 @@ namespace module::vision {
...
+                        const float min_b = std::atan2(rays(1, *range_b.first), rays(0, *range_b.first));
+                        const float max_b = std::atan2(rays(1, *range_b.second), rays(0, *range_b.second));
+
+                        // The clusters are overlapping, merge them
+                        if (((min_a <= min_b) && (min_b <= max_a)) || ((min_b <= min_a) && (min_a <= max_b))) {
+                            // Append the second cluster on to the first
+                            it->insert(it->end(), it2->begin(), it2->end());
+                            // Delete the second cluster
+                            it2 = clusters.erase(it2);
+                        }
```
wongjoel, line 148: This is using C++ features I'm not familiar with, but I'll take a stab at the logic:

`a` and `b` are clusters, and we've already calculated the max and min:
If min_a <= min_b <= max_a, then the clusters overlap because b contains point(s) in a
If min_b <= min_a <= max_b, then the clusters overlap because a contains point(s) in b

I don't know what the `insert` and `erase` functions do exactly, but from the comments the intent seems fine to me.

tayloryoung6396: It's essentially the same as https://github.com/NUbots/NUbots/blob/c30d4dd8d73964f2f3f6dd45b181bfe2c1be35d9/module/vision/GreenHorizonDetector/src/GreenHorizonDetector.cpp#L91 where we try to join near clusters together that we believe is actually the same object

wongjoel:
I'm confused, has all the code for Culling colinear goals been removed from this PR?

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -262,105 +292,6 @@ namespace module::vision {
...
-                        // Vector from camera to the bottom center of the post, projected to the observation
-                        Eigen::Matrix<float, 2, 1> rGOo_a =
-                            (goal_post_a->post.bottom.head<2>() * goal_post_a->post.distance);
-
-                        // Length of that vector
-                        const float distance_a = rGOo_a.norm();
-
-                        rGOo_a /= distance_a;
-
-                        for (auto goal_post_b = goals->goals.begin(); goal_post_b != goals->goals.end();) {
```
wongjoel: Was all this supposed to be removed?

tayloryoung6396: Yeah, we changed the metric and decided merging clusters would be better than a colinear solution

module/vision/GoalDetector/src/GoalDetector.cpp
wongjoel: makes sense - I didn't realise you could call insert on an iterator

wongjoel (approved):
This looks like a sensible approach to dealing with overlapping objects - LGTM

module/vision/GoalDetector/src/GoalDetector.cpp
wongjoel: got it

wongjoel (approved):
This most recent commit means that we now look at the last cluster, though since the inner loop starts from the iterator's position, it won't have any elements to compare against on that iteration. Looks cleaner to me, and deals better with the case where clusters is only size 1.

PR #590, opened 2021-06-25
Implements a new particle filter

Bidski:
The new filter achieves better performance on the unit test than the old filter did


module/localisation/BallLocalisation/src/BallModel.hpp
```diff
@@ -93,7 +93,7 @@ namespace module::localisation {
         inline int getRogueCount() const {
             return n_rogues;
         }
-        Eigen::Matrix<Scalar, 2, 1> getRogueRange() const {
+        [[nodiscard]] Eigen::Matrix<Scalar, 2, 1> get_rogue() const {
```
ysims, line 96: whats a rogue range

ysims: oh like rogue particles and how large of a range do we allow them to stick about?

module/localisation/RobotParticleLocalisation/src/RobotParticleLocalisation.cpp
```diff
@@ -185,8 +185,9 @@ namespace module::localisation {
             "Reset Robot Hypotheses",
             [this](const ResetRobotHypotheses& locReset, const Sensors& sensors) {
                 if (locReset.hypotheses.empty()) {
-                    filter.set_state(config.start_state,
-                                     std::vector<Eigen::Vector3d>(config.start_state.size(), config.start_variance));
+                    filter.set_state(
+                        config.start_state,
+                        std::vector<Eigen::Matrix3d>(config.start_state.size(), config.start_variance.asDiagonal()));
```
ysims, line 190: what's the benefit of diagonal?

ysims (commented):
A short review while I was waiting for my uber. This is a decent PR. In a nutshell, what is the update from the old one to this one that makes it better? Is this field line stuff?

module/localisation/BallLocalisation/src/BallModel.hpp
Bidski: `rogue_range` is old terminology, not sure what it is referring to. A rogue particle can be whatever you want (i.e. sample new state from a uniform distribution spread out over the field).

module/localisation/RobotParticleLocalisation/src/RobotParticleLocalisation.cpp
Bidski: covariance matrices are square, symmetric matrices. We generally only set the main diagonal elements because determining the off-diagonals is painful. Using `asDiagonal` takes a vector and makes a matrix that contains that vector as its main diagonal

tayloryoung6396 (commented):
The behaviour in localisation looks the same as the previous filter. This PR doesn't break thing, but the unit test does perform better.

module/localisation/BallLocalisation/src/BallModel.hpp
TrentHouliston: Rogue range was basically the range of values for each dimension that a rogue could be generated in. Now the rogues are just generated by the model directly

shared/utility/math/stats/multivariate.hpp
```diff
@@ -0,0 +1,119 @@
...
+    private:
+        /// @brief The mean (\mu) of the multivariate normal distribution
+        Eigen::Matrix<Scalar, N, 1> mean = Eigen::Matrix<Scalar, N, 1>::Zero();
+        /// @brief The covariance (\Sigma) of the multivariate normal distribution
+        Eigen::Matrix<Scalar, N, N> covariance = Eigen::Matrix<Scalar, N, N>::Zero();
+        /// @brief A matrix A such that AA' = \Sigma
+        Eigen::Matrix<Scalar, N, N> transform = Eigen::Matrix<Scalar, N, N>::Zero();
+
+        /// @brief Standard normal distribution. Used in MultivariateNormal::sample
+        std::random_device rd{};
```
TrentHouliston, line 112:
```suggestion
```

shared/utility/math/stats/multivariate.hpp
```diff
@@ -0,0 +1,119 @@
...
+            : MultivariateNormal(Eigen::Matrix<Scalar, N, 1>::Zero(), covariance_) {}
+
+        /**
+         * @brief Construct a multivariate normal distribution of dimension N with the provided mean and covariance
+         *
+         * @param mean_ The mean for this mean multivariate normal distribution
+         * @param covariance_ The covariance matrix for this multivariate normal distribution
+         */
+        MultivariateNormal(const Eigen::Matrix<Scalar, N, 1>& mean_, const Eigen::Matrix<Scalar, N, N>& covariance_)
+            : mean(mean_), covariance(covariance_), gen(rd()) {
```
TrentHouliston, line 62: No need to store a random device we just need to use it once to seed the pseudorandom number generator
```suggestion
            : mean(mean_), covariance(covariance_), gen(std::random_device()()) {
```

shared/utility/math/stats/multivariate.hpp
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * The NUbots Codebase is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright 2013 NUbots <nubots@nubots.net>
```
TrentHouliston, line 17:
```suggestion
 * Copyright 2021 NUbots <nubots@nubots.net>
```

module/localisation/RobotParticleLocalisation/src/RobotModel.hpp
```diff
@@ -112,7 +112,7 @@ namespace module::localisation {
         int getRogueCount() const {
             return n_rogues;
         }
-        Eigen::Matrix<Scalar, Eigen::Dynamic, 1> getRogueRange() const {
+        [[nodiscard]] Eigen::Matrix<Scalar, Eigen::Dynamic, 1> get_rogue() const {
             return resetRange;
```
TrentHouliston, line 116: Is returning a range here correct?
Shouldn't you be generating a rogue here based on the rage as a uniform value (I think eigen can do it)
Also, pretty sure the return type here rather than dynamic can just be the model size * 1

module/localisation/BallLocalisation/src/BallModel.hpp
```diff
@@ -93,7 +93,7 @@ namespace module::localisation {
...
             return resetRange;
```
TrentHouliston, line 97: Pretty sure this rogue range should also be just generating rogues with the correct distribution

shared/utility/math/filter/ParticleFilter.hpp
```diff
@@ -26,8 +26,42 @@
...
+     * @author Alex Biddulph, Trent Houliston
+     *
+     * @details Uses a set of particles (or samples) to represent possible states in a system. Each particle is randomly
+     * selected from a multivariate normal distribution. Each particle is weighted by how likely it is to be the real
+     * state of the system. Particles with a higher weight are more likely to be resampled (with noise) during the time
+     * update.
+     *
+     * Time Update: Resamples all particles, perturbing the resampled particles with the model's process noise, and asks
+     * the model to predict what should happen to each particle after an elapsed amount of time. Also known as the
+     * predict step.
```
TrentHouliston, line 54: Is this really the predict step?

shared/utility/math/filter/ParticleFilter.hpp
```diff
@@ -26,8 +26,42 @@
...
+     *
+     * Measurement Update: Asks the model to predict what the next measurement will be (based on current state/particle)
+     * and the finds the difference between the model's prediction and the actual measurement. The likelihood of this
+     * measurement occuring is then determined using a multivariate normal distribution probability density function.
```
TrentHouliston, line 58: It's actually the other way around, it's calculating the likelihood of this model's prediction given the measurement

shared/utility/math/filter/ParticleFilter.hpp
```diff
@@ -38,242 +72,396 @@ namespace utility::math::filter {
...
+         * @param covariance The covariance of the multivariate normal distribution. Used for sampling the initial
+         * particles in the filter.
+         * @param resample_method_ The method to use when resampling particles during the time update step.
+         * @param residual_method_ When using ResampleMethod::RESIDUAL for the resample method, this method will be used
+         * when the residual method needs to resample residual particles.
+         */
+        ParticleFilter(const StateVec& mean,
+                       const StateMat& covariance,
+                       const ResampleMethod& resample_method_ = ResampleMethod::RESIDUAL,
+                       const ResampleMethod& residual_method_ = ResampleMethod::SYSTEMATIC) {
```
TrentHouliston, line 102: Don't know how I feel about this one given that sometimes this parameter does nothing. Honestly it might make more sense to use this as a bitfield and just have a `flags` field that you set to `ResampleMethod::SYSTEMATIC | ResampleMethod::RESIDUAL`

shared/utility/math/filter/ParticleFilter.hpp
```diff
@@ -38,242 +72,396 @@ namespace utility::math::filter {
...
+         * @param mean The mean of the multivariate normal distribution. Used for sampling the initial particles in the
+         * filter. There is one mean for each initial hypothesis.
+         * @param covariance The covariance of the multivariate normal distribution. Used for sampling the initial
+         * particles in the filter.  There is one covariance for each initial hypothesis.
+         * @param resample_method_ The method to use when resampling particles during the time update step.
+         * @param residual_method_ When using ResampleMethod::RESIDUAL for the resample method, this method will be used
+         * when the residual method needs to resample residual particles.
+         */
+        ParticleFilter(const std::vector<StateVec>& mean,
+                       const std::vector<StateMat>& covariance,
```
TrentHouliston, line 135: If they need to be paired, then make them paired
```suggestion
        ParticleFilter(const std::vector<std::pair<StateVec, StateMat>>& pairs,
```
Never rely on the user for anything!

shared/utility/math/filter/ParticleFilter.hpp
```diff
@@ -38,242 +72,396 @@ namespace utility::math::filter {
...
+            // Get indexes to the particles that are being resampled
+            // Some particles may be resampled multiple times
+            std::vector<int> idx = resample_particles();
+
+            // Resample the particles
+            // For each index get the corresponding particle and perturb it with the model's process noise
+            ParticleList resampled_particles(Model::size, particles.cols());
+            for (int i = 0; i < int(idx.size()); ++i) {
+                // Create a new particle by using the selected particle as the mean + value from process noise
+                resampled_particles.col(i) = particles.col(idx[i]) + multivariate.sample();
```
TrentHouliston, line 349: Might be a good idea to do this perturbing of the state by the multivariate and the limiting in the time update rather than here. It's not really part of resampling and it will make it clearer each step that we are doing

shared/utility/math/stats/multivariate.hpp
KipHamiltons: What's the thinking behind this suggestion

module/localisation/RobotParticleLocalisation/src/RobotModel.hpp
Bidski: I dont know if it is correct or not. Fixing that is for a later PR (this model doesn't currently use rogues anyway so it doesn't make a difference at the moment)

Bidski: The range (in the config) is also current `(2, 2, 0)`

shared/utility/math/filter/ParticleFilter.hpp
Bidski: According to some matlab docs I read, what they call the predict step sounds like our time update step. But if you can word this whole thing better I am open to suggestions

shared/utility/math/filter/ParticleFilter.hpp
Bidski: That also sounds kinda clunky tbh

shared/utility/math/filter/ParticleFilter.hpp
Bidski: Oh yea, I forgot that was a thing lol

module/localisation/RobotParticleLocalisation/src/RobotModel.hpp
Bidski: Which doesn't actually make sense as a range either

shared/utility/math/filter/ParticleFilter.hpp
tayloryoung6396: I mean I don't have a problem with the naming.. kinda makes sense

TrentHouliston: It's just a pain that our measurement function is called "predict" lol

shared/utility/math/filter/ParticleFilter.hpp
```diff
@@ -38,242 +74,393 @@ namespace utility::math::filter {
...
+            // The residual resampling method requires a secondary method to resample the residual particles. This
+            // secondary method cannot be the residual method.
+            if (residual_method == ResampleMethod::RESIDUAL) {
+                throw std::runtime_error(
+                    "Invalid setting for residual method. Can't use the residual resampling method to sample "
+                    "residuals.");
             }
-
-            return new_particles;
+            residual_method = resample_method_;
```
tayloryoung6396, line 115: I think you need to not store a in b and b in a...

tayloryoung6396 (changes requested):

shared/utility/math/filter/ParticleFilter.hpp
TrentHouliston: Lol @Bidski

shared/utility/math/filter/ParticleFilter.hpp
Bidski: It is, I can remove that part of the comment if you think it will lead to confusion?

shared/utility/math/filter/ParticleFilter.hpp
```diff
@@ -38,242 +74,393 @@ namespace utility::math::filter {
...
+            // Return the mean log probability
             return logits.mean();
         }
 
-        // Take the mean of all particles, pay no attention to the type of the data
-        StateVec get() const {
+        /**
+         * @brief Calculates and returns the mean of all particles.
+         */
+        [[nodiscard]] StateVec get() const {
```
tayloryoung6396, line 439: I feel like `get()` should be `getMean()` given that the other getter functions are specified

shared/utility/math/filter/ParticleFilter.hpp
Bidski: Lol @TrentHouliston you missed it too

shared/utility/math/filter/ParticleFilter.hpp
TrentHouliston: Yeah I would just to avoid future confusion

shared/utility/math/filter/ParticleFilter.hpp
TrentHouliston: Yes, but less clunky until we find a better alternative

tayloryoung6396 (approved):

TrentHouliston (approved):
Sure

tayloryoung6396:
@TrentHouliston You wrote half of this :stuck_out_tongue:  and missed the simple critical errors

CampbellRoss (commented):
Had a read, looks good to me

PR #596, opened 2021-06-25
Changes target flags to pull more intuitively

KipHamiltons:


tools/target.py
```diff
@@ -53,19 +51,31 @@ def run(target, build, should_pull, **kwargs):
...
-            build = True
-            should_pull = False
-
-        if should_pull:
+        if pull:
             # Pull the image from dockerhub
             platform.pull(defaults.image, target)
 
-        # If user wants or needs to build it locally, do that
+        # If user wants or needs to build it locally, do that unless the build context is the same as origin/master
```
abihall, line 58: Where does it check if the build context is the same as origin/master? I can't see where it does that

tools/target.py
```diff
@@ -21,21 +21,19 @@ def register(command):
         "--build",
         dest="build",
         action="store_true",
-        default=False,
         help="build the image locally, instead of pulling and tagging it",
     )
 
     command.add_argument(
```
Crafty15, line 27: Nice comments down further. Would be super cool to have similar comments up here.

tools/target.py
Bidski: https://github.com/NUbots/NUbots/blob/a85369bf060cb27c54dfd4476c6e24e109214a48/tools/target.py#L45-L52

KipHamiltons: just above that in the file (in parts we didn't change), we run `git diff` to check if there are changes in the `docker` directory. If there are, the "build context" has been changed (and is "dirty")

tools/target.py
```diff
@@ -53,19 +51,31 @@ def run(target, build, should_pull, **kwargs):
...
+                cprint(
+                    "There aren't any changes to be made to the docker image. "
+                    "Skipping build, and using tagged image instead",
+                    "red",
+                    attrs=["bold"],
+                )
+            else:
+                platform.build(defaults.image, target)
+        # If the user hasn't selected build but they have changed the docker directory, then we warn them that the image
+        # isn't going to be built, even though the dirty diff implies they might want it to be built
```
Crafty15, line 70: What is a diff? and why is a dirty one bad?

Bidski (approved):
Seems more logical to me (but then again this was my idea)

tools/target.py
```diff
@@ -53,19 +51,31 @@ def run(target, build, should_pull, **kwargs):
...
+        else:
+            if changed:
```
Crafty15, line 72: What is the construct that changed is initialised with (line 45). Not in your changes, just wondering.

Crafty15 (commented):
Nice one bruv. Mind giving me a quick run down on what this does?

tools/target.py
abihall: Okay thanks I see it now

tools/target.py
Y0sh1dk: Dirty diff here is meaning that your local `docker` directory is different to the `docker` directory that is on `origin/master`. In this case you would usually want to rebuild a docker image with your local changes, if you decide not too, a warning is displayed.

tools/target.py
Y0sh1dk: `changed` gets set to `true` on line 45 if the git diff returns a exit code other than 0 (this happens when your local docker directory is different to the docker directory that is on origin/master

PR #598, opened 2021-06-25
Sets Sensorfilter log level using new construct to NUClear::ERROR

KipHamiltons:
Also changes cholesky decomps to be warnings, so that they don't clog up the robocup logs. Given that we can now recover from cholesky decomp errors, this seems like the right log level anyway


KipHamiltons:
Updated to use new log_level nice thing

CMurtagh-LGTM (approved):
Looks Good To Me

TrentHouliston (approved):
And by new, you mean only introduced about 8 years ago or so :P

Bidski (approved):
Seems reasonable

miikyla (commented):
Looks good!

PR #599, opened 2021-06-25
Fixes NUsight messages

tayloryoung6396:


Bidski (changes requested):
Needs formatting changes, but otherwise it looks good

Bidski (approved):

PR #604, opened 2021-06-25
Fix webots run

Y0sh1dk:
Forgot default value for `clean` when using `./b webots run`


KipHamiltons:
To clarify (when I question this once I have had some sleep), we need this because when we run another subcommand which isn't `webots build`, it doesn't set the flag, and thus we are missing a positional argument when `run` is called.

KipHamiltons (approved):
Tested. LGTM

PR #606, opened 2021-06-25
Reduce logging in the `webotsrobocup` role

JosephusPaye:
This turns of logging off `CompressedImage`, `VisualMesh`, and `GreenHorizon` messages in the webotsrobocup role. This is needed to not exceed the 10GB per robot limit, which we've exceeded so far in our first few games.


tayloryoung6396 (approved):
LGTM

Y0sh1dk (approved):
LGTM. At one point we should benchmark the storage requirments of each message type.

PR #614, opened 2021-07-03
Add a priority comparison algorithm for the Director

TrentHouliston:
This implements the priority comparison algorithm which determines which of any two tasks should be executed


module/extension/Director/src/Director.hpp
```diff
@@ -62,6 +62,22 @@ namespace module::extension {
...
 
+        /**
+         * Compares the priorities of two director tasks and returns true if the challenger has priority over the
+         * incumbent.
+         *
+         * The function requires that the challengers precendence is strictly greater than the incumbent.
+         * This ensures that we don't change tasks unnecessarily when the priority is equal.
+         *
+         * @param incumbent     the task to compare which is currently the active running task
+         * @param challenger    the task to compare which wants to run but is not currently
```
ysims, line 73:
```suggestion
         * @param challenger    the task to compare which wants to run but is not currently running
```

module/extension/Director/src/Director.hpp
```diff
@@ -62,6 +62,22 @@ namespace module::extension {
...
+         *
+         * @return true     if the challenger has strictly higher priority than the incumbent
+         * @return false    if the incumbent task has equal or higher priority
+         */
+        bool compare_priority(const std::shared_ptr<const ::extension::behaviour::commands::DirectorTask>& incumbent,
```
KipHamiltons, line 78: I would also consider changing this variable name. This is not a commutative operation, but the name implies it could be? I could also be reading too far into this

module/extension/Director/src/bits/compare_priority.cpp
```diff
@@ -0,0 +1,86 @@
...
+namespace module::extension {
+
+    using ::extension::behaviour::commands::DirectorTask;
+
+    bool Director::compare_priority(const std::shared_ptr<const DirectorTask>& incumbent,
+                                    const std::shared_ptr<const DirectorTask>& challenger) {
+
+        // Compare to yourself and you get that you shouldn't change tasks
+        if (incumbent->requester_id == challenger->requester_id) {
+            return false;
```
KipHamiltons, line 37: Should we try to log this? It seems like something we should avoid if possible.

module/extension/Director/src/Director.hpp
```diff
@@ -62,6 +62,22 @@ namespace module::extension {
...
+         * incumbent.
+         *
+         * The function requires that the challengers precendence is strictly greater than the incumbent.
+         * This ensures that we don't change tasks unnecessarily when the priority is equal.
+         *
+         * @param incumbent     the task to compare which is currently the active running task
+         * @param challenger    the task to compare which wants to run but is not currently
+         *
+         * @return true     if the challenger has strictly higher priority than the incumbent
+         * @return false    if the incumbent task has equal or higher priority
```
KipHamiltons, line 76: add a @throws too please

module/extension/Director/src/bits/compare_priority.cpp
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        // Function to get the priorities of the ancestors of this task
+        auto get_ancestor_priorities = [this](const std::shared_ptr<const DirectorTask>& task) {
+            // Loop up through the providers until we reach a point where a task was emitted by a non provider
+            std::vector<std::tuple<uint64_t, int, bool>> ancestors;
+            for (auto t = task; providers.count(t->requester_id) != 0;) {
+                auto provider = providers[t->requester_id];
```
KipHamiltons, line 45:
```suggestion
                const auto& provider = providers[t->requester_id];
```

module/extension/Director/src/bits/compare_priority.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+        // Remove all of the common ancestors
+        while (std::get<0>(a_p.back()) == std::get<0>(b_p.back())) {
+            a_p.pop_back();
+            b_p.pop_back();
+        }
+
+        // Work out if there are any optionals in either of the tasks parentage
+        bool a_optional = std::any_of(a_p.begin(), a_p.end(), [](const auto& v) { return std::get<2>(v); });
+        bool b_optional = std::any_of(b_p.begin(), b_p.end(), [](const auto& v) { return std::get<2>(v); });
```
KipHamiltons, line 75:
```suggestion
        const bool a_optional = std::any_of(a_p.begin(), a_p.end(), [](const auto& v) { return std::get<2>(v); });
        const bool b_optional = std::any_of(b_p.begin(), b_p.end(), [](const auto& v) { return std::get<2>(v); });
```

module/extension/Director/src/bits/compare_priority.cpp
```diff
@@ -0,0 +1,86 @@
...
+        auto get_ancestor_priorities = [this](const std::shared_ptr<const DirectorTask>& task) {
+            // Loop up through the providers until we reach a point where a task was emitted by a non provider
+            std::vector<std::tuple<uint64_t, int, bool>> ancestors;
+            for (auto t = task; providers.count(t->requester_id) != 0;) {
+                auto provider = providers[t->requester_id];
+                auto& group   = groups[provider->type];
+
+                if (group.active_task != nullptr) {
+                    ancestors.emplace_back(t->requester_id, t->priority, t->optional);
+                    t = group.active_task;
```
ysims, line 50: I sort of get this function but not sure how it actually achieves what I think it's meant to do? I assume it goes up through the chain of provider task calls and stores all those tasks in `ancestors`. Is it also meant to include the input task `task`? 
And how does this line, `t=group.active_task` actually go up the chain? To me it looks like it just grabs the task that is active in this tasks group - surely that's the same task every time (except maybe not the first time)? 
My understanding of task groups is that you might have `statically_stable_walk` and `dynamically_stable_walk` and they're in a `walk` group together. Or maybe I'm getting confused with something else?
Definitely not getting something here, but I am curious about if the first task is an ancestor as well?

module/extension/Director/src/bits/compare_priority.cpp
```diff
@@ -0,0 +1,86 @@
...
+            b_p.pop_back();
+        }
+
+        // Work out if there are any optionals in either of the tasks parentage
+        bool a_optional = std::any_of(a_p.begin(), a_p.end(), [](const auto& v) { return std::get<2>(v); });
+        bool b_optional = std::any_of(b_p.begin(), b_p.end(), [](const auto& v) { return std::get<2>(v); });
+
+        // If one is optional and not the other, the non optional one is true
+        if (a_optional != b_optional) {
+            return a_optional;  // If they are not equal and a is optional, a < b
```
KipHamiltons, line 79: this is cool and correct, but its a headtwister to think about lol

TrentHouliston: I agree... But I really really didn't want to have to write the 4 if statements I'd need otherwise :P

KipHamiltons: I like the change :+1:

module/extension/Director/src/bits/compare_priority.cpp
TrentHouliston: It is something we should avoid. I had to put that in there because if you put the task in twice you end up with a segfault because of the ancestry being empty

module/extension/Director/src/Director.hpp
Bidski: which variable are you talking about?

TrentHouliston: Open to name suggestions for this function, I've changed it like 3 times before it even made it to a PR

module/extension/Director/src/bits/compare_priority.cpp
```diff
@@ -0,0 +1,87 @@
...
+ * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright 2021 NUbots <nubots@nubots.net>
+ */
+
+#include <algorithm>
+#include <tuple>
+#include <vector>
+
+#include "../Director.hpp"
```
Bidski, line 24: Pretty sure this should already be on the search path

module/extension/Director/src/bits/compare_priority.cpp
```diff
@@ -0,0 +1,87 @@
...
+    bool Director::compare_priority(const std::shared_ptr<const DirectorTask>& incumbent,
+                                    const std::shared_ptr<const DirectorTask>& challenger) {
+
+        // Compare to yourself and you get that you shouldn't change tasks
+        if (incumbent->requester_id == challenger->requester_id) {
+            return false;
+        }
+
+        // Function to get the priorities of the ancestors of this task
+        auto get_ancestor_priorities = [this](const std::shared_ptr<const DirectorTask>& task) {
```
Bidski, line 41: More comments on how this function is actually traversing the tree would be good. It took a second of thinking to figure out why this should work

Something that spells this process out (like this) but written better and aligned with appropriate lines
> so you start with task A
> task A has a requester_id
> that requester_id is part of some provider group
> if that group is active then there is a task that initiated that group?
> and that task has a requester_id, and loop?
> until you find a requester_id that isn't part of a provider group

module/extension/Director/src/Director.hpp
KipHamiltons: I was indeed talking about the function, wrong word. `challenge_priority` perhaps?

module/extension/Director/src/bits/compare_priority.cpp
```diff
@@ -0,0 +1,87 @@
...
+                else {
+                    throw std::runtime_error("Task has broken parentage");
+                }
+            }
+            return ancestors;
+        };
+
+        // Convenience references for the two to make the code shorter
+        const auto& a = incumbent;
+        const auto& b = challenger;
```
ysims, line 61: I'm not sure if this is that useful, it's not like there is a lot of code. But I guess at the same time, there isn't a lot of code so it's not hard to look back and see what these mean... I'm mostly thinking because `a` and `b` are not things that can be swapped around, like in an `add(int a, int b)` function. They have specific meanings where one is the currently running task and the other is not.

module/extension/Director/src/bits/compare_priority.cpp
```diff
@@ -0,0 +1,87 @@
...
+        // Compare to yourself and you get that you shouldn't change tasks
+        if (incumbent->requester_id == challenger->requester_id) {
+            return false;
+        }
+
+        // Function to get the priorities of the ancestors of this task
+        auto get_ancestor_priorities = [this](const std::shared_ptr<const DirectorTask>& task) {
+            // Loop up through the providers until we reach a point where a task was emitted by a non provider
+            std::vector<std::tuple<uint64_t, int, bool>> ancestors;
+            for (auto t = task; providers.count(t->requester_id) != 0;) {
```
Bidski, line 44: Since loops are possible in the provider tree, this could loop inifinitely. You should either add a visitor list to prevent this or remove the possibility of loops

module/extension/Director/src/bits/compare_priority.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+        // Compare to yourself and you get that you shouldn't change tasks
+        if (incumbent->requester_id == challenger->requester_id) {
+            return false;
+        }
+
+        // Function to get the priorities of the ancestors of this task
+        auto get_ancestor_priorities = [this](const std::shared_ptr<const DirectorTask>& task) {
+            // Loop up through the providers until we reach a point where a task was emitted by a non provider
+            std::vector<std::tuple<uint64_t, int, bool>> ancestors;
```
Bidski, line 43: Is it worth making a struct to replace this tuple to make the rest of the code easier to read? `std::get` is really not the most readable construct

module/extension/Director/src/Director.hpp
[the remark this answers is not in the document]
TrentHouliston: I mean if you had clang-tidy they would literally write themselves

Bidski (changes requested):

module/extension/Director/src/bits/compare_priority.cpp
TrentHouliston: This shoudn't be `const auto&` here it's a shared_ptr that I'm grabbing, I want my own copy of it to ensure that it won't get deleted while using it. A reference won't do that for me.

ysims (commented):
Mostly just me trying to figure out what it all means. 

Do conditions like the robot having to be stable or something apply here? Is this done after checking that kind of thing, or is it not applicable here?

module/extension/Director/src/Director.hpp
TrentHouliston: Sure, unique words make it easier to understand concepts

Bidski: I think `compare_priority` is right. Comparisons dont have to be commutative (strictly speaking), but I think it might be in this case?

CMurtagh-LGTM:
Why is `compare_priority.cpp` in a folder called `bits`?

Bidski:
> Why is `compare_priority.cpp` in a folder called `bits`?

I am also not hugely happy about this, but I get what he is trying to do. Maybe `bits` is not the right word?

module/extension/Director/src/bits/compare_priority.cpp
TrentHouliston: I mainly did this to make the following lines one liners. which made it easier for me to read because the function ends up less lines of code. If I change a and b back it gains about 10ish lines of code

ysims: Ah ok. I'd suggest `c` and `i` but those are probably the worst letters to have 😆

module/extension/Director/src/bits/compare_priority.cpp
TrentHouliston: I think there was a bug here, so I understand your confusion, I tried to explain it a little better in the comments.

Basically you get each tasks emitter, which will be different providers and you use them to go up the tree

module/extension/Director/src/bits/challenge_priority.cpp
```diff
@@ -0,0 +1,93 @@
...
+                                      const std::shared_ptr<const DirectorTask>& challenger) {
+
+        // Compare to yourself and you get that you shouldn't change tasks
+        if (incumbent->requester_id == challenger->requester_id) {
+            return false;
+        }
+
+        // Function to get the priorities of the ancestors of this task
+        auto get_ancestor_priorities = [this](const std::shared_ptr<const DirectorTask>& task) {
+            // We are our first anscestor
```
ysims, line 42:
```suggestion
            // We are our first ancestor
```

TrentHouliston:
> Mostly just me trying to figure out what it all means.
> 
> Do conditions like the robot having to be stable or something apply here? Is this done after checking that kind of thing, or is it not applicable here?

That's another part of the algorithm. This determines if a task should even run. Working out if it has other conditions that need to be satisfied is done after this in the algorithm.
There is no point checking those other things if it doesn't even have the priority to run in the first place.

> > Why is compare_priority.cpp in a folder called bits?
>
> I am also not hugely happy about this, but I get what he is trying to do. Maybe bits is not the right word?

I imagine this director algorithm will get quite large, splitting up each part of the algorithm into separate files can make it easier to understand the individual parts.
Changing the folder name from bits to something else is fine, it's just a way to break off individual parts of the algorithm

module/extension/Director/src/Director.hpp
```diff
@@ -62,6 +62,25 @@ namespace module::extension {
...
+         * The function requires that the challengers precendence is strictly greater than the incumbent.
+         * This ensures that we don't change tasks unnecessarily when the priority is equal.
+         *
+         * @param incumbent     the task to compare which is currently the active running task
+         * @param challenger    the task to compare which wants to run but is not currently running
+         *
+         * @return true     if the challenger has strictly higher priority than the incumbent
+         * @return false    if the incumbent task has equal or higher priority
+         *
+         * @throws std::runtime_error if the directors provider anscestory is broken
```
KipHamiltons, line 78:
```suggestion
         * @throws std::runtime_error if the directors provider ancestry is broken
```

ysims:
The algorithm makes sense to me now. 

I thought the folder name was like bits/bytes which made no sense but now I realise it's probably like 'bits and pieces'. 
Maybe `algorithms` would work better? Or `utility`?

module/extension/Director/src/bits/compare_priority.cpp
TrentHouliston: Hopefully it's all good

module/extension/Director/src/bits/compare_priority.cpp
TrentHouliston: I'll add loop prevention earlier in the algorithm.
If it makes it here with a loop then crashing is the appropriate result.

module/extension/Director/src/Director.hpp
```diff
@@ -62,6 +62,25 @@ namespace module::extension {
          */
         void add_causing(const ::extension::behaviour::commands::CausingExpression& causing);
 
+        /**
+         * Compares the priorities of two director tasks and returns true if the challenger has priority over the
+         * incumbent.
+         *
+         * The function requires that the challengers precendence is strictly greater than the incumbent.
```
KipHamiltons, line 69:
```suggestion
         * The function requires that the challenger's precendence is strictly greater than the incumbent's.
```

module/extension/Director/src/Director.hpp
```diff
@@ -62,6 +62,25 @@ namespace module::extension {
...
+         * This ensures that we don't change tasks unnecessarily when the priority is equal.
+         *
+         * @param incumbent     the task to compare which is currently the active running task
+         * @param challenger    the task to compare which wants to run but is not currently running
+         *
+         * @return true     if the challenger has strictly higher priority than the incumbent
+         * @return false    if the incumbent task has equal or higher priority
+         *
+         * @throws std::runtime_error if the directors provider ancestry is broken
```
KipHamiltons, line 78:
```suggestion
         * @throws std::runtime_error if the director's provider ancestry is broken
```

module/extension/Director/src/bits/challenge_priority.cpp
```diff
@@ -0,0 +1,104 @@
...
+        const bool a_optional = std::any_of(a_p.begin(), a_p.end(), [](const auto& v) { return v.optional; });
+        const bool b_optional = std::any_of(b_p.begin(), b_p.end(), [](const auto& v) { return v.optional; });
+
+        // If both or neither are optional then we compare at the point where they were siblings.
+        // If both are optional then we would rather that whichever ancestor had higher priority have its optional tasks
+        if (a_optional == b_optional) {
+            return a_p.back().priority < b_p.back().priority;
+        }
+
+        // If the optionals are not equal then return if a_optional is true. a being optional would make it less than b
```
LachlanCourt, line 100: This comment doesn't seem to make sense to me - it's not only returning if a_optional is true it's just returning the value of a_optional? I'm guessing "it" in this case you are referring to the priority of task a?

module/extension/Director/src/bits/challenge_priority.cpp
```diff
@@ -0,0 +1,104 @@
...
+                }
+
+                // Add this new parent we found and set t so we can loop up further
+                t = group.active_task;
+                ancestors.emplace_back(t->requester_id, t->priority, t->optional);
+            }
+            return ancestors;
+        };
+
+        // Convenience references for the two to make the code shorter
```
LachlanCourt, line 76: These variables are only used once in the following line it seems unnecessary to declare them explicitly unless you plan to extend this algorithm further

module/extension/Director/src/bits/challenge_priority.cpp
KipHamiltons: This is in fact a correct suggestion :p

KipHamiltons (commented):
Seems fine I guess. Will approve after these spelling/grammar suggestions
EDIT: and by "Seems fine I guess", I mean "I can't see anything wrong with this, so, nice one. Keep it up :grin:"

module/extension/Director/src/bits/challenge_priority.cpp
TrentHouliston: Fine I'll remove them!

LachlanCourt: I just meant those two lines but getting rid of a and b entirely works too I guess 😆

module/extension/Director/src/bits/challenge_priority.cpp
TrentHouliston: I made the comment more involved for you

LachlanCourt: Thankyou for putting up with me being big dum 😁

LachlanCourt (dismissed):
This LGTM

module/extension/Director/src/bits/challenge_priority.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+#include "Director.hpp"
+
+#include "extension/Behaviour.hpp"
+
+namespace module::extension {
+
+    using ::extension::behaviour::commands::DirectorTask;
+
+    namespace {
```
JosephusPaye, line 32: Is the point of this anonymous namespace to scope the scruct defined therein to this translation unit? If so, a comment would be nice.

module/extension/Director/src/bits/challenge_priority.cpp
```diff
@@ -0,0 +1,103 @@
...
+                }
+
+                // Add this new parent we found and set t so we can loop up further
+                t = group.active_task;
+                ancestors.emplace_back(t->requester_id, t->priority, t->optional);
+            }
+            return ancestors;
+        };
+
+        // Get the ancestor priorities of a and b
```
ysims, line 77: Update the comment now that you've removed `a` and `b`.

LachlanCourt (approved):
Gorgeous. Iconic. Love it

JosephusPaye (commented):
Looks pretty good, the previous reviews have been good suggestions.

---

Couldn't comment on these lines, since they haven't been changed in this PR, but:

https://github.com/NUbots/NUbots/blob/houliston/director_priority/module/extension/Director/src/Director.hpp#L36 has Behavior spelt wrong (these spelling suggestions are probably getting tedious, you should get a spell check extension 😛)

https://github.com/NUbots/NUbots/blob/houliston/director_priority/module/extension/Director/src/Director.hpp#L45 Does the comment here with plural Providers imply `remove_provider` could remove more than one provider?

module/extension/Director/src/bits/challenge_priority.cpp
TrentHouliston: That is what anonymous namespaces do in c++ files.
Don't use them in headers or every single c++ file that includes it will make their own version of the struct

JosephusPaye: At first glance it seemed superfluous. Had to look it up to see what it was doing.

LachlanCourt (dismissed):
Ready to take the world by storm

JosephusPaye (dismissed):
𝕷𝖔𝖔𝖐𝖘𝕲𝖔𝖔𝖉𝕿𝖔𝕸𝖊

ysims:
I still don't like the folder name.

KipHamiltons (approved):
L(et's) G(et) T(his) M(erged)

ysims (approved):
This is acceptable.

module/extension/Director/src/Director.cpp
```diff
@@ -31,6 +31,8 @@ namespace module::extension {
     using ::extension::behaviour::commands::ProvidesReaction;
     using ::extension::behaviour::commands::WhenExpression;
     using ::NUClear::threading::Reaction;
+    using provider::Provider;
```
Bidski, line 34: Formatting?

Bidski (changes requested):

module/extension/Director/src/Director.cpp
TrentHouliston: I don't understand what you're mad about

Bidski (approved):
Go ahead. Do your worst

PR #617, opened 2021-07-03
Add a task removal function for the Director

TrentHouliston:


module/extension/Director/src/director/remove_task.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+            // Remove this task, we are no longer doing it
+            group.active_task = nullptr;
+
+            // Re-evaluate the queue of tasks for this group
+            reevaluate_queue(group);
+
+            // If nothing in the queue updated the active task to a new task we are now idle
+            // That also means we need to remove any subtasks this group had recursively
+            if (group.active_task == nullptr) {
```
LachlanCourt, line 43: When does `group.active_task` get updated? Is that something that will be happening in `run_task_pack`? Otherwise if I understand correctly it will set it to `nullptr` and never run the `else`

TrentHouliston: Yeah it should get set in `run_task_pack`, hopefully when the task pack runs it will become the active task

module/extension/Director/src/Director.cpp
```diff
@@ -174,9 +175,9 @@ namespace module::extension {
         on<Trigger<DirectorTask>, Sync<DirectorTask>>().then(
             "Director Task",
             [this](std::shared_ptr<const DirectorTask> task) {
-                // Root level task, make the TaskPack immediately and send it off to be executed as a root task
+                // Root level task, make the pack immediately and send it off to be executed as a root task
                 if (providers.count(task->requester_id) == 0) {
-                    emit(std::make_unique<TaskPack>(TaskPack({task})));
+                    emit(std::make_unique<TaskList>(TaskList({task})));
```
Bidski, line 180: Do you really need to construct it? Doesnt make_unique call the constructor for you?

module/extension/Director/src/Director.cpp
```diff
@@ -213,24 +214,19 @@ namespace module::extension {
...
+            // Go check if this state update has changed any of the tasks that are queued
+            reevaluate_queue(g);
+        });
 
-            for (const auto& task : pack) {
-                // TODO run this task through the director's algorithm to work out what to do with it
-                // TODO we need to process this pack as a whole, if one of the required actions fail they all do
-            }
+        // We have a new task pack to run
+        on<Trigger<TaskList>, Sync<Director>>().then("Run Task Pack", [this](const TaskList& pack) {  //
```
Bidski, line 228:
```suggestion
        on<Trigger<TaskList>, Sync<Director>>().then("Run Task List", [this](const TaskList& list) {  //
```

module/extension/Director/src/Director.cpp
```diff
@@ -213,24 +214,19 @@ namespace module::extension {
...
+            run_task_pack(pack);
```
Bidski, line 229:
```suggestion
            run_task_list(list);
```

module/extension/Director/src/Director.hpp
```diff
@@ -29,6 +31,10 @@
 namespace module::extension {
 
     class Director : public NUClear::Reactor {
+    public:
+        /// A task list is just a list of tasks
```
Bidski, line 35: A redundant statement is redundant

Bidski (changes requested):
I think you need to decide if its a task list or a task pack

module/extension/Director/src/Director.cpp
TrentHouliston: It wouldn't compile when I did that. I did try

module/extension/Director/src/Director.hpp
TrentHouliston: This is true, this statement used to read `a task pack is just a list of tasks`

module/extension/Director/src/Director.cpp
TrentHouliston: I'm trying to separate the concepts of a task pack and a list of tasks.
It just so happens that they are both just a list of tasks.

I want a task pack to be a set of tasks that are all run together.
While a list of tasks is just a list of tasks.
So the queue of tasks that are waiting to run on a provider group are a list of tasks, but the subtasks that it's trying to run are a task pack.

One you can split up and delete parts, the other should always be taken as a whole

module/extension/Director/src/Director.cpp
TrentHouliston: likewise, I want to separate the concept of a task list and a task pack

ysims:
I had a look and couldn't see any issues.

module/extension/Director/src/Director.hpp
LachlanCourt: It also has 3 / 👀

module/extension/Director/src/Director.cpp
CMurtagh-LGTM: c++ cannot work out the type of `{task}` is meant to be. We can either tell it or avoid using the initializer_list.
```suggestion
                    emit(std::make_unique<TaskList>(std::initializer_list({task})));
```
```suggestion
                    emit(std::make_unique<TaskList>(1, task));
```
The way you have it currently is the most readable I recon.

module/extension/Director/src/Director.hpp
CMurtagh-LGTM: > It also has 3 /

@LachlanCourt  A comment that starts with `///` or `/**` is a documenting comment, programs like doxygen will treat this as a brief description.

module/extension/Director/src/Director.hpp
```diff
@@ -29,6 +31,10 @@
...
+        using TaskList = std::vector<std::shared_ptr<const ::extension::behaviour::commands::DirectorTask>>;
```
CMurtagh-LGTM, line 36:
```suggestion
        using TaskList = std::vector<std::shared_ptr<const ::extension::behaviour::commands::DirectorTask>>;
        /// Add the reason why a pack of tasks is different to a list of tasks.
        using TaskPack = std::vector<std::shared_ptr<const ::extension::behaviour::commands::DirectorTask>>;
```
If you want to separate `TaskList` and `TaskPack` either do it thoroughly to avoid confusion.

module/extension/Director/src/Director.hpp
LachlanCourt: Ohhhh I see I see I've only ever seen it as /**

Bidski: Remove this or reword it. This comment really adds nothing

module/extension/Director/src/director/remove_task.cpp
```diff
@@ -0,0 +1,65 @@
...
+        // Get the group for this task
+        auto& group = groups[task->type];
+
+        // Check if this task is the currently active task this provider is running
+        if (task == group.active_task) {
+
+            // Remove this task, we are no longer doing it
+            group.active_task = nullptr;
+
+            // Re-evaluate the queue of tasks for this group
```
Bidski, line 38:
```suggestion
            // Re-evaluate the queue of tasks for this group
            // This may set `group.active_task` to a valid value
```

module/extension/Director/src/provider/ProviderGroup.hpp
```diff
@@ -54,8 +54,8 @@ namespace module::extension::provider {
         std::shared_ptr<const ::extension::behaviour::commands::DirectorTask> active_task;
         /// The queue of tasks waiting to run if the situation changes
         std::vector<std::shared_ptr<const ::extension::behaviour::commands::DirectorTask>> task_queue;
-        /// List of subordinate provider groups
-        std::vector<std::map<std::type_index, ProviderGroup>::iterator> subordinates;
+        /// List of current subtasks that have been emitted by this provider group
+        std::vector<std::shared_ptr<const ::extension::behaviour::commands::DirectorTask>> subtasks;
```
Bidski, line 58:
```suggestion
        TaskList subtasks;
```

Bidski (changes requested):

module/extension/Director/src/director/remove_task.cpp
```diff
@@ -0,0 +1,65 @@
...
+            reevaluate_queue(group);
+
+            // If nothing in the queue updated the active task to a new task we are now idle
+            // That also means we need to remove any subtasks this group had recursively
+            if (group.active_task == nullptr) {
+                group.state = provider::ProviderGroup::IDLE;
+
+                // TODO if there is a pure leaving provider, run it
```
KipHamiltons, line 46: please add @ TrentHouliston to your todos

module/extension/Director/src/Director.hpp
TrentHouliston: I've been trying to document all the variables/types I make so I needed a comment here. I will make a better one

CMurtagh-LGTM (approved):
Looks Good To Me

Bidski (approved):
Cant see anything too offensive

PR #624, opened 2021-07-15
Removes GoalSaver

KipHamiltons:
This module is unused and I don't think we have plans to use it. This PR removes it. 
Relates to #398


ysims (approved):
Sounds good to me. I, for one, would not like to see our robot try to dive 😵 🤕

LachlanCourt (approved):
I feel like this would be fun to see at some point but I think she needs to be a bit more robust first

TrentHouliston (approved):
We've actually developed dives for the Darwin and iGus before that don't cause damage to it.
The best method we found was the banana crumble. @Bidski probably remembers it well enough. Make robot a banana, then crumble (turn of servo torques)

That said I think we are nowhere near the state of play that this is useful, also this module wasn't that great to begin with so I'm in favour of removing it.

Bidski:
Yea, goalie behaviour has never really been a thing and I think any attempt at actually implementing behaviour properly should start from scratch

PR #625, opened 2021-07-15
Removes OdometryLocalisation module

KipHamiltons:
This module is unused and I don't think we have plans to use it. This PR removes it.
Relates to #398


TrentHouliston (commented):
I'd keep this one.

What it's used for is when you're trying to test behaviours etc and localisation is broken (which is not that odd a state)

In that case what this module is supposed to do is just replicate the localisation messages, but make them just be a copy of whatever odometry is saying (so place the field at 0,0)

So it's unused because all the modules have localisation in them. What you would do with this module is when you're trying to test behaviours or similar you'd swap real localisation for this module

ysims:
This just makes me want to make isolated webots module tests even more.

KipHamiltons:
If this has a use then we can keep it. 
(@abihall make sure you include something about this module in the NUbook localisation write-up)

abihall:
I still don't quite understand how you would actually use this module in replacement of 'real' localisation. What would someone testing behaviour have to do to use this instead of actual localisation. It would be great to add some documentation with it. Any guidance would be great

KipHamiltons:
I think the idea is that this acts as a stand-in for localisation, when you want to test the system without it. So it just receives and forwards the odometry outputs as if they were localisation outputs, standing in for localisation.

TrentHouliston:
Yeah, basically instead of doing localisation it uses the odometry system and makes the 0,0 point in the world be the location for the field.
I believe it is also set up so you can push a button on the robot to make wherever the robot is "now" be the centre of the field facing the opponent.

If odometry works properly it should make for a dead reckoning style localisation without needing vision. Can be good for testing short walk path behaviours to walk to various spots on the field when vision/localisation isn't working properly

PR #626, opened 2021-07-15
Remove eSpeak module

KipHamiltons:
This module is unused and I don't think we have plans to use it in the future. This PR removes it, as well as removing the dependencies it required.
Relates to #398


ysims (commented):
I'm not opposed to removing it since we really don't use it right now. But we eventually do want to do something like this with the @home robot.

TrentHouliston (commented):
You should keep this, and make it work.
It was super useful to be able to have the robot say things when it's running on the field and you can't just terminal into it.

Bidski:
> You should keep this, and make it work.
> It was super useful to be able to have the robot say things when it's running on the field and you can't just terminal into it.

While I agree we should keep this, it is currently dependant on hardware that we no longer have

KipHamiltons:
> > You should keep this, and make it work.
> > It was super useful to be able to have the robot say things when it's running on the field and you can't just terminal into it.
> 
> While I agree we should keep this, it is currently dependant on hardware that we no longer have

So you're advocating keeping this @Bidski?

TrentHouliston:
I mean the hardware you no longer have is literally a speaker... It's not like it's some esoteric weird hardware that we will never have again.

Plug something into the NUC's headphone port and boom you have hardware again

We actually had plans to replace that little radiation symbol on the chest with a little speaker already right @tayloryoung6396

Bidski:
> So you're advocating keeping this @Bidski?

I am. Just because a module is not being used, doesn't mean it is useless (or not worth keeping). Just putting it out there that there is a hardware dependency for this module

KipHamiltons:
Roger that. If we think this is worth keeping, that's all good by me. Closing

Issue #622, opened 2021-07-14
Benchmarking Our Code


KipHamiltons:
At the moment, we don't have a way of finding out if our performance improves or changes when we do things. We should devise some way of benchmarking our system.
To benchmark our modules, we'd have to run them doing something. Benchmarking the time it takes to install all of the modules and start the powerplant is one (probably easy to implement) idea I think of immediately, although that's more benchmarking NUClear than our code directly.

KipHamiltons:
This might actually be not as bad using catch [relevant docs](https://github.com/catchorg/Catch2/blob/devel/docs/benchmarks.md). This is certainly worth investigating!

PR #627, opened 2021-07-15
Removes ReactionTimer module

KipHamiltons:
This module is unused and I don't think we have any plans to use it. It seems like it could be useful, perhaps (#622 ?), but we can reimplement it fairly easily if we decide we need to. This PR removes the module.
Relates to #398


TrentHouliston (commented):
This module is unused because we never actually had it in a real role. We would use it for things like timing how long vision was taking by monitoring it's reaction.

I think it's worth keeping, it's more a documentation thing that it's here so it can be used to time reactions than deleting it because this should never be inside a role that you leave permanently on the repo.

Either that or get NUsights reaction view working so you can see this information elsewhere

KipHamiltons:
Roger that. If it's useful, we can keep it. Closing

PR #628, opened 2021-07-15
Removes WalkPathFollower

KipHamiltons:
This module is unused and I don't think we have plans to use it (see discussion on #398). This PR removes it.


TrentHouliston (commented):
I mean, this was the code that let us follow a walk path plan, do we no longer need to do that?

ysims:
> I mean, this was the code that let us follow a walk path plan, do we no longer need to do that?

The walk path planner emits a walk command which is used by the walk engine - we don't use this follower module.

KipHamiltons:
> > I mean, this was the code that let us follow a walk path plan, do we no longer need to do that?
> 
> The walk path planner emits a walk command which is used by the walk engine - we don't use this follower module.

Would the static walk use it?

ysims:
> > > I mean, this was the code that let us follow a walk path plan, do we no longer need to do that?
> > 
> > 
> > The walk path planner emits a walk command which is used by the walk engine - we don't use this follower module.
> 
> Would the static walk use it?

Any walk engine, including the static walk, will use a walk command. 
What Trent is saying sounds more like
walk path planner --(on some message)--> walk follower --(on walk command)--> walk engine
The walk engine doesn't care if the planner or the follower gives it the walk command. It's more about what the planner emits. 

This follower module sounds like something @TheCoolKidsTable was after for his planner idea.

TrentHouliston:
Yeah, this takes a list of waypoints and walks to them in sequence.
You can just make your walk path planner skip this module

KipHamiltons:
What's the consensus then? are we keen to keep this one?

ysims (approved):
Delete 🤖 
It's not used and we plan to rewrite the walk path planner anyway.

PR #630, opened 2021-07-16
Move NetworkForwarder to a more sensible location

TrentHouliston:
For some reason I put NetworkForwarder in output instead of in network.
This location makes much more sense.


KipHamiltons (commented):
Will approve if builds passing :+1:

KipHamiltons (approved):

PR #631, opened 2021-07-17
Add info to robot selector panel

LachlanCourt:
Adds to the robot selector panel to include IP Address and Port

To test, run the NUbots codebase alongside NUsight and click on the "Select Robots" button on the top right, click on the dropdown and see the IP Address and Port of the connection.

test


BrendanAnnable (commented):
Nice one!

Please update the storybook as it currently looks like this:

Screen Shot 2021-07-17 at 10 03 00 pm

Run `yarn storybook` and navigate to http://localhost:9001/?path=/story/components-robot-selector--renders-with-robots

Code to update:

https://github.com/NUbots/NUbots/blob/6990ca84445f6aa5abd9117d6a090b67a1eecb5a/nusight2/src/client/components/robot_selector/stories/robot_selector.stories.tsx#L45-L72

LachlanCourt:
Nice, I didn't even know that existed! I just added some random addresses and ports in there

BrendanAnnable (approved):
LGTM

nusight2/src/client/components/robot_selector/robot_label/style.css
```diff
@@ -61,7 +61,7 @@
 
 .dataTable {
   display: grid;
-  grid-template-columns: 1fr 1fr;
+  grid-template-columns: 220px 180px;
```
JosephusPaye, line 64:
```suggestion
  grid-template-columns: 180px 180px;
```

Pairs with the suggestion below.

nusight2/src/client/components/robot_selector/style.css
```diff
@@ -69,5 +69,5 @@
     0px 0px 3px 0px rgba(0, 0, 0, 0.15),
     0px 3px 8px 0px rgba(0, 0, 0, 0.15);
   overflow: hidden;
-  width: 100%;
+  width: 400px;
```
JosephusPaye, line 72: We could get away with making this a little more compact.

<details>
<summary>Compare before/after</summary>

Screenshot of selector panel before and after changes

</details>

```suggestion
  width: 330px;
```

LachlanCourt: @ysims wanted a bit of a bigger gap hence the change but I can do this instead

JosephusPaye: The template areas can be bigger than the parent size, and will be drawn as if the parent was that large.

In this case it's better to do `180px 1fr`, to make the first column 180px and have the second fill the remaining space.

LachlanCourt: Gotcha okay

JosephusPaye (dismissed):
LGTM

nusight2/src/client/components/robot_selector/stories/robot_selector.stories.tsx
```diff
@@ -49,24 +49,24 @@ function getRobots(): RobotModel[] {
       name: 'Virtual Robot 1',
       connected: true,
       enabled: true,
-      address: '',
-      port: 0,
+      address: '10.222.10.14',
```
KipHamiltons, line 52: I don't understand this change - could you please give context? It seems to me that these are arbitrary addresses that are strange to commit, but I have no idea what's happening here to start with. Is this "story" an example?

nusight2/src/client/components/robot_selector/style.css
```diff
@@ -69,5 +69,5 @@
     0px 0px 3px 0px rgba(0, 0, 0, 0.15),
     0px 3px 8px 0px rgba(0, 0, 0, 0.15);
   overflow: hidden;
-  width: 100%;
+  width: 330px;
```
KipHamiltons, line 72: It seems strange to me to select an absolute width. Will this break scaling on some monitors?

.github/dependabot.yml
```diff
@@ -5,11 +5,6 @@ updates:
   # Package ecosystem for yarn is npm
   - package-ecosystem: npm
     directory: /nusight2
-    commit-message:
-      # include: "scope" tells it to add the names of the
-      # upgraded deps to the name, after the prefix
-      include: "scope"
```
KipHamiltons, line 11: I think you should keep `commit-message` and `include`, scrapping `prefix`, which was the problem field.

KipHamiltons (commented):
A couple of questions

.github/dependabot.yml
LachlanCourt: I removed `prefix` and it broke because of some other issue

nusight2/src/client/components/robot_selector/stories/robot_selector.stories.tsx
LachlanCourt: [Storybook](https://storybook.js.org/docs/react/get-started/introduction) is a UI development tool it allows you to test small components of NUsight and see what they look like without having to implement everything into NUsight in one go. These are just some test cases, some virtual robots, that let you see what it might look like with active connections. As such I just picked some random ip addresses and ports. You can check it out by running `./b yarn storybook`

nusight2/src/client/components/robot_selector/style.css
LachlanCourt: Potentially, although I think most monitors that you will be viewing NUsight on will be larger than 330px. The issue is a total width of 100% will automatically wrap to the smallest it can be, whereas adding this width absolutely means that it will render with some padding that will look more readable

LachlanCourt: This panel overlays over the existing display of NUsight so it will not be pushing any other elements out of the way

nusight2/src/client/components/robot_selector/stories/robot_selector.stories.tsx
KipHamiltons: Thanks for the explanation. That makes sense

.github/dependabot.yml
KipHamiltons: what was the other issue?

nusight2/src/client/components/robot_selector/style.css
KipHamiltons: seems reasonable enough

KipHamiltons (approved):
I am approving on the basis that it was approved and you only changed dependabot configs since then.

PR #633, opened 2021-07-18
Compile unused modules as part of pipeline

KipHamiltons:
This PR adds a `./b unused` option to generate a role (which is git ignored) which contains all of the unused modules. This role is then created and built as part of the standard build pipeline. This means that changes which break unused modules will break the buildkite builds. This is good.


CMurtagh-LGTM (approved):
There's some unrelated changes, but I trust those make the unused modules build.

CMurtagh-LGTM:
Rather than making the fs writable, would it be possible to instead write the unused role inside the build folder?

KipHamiltons:
@CMurtagh-LGTM Interesting idea. It would need a fair bit more work, because I would probably have to generate the actual C++ role manually, but I don't see anything which would stop me. I'll give it a go.

CMurtagh-LGTM:
> @CMurtagh-LGTM Interesting idea. It would need a fair bit more work, because I would probably have to generate the actual C++ role manually, but I don't see anything which would stop me. I'll give it a go.

Is it not possible to add an extra folder to look for role files?

KipHamiltons:
It is, with a bit more work than just adding the string though.
But, I was just playing around with this idea trying to make it work and I don't really like it because the role file becomes hidden in the build directory, so you won't be able to see it or access it without `./b shell`, which is bad. People will need to use `./b unused -g` to generate the role if their build is failing, so they should be able to see it and know if it's being built or not easily, without having to know about the file obscured in the build directory.

CMurtagh-LGTM:
> It is possible, with a bit more work than just adding the string though.
> But, I was just playing around with this idea trying to make it work and I don't really like it because the role file becomes hidden in the build directory, so you won't be able to see it or access it without `./b shell`, which is bad. People will need to use `./b unused -g` to generate the role if their build is failing, so they should be able to see it and know if it's being built or not easily, without having to know about the file obscured in the build directory.

:shrug: Maybe add a cmake option to build unused modules, then they can clearly see what is going on when they run `./b configure -i`.

KipHamiltons:
I think what I have just done is pretty clean. All that changes is the fs is not read-only and the roles directory is writeable by anyone, and then it'll be like a normal (gitignored) role for anyone who wants to generate and use it, including having an option to enable/disable it with the rest of the roles in `./b configure -i`

CMurtagh-LGTM:
> I think what I have just done is pretty clean. All that changes is the fs is not read-only and the roles directory is writeable by anyone, and then it'll be like a normal (gitignored) role for anyone who wants to generate and use it, including having an option to enable/disable it with the rest of the roles in `./b configure -i`

I'm thinking that we should put it in the build folder when we do CI and let people do it your way if they want to do it manually.

KipHamiltons:
Is there a problem doing it this way? Having two separate spots to look for role files increases the complexity of `NUClearRole.cmake` a lot.

Y0sh1dk:
I think It would be nice to test the unused modules as a separate step in the pipline instead of testing It all together. Then atleast if the unsed doesnt pass we know it won't break anything in use.

.buildkite/pipeline.yml
```diff
@@ -52,7 +52,15 @@ steps:
 
   # Build generic target and test
   - label: "Build generic and Test"
-    command: "./b configure -- -DBUILD_TESTS=ON && ./b build && ./b tests run"
+    # We have to configure twice. The first configuration generates the cache for `./b unused`, then the next one
```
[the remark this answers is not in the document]
KipHamiltons: Cmake cache. It's stored as a file in the build directory. That's how the `./b configure` options are persistent - they're stored in that file.

.buildkite/NUbots/docker-compose.yml
```diff
@@ -12,7 +12,8 @@ services:
       - type: bind
         source: ./../..
         target: /home/nubots/NUbots
-        read_only: true
+        # False, so we can generate a role with unused modules and write it
+        read_only: false
```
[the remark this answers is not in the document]
KipHamiltons: I am not sure. I would _probably_ err on the side of keeping it, given that we specify `true` on the others. Idk. Are you thinking I should remove it?

.buildkite/pipeline.yml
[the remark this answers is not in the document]
KipHamiltons: I clarified it in the comment too :+1:

KipHamiltons:
@Y0sh1dk Great idea. I have made it its own pipeline step :+1:

.buildkite/NUbots/docker-compose.yml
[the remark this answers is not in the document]
Y0sh1dk: It definitely defaults to false if you dont specify so we can just remove it, but I guess its good to keep it to document why we need to write to the mount.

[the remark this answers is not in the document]
KipHamiltons: There you go. 

(If it's any consolation @CMurtagh-LGTM, the fs is not read-only by default)

tools/unused.py
```diff
@@ -80,3 +91,12 @@ def run(**kwargs):
     cprint("Unused Modules", "red", attrs=["bold"])
     cprint("\n".join(unused_modules), "red", attrs=["bold"])
     print("\n")
```
Bidski, line 93: Only print if we aren't generating?

tools/unused.py
```diff
@@ -80,3 +91,12 @@ def run(**kwargs):
...
+
+    # If we want to make a role with the unused modules to check if they compile, we do that now
+    if generate_role:
+        with open(os.path.join(roles_path, UNUSED_ROLE_NAME), "w") as unused_role:
+            unused_role.write("nuclear_role(\n  ")
+            unused_role.write("\n  ".join(unused_modules))
+            unused_role.write("\n)\n")
+        # Warn the user that the new role won't build unless they reconfigure
+        cprint("You must run ./b configure after generating a new role", "red", attrs=["bold"])
```
Bidski, line 102:
```suggestion
        cprint("You need to run ./b configure now", "red", attrs=["bold"])
```

Bidski (changes requested):

TrentHouliston:
Rather than doing this in python, why not just do it in CMake, in CMake you could just list all the folders in the modules dir, and then call the nuclear_role function. No need for generating anything

KipHamiltons:
I have retooled this PR so that it uses cmake to call `nuclear_role(${ALL_MODULES})`. This did end up cleaner - thanks for the suggestion @TrentHouliston. 

All the unrelated changes were necessary to make the role compile. The change to the `RawSensors` enum value name was a strange case where it seems that something had defined a macro or preprocessor definition for `OK`. I renamed that to `EVERYTHING_OK`, which resolved the error.

CMurtagh-LGTM (approved):
LGTM

cmake/Modules/FindeSpeak.cmake
```diff
@@ -3,13 +3,12 @@ include(ToolchainLibraryFinder)
 # Find ALSA for eSpeak
 find_package(ALSA REQUIRED)
 
-# Find PortAudio for eSpeak
-find_package(PortAudio REQUIRED)
+find_package(pcaudio REQUIRED)
```
Bidski, line 6: How is this used?

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -358,16 +358,16 @@ namespace module::input {
 
 
                         // This checks for an error on the platform and reports it
-                        if (input.platform_error_flags != RawSensors::Error::OK) {
+                        if (input.platform_error_flags != RawSensors::Error::EVERYTHING_OK) {
```
Bidski, line 361: Good god! Why?

Bidski (changes requested):

cmake/Modules/FindeSpeak.cmake
KipHamiltons: Well spotted. Will make a PR to remove that.

KipHamiltons: #720

module/input/SensorFilter/src/SensorFilter.cpp
KipHamiltons:
```
In file included from /home/nubots/NUbots/module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.hpp:25,
                 from /home/nubots/build/nuclear/roles/allmodules.cpp:14:
/home/nubots/build/nuclear/message/protobuf/message/platform/RawSensors.pb.h:1903:26: error: expected unqualified-id before numeric constant
 1903 |   static constexpr Error OK =
      |                          ^~
```
Something (outside of our codebase) has defined `OK`. I think it's ncurses, because [here](https://github.com/NUbots/NUbots/blob/main/module/behaviour/tools/ScriptTuner/src/ScriptTuner.cpp#L24) we undefine it. In keyboardwalk we [include it](https://github.com/NUbots/NUbots/blob/main/module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.hpp#L25) without undefining it though, and I think that's _probably_ the clash.

Bidski: find the offending header file and then `#undef OK` rather than changing our enum values as changing our enum values doesnt actually solve the problem

cmake/Modules/FindeSpeak.cmake
Bidski: But you added `pcaudio` to replace `PortAudio` in this PR? Why make another PR that removes `PortAudio` and doesn't address the question of how `pcaudio` is used?

Bidski: it would also appear that pcaudio is a dependency of espeawould also seem that pcaudio is a dependency of espeak, so removing it seems like a bad idea

KipHamiltons: `pcaudio` which replaces `PortAudio` isn't actually used. They refer to the same functionality/are meant to do the same thing, which we apparently don't need

module/input/SensorFilter/src/SensorFilter.cpp
KipHamiltons: sure

cmake/Modules/FindeSpeak.cmake
Bidski: we use it transitively via espeak, so I think we should still be building it we just dont need to `find_package` it (unless we want the sanity check that it is part of our toolchain)

KipHamiltons: Trueeeee. [looks like you should have it hahaha](https://github.com/NUbots/NUbots/pull/691#discussion_r713786179). Will close that PR. 


On a related note, should we be linking `pcaudio` explicitly with `target_link_libraries`?

Bidski: > On a related note, should we be linking pcaudio explicitly with target_link_libraries?

It shouldn't be necessary. Linking against `espeak-ng` should cause any necessary symbols to be loaded from `pcaudio` (assuming there are any at link time). The more likely behaviour will be the loading of `pcaudio` at runtime when we load `espeak-ng`. The only reason we would want to link against `pcaudio` explicitly is as a "sanity check" on its existence (which is the only role that `find_package` is playing in this instance). I think we should just remove the `find_package` line and the find module, but keep it in the `Dockerfile`

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.hpp
```diff
@@ -23,6 +23,7 @@
 #include <Eigen/Core>
 #include <mutex>
 #include <ncurses.h>
+#undef OK  // ncurses defines OK. We don't need it
```
Bidski, line 26: Should wrap this in `// clang-format off` just to maie sure clang-format never does something weird with this

Bidski (changes requested):

module/behaviour/strategy/KeyboardWalk/src/KeyboardWalk.hpp
KipHamiltons: I've moved it above the undef - this seemed simpler than adding the boilerplate for the formatting

PR #634, opened 2021-07-19
Removes gyro stuff from hardware simulator

KipHamiltons:
At the moment, hardware simulator has a mechanism for handling gyroscope data using a queue of gyro readings. However, (I am reasonably sure) that nothing emits gyro readings. 
If they were emitted, we think the code would break because as @Bidski  pointed out to me, if it was executed, it would attempt to acquire the same mutex twice in a row.

This PR removes that stuff which handles gyroscope readings in hardware simulator. We might want to fix it instead, in which case I am happy to close this PR, but I figure it might as well go if we used it this long without that function anyway lol

### DIFF
--- a/module/platform/darwin/HardwareSimulator/src/HardwareSimulator.cpp
+++ b/module/platform/darwin/HardwareSimulator/src/HardwareSimulator.cpp
@@ -21,7 +21,6 @@
 
 #include <Eigen/Core>
 #include <Eigen/Geometry>
-#include <iostream>
 #include <limits>
 #include <mutex>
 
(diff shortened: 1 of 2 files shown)


module/platform/darwin/HardwareSimulator/src/HardwareSimulator.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include <Eigen/Core>
 #include <Eigen/Geometry>
+#include <iostream>
```
ysims, line 24: ?

KipHamiltons: ahh yes, the include for the classic print statements I used to validate my findings :joy:

ysims:
Doesn't make sense to me that it was there in the first place. Why would it be getting gyro data from somewhere else anyway. HardwareSimulator is meant to be the 'robot' and gyro readings should come from HardwareSimulator. Maybe it's part of an old design for the module that somehow got left in 🤷

TrentHouliston (approved):
Yeah this was super old code that was used to make noisy gyro measurements somehow.

PR #635, opened 2021-07-19
Gives Walk Engine logs punctuation for readability

KipHamiltons:
I found these logs more confusing than they needed to be. This PR punctuates them, making them clearer IMO


ysims (approved):
lgtm

PR #637, opened 2021-07-19
Refactors vector style messages into vectors

KipHamiltons:
`RawSensors::{Accelerometer, Gyroscope}` are both message types wrapping vectors. Now that we have nice vector types in proto + neutrons, these are unnecessary. 
This PR removes those message types and changes the code where they were used.


CMurtagh-LGTM (approved):
:rofl:

PR #638, opened 2021-07-19
Fix width of single robot selector

LachlanCourt:
The image was cut off in odometry and vision view if there were no robots connected. It bothered me. This fixes it :)

To test, run either `./b yarn storybook` and navigate to robot selector single and look at the different options, or run `./b yarn build && ./b yarn prod` and don't start NUbots alongside to see without any robots connected

I am aware there is a typo in the branch name don't @ me :(

image


JosephusPaye (approved):
LGTM

PR #641, opened 2021-07-20
Fixes dependabot config by adding required field

KipHamiltons:
I screwed up the config. #631 is having issues because of a missing field, so I added it in this PR


LachlanCourt (approved):
Iconic

PR #647, opened 2021-07-21
Add debug statments to nugus_sim image entrypoint

Y0sh1dk:
Adds some debugging statements to nugus_sim entrypoint, specifically for finding out if our binary crashed.

### DIFF
--- a/docker/nugus_sim/run.py
+++ b/docker/nugus_sim/run.py
@@ -143,10 +143,13 @@ def run_role(role: str, binaries_dir: str, env_vars: dict) -> None:
 
     # Run the role binary
     while True:
-        subprocess.run(f"./{role}", env=modified_env)
+        print(f"Binary '{role}' starting...")  # For debugging
+        exit_code = subprocess.run(f"./{role}", env=modified_env).returncode
+        print(f"Binary '{role}' crashed! Exit code: {exit_code}")  # For debugging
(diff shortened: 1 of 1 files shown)


docker/nugus_sim/run.py
```diff
@@ -143,10 +143,13 @@ def run_role(role: str, binaries_dir: str, env_vars: dict) -> None:
 
     # Run the role binary
     while True:
-        subprocess.run(f"./{role}", env=modified_env)
+        print(f"Binary '{role}' starting...")  # For debugging
+        exit_code = subprocess.run(f"./{role}", env=modified_env).returncode
+        print(f"Binary '{role}' crashed! Exit code: {exit_code}")  # For debugging
```
LachlanCourt, line 148: If this exits with a code of 0 will it still run this line saying that it crashed? Maybe reword as it may be a valid exit code?

KipHamiltons: When you run roles, the only way they stop is if they're killed or they crash (unless you're doing some super funky stuff triggering shutdown yourself, which we don't do). Basically they won't ever `exit(0)`

LachlanCourt: Sounds good

LachlanCourt (approved):
LGTM

PR #652, opened 2021-07-22
Adds flags to enable or disable a set of roles, doing the opposite to the other roles

### LINKED ISSUE #651: Flag for configuring only a set of roles
If you want to build only one role atm it's a bit painful, because you have to go through the full list in `./b configure -i` spamming space bar. Selecting a subset of the roles and configuring using those only is fairly straightforward (as demonstrated [here](https://github.com/NUbots/NUbots/blob/d89c7a2e16b6005a1663e4ac70c063de7a732966/tools/webots.py#L65) by an impressively neat solution from @JosephusPaye). Making this a proper thing for `./b configure` would be nice.

KipHamiltons:
This PR resolves #651 - basically I am tired of running `./b configure -i` and spamming space to disable all roles except one.
The PR does that by using the role selecting capability @JosephusPaye wrote for `./b webots` - I moved that function into `b` itself.

This PR might be contentious because I moved the function into `b`. I am willing to budge on that of course @TrentHouliston; it just seemed like the most reasonable solution for a `b` tool utility function.

### Reviewers, note

To test, it's `./b configure -r <roles> # enables <roles>` or `./b configure -R <roles> # disables <roles>`, where `<roles>` is a colon-separated list of role names. You can also do `./b configure -r all` to enable all roles.

### DIFF
--- a/nuclear/b.py
+++ b/nuclear/b.py
@@ -1,9 +1,12 @@
 #!/usr/bin/env python3
+
 import argparse
+import glob
 import os
 import pkgutil
 import re
 import sys
+from pathlib import Path
 
 # Don't make .pyc files
 sys.dont_write_bytecode = True
@@ -69,6 +72,28 @@ try:
     source_dir = cmake_cache[cmake_cache["CMAKE_PROJECT_NAME"] + "_SOURCE_DIR"]
 except:
     source_dir = project_dir
+
+
+def get_cmake_role_flags(roles_to_build):
+    """Utility function to get list of CMake flags corresponding to the roles passed in"""
+
+    roles_dir = os.path.join(project_dir, "roles")
+    roles_glob = os.path.join(roles_dir, "*.role")
+
+    available_roles = [Path(role_path).stem for role_path in glob.glob(roles_glob)]
+
+    # Ensure that all the roles requested are available
+    for role in roles_to_build:
+        if role not in available_roles:
+            print(f"role '{role}' not found", color="red", attrs=["bold"])
+            sys.exit(1)
+
+    role_flags = [f"-DROLE_{role}=ON" for role in available_roles if role in roles_to_build] + [
+        f"-DROLE_{role}=OFF" for role in available_roles if role not in roles_to_build
+    ]
+
+    return role_flags
+
 
 if __name__ == "__main__":
 
--- a/nuclear/b.py
+++ b/nuclear/b.py
@@ -74,8 +74,14 @@ except:
     source_dir = project_dir
 
 
-def get_cmake_role_flags(roles_to_build):
-    """Utility function to get list of CMake flags corresponding to the roles passed in"""
+def get_cmake_role_flags(roles, disable_instead=False):
+    """
+    @brief Utility function to get list of CMake flags corresponding to the roles passed in.
+    If disable_instead == True, then those roles are the only ones not enabled in the
+    resulting flag list
+    @param roles The roles to enable (or disable, if disable_instead == True)
+    @param disable_instead Indicates whether the roles passed in should be enabled or disabled
+    """
 
     roles_dir = os.path.join(project_dir, "roles")
     roles_glob = os.path.join(roles_dir, "*.role")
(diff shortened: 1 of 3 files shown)


TrentHouliston:
I am still of the opinion this should be done in CMake and you do `./b configure -DSOMETHINGHERE=ROLE_X

KipHamiltons:
@TrentHouliston what's the advantage to keeping that within cmake as opposed to in a `b` script?

nuclear/b.py
```diff
@@ -70,6 +73,39 @@
...
+    # If the roles passed in are meant to be disabled instead of enabled then
+    # we take the complement of our roles list, in the universe of available_roles
+    if disable_instead:
+        roles = [role for role in available_roles if role not in roles]
+
+    role_flags = [f"-DROLE_{role}=ON" for role in available_roles if role in roles] + [
+        f"-DROLE_{role}=OFF" for role in available_roles if role not in roles
+    ]
+
+    return role_flags
```
Bidski, line 106: Maybe something like this?

```suggestion
def get_cmake_role_flags(roles, disable_roles=False):
    """
    @brief Utility function to get list of CMake flags corresponding to the roles passed in.
    If disable_roles == True, then those roles are the only ones not enabled in the
    resulting flag list
    @param roles The roles to enable (or disable, if disable_roles == True)
    @param disable_roles Indicates whether the roles passed in should be enabled or disabled
    """

    roles_dir = os.path.join(project_dir, "roles")
    roles_glob = os.path.join(roles_dir, "*.role")

    available_roles = set([Path(role_path).stem for role_path in glob.glob(roles_glob)])

    # Ensure that all the roles requested are available
    failure = False
    for role in roles:
        if role not in available_roles:
            print(f"role '{role}' not found")
            failure = True
    if failure:
        sys.exit(1)
        
    # Deal with the roles that were passed in first
    toggle = "OFF" if disable_roles else "ON"
    roles = set(roles)
    role_flags = [f"-DROLE_{role}={toggle}" for role in roles]

    # Now deal with the roles that weren't passed in
    toggle = "ON" if disable_roles else "OFF"
    available_roles = available_roles - roles
    role_flags.extend([f"-DROLE_{role}={toggle}" for role in available_roles])

    return role_flags
```

tools/configure.py
```diff
@@ -20,21 +22,63 @@ def register(command):
...
+
+    command.add_argument(
+        "-R",
+        "--disable-roles",
+        action="store",
+        dest="disabled_roles",
+        default=None,
+        help="Colon-separated list of roles to disable. If '-R' is set, '-r' is ignored",
+    )
+
```
Bidski, line 43: Add a command to list all known roles to make it easier to list enabled/disabled roles
```suggestion
    command.add_argument("-l", "--list-roles", action="store_true", dest="list_roles", default=False, help="List all known roles in the system")
```

tools/configure.py
```diff
@@ -20,21 +22,63 @@ def register(command):
...
+    # Change into the build directory
     os.chdir(os.path.join(b.project_dir, "..", "build"))
 
+    # If specific roles are disabled, then we enable all of the others
+    if disabled_roles:
+        disabled_roles = disabled_roles.strip()
+        # roles are a colon separated list, so we split on that character
+        disabled_roles = disabled_roles.split(":")
+        cprint("Disabling all roles except these: ", end=" ", color="red", attrs=["bold"])
+        cprint(str(disabled_roles), color="green", attrs=["bold"])
```
Bidski, line 60: `from termcolor import colored`
```suggestion
        desc_str = colored("Disabling all roles except these:", color="red", attrs=["bold"])
        roles_str = colored(", ".join([role for role in disabled_roles]), color="green", attrs=["bold"])
        print(f"{desc_str} {roles_str}")
```

Bidski (changes requested):

TrentHouliston:
> @TrentHouliston what's the advantage to keeping that within cmake as opposed to in a `b` script?

Cohesion, so that the code that configures which roles to build is all in the same place, not split into two places

ysims:
Superceded by #1031

PR #654, opened 2021-07-24
Adds jobs flag to the webots b tool

Crafty15:
Adds a jobs flag for building webots image.
For example, run: `./b webots build <role> -j <jobs>`, where `<role>` is a role name and `<jobs>` is the desired number of jobs to build in parallel.
Edit: Removed passing ninja a default parallel job count in favour of letting it choose it's own, which is building cores + 2.
Also added a build args list to make it easier to add more args later.


KipHamiltons:
Now that I am reading this, I am thinking maybe instead of defaulting to `cpu_count()` for `-j`, we should not pass anything, and let ninja pick the default for us (which I learnt recently was cpus+2). What do you think?

tools/webots.py
```diff
@@ -40,6 +41,10 @@ def register(command):
         default=True,
         help="Disables cleaning the build volume before compiling",
     )
+    #
+    build_subcommand.add_argument(
+        "-j", "--jobs", dest="jobs", action="store", help="Dictates the number of jobs to run in parrallel"
```
moosahassanx, line 46: Incorrect spelling of "parrallel" -> should be "parallel" :p

tools/webots.py
```diff
@@ -40,6 +41,10 @@ def register(command):
         default=True,
         help="Disables cleaning the build volume before compiling",
     )
+    #
```
moosahassanx, line 44: I think you might have forgotten to finish this comment - or maybe it was a WIP until the branch is finished 🤷🏿‍♀️

Crafty15:
> Now that I am reading this, I am thinking maybe instead of defaulting to `cpu_count()` for `-j`, we should not pass anything, and let ninja pick the default for us (which I learnt recently was cpus+2). What do you think?
> 
> Edit: also, please edit the title of the pr

Title edited, is that ok?
As for the default value, would that work better? Also, could you direct me to the appropriate Ninja docs so i can have a read plz?

KipHamiltons:
@Crafty15 Title should be something like "Adds jobs flag to webots `b` tool" or something to that effect. This serves two purposes: it tells people exactly what the PR does, and secondly, it becomes the commit message of all the changes when they're squashed into 1 commit when you "squash and merge". That means that when people use `git blame` or `git blame`-using tools they can see what the commit which changed the line was trying to do easily.

[ninja docs](https://ninja-build.org/manual.html) are pretty bare-bones. [Here's a source](https://blog.kitware.com/cmake-building-with-all-your-cores/) for the `cores + 2` assertion.

I think it makes more sense to let ninja do its thing if unspecified. Otherwise, for systems that don't run out of ram, this would  slow builds down slightly, because they would be reducing the number of jobs by 2 (I say only slightly, because more jobs than cpu threads has limited perf increases afaik).

Crafty15:
@KipHamiltons That was a good read. I'm convinced.

TrentHouliston:
More generally when it comes to commit messages in git you should write them as "what will happen if I hit the merge button"

So for your title (and therefore commit message) complete this sentence.

If you hit the merge button this will <TITLE_HERE>.

KipHamiltons:
iirc @JosephusPaye is an advocate of present tense commit messages like this, because it is the complete sentence of what the commit is doing, without the need for an implied sentence

TrentHouliston:
> iirc @JosephusPaye is an advocate of present tense commit messages like this, because it is the complete sentence of what the commit is doing, without the need for an implied sentence

Doesn't matter so much to me so long as you pick a standard.
For me, I prefer the other way because most of the time when I am looking at commit history I want to know what a specific commit does. If I were to cherry-pick this commit what would it do.

I think of each commit as an action that's applied to the codebase

JosephusPaye:
I'm for present tense, singular verb, like I added to @KipHamiltons's PR in NUbots/NUbook#54:

image

In addition to completing the sentence, "When applied, this commit will ...", it's also the default for comments generated by Git and GitHub. E.g. when you merge a branch, you get "Merge branch ..." instead of "Merges branch ..." or "Merged branch ..."

tools/webots.py
```diff
@@ -41,6 +42,10 @@ def register(command):
         help="Disables cleaning the build volume before compiling",
     )
 
+    build_subcommand.add_argument(
+        "-j", "--jobs", dest="jobs", action="store", help="Dictates the number of jobs to run in parallel"
```
Y0sh1dk, line 46: https://blog.kitware.com/cmake-building-with-all-your-cores/
```suggestion
        "-j", "--jobs", dest="jobs", action="store", help=f"Dictates the number of jobs to run in parallel ({multiprocessing.cpu_count() + 2} by default"
```

tools/webots.py
```diff
@@ -100,7 +105,11 @@ def exec_build(target, roles, clean):
 
     # Compiles code for correct target
     print("Building code...")
-    exit_code = subprocess.run(["./b", "build"]).returncode
+    build_command = ["./b", "build"]
+    # Check if a -j value has been given
+    if jobs != None:
```
Y0sh1dk, line 110: Not a big deal, but just to be consistent with the rest of our codebase.
```suggestion
    if jobs:
```

tools/webots.py
KipHamiltons: @Y0sh1dk I like this suggestion. It does make me consider if we should go further though, and tell people how we get the number, so that people can be sure it's different on different machines and what it will be. Basically, mention that it's number of threads plus 2, so we don't have the ambiguity that `ninja --help` introduces.

tools/webots.py
```diff
@@ -41,6 +42,10 @@ def register(command):
         help="Disables cleaning the build volume before compiling",
     )
 
+    build_subcommand.add_argument(
+        "-j", "--jobs", dest="jobs", action="store", help=f"Dictates the number of jobs to run in parallel ({multiprocessing.cpu_count() + 2} by default"
+    )
```
Bidski, line 47:
```suggestion
    build_subcommand.add_argument(
        "-j", "--jobs", dest="jobs", action="store", default=multiprocessing.cpu_count() + 2, help=f"Dictates the number of jobs to run in parallel ({multiprocessing.cpu_count() + 2} by default"
    )
```

tools/webots.py
```diff
@@ -100,7 +105,11 @@ def exec_build(target, roles, clean):
 
     # Compiles code for correct target
     print("Building code...")
-    exit_code = subprocess.run(["./b", "build"]).returncode
+    build_command = ["./b", "build"]
+    # Check if a -j value has been given
+    if jobs:
+        build_command.extend(["-j", jobs])
```
Bidski, line 111:
```suggestion
    build_command = ["./b", "build", "-j", jobs]
```

Bidski (changes requested):

tools/webots.py
KipHamiltons: We had it like this before, but it seemed more proper to me to just let ninja handle it in the event that we don't supply a flag. What's your thinking behind specifying it every time?

tools/webots.py
Crafty15: We had a discussion about this at the top of the page. We left out the default value because ninja chooses it's own default, which is CPU count + 2. 
Is there another reason we should be passing it in?

tools/webots.py
Crafty15: The idea here was to keep things clean if we need to add more flags, rather than having to write out the build command each time, just extend each time we add another.
I think @KipHamiltons could explain it a bit better.

KipHamiltons: This suggestion leaves us open to that. It'll have to be done differently if there were flags we wanted to add which didn't have sensible defaults though - they would have to be added separately with an if statement (inline `if`s are not ok, so they don't count). However, I am not sure if that case is relevant/would ever happen.

Crafty15: > 
> 
> We had it like this before, but it seemed more proper to me to just let ninja handle it in the event that we don't supply a flag. What's your thinking behind specifying it every time?

Is this comment meant for the other discussion?

tools/webots.py
```diff
@@ -41,6 +42,14 @@ def register(command):
         help="Disables cleaning the build volume before compiling",
     )
 
+    build_subcommand.add_argument(
+        "-j",
+        "--jobs",
+        dest="jobs",
+        action="store",
+        help=f"Dictates the number of jobs to run in parallel",
```
KipHamiltons, line 50:
```suggestion
        help="Dictates the number of jobs to run in parallel",
```

KipHamiltons (commented):
Have you tested the other sub commands besides `build`? Have you tested without jobs set?

The former question is important, because we were having issues when we ran other sub commands which didn't have a given flag, if the flag didn't have a default in `run`. That's why all of the other flags have default values set in the function signature - so that when other subcommands are run it doesn't throw a fit because you didn't pass an argument

KipHamiltons (approved):
LGTM

ysims (approved):
This helped me when I was fixing the robot freezing stuff last night. No crashes! 😄 Thanks for getting this done 🎉

PR #663, opened 2021-08-08
Matching User Input to Available Tests

moosahassanx:
Running scripts such as `./b run keyboardWalkFake` can be typed as `./b run keyboardwalkfake`.


shared/tests/CMakeLists.txt
```diff
@@ -2,15 +2,15 @@
 file(GLOB_RECURSE test_src "*/**.cpp" "*/**.hpp")
 
 if(test_src)
-  add_executable(TestShared "${CMAKE_SOURCE_DIR}/${NUCLEAR_SHARED_DIR}/tests/test.cpp" ${test_src})
-  target_link_libraries(TestShared PRIVATE nuclear::message nuclear::utility)
+  add_executable(testshared "${CMAKE_SOURCE_DIR}/${NUCLEAR_SHARED_DIR}/tests/test.cpp" ${test_src})
+  target_link_libraries(testshared PRIVATE nuclear::message nuclear::utility)
```
ysims, line 6: What do all these changes to lowercase `testshared` achieve?

KipHamiltons: a binary named `testshared`. The advantage of having lower case names is that we can `to_lower()` string inputs people give to run specific tests, which means they can be case insensitive.

KipHamiltons:
I presume that you're going to have a follow up PR with the change to the `tests` tool, so that it converts those strings to lower case?

TrentHouliston (changes requested):
Why do we want to make everything lowercase? This isn't so much making it case insensitive as it is making it lowercase

KipHamiltons:
@TrentHouliston this is the simplest solution we could think of for making it case insensitive. When you're running specific tests with `ctest -R <regex>` as we are, you have to have the case correct, or you have to compensate for that in the regex. The tests are named `Test<lowercasemetamodule><PascalCaseModule>` which isn't widely known or understood, so by making them all lower case, we can transform the inputs with a blanket rule and have easy use.

PR #666, opened 2021-08-13
Remove subsumption.proto

ysims:
This isn't used, so I've removed it.


TrentHouliston (approved):
I understand the issue. id is a special name for recordings. However I feel like there should be some code changes that go along with this (my guess is it won't compile)

shared/message/behaviour/Subsumption.proto
```diff
@@ -29,9 +29,9 @@ message Subsumption {
     }
 
     message ActionRegister {
```
KipHamiltons, line 31: This message is unused. I say we remove it altogether

KipHamiltons:
> I understand the issue. id is a special name for recordings. However I feel like there should be some code changes that go along with this (my guess is it won't compile)

You'd think so, but... literally every time we use `ActionPriorities` we construct it with a pair of magic numbers, not naming the field directly.

ysims:
...I think this whole file is unused... what we thought this does it actually done by [Action.hpp](https://github.com/NUbots/NUbots/blob/master/shared/utility/behaviour/Action.hpp) (from what I can tell).

ysims:
Or maybe it is. The build failed. Not sure why mine was working locally. Will look at it later.

KipHamiltons:
The file is included in simplewalkpathplanner,but it's not used. Removing the include should fix that error

ysims:
`TestextensionFileWatcher` segfaulted... why :cry:

KipHamiltons:
@ysims  the rebuild/retest worked. I posit it was some fs or kernel intrusion that made it fault. Could be worth looking into though. (@CMurtagh-LGTM see above comment)

KipHamiltons (approved):
LGTM

CMurtagh-LGTM:
> `TestextensionFileWatcher` segfaulted... why

@KipHamiltons What tests would be run here?

KipHamiltons:
@CMurtagh-LGTM how do you mean? All tests are ran. The test named `TestextensionFileWatcher` segfaulted. Here is the log https://buildkite.com/nubots/nubots/builds/2862#060de692-2ae5-469a-9d12-bc36dbc4aa47

CMurtagh-LGTM:
@KipHamiltons What file would that test be? Would it be either `FileWatcherTestFile.cpp` or `FileWatcherTestDirectory.cpp`?

KipHamiltons:
They get put into one test binary. 
```cpp
#define CATCH_CONFIG_MAIN  // This tells Catch to provide a main() - only do this in one cpp file
```
is in `FileWatcherTestDirectory.cpp`, and commented out in the other one. Basically, how it works is the files are considered together, with catch putting all the relevant test cases into a `main` it defines for you. 
Long story short - I don't know. They're one binary, and one set of test cases when we call them like this.

TrentHouliston:
FYI the reason that this proto exists is that once upon a time the c++ NUsight module used to construct these based on the actual subsumption messages that went through.

It would then send these to NUsight so it could show you what subsumption things had control at any time

PR #671, opened 2021-08-14
Purge build volume files rather than re-creating it

Bidski:
If vscode has created a container then deleting the build volume will because the volume is in use. Deleting the vscode container is undesirable as vscode installs extensions. etc in this container and they can take a while to download.

This PR removes the "delete build volume on clean" behaviour and replaces it with a "delete all files/folder in build volume" behaviour.

### DIFF
--- a/tools/utility/dockerise/run.py
+++ b/tools/utility/dockerise/run.py
@@ -156,14 +156,12 @@ def run(func, image):
 
         # If we are cleaning then delete all files and directories in the build volume
         if kwargs["clean"]:
-            if (
-                pty.spawn(
-                    docker_args + ["/bin/bash", "-c", "find /home/{}/build -mindepth 1 -delete".format(defaults.user)]
-                )
-                != 0
-            ):
+            exit_code = pty.spawn(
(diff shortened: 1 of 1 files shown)


tools/utility/dockerise/run.py
```diff
@@ -162,6 +154,17 @@ def _run(**kwargs):
...
+        # If we are cleaning then delete all files and directories in the build volume
+        if kwargs["clean"]:
+            if (
+                pty.spawn(
+                    docker_args + ["/bin/bash", "-c", "find /home/{}/build -mindepth 1 -delete".format(defaults.user)]
+                )
+                != 0
+            ):
+                cprint("Failed to clean the build volume", "red", attrs=["bold"])
+                exit(1)
```
KipHamiltons, line 166: Personally, I prefer the style of naming and capturing the exit code that we used in `webots.py`. I think it reads a little better
```suggestion
            exit_code = pty.spawn(
                docker_args + ["/bin/bash", "-c", "find /home/{}/build -mindepth 1 -delete".format(defaults.user)]
            )
            if exit_code != 0:
                cprint("Failed to clean the build volume", "red", attrs=["bold"])
                exit(exit_code)
```

KipHamiltons (approved):
LGTM

PR #672, opened 2021-08-14
Stops the robot from freezing after trying to get up twice

ysims:
This PR removes the ReactionHandles in Getup to prevent a race condition that resulted in Getup taking control and never giving it up. 

The problem was
- Get up is detected, priority is increased :arrow_up: 
- Because priority is increased, Getup gets control and ExecuteGetup is emitted :envelope: 
- getUp reactor is enabled by default so it runs fine :+1: 
- KillGetup runs after get up finishes and disables the getUp reactor :x: 
- Get up is detected, priority is increased :arrow_up: 
- Because priority is increased, Getup gets control and ExecuteGetup is emitted :envelope: 
- getUp reactor was disabled, so it doesn't run :x: 
- getUp reactor is enabled, but ExecuteGetup has already disappeared :cry:

### DIFF
--- a/module/behaviour/skills/Getup/src/Getup.cpp
+++ b/module/behaviour/skills/Getup/src/Getup.cpp
@@ -67,7 +67,7 @@ namespace module::behaviour::skills {
             EXECUTION_PRIORITY = config["EXECUTION_PRIORITY"].as<float>();
         });
 
-        fallenCheck = on<Last<20, Trigger<RawSensors>>, Single>().then(
+        on<Last<20, Trigger<RawSensors>>, Single>().then(
             "Getup Fallen Check",
             [this](const std::list<std::shared_ptr<const RawSensors>>& sensors) {
                 Eigen::Vector3d acc_reading = Eigen::Vector3d::Zero();
@@ -84,11 +84,10 @@ namespace module::behaviour::skills {
                     isFront = (M_PI_2 - std::acos(Eigen::Vector3d::UnitX().dot(acc_reading)) <= 0.0);
 
                     updatePriority(GETUP_PRIORITY);
-                    getUp.enable();
                 }
             });
 
-        getUp = on<Trigger<ExecuteGetup>, Single>().then("Execute Getup", [this]() {
+        on<Trigger<ExecuteGetup>, Single>().then("Execute Getup", [this]() {
             gettingUp = true;
 
             // Check with side we're getting up from
@@ -108,7 +107,6 @@ namespace module::behaviour::skills {
         on<Trigger<KillGetup>>().then([this] {
             gettingUp = false;
             updatePriority(0);
-            getUp.disable();
         });
 
         emit<Scope::INITIALIZE>(std::make_unique<RegisterAction>(RegisterAction{
--- a/module/behaviour/skills/Getup/src/Getup.cpp
+++ b/module/behaviour/skills/Getup/src/Getup.cpp
@@ -51,8 +51,6 @@ namespace module::behaviour::skills {
         , id(size_t(this) * size_t(this) - size_t(this))
         , isFront(true)
         , gettingUp(false)
-        , fallenCheck()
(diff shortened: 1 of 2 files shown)


module/behaviour/skills/Getup/src/Getup.cpp
```diff
@@ -67,7 +65,7 @@ namespace module::behaviour::skills {
             EXECUTION_PRIORITY = config["EXECUTION_PRIORITY"].as<float>();
         });
 
-        fallenCheck = on<Last<20, Trigger<RawSensors>>, Single>().then(
+        on<Last<20, Trigger<RawSensors>>, Single>().then(
```
KipHamiltons, line 68: Is there a real benefit to removing the reaction handle names? It seems to me like the thing that was causing the issues was just line 87, not the named reaction handles.

TrentHouliston: Yeah but it looks like it's not even used anymore now right? This assignment is gone

tayloryoung6396: We should be using it, was mean't to stop it getting up twice, removing this SHOULD introduce an old problem

ysims: Yeah neither of them are used once you remove that line. 
I don't see why the robot would try to get up again, there's only one ExecuteGetup emitted (when it takes priority). It would only run that again if it took priority again, which it would only do if it ran into that if statement in the check, which it would only do if it thought it was fallen over. 
I tested it a bit with this change and it was fine, but feel free to test it more 🤷

tayloryoung6396: @Bidski and @TrentHouliston are probably the people who would properly remember why it was added. Just one of those things, trying to understand why it was there before it's removed things

Bidski: Why would the robot try to get up again? Because its halfway through getting up and the check for falling triggers again? Our scripts are not exactly stellar in this area

tayloryoung6396: Not sure, just remember that was the issue, this was a number of years ago, It would try and get up after it got up

ysims: While it's getting up, there is a gettingUp boolean that stopped us running into the getup check again. Maybe that is what was missing in the old version? We'll only be able to check again if we lose priority and the KillGetup is run.

ysims: Tbh though what you're saying sounds different to what I've done in this PR. This old issue with double get ups would require disabling the check reactor, not the execute reactor. This PR stops us disabling the execute reactor.

ysims: Or maybe that was broken odometry @tayloryoung6396 😝

tayloryoung6396: I mean.. it's possible, but i feel like past us would have thought of that.. this PR removes both checks.. but if it works then it works.. we'll find out soon if it breaks things again

ysims: fallenCheck isn't used to enable/disable even though it's there, this PR only stops us disabling/enabling the getUp

ysims:
@KipHamiltons  `Expected — Waiting for status to be reported` 😢

KipHamiltons:
@ysims https://buildkite.com/nubots/nubots/builds/2865#3ffe9688-89a0-4249-9362-5ee810c26db1
It failed. Not sure why it didn't update here thoug

ysims:
That's an old build, there's another commit since then that fixes that.

KipHamiltons:
I tested and it seems to work :shrug:

KipHamiltons (approved):

PR #674, opened 2021-08-24
Removes Overview field I removed from the proto file

KipHamiltons:
Fixes a boo-boo I made

PR #675, opened 2021-08-25
Adds error handling code to nusight build script

KipHamiltons:
If this works, the CI check should fail until #674 is merged.


nusight2/build_scripts/build.ts
```diff
@@ -15,6 +13,20 @@ const opts: ConfigOptions = {
...
+    }
+  }
+  if (err) {
+    process.stderr.write(err.message)
+    if (err.stack) {
+      process.stderr.write(err.stack)
+    }
+    process.exitCode = 1
+  }
+  close()
```
JosephusPaye, line 31: Where's `close` from? Should it be `webpack.close()`?

```suggestion
  webpack.close()
```

nusight2/build_scripts/build.ts
```diff
@@ -15,6 +13,20 @@ const opts: ConfigOptions = {
...
+      process.exitCode = 1
+    }
+  }
+  if (err) {
+    process.stderr.write(err.message)
+    if (err.stack) {
+      process.stderr.write(err.stack)
+    }
+    process.exitCode = 1
+  }
```
JosephusPaye, line 30: Is idiomatic to handle `err` first thing in the callback, so I'd move this up, with an early return if there's an error, like the example shows here: https://webpack.js.org/api/node/#error-handling.

nusight2/build_scripts/build.ts
```diff
@@ -15,6 +13,23 @@ const opts: ConfigOptions = {
...
+    return
+  }
+  if (stats) {
+    process.stdout.write(stats.toString({ colors: true }) + '\n')
+    if (stats.hasErrors()) {
+      process.exitCode = 1
+    }
+  }
+  // Calling close() allows low-priority work (like persistent caching) the opportunity to complete
+  compiler.close((closeErr) => {})
```
LachlanCourt, line 34: Is this a variable? It is not assigned anywhere

KipHamiltons: As in, is `compiler` a variable?

KipHamiltons: I was basically trying to follow these docs:

https://webpack.js.org/api/node/#error-handling

https://webpack.js.org/api/node/#statshaserrors

And joey's comment, regarding `close`

https://nubotsteam.slack.com/archives/C3T1J8SGL/p1629878566062600

LachlanCourt: Sorry should have been more specific, I don't see where `closeErr` gets it's value, I'm assuming it is a variable

KipHamiltons: It's an anonymous function/callback. That's the parameter for it. Basically when you call close you give it that function. It will then call that function with the parameter it uses for the callback. So that's the internal object it passes to the callback when it executes it.

KipHamiltons: In this case, i did not know of anything specific which needed to be done in it, so I did nothing.

JosephusPaye: You should probably write it to `stderr`. We'd want to know if there was an error closing the compiler.

JosephusPaye (approved):
LGTM if it works

PR #678, opened 2021-08-25
Emit vision balls even if there are no balls

ysims:
At the moment we skip emitting a `vision::balls` message in `BallDetector` if there are no balls found. The rest of the codebase seems to assume otherwise - that we still send `vision::balls` when there are no balls. Examples in the codebase:
- Head behaviour gets vision balls on a 'with' statement, and it updates based on the last balls message. This ends up with behaviour chasing a ball that it can no longer see since it never gets the update that balls are gone. 
- Soccer strategy has an update that triggers when a balls message comes in. It already checks that the balls message is empty before continuing on and acting on that balls message.


TrentHouliston (approved):
This is very important! Not seeing a ball is in itself information

PR #681, opened 2021-08-31
Add extra flag to run

CMurtagh-LGTM:
I want to run commands in the docker image.

### DIFF
--- a/tools/run.py
+++ b/tools/run.py
@@ -25,7 +25,12 @@ def register(command):
         help="Run the specified program using valgrind",
     )
     command.add_argument(
-        "--dont-append-bin", dest="use_bin", action="store_false", default=True, help="(Advanced) run a shell command"
+        "-p",
+        "--run-program",
+        dest="use_bin",
+        action="store_false",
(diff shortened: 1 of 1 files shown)


tools/run.py
```diff
@@ -24,11 +24,14 @@ def register(command):
         default=False,
         help="Run the specified program using valgrind",
     )
+    command.add_argument(
+        "--dont-append-bin", dest="use_bin", action="store_false", default=True, help="(Advanced) run a shell command"
```
KipHamiltons, line 28: You're prepending `bin`, not appending. Also, I think this would be better if it were something like `run-command` or `run-program` or `execute-command` or something, rather than the logically negative, implementation-specific `dont-prepend-bin`.
Also, please add a short version to this argument.

KipHamiltons (commented):
I think this would be better served as an extension to `./b shell` tbh. I am not 100% sure this is necessary either

CMurtagh-LGTM:
I'm being dumb, this can be done with `./b shell`.

PR #682, opened 2021-08-31
Add reaction statistics view to NUsight

### LINKED ISSUE #629: NUsight Reaction View
Having a reaction view (like @TrentHouliston mentioned in #627) with reaction statistics about individual reactions and aggregate statistics would be sweet. We could perhaps even have charts indicating the performance of regular reactions.

LachlanCourt:
This view will allow a user to see a visual output of reaction statistics in a format that would be easier to see than a classic debug log message in terminal

To test, run NUbots codebase with the role `testreactionstats` which runs a module copied from `keyboardwalkfake` which emits the ReactionStatistics protobuf message. Alongside, run `yarn build` and `yarn prod --address <address>` and navigate to NUsight reaction statistics tab to see reactions showing up.

To do:
- [ ] Have NUClear send time properly in emit message
- [ ] Design UI
- [ ] Add storybook for UI
- [ ] Separate reaction stats out by reactionID to make them more readable
- [ ] Implement UI
- [ ] Add simulation for `yarn dev`


module/support/ReactionStats/src/ReactionStats.cpp
```diff
@@ -0,0 +1,46 @@
...
+            reactionData->cause_task_id     = stats.cause_task_id;
+            reactionData->emitted           = 0;
+            reactionData->started           = 0;
+            reactionData->finished          = 0;
+            // TODO fix this
+            /*reactionData.emitted           = getUtcTimestamp<std::chrono::microseconds>(stats.emitted);
+            reactionData.started           = getUtcTimestamp<std::chrono::microseconds>(stats.started);
+            reactionData.finished          = getUtcTimestamp<std::chrono::microseconds>(stats.finished);*/
+
+            emit(reactionData);
```
TrentHouliston, line 36: It might be better here to just directly emit to NUsight?
Not 100% sure on this one as the NetworkForwarder could also do it of course

module/support/ReactionStats/src/ReactionStats.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+
+            log(stats.identifier[0],
+                1000.0
+                    * (double((stats.finished - stats.started).count())
+                       / double(NUClear::clock::duration::period::den)),
+                "ms");
```
TrentHouliston, line 43: If you're doing math with times, just use the chrono tools. In this case it looks like you're after the time in milliseconds

`std::chrono::duration_cast<std::chrono::milliseconds(stats.finished - stats.started).count()`

If you really wanted it as a double, might be good to do this
`std::chrono::duration_cast<std::chrono::duration<double>>(stats.finished - stats.started).count() * 1000`

module/support/ReactionStats/src/ReactionStats.cpp
```diff
@@ -0,0 +1,46 @@
...
+            reactionData->task_id           = stats.task_id;
+            reactionData->cause_reaction_id = stats.cause_reaction_id;
+            reactionData->cause_task_id     = stats.cause_task_id;
+            reactionData->emitted           = 0;
+            reactionData->started           = 0;
+            reactionData->finished          = 0;
+            // TODO fix this
+            /*reactionData.emitted           = getUtcTimestamp<std::chrono::microseconds>(stats.emitted);
+            reactionData.started           = getUtcTimestamp<std::chrono::microseconds>(stats.started);
+            reactionData.finished          = getUtcTimestamp<std::chrono::microseconds>(stats.finished);*/
```
TrentHouliston, line 34: You should be able to just attach these directly to the message with no function, each of the `stats.x` times are just NUClear::clock::time_point objects.
These can be directly assigned into the protobuf so long as the protobufs type is a timestamp.

So change the protobuf type to `google.protobuf.Timestamp` and Neutron will figure out the rest.

TrentHouliston (commented):
A review of some of the parts of the c++ code

module/support/ReactionStats/src/ReactionStats.cpp
LachlanCourt: I thought that this was the only way to do it, what are you suggesting sorry?

module/support/ReactionStats/src/ReactionStats.cpp
LachlanCourt: Gotcha, this was just leftover copypasta from the file I used as a template, Once I've finished testing I'll be removing that line anyway but I'll do it for now while I verify 👍

module/support/ReactionStats/src/ReactionStats.cpp
LachlanCourt: Sweet, I couldn't figure it out and decided it was a future me problem as it wasn't a priority 😆 I'll chuck it in, thanks 🙂

module/support/ReactionStats/src/ReactionStats.cpp
TrentHouliston: I'm flip-flopping on if it's better to use network forwarder here :P
But basically, instead of this you could do `emit<Scope::NETWORK>(reactionData, "nusight")`; and send it straight to NUsight (that's how network forwarder works internally)

LachlanCourt: Gotcha, I think that makes sense, because the only reason you would include this module is if you were planning on using it on NUsight - it doesn't have any other purpose

LachlanCourt: If in the future it is needed by some other module it could be changed back

TrentHouliston: The reason I keep flip-flopping is that it lets you control all the communication to NUsight from the NetworkForwarder module. So maybe it's better to do it this way. Now that I think about it I think that's the same thing I do in the Overview module. I send it on so the network forwarder can deal with it.

LachlanCourt: Yeah okay fair, keep it consistent

PR #684, opened 2021-09-01
Transform Class with Templated Spaces - RFC

KipHamiltons:
Having submitted the PRs which removed the old `Transform` utility classes, I may have come full circle.
This Transform class won't compile if you try to use incompatible spaces.
This is just an idea. I think it could be worth pursuing, but I am keen to see what people think.
Another benefit is that we get to keep the concise variable names, without losing ambiguity as to what a letter means - for example, `f` can be foot or field, but it's clear using this system.

One thing I am considering is (and this does sound ghastly, but hear me out), instead of using an `enum` for the spaces, making a namespace called `space`, then defining `constexpr int`s, one for each space. This feels a bit dirty, but the benefit is that you can do
```cpp
using utility::math::space::TORSO;
// instead of:
using utility::math::Space;
```
for the current `enum` setup.

The difference is that when you use the `enum` values, you always have to prepend `Space`, making it like
```cpp
auto Htc = Transform<double, Space::TORSO, Space::CAMERA>(); // TORSO, CAMERA are enum values from the Space enum
// instead of (using ints):
auto Htc = Transform<double, TORSO, CAMERA>(); // TORSO, CAMERA are defined as ints
```

If we did opt to keep using the enum, I would make it into a smart enum :tm: too.


ysims:
How do we feel about people not using Eigen normally in terms of learning how to use Eigen bc it's a popular library?

CMurtagh-LGTM:
Can we use https://en.cppreference.com/w/cpp/language/enum#Scoped_enumerations instead of constexpr ints?

KipHamiltons:
> Can we use https://en.cppreference.com/w/cpp/language/enum#Scoped_enumerations instead of `constexpr int`s?

If you can use `using` declarations on those values, for sure. Otherwise, there'd be no point

KipHamiltons:
@TrentHouliston has raised an interesting alternative - we could enforce that transforms have compatible spaces with a clang-tidy check. Basically just a regex rule which says that the letters have to match up. This gets around potential concerns of having a mega-size `Space` enum, whilst still enforcing correctness (down to the correctness of the letters, which can be incorrect or ambiguous themselves).
Benefits:
- No mega `Space` enum
- Zero friction from having to add an extra field if you want to add a space. (IMO this isn't an issue, because I personally don't think that there are that many spaces that we use. For cases like `Htx`, which is a list of transforms each with different spaces, we could use `Space::OTHER`, or some other solution)

Downsides:
- Overloaded letters like `f` aren't covered and they could introduce bugs the proposed transform class would catch
- It has to wait on clang-tidy being a thing we regularly use (this isn't so much a downside as a "we can't do this until that roadblock is removed")
- We don't get the added benefits of the proposed Transform class disambiguating what a given letter means (again, not necessarily a downside, because this is something that _should_ be addressed properly in documentation)\

@TrentHouliston, does that steelman your point sufficiently? Anything to add?

KipHamiltons:
@Bidski @TheCoolKidsTable @CMurtagh-LGTM @ysims I am keen to hear if you think this is actually a good idea. Is this something you think we should do? Do you prefer the clang-tidy check?

KipHamiltons:
> How do we feel about people not using Eigen normally in terms of learning how to use Eigen bc it's a popular library?

I am ambivalent about this. A plus of this is that it could insulate us if Eigen's interface updates or we stop using Eigen (both seem unlikely right now of course). A minus is that Eigen is "worth"
 learning if people want to do more stuff in their career outside NUbots though. That doesn't seem like it's a concern rivaling potential bugs caught by a Transform class like this though, IMO.

Bidski:
I dont see a reason not to use an enum to be honest. You arr basically trying to use one without explicitly using one anyway

CMurtagh-LGTM:
I think that having a Transform class is more explicit and that the clang-tidy check may confuse people.

KipHamiltons:
> I dont see a reason not to use an enum to be honest. You arr basically trying to use one without explicitly using one anyway

It would strictly be for more concise usage. `Space::TORSO` is simply more verbose than `TORSO`. The cost is just a slightly dirty list of constants and extra `using` declarations

Bidski:
> > I dont see a reason not to use an enum to be honest. You arr basically trying to use one without explicitly using one anyway
> 
> It would strictly be for more concise usage. `Space::TORSO` is simply more verbose than `TORSO`. The cost is just a slightly dirty list of constants and extra `using` declarations

Enums put their members into the parent scope. Wrap that in a namespace and you have the same effect as your constexprs

Bidski:
> > How do we feel about people not using Eigen normally in terms of learning how to use Eigen bc it's a popular library?
> 
> I am ambivalent about this. A plus of this is that it could insulate us if Eigen's interface updates or we stop using Eigen (both seem unlikely right now of course). A minus is that Eigen is "worth"
>  learning if people want to do more stuff in their career outside NUbots though. That doesn't seem like it's a concern rivaling potential bugs caught by a Transform class like this though, IMO.

As a reminder of how widely used eigen is industry. The people who made armadillo (csiro/data61) use eigen

ysims:
Code that makes sense without documentation sounds better to me. The Transform class has high readability imo, which is part of the point? I dunno, I'm kind of leaning towards Transform class. The learning thing isn't a big deal if it makes our systems easier to use (there's already plenty other stuff for peeps to learn lol).

KipHamiltons:
> > > I dont see a reason not to use an enum to be honest. You arr basically trying to use one without explicitly using one anyway
> > 
> > 
> > It would strictly be for more concise usage. `Space::TORSO` is simply more verbose than `TORSO`. The cost is just a slightly dirty list of constants and extra `using` declarations
> 
> Enums put their members into the parent scope. Wrap that in a namespace and you have the same effect as your constexprs

You can't do `using` declarations for `enum` members though. You have to qualify them

shared/utility/math/Transform.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+        template <Space OtherTo, Space OtherFrom>
+        Transform<Scalar, To, OtherFrom> operator*(const Transform<Scalar, OtherTo, OtherFrom>& other) {
+            static_assert(From == OtherTo,
+                          "Incompatible spaces used in transform multiplication. "
+                          "Left Transform's From Space does not match right Transform's To Space.");
+
+            return Transform<Scalar, To, OtherFrom>(
+                Eigen::Transform<Scalar, 3, Eigen::Affine>(this->matrix() * other.matrix()));
+        }
```
CMurtagh-LGTM, line 53: :laughing: 
```suggestion
        template <Space OtherTo, Space OtherFrom>
        [[nodiscard]] Transform<Scalar, To, OtherFrom> operator*(const Transform<Scalar, OtherTo, OtherFrom>& other) {
            static_assert(From == OtherTo,
                          "Incompatible spaces used in transform multiplication. "
                          "Left Transform's From Space does not match right Transform's To Space.");

            return Transform<Scalar, To, OtherFrom>(
                Eigen::Transform<Scalar, 3, Eigen::Affine>(this->matrix() * other.matrix()));
        }
```

Bidski:
> > > > I dont see a reason not to use an enum to be honest. You arr basically trying to use one without explicitly using one anyway
> > > 
> > > 
> > > It would strictly be for more concise usage. `Space::TORSO` is simply more verbose than `TORSO`. The cost is just a slightly dirty list of constants and extra `using` declarations
> > 
> > Enums put their members into the parent scope. Wrap that in a namespace and you have the same effect as your constexprs
> 
> You can't do `using` declarations for `enum` members though. You have to qualify them

You `using` the surrounding namespace which gives you access to the symbols that are in that namespace which includes the enum amd its members

TrentHouliston:
I'm not a big fan of the Transform class idea for a number of reasons.

I think it's going to add maintenance overhead and not really pay for itself.
So often we have new spaces that we need to define, and they're not always fixed.
For example @ysims we would have a space for the swing foot and the support foot. Also one for the left foot, right foot. Ground space, torso space etc.
Then you end up with other intermediate spaces like observation space in the visual mesh.
Every module that is written could end up adding new spaces to this giant list. And since the enums aren't local to the file I can guarantee that this file will grow without bound and cause more confusion.
Basically, you'll have a god file with hundreds of spaces which won't help people code and look up what space they're using, it'll just make things more confusing. Now instead of having to just look at one file, I have to look at multiple files.

A very important concept in code is to keep the definitions and meanings of what you're doing as close to the code it's relevant to as possible. In the case of the spaces, define them in the module that you're writing, the only spaces that have meaning across multiple files are the ones in the messages, which should be documented well because they're messages!

I also don't like that you'll constantly be having to "break the rules" when you try to assign the "left foot space" to a "swing foot space". With a clang-tidy rule that's a warning, you can silence to say "I know what I'm doing". With a Transform class, it's a compile error that you have to deal with any time you want to do spaces, and therefore it will be more code you have to write every time you want to do something, it will make it much harder to rapidly prototype since you'll need casts everywhere. You can't just do the whole, I'll write this quickly and fix the warnings when I'm done.

And if you solve this by having an "Other" space you're falling into the classic trap of software engineering. You cannot make code better by making exceptions, if it's too hard people will just use Other for everything because it's too hard to work out the spaces. Or you'll end up with problems because people will use Other for two separate spaces and think they can use them just fine. With a clang-tidy warning you can leave the warnings for the end. With `Other` hacks you might miss them in review and they'll end up in the codebase.

The way I see it there are two problems that need to be solved here.
1. We need to make sure that people are doing the math according to the rules on the math page
2. People need to know what each letter represents in terms of spaces

I think 1 should be solved using a clang-tidy rule that enforces our math rules based on the variable names. Once you write some code that takes the variable names and understands them, it's fairly straightforward to work out what operations would convert them to and make sure they're correct.

I think 2 is a documentation problem, which you are trying to solve by having a god file with all the spaces documented there as code enums.
I personally think that approach is worse than just having a NUbook page or just a comment at the top of a module saying what each letter represents. You're taking local information to a module and pushing it out to a global spaces file.

KipHamiltons:
Thanks for C++20 magic, I have made it so that the `Space`s are literal strings. This way they are completely local to the class, while still retaining the expressiveness of the `enum` approach. I think this mostly addresses the issues @TrentHouliston had, but I am keen for more feedback.

Bidski:
I'm still not really a fan of this. While I agree that something like this is probably needed, I don't think this is actually going to solve the problem. I also don't think you have really addressed all of the issues that @TrentHouliston has raised.

Using the string literals isn't all that much more expressive than using a single character and doesn't resolve the issue of space aliasing (not that I think there is going to be a clean solution to this). So we will still have to have explicit casts in any place that we are doing this. I think it also needs to be remembered that embedding the spaces into template-space is going to create a lot of icky template compiler errors.

While I think Trent's idea of a clang-tidy regex-based check would resolve this issue better, I am not convinced that we could craft a regular expression that is expressive enough to properly meet our usage requirements. On top of this, if you only make such a check  a warning then people will likely ignore it as it is only a warning. If you make it an error then I am worried that people will just `// NOLINTNEXTLINE` everything for exactly the same reasons that people would just use `Space::Other` everywhere.

We also need to remember that unless me move to clang as our main compiler the only ways people will be able to see that they have got the naming wrong is by either explicitly running clang-tidy via cmake (needing a complete rebuild using clang/clang-tidy) or by pushing their changes and making a (draft) PR and waiting for the CI checks to complete. Both of these options are annoying and painful, and if they are developing in one of the NUC docker images then they can't actually run clang-tidy until they swap to the generic image.

Ultimately, I think more thorough documentation, both about the common spaces we use and how our mathematical notation is used, is the only thing that will "resolve" this issue.

KipHamiltons:
> Using the string literals isn't all that much more expressive than using a single character and doesn't resolve the issue of space aliasing (not that I think there is going to be a clean solution to this). So we will still have to have explicit casts in any place that we are doing this. I think it also needs to be remembered that embedding the spaces into template-space is going to create a lot of icky template compiler errors.

I agree template compiler errors are annoying, however I personally would prefer a compiler error when you do an incompatible multiplication than no error at all.

In terms of the concern of casting in places where we have more dynamic spaces like the swing-foot case, I would argue that the current solution is as bad or worse, in those limited cases. The letters have to change (or the letters suffer from the exact same issue of casting/reassignment) when you change from `left-foot` to `swing-foot` anyway. I don't understand how having it templated changes this.
 
> While I think Trent's idea of a clang-tidy regex-based check would resolve this issue better, I am not convinced that we could craft a regular expression that is expressive enough to properly meet our usage requirements. On top of this, if you only make such a check a warning then people will likely ignore it as it is only a warning. If you make it an error then I am worried that people will just `// NOLINTNEXTLINE` everything for exactly the same reasons that people would just use `Space::Other` everywhere.

I think the clang-tidy check is a great idea too. I think it would sit on top of this solution quite well - we don't have to have an exclusive-or choice between the two. The naming convention of `Hab` transforms is retained with this solution and the clang-tidy solution is valid whether we add this or not.

In terms of people just ignoring the check: that's what code review is for. Once we fix all of the clang-tidy warnings existing in the codebase, we'll turn clang-tidy on in the CI (along with -Werror, soon^TM), which means that people will be forced to use `// NOLINT` if they want to do naughty things, which is super easy for reviewers to catch. I don't think this will be an issue once clang-tidy and -Werror are on for CI builds.

> We also need to remember that unless we move to clang as our main compiler the only ways people will be able to see that they have got the naming wrong is by either explicitly running clang-tidy via cmake (needing a complete rebuild using clang/clang-tidy) or by pushing their changes and making a (draft) PR and waiting for the CI checks to complete. Both of these options are annoying and painful, and if they are developing in one of the NUC docker images then they can't actually run clang-tidy until they swap to the generic image.

People _should_ be running clang-tidy before they merge things. They can turn on `CI_BUILD` in configuration and get a build which perfectly matches the CI build - that's the benefit of that CMake variable. Yes, it's annoying! I completely agree. But, it's way way better if we go through that effort to catch potential bugs before they get merged and are in the codebase for 10 years.
 
> Ultimately, I think more thorough documentation, both about the common spaces we use and how our mathematical notation is used, is the only thing that will "resolve" this issue.

100% with you on documentation. It should certainly be improved. I am thinking a NUbook section with the list of spaces and their corresponding letters is necessary, so people can do a look-up of a letter if they're not sure.

One benefit of this class is that it gives a direct indication of what the Transform objects are. It's self-documenting in multiple ways - it clearly states the spaces involved, giving an intuitive look at the mechanics of the naming scheme (the naming scheme makes a lot more sense with a bit more context, which this provides, right at the location of the code), and it clearly indicates what the transform is - `Transform<double, "Camera", "Torso">` is much more helpful that the typename `Eigen::Affine3d`.

At the end of the day, all of your points are valid. Template errors are annoying (the `static_assert` for the * operator does provide a helpful error though - if I do say so myself - and that's where nearly all of the template errors will come from). The hypothetical clang-tidy check will prevent bugs, even if clang-tidy requires extra builds. Documentation should be improved.

I don't think any of those really outweigh the immediate benefits that this construction could give us though: no incorrect space manipulations (huge!!), and clear, self-documenting space assignment, as well as providing insight into the naming scheme.

IMO, those factors combined outweigh the issue of a potentially having a few extra casts in a couple of spots in the codebase, and _potentially_ having verbose compiler errors (which also _prevent_ you from making mistakes). I started lukewarm on this, but now I am more and more keen the more I think about it. If you @Bidski and @TrentHouliston both think that this isn't worth it, I'll close the PR, but I really think this could help us.

KipHamiltons:
I'll convert a couple of reactors on a branch off this one, and we can see if it seems good when we use it. I will do it for sensorfilter and the Quinticwalk

Bidski:
> I agree template compiler errors are annoying, however I personally would prefer a compiler error when you do an incompatible multiplication than no error at all.

Compiler errors are good until they are unreadable. I'm not saying this will create such errors (perhaps your `static_assert` will mask the pain, this waits to be seen), I'm just cautious of this.

> In terms of the concern of casting in places where we have more dynamic spaces like the swing-foot case, I would argue that the current solution is as bad or worse, in those limited cases. The letters have to change (or the letters suffer from the exact same issue of casting/reassignment) when you change from `left-foot` to `swing-foot` anyway. I don't understand how having it templated changes this.

Yes, the current "solution" sucks. This is why we are here. Honestly, I don't think there is a workable solution to this, apart from refactoring the offending code, but I don't think that is possible either. I am mainly venting and hoping someone else has an awesome brainwave.

> I think the clang-tidy check is a great idea too. I think it would sit on top of this solution quite well - we don't have to have an exclusive-or choice between the two. The naming convention of `Hab` transforms is retained with this solution and the clang-tidy solution is valid whether we add this or not.

Having both may not be smart. It means you have two tightly-coupled systems that need to be maintained that are both trying to do the same job.

> In terms of people just ignoring the check: that's what code review is for. Once we fix all of the clang-tidy warnings existing in the codebase, we'll turn clang-tidy on in the CI (along with -Werror, soon^TM), which means that people will be forced to use `// NOLINT` if they want to do naughty things, which is super easy for reviewers to catch. I don't think this will be an issue once clang-tidy and -Werror are on for CI builds.

Code review is only as good as the reviewers. If the reviewers aren't downloading and compiling the code themselves to check the warnings, or if they aren't trawling through the buildkite (or github actions) output to check for any warnings, and they miss it in the code then all of our "checks" have come to naught. If this issue is as severe as we think it is (or want it to be) then it can't be something that is easily bypassed/missed.
 
> People _should_ be running clang-tidy before they merge things. They can turn on `CI_BUILD` in configuration and get a build which perfectly matches the CI build - that's the benefit of that CMake variable. Yes, it's annoying! I completely agree. But, it's way way better if we go through that effort to catch potential bugs before they get merged and are in the codebase for 10 years.

That is my point though. If it's hard/annoying to do people won't do it. The only "nice" part about this situation is that we have a CMake variable for it.


> 100% with you on documentation. It should certainly be improved. I am thinking a NUbook section with the list of spaces and their corresponding letters is necessary, so people can do a look-up of a letter if they're not sure.

No, that is way too much. That would make the documentation page as bad as what @TrentHouliston said the `Spaces` namespace was going to be. You would have this one documentation page that could be getting updated frequently and which is a "long distance" from the code you are looking at for you to look it up. The documentation page should be short, simple, and sweet. Just list some of the most common spaces we use (`torso`, `camera`, `field`, `world`, not sure there are too many others worth a special mention). The maths page can also expand more on the notation with how and why we use it (maybe even an extra page that just lists the notation we use so it isn't lost in the rest of the maths).

> One benefit of this class is that it gives a direct indication of what the Transform objects are. It's self-documenting in multiple ways - it clearly states the spaces involved, giving an intuitive look at the mechanics of the naming scheme (the naming scheme makes a lot more sense with a bit more context, which this provides, right at the location of the code), and it clearly indicates what the transform is - `Transform<double, "Camera", "Torso">` is much more helpful that the typename `Eigen::Affine3d`.

Yes, that is nice. However, clear commenting also serves that purpose (which people _should_ be doing anyway). Also understand that `Transform` can also be considered to be vague. What sort of transform? Is it affine? projective? isometry? rotation? translation? scaling? perspective? Some other sort of non-affine transformation? Is it 2D transformation? 3D? 4D? ND? `Transform` could apply equally to all of these while `Eigen::Affine3d` can only apply to 3D affine transformations. Your `Transform` class also only wraps 3D affine transformations, but we have been trying to introduce 2D affine transformations as well, will you create a wrapper for that? How will it differ from this one?

> I don't think any of those really outweigh the immediate benefits that this construction could give us though: no incorrect space manipulations (huge!!), and clear, self-documenting space assignment, as well as providing insight into the naming scheme.

We clearly are disagreeing on this point. While there _may_ potentially be some benefits to this construct I can also see some detriments and I am not convinced that the benefits outweigh the detriments. I can also see other ways to achieve the same/similar benefits and I am seeing no clear winners.

> IMO, those factors combined outweigh the issue of a potentially having a few extra casts in a couple of spots in the codebase, and _potentially_ having verbose compiler errors (which also _prevent_ you from making mistakes). I started lukewarm on this, but now I am more and more keen the more I think about it. If you @Bidski and @TrentHouliston both think that this isn't worth it, I'll close the PR, but I really think this could help us.

At the moment I am on the fence about this. As much as I am for the clang-tidy check. I can't see that this extra complexity (in either case) is going to provide clear benefits that make maintaining the complexity worth it. I am also concerned that our space aliasing is going to result in a lot of casts/`NOLINTNEXTLINE`s (I don't know how extensively we alias spaces so this may not be an important point, but if there are a lot then that will just be annoying).

> I'll convert a couple of reactors on a branch off this one, and we can see if it seems good when we use it. I will do it for sensorfilter and the Quinticwalk

Sure, lets see where this leads us.

I am also really annoyed that the C++20 "improvement" is only for string literals .....

KipHamiltons:
I have pushed a branch based off this branch which replaces `AffineXd` usage with `Transform` usage for `Sensorfilter` and `QuinticWalk`. The branch compiles as is.
Heres `SensorFilter.cpp`: [link](https://github.com/NUbots/NUbots/blob/hamiltons/transform-class-trial/module/input/SensorFilter/src/SensorFilter.cpp)
and `QuinticWalk.cpp`: [link](https://github.com/NUbots/NUbots/blob/hamiltons/transform-class-trial/module/motion/QuinticWalk/src/QuinticWalk.cpp)
and the branch, for general browsing: [all relevant changes in last two commits](https://github.com/NUbots/NUbots/compare/hamiltons/transform-class-trial)

I am a fan :stuck_out_tongue_closed_eyes: 

For reference, here is a picture of what it looks like when the spaces are incompatible for a multiplication. The line number is directly above the word `error` in red. It was a very quick fix with the line number.
static_assert_failure

KipHamiltons:
It seems like we won't reach a consensus in favour of this change. Closing.

CMurtagh-LGTM:
I think that there is still merit in this.

> Code review is only as good as the reviewers. If the reviewers aren't downloading and compiling the code themselves to check the warnings, or if they aren't trawling through the buildkite (or github actions) output to check for any warnings, and they miss it in the code then all of our "checks" have come to naught.

The build fails, the creator of the code reads the error log, fixes the error. No need for reviewer to read the logs.

> We also need to remember that unless me move to clang as our main compiler the only ways people will be able to see that they have got the naming wrong is by either explicitly running clang-tidy via cmake (needing a complete rebuild using clang/clang-tidy) or by pushing their changes and making a (draft) PR and waiting for the CI checks to complete. Both of these options are annoying and painful, and if they are developing in one of the NUC docker images then they can't actually run clang-tidy until they swap to the generic image.

Better than guessing :shrug:, and I think it is important that broken code should fail in CI.

> Compiler errors are good until they are unreadable.

A compiler error is better than a runtime error or logic error, even if it is unreadable.

Now I do think that we should move away from the Transformation name, as that name is too generic.

Some new ideas :stuck_out_tongue_closed_eyes: 
We should make explicit space objects rather than having string literals.
We could have common spaces as constant expressions e.g. `constexpr Space torso{};`,
have temporary spaces like `Space cool_space_only_in_this_scope{}`
And have space aliases like `Space swing_foot = SpaceAlias(Space::foot)`
I'll make a PR

PR #685, opened 2021-09-01
Removes Unused Utilities

KipHamiltons:
These utilities aren't used. They're never even included anywhere. This PR removes them.

(I thought that `euler.hpp` was in a similar boat, but I had a filter on my search on vscode :facepalm: )


shared/utility/strutil/banner.hpp
```diff
@@ -1,559 +0,0 @@
...
-                                          R"(|_|    )",    R"(    |_|)",
-                                          R"(      )",     R"(     )",
-                                          R"(     )",      R"(       )",
-                                          R"(       )",    R"(          )",
-                                          R"(      )",     R"( |___/ )",
-                                          R"(     )",      R"(     )",
-                                          R"(|_|)",        R"(     )",
-                                          R"(      )",     R"()"}};
-
-    std::string banner(const std::string& input) {
```
TrentHouliston, line 542: This file is very much used!
Where do you think the text at the start of literally every role comes from?

shared/utility/support/Timer.hpp
```diff
@@ -1,30 +0,0 @@
-#ifndef TIMER_HPP
-#define TIMER_HPP
-
-#include <chrono>
-
-class Timer {
```
TrentHouliston, line 6: I would keep the timer, It will never really stay in code, but you use it when you want to time a section of code (and then don't commit your performance prints when you finish)

TrentHouliston (changes requested):

shared/utility/strutil/banner.hpp
KipHamiltons: I thought that, but everything compiles without it, and the term `banner.hpp` doesn't appear in the codebase. :shrug:

shared/utility/support/Timer.hpp
KipHamiltons: Fair. Will remove this change

shared/utility/strutil/banner.hpp
CMurtagh-LGTM: @TrentHouliston We don't use `banner.hpp` for banners, we use `nuclear/roles/banner/bigtext.py`.

CMurtagh-LGTM (approved):
Looks Good To Me

shared/utility/strutil/banner.hpp
TrentHouliston: :O

shared/utility/support/Timer.hpp
TrentHouliston: I say this because you have no idea how many times I've written a timer module and just hacked it and wished I had something easier to use :P

TrentHouliston (approved):

PR #690, opened 2021-09-04
Updates Archlinux date and migrates to LLVM12

KipHamiltons:
Basically, I want GCC11, because they upgraded template type deduction, which will allow string literals for template types - you can probably guess why I want that :wink: .
I updated the arch date to a new date, but then LLVM didn't want to build. So, I updated the LLVM version, which required many downstream dependency changes. In related news, I officially hate intel's open source packages.


docker/Dockerfile
```diff
@@ -95,55 +95,62 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr/local
 RUN install-from-source https://www.zlib.net/zlib-1.2.11.tar.gz
 
 # LLVM and Clang
-RUN BUILD_FILE_DIR='llvm' install-cmake-from-source https://github.com/llvm/llvm-project/archive/llvmorg-11.0.1.tar.gz \
-    -DLLVM_ENABLE_PROJECTS="clang;clang-tools-extra" \
+RUN BUILD_FILE_DIR='llvm' install-cmake-from-source https://github.com/llvm/llvm-project/archive/llvmorg-12.0.1.tar.gz \
+    -DLLVM_ENABLE_PROJECTS="clang;clang-tools-extra;lld" \
```
KipHamiltons, line 99: `lld` is now an IGC dep

docker/Dockerfile
```diff
@@ -95,55 +95,62 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr/local
...
 
 # opencl-clang
 COPY usr/local/package/opencl-clang.sh /usr/local/package/opencl-clang.sh
-RUN /usr/local/package/opencl-clang.sh https://github.com/intel/opencl-clang/archive/v11.0.0.tar.gz
+RUN /usr/local/package/opencl-clang.sh https://github.com/intel/opencl-clang/archive/v12.0.0.tar.gz
 
 # intel graphics compiler
 COPY usr/local/package/intel-graphics-compiler.sh /usr/local/package/intel-graphics-compiler.sh
-RUN /usr/local/package/intel-graphics-compiler.sh https://github.com/intel/intel-graphics-compiler/archive/igc-1.0.5435.tar.gz \
+RUN /usr/local/package/intel-graphics-compiler.sh https://github.com/intel/intel-graphics-compiler/archive/igc-1.0.8173.tar.gz \
```
KipHamiltons, line 114: This version isn't the newest, but it's what the arch packages are up to, and it's the commit which their patches are valid on.

docker/Dockerfile
```diff
@@ -258,7 +265,7 @@ COPY usr/local/package/espeak.sh /usr/local/package/espeak.sh
 RUN /usr/local/package/espeak.sh https://github.com/espeak-ng/espeak-ng/releases/download/1.50/espeak-ng-1.50.tgz
 
 # mio memory mapping library
-RUN install-from-source https://github.com/mandreyel/mio/archive/master.tar.gz
+RUN install-from-source https://github.com/mandreyel/mio/archive/3f86a95c0784d73ce6815237ec33ed25f233b643.tar.gz
```
KipHamiltons, line 268: Techniically unrelated. I figured it was best to pin this while I was here

docker/usr/local/package/intel-graphics-compiler.sh
```diff
@@ -33,6 +33,26 @@ echo "Configuring using cmake file ${CMAKELISTS_FILE}"
...
+
+# Apply the patches
+patch -p1 < llvm12-porting.patch
+patch -p1 -R < revert-stack-analysis.patch
+patch -p1 < igc-missing-limits-headers.patch
+patch -p1 < fix-dce.patch
+
+# The build requires a binary in here to be run (CMCLTranslatorTool)
+# Without this directory in the path, bash can't find it
+PATH="${PATH}:$(pwd)/build/IGC/Release"
```
KipHamiltons, line 54: I will admit this is a little hacky. [this call](https://github.com/intel/intel-graphics-compiler/blob/30cf926bc1f69761a774c2475208865099c94c4b/IGC/VectorCompiler/lib/BiF/CMakeLists.txt#L85) is the one that fails without this on the path. Welcoming better ideas here for sure.

KipHamiltons:
We should definitely monitor IGC for updates about when they officially start llvm12 or 13, because I'll be the first to admit that 4 patches isn't ideal

docker/usr/local/package/intel-graphics-compiler.sh
```diff
@@ -33,6 +33,26 @@ echo "Configuring using cmake file ${CMAKELISTS_FILE}"
 # Clone the dependency
 git clone https://github.com/intel/vc-intrinsics vc-intrinsics
 
+# Download the patches
```
Y0sh1dk, line 36: Cant we just pipe the output of `curl` into `patch` instead of saving to file?

KipHamiltons: We could. That seems reasonable. Thinking about it, there is a benefit to doing it like this though: the patches are named clearly, so we know which one is which, which might be convenient later. We could do that with comments though I suppose. Good point!

Y0sh1dk: Was just thinking for consistency with our other scripts (`install-from-source` and `espeak.sh`).

CMurtagh-LGTM:
I like that we're updating things, but I'm not keen to review this whilst it fails CI.

KipHamiltons:
My last change screwed it up somehow :weary: I'll fix that.

Y0sh1dk (dismissed):
LGTM once https://github.com/NUbots/NUbots/pull/697 is ready to be merged also.

KipHamiltons:
@Y0sh1dk the clang build is broken already. Clang12 just has a couple of extra errors to fix 😝
I think this pr is independent of that one

Y0sh1dk:
@KipHamiltons Ah true. We should probably update the archlinux tag in [nugus_sim.Dockerfile](https://github.com/NUbots/NUbots/blob/main/docker/nugus_sim.Dockerfile#L1) at the same time as this.

Y0sh1dk (approved):
LGTM.

PR #691, opened 2021-09-05
Modernises Configuration.hpp

KipHamiltons:
This makes "Modern C++" changes to Configuration. I got sidetracked again hahaha


module/platform/Webots/src/Webots.cpp
```diff
@@ -252,7 +252,8 @@ namespace module::platform {
 
         on<Configuration>("WebotsCameras").then([this](const Configuration& config) {
             // Strip the .yaml off the name of the file to get the name of the camera
-            const std::string name = ::basename(config.fileName.substr(0, config.fileName.find_last_of('.')).c_str());
+            const std::string config_file = config.fileName.string();
+            const std::string name        = ::basename(config_file.substr(0, config_file.find_last_of('.')).c_str());
```
CMurtagh-LGTM, line 256: I think this logic should go in a method of Configuration, something like `[[nodiscard]] string get_name() const;`

shared/extension/Configuration.hpp
```diff
@@ -235,11 +241,11 @@ namespace NUClear::dsl {
                 const auto* binary = basename(data.data());
 
                 // Set paths to the config files.
-                auto defaultConfig = "config/" + path;
-                auto robotConfig   = "config/" + hostname + "/" + path;
-                auto binaryConfig  = "config/" + std::string(binary) + "/" + path;
+                const auto defaultConfig = "config" + path;
+                const auto robotConfig   = "config" + hostname + path;
+                const auto binaryConfig  = "config" + std::string(binary) + path;
```
CMurtagh-LGTM, line 246:
```suggestion
                const auto defaultConfig = fs::path("config") / path;
                const auto robotConfig   =  fs::path("config") / hostname / path;
                const auto binaryConfig  =  fs::path("config") / std::string(binary) / path;
```

shared/extension/Configuration.hpp
```diff
@@ -294,24 +300,21 @@ namespace NUClear::dsl {
...
                         for (const auto& component : components) {
                             // Ignore the hostname/binary name if they are present.
                             if (flag && (component.compare(hostname) != 0) && (component.compare(binary) != 0)) {
-                                relativePath.append(component + "/");
+                                relativePath = relativePath / component;
                             }
 
                             // We want out paths relative to the config folder.
-                            if (component.compare("config") == 0) {
+                            if (component.compare(fs::path("config")) == 0) {
```
CMurtagh-LGTM, line 313: Why are we wrapping config in `fs::path` here?

CMurtagh-LGTM (changes requested):

shared/extension/Configuration.hpp
KipHamiltons: `ctrl + f` took the wheel and added one too many here :joy: 
Thank ya sir

module/platform/Webots/src/Webots.cpp
KipHamiltons: Great suggestion. Doing that now. Thanks

shared/extension/Configuration.hpp
```diff
@@ -223,27 +223,31 @@ namespace extension {
...
     namespace operation {
+        namespace {
+            [[nodiscard]] std::string getFirstCommandLineArg() {
+                std::shared_ptr<const message::CommandLineArguments> args =
+                    store::DataStore<message::CommandLineArguments>::get();
+                std::vector<char> data(args->at(0).cbegin(), args->at(0).cend());
+                data.push_back('\0');
+                return std::string(data.data());
+            }
+        }  // namespace
```
CMurtagh-LGTM, line 234: Isn't anon namespace in hpp bad?

module/input/Camera/src/Camera.cpp
```diff
@@ -54,7 +54,8 @@ namespace module::input {
              ************************************************************************/
             if (it == cameras.end()) {
                 // Strip the .yaml off the name of the file to get the name of the camera
-                std::string name = ::basename(config.fileName.substr(0, config.fileName.find_last_of('.')).c_str());
+                const std::string config_file = config.fileName.string();
+                const std::string name = ::basename(config_file.substr(0, config_file.find_last_of('.')).c_str());
```
Bidski, line 58: Why? Perhaps [this](https://en.cppreference.com/w/cpp/filesystem/path/stem) or similar would be better?

shared/extension/Configuration.hpp
```diff
@@ -29,93 +30,98 @@
 #include "utility/support/hostname.hpp"
 #include "utility/support/yaml_log_level.hpp"
 
+namespace fs = ::std::filesystem;
```
Bidski, line 33: I am generally quite opposed to `using`s on the `std` namespace (and its sub-spaces). And I think in this case the code can be refactored in a way that renders this using quite pointless.

shared/extension/Configuration.hpp
```diff
@@ -217,42 +223,45 @@ namespace extension {
 // NUClear configuration extension
 namespace NUClear::dsl {
     namespace operation {
+        namespace {
+            [[nodiscard]] std::string getFirstCommandLineArg() {
+                std::shared_ptr<const message::CommandLineArguments> args =
+                    store::DataStore<message::CommandLineArguments>::get();
+                std::vector<char> data(args->at(0).cbegin(), args->at(0).cend());
+                data.push_back('\0');
+                return std::string(data.data());
```
Bidski, line 232: Also, don't assume that `CommandLineArguments` has something in it, it is possible for it to be empty
```suggestion
                return std::string(args->at(0).cbegin(), args->at(0).cend());
```

Bidski (changes requested):

shared/extension/Configuration.hpp
TrentHouliston: I'm ok with this one, std::filesystem and std::chrono can make for every long things. That said I would prefer that the using of namespace is as close to the useage scope as possible

shared/extension/Configuration.hpp
TrentHouliston: It makes and compiles a separate version of it for every c++ file that includes this. So for Configuration.hpp that almost everything includes it's probably not great :P

Also, this is clearly not a templated or inline function so what's it even doing in the header in the first place!

shared/extension/Configuration.hpp
TrentHouliston: Pretty sure it's not possible for it to be empty? You always at least have the actual binary name itself that was executed (which is what this is accessing).

I suppose if you made a custom PowerPlant instance rather than via the roles system you could do it?

Bidski: When running unit tests the powerplant is started in an interesting way

KipHamiltons: Thanks for pointing this out @Bidski, that's exactly what I came to configuration to do (and promptly forgot 😂), precisely for that reason

shared/extension/Configuration.hpp
KipHamiltons: [it's bad](https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL59-CPP.+Do+not+define+an+unnamed+namespace+in+a+header+file#:~:text=Do%20not%20define%20an%20unnamed%20namespace%20in%20a%20header%20file&text=When%20an%20unnamed%20namespace%20is,used%20within%20that%20translation%20unit.). I forgot lol. there's a [clang-tidy check](https://clang.llvm.org/extra/clang-tidy/checks/google-build-namespaces.html) for it.

shared/extension/Configuration.hpp
```diff
@@ -217,42 +223,52 @@ namespace extension {
...
+        [[nodiscard]] inline std::string getFirstCommandLineArg() {
+            std::shared_ptr<const message::CommandLineArguments> args =
+                store::DataStore<message::CommandLineArguments>::get();
+            if (!args->empty()) {
+                std::vector<char> data(args->at(0).cbegin(), args->at(0).cend());
+                data.push_back('\0');
+                return std::string(data.data());
+            }
+            return std::string{};  // Empty string
+        }
```
CMurtagh-LGTM, line 238: Can this go in `shared/utility/support/`, it seems a much better home?

shared/extension/Configuration.hpp
```diff
@@ -217,42 +223,52 @@ namespace extension {
...
+
         template <>
         struct DSLProxy<::extension::Configuration> {
+            /// @brief Binds config files, in order from least specific to most specific - default, per-robot, binary
```
CMurtagh-LGTM, line 242: It does not bind config files, it is a bind call from nuclear. We're setting up the reaction here.

CMurtagh-LGTM (changes requested):

shared/extension/Configuration.hpp
TrentHouliston: It's 100% an extension not a utiltiy

CMurtagh-LGTM: How does `getFirstCOmmandLineArg()` extend nuclear?

Also can this be snake_case?

TrentHouliston: It's mainly that I don't see this as a utility, but part of the configuration. The fact that it's screwing with the internals of NUClear just makes it feel wrong to consider it anything that's not super tightly coupled to NUClear itself

KipHamiltons: It's also not used anywhere except this file, which goes against our (albeit not strict) policy of making things utilities if they are used by more than one thing

KipHamiltons: Regarding snake case, I was trying to be consistent with the other functions, but now I see that it's only mergeYAML which isn't snake case, in terms of functions in this file. I'll change it too

shared/extension/Configuration.hpp
KipHamiltons: Could you please elaborate what you mean by "it is a bind call from nuclear"? I don't-understand the distinction

CMurtagh-LGTM: When you ask NUClear to subscribe to an event (e.g. calling `then`), represented as a type, NUClear calls the `bind` function of the type representing the event. The `bind` function is called once and is used to setup the reaction handle. This reaction handle is then returned from `then`.

`DSLProxy` is a magical type that allows you to add nuclear specific functions to already defined types.

CMurtagh-LGTM (changes requested):

CMurtagh-LGTM (approved):

shared/extension/Configuration.hpp
Bidski: This is still ugly and can be made not ugly

shared/extension/Configuration.hpp
```diff
@@ -217,90 +223,113 @@ namespace extension {
...
                     DSLProxy<::extension::FileWatch>::bind<DSL>(reaction, robotConfig, flags);
                 }
 
-                // Bind our binary specific path if it exists
-                if (utility::file::exists(binaryConfig)) {
-                    DSLProxy<::extension::FileWatch>::bind<DSL>(reaction, binaryConfig, flags);
+                // If there were command line arguments, we can get the binary name, and check for a binary config
+                // If not, we don't bother checking for a binary config to bind
+                const auto binaryName = get_first_command_line_arg();
+                if (binaryName != "") {
```
Bidski, line 293:
```suggestion
                if (!binaryName.empty()) {
```

Bidski (changes requested):

shared/extension/Configuration.hpp
KipHamiltons: It has been beautified :heart_eyes:

Bidski (changes requested):

shared/extension/Configuration.hpp
```diff
@@ -217,90 +223,113 @@ namespace extension {
...
+
+        /// @brief utility function to fetch the first command line argument, used to get the binary name
+        /// @returns If the first command line argument exists, it is returned. Otherwise, the empty string is returned
+        [[nodiscard]] inline std::string get_first_command_line_arg() {
+            std::shared_ptr<const message::CommandLineArguments> args =
+                store::DataStore<message::CommandLineArguments>::get();
+
+            // args is effectively a shared_ptr<std::vector<std::string>>
+            if (!args->empty()) {
+                return (*args)[0];
```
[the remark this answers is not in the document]
KipHamiltons: The treatment is worse than the illness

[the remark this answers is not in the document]
Bidski: yea, it might be

Bidski (dismissed):
:shrug:

Bidski (approved):
Not sure I approve of the changed made in the last couple of commits

PR #692, opened 2021-09-06
Removes unnecessary casts from QuinticWalk reaction

KipHamiltons:
It casts to a matrix of doubles, before immediately casting back to floats. This is silly


ysims (approved):
Lol. LGTM.

PR #693, opened 2021-09-07
Make it technically possible to respond to voice commands

Mrwerdo:
This pull request implements intention recognition (i.e. speech to text, then text to intention), and is James and I's work integrated learning project. The project is built on top of [voice2json](http://voice2json.org) which is a collection of command line utilities that perform open speech recognition and intent recognition (among other things).

This pull request contains all the necessary changes for building voice2json on Arch Linux and emitting messages in NUclear when voice2json recognises an intention (a module called NUhear).


tayloryoung6396:
Where's the actual code files? There's only the docker files?

You might want to change the title of this PR as well. It should be of the form of completing this sentence
If you push the merge button this will <INSERT TITLE HERE>

CMurtagh-LGTM:
What is the `base.Dockerfile`? Why do we need it? Is it necessary for this PR?

Mrwerdo:
> Where's the actual code files? There's only the docker files?
> 
> You might want to change the title of this PR as well. It should be of the form of completing this sentence
> If you push the merge button this will

Yup, at the moment. I haven't included all the files yet because I'm still working on the setup. At the moment they are in a separate repository James and I have been using but I'll be copying them into this branch later this week.

> What is the `base.Dockerfile`? Why do we need it? Is it necessary for this PR?

It's not necessary and I'm putting everything into a single `Dockerfile`. I initially wrote it that way so I had an idea of what the necessary things voice2json needed to run on Arch Linux.

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,359 @@
...
+        (char*) "/home/nubots/voice2json/",
+        (char*) "-p",
+        (char*) "en",
+        (char*) "recognize-intent",
+        NULL
+    };
+    
+    MicProcHandles handles = {};
+    if(!spawn_process(&handles, args)) {
+        goto cleanup;
```
KipHamiltons, line 193: did you write this? `goto` :eyes: :eyes:

Mrwerdo: [git blame **/src/Microphone.cpp](https://xkcd.com/292/) 😉

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+#if 1
+    //this is just for testing
+    //we need this as we do not have an
+    //environment for testing the microphone
+    on<IO>(STDIN_FILENO, IO::READ).then([this] {
+        char b = 0;
+        ssize_t bytes_read = read(STDIN_FILENO, &b, 1);
+        assert(bytes_read == 1);
+        switch(b) {
```
tayloryoung6396, line 339: b?

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,369 @@
...
+            case 'q':
+                char *string = (char*)malloc(100);
+                size_t size = 100;
+                bytes_read = getline(&string, &size, stdin);
+                
+                uint32_t num = atoi(string);
+                printf("%ld,%d - %s \n", bytes_read, num, string);
```
tayloryoung6396, line 346: NUClear::log

tayloryoung6396 (changes requested):

module/input/Microphone/src/Microphone.cpp
KipHamiltons: probably want to use `fmt` too:
`#include <fmt/format.h>`
then when you use it:
`[NUClear::]log(fmt::format("{},{} - {}", bytes_read, num, string));`

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+    posix_spawn_file_actions_adddup2(&actions, stdout_pipe[1], STDOUT_FILENO);
+    posix_spawn_file_actions_adddup2(&actions, stderr_pipe[1], STDERR_FILENO);
+    posix_spawn_file_actions_adddup2(&actions, stdin_pipe[0], STDIN_FILENO);
+    
+    posix_spawn_file_actions_addclose(&actions, stdout_pipe[1]);
+    posix_spawn_file_actions_addclose(&actions, stderr_pipe[1]);
+    posix_spawn_file_actions_addclose(&actions, stdin_pipe[0]);
+        
+    if(posix_spawnp(&pid, args[0], &actions, 0, args, envp) != 0) {
```
Mrwerdo, line 129: What file descriptors are inherited in the child process?

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,367 @@
...
+    
+    success = true;
+cleanup:
+    if(stdout_pipe[1]) close(stdout_pipe[1]);
+    if(stderr_pipe[1]) close(stderr_pipe[1]);
+    if(stdin_pipe[0]) close(stdin_pipe[0]);
+
+    *handles = {stdout_pipe[0], stderr_pipe[0], stdin_pipe[1]};
+    
+    if(actions.__actions) {
```
Mrwerdo, line 141: Isn't the double underscore a convention used to indicate that a field is private? Is this the right way of checking for whether the value is present?

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,367 @@
...
+        switch(msg.type) {
+            case MIC_MSG_ENABLE: 
+                enabled = true;
+            break;
+            case MIC_MSG_DISABLE:
+                enabled = false;
+            break;
+            case MIC_MSG_TEST_AUDIO:
+                char filename[1024];
+                sprintf(filename, "/home/nubots/nuhear/audio/voice_%03d.wav", msg.data);
```
Mrwerdo, line 267: Could we perhaps enter the full path to a file instead of an integer for the number? I'd like to be able to test it with files located somewhere else on the file system.

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+    
+    on<IO>(handles.stdout, IO::READ).then([this] {
+        printf("fds[0].revents & POLLIN\n");
+            
+        //TODO: need to read and append to buffer until we reach a newline...
+        char buffer[0x1000];
+        ssize_t bytes_read = read(handles.stdout, buffer, sizeof(buffer)-1);
+        if(bytes_read > 0) {
+            assert((size_t)bytes_read < sizeof(buffer)); //TODO
```
Mrwerdo, line 294: Can this assert ever be triggered? I thought `read` will guarantee not to overfill the buffer.

module/input/Microphone/src/Microphone.cpp
Mrwerdo: `b` for byte, obviously :wink:

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,367 @@
...
+            
+        char buffer[0x1000];
+        ssize_t bytes_read = read(handles.stderr, buffer, sizeof(buffer)-1);
+        if(bytes_read > 0) {
+            assert((size_t)bytes_read < sizeof(buffer)); //TODO
+            buffer[bytes_read] = 0;
+            
+            printf("STDERR (%ld) = %s\n", bytes_read, buffer);
+            
+            perror("stderr");
```
Mrwerdo, line 325: Have you found this to report the correct error message? I'm skeptical about whether this will return a valid value in a multithreaded environment.

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,367 @@
...
+                std::unique_ptr<SpeechIntentMessage> intent = parse_voice2json_json(intent_json, strlen(intent_json));
+                emit(std::move(intent));
+
+                free(intent_json);
+            }
+            
+        }
+    });
+    
+    //TODO: have a better story on how we handle this
```
Mrwerdo, line 312: We really don't know why the subprocess is printing something to the standard error stream. I think the most appropriate behaviour is to just print what it displays, maybe prefixed with the subprocess that's being ran. If they have a logging facility, it would probably also be a good idea to send it there instead of the standard error stream.

module/input/Microphone/README.md
```diff
@@ -0,0 +1,14 @@
+Microphone
+==========
+
+## Description
+
```
Mrwerdo, line 5: Something like, "starts a subprocess and emits messages when the subprocess outputs something to `stdout`".

module/input/Microphone/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Usage
+
```
Mrwerdo, line 8: How do you use this module? What's the syntax? What parameters can I change?

module/input/Microphone/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Emits
+
```
Mrwerdo, line 11: What messages can I react to? What's the format?

module/input/Microphone/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Dependencies
+
```
Mrwerdo, line 14: `voice2json`

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,367 @@
...
+    env_path[written] = 0;
+    
+    char *const envp[] = { 
+        (char*) "KALDI_DIR=", //KALDI_DIR env var must be blank!
+        (char*) env_path, 
+        NULL 
+    };
+    
+    if((pipe(stdout_pipe) == -1) || (pipe(stderr_pipe) == -1) || (pipe(stdin_pipe) == -1)) {
+        goto cleanup;
```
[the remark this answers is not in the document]
KipHamiltons: There are other ways of dealing with this. One example is throwing exceptions. `goto` is a nasty code smell :stuck_out_tongue_closed_eyes:

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,367 @@
...
+        free(env_path);
+    }
+    
+    return success;
+}
+
+//python -m voice2json --base-directory /home/nubots/voice2json/ -p en transcribe-wav
+bool voice2json_transcribe_stream(MicProcHandles *handles) {
+    char *const args[] = {
+        (char*) "python",
```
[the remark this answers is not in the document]
KipHamiltons: string literals are of type `const char*` in C++, so I am fairly sure that this is effectively a `const_cast`. That's a bad code smell. I would try to find another way for this if possible.
Also, avoid `C-style` casts in general. the basic (very naive) tier list is `static_cast` > "functional style cast" >> `reinterpret_cast` > `const_cast`. There are times that reinterpret casts are necessary, but they are few. `C-style` casts basically throw the type at the wall and see which cast is necessary, using any of those casts it has to in the process.

module/input/Microphone/src/Microphone.cpp
KipHamiltons: call it `byte` then :P

module/input/Microphone/src/Microphone.cpp
KipHamiltons: we do have a logging facility. `NUClear::log(message)`. If you want, you can change the "log level" it gets logged at - as in, log it as a trace, debug, info, warning, error, or fatal message (in order of badness).

module/input/Microphone/src/Microphone.cpp
c3144856: guilty as charged

module/input/Microphone/src/Microphone.cpp
```diff
@@ -108,37 +110,33 @@ bool spawn_process(MicProcHandles *handles, char *const *args) {
...
+        posix_spawn_file_actions_addclose(&actions, stdin_pipe[1]);
 
-    posix_spawn_file_actions_adddup2(&actions, stdout_pipe[1], STDOUT_FILENO);
-    posix_spawn_file_actions_adddup2(&actions, stderr_pipe[1], STDERR_FILENO);
-    posix_spawn_file_actions_adddup2(&actions, stdin_pipe[0], STDIN_FILENO);
-    
-    posix_spawn_file_actions_addclose(&actions, stdout_pipe[1]);
-    posix_spawn_file_actions_addclose(&actions, stderr_pipe[1]);
-    posix_spawn_file_actions_addclose(&actions, stdin_pipe[0]);
+        posix_spawn_file_actions_adddup2(&actions, stdout_pipe[1], STDOUT_FILENO);
```
tayloryoung6396, line 125: What even are these functions names...

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,380 @@
+#include <alsa/asoundlib.h>
+#include <spawn.h>
+#include <fmt/format.h>
+#include "Microphone.hpp"
+
+#include "extension/Configuration.hpp"
+#include "json.h"
```
Mrwerdo, line 7: @KipHamiltons, I'm just wondering about this include. It's a single header file located at: https://github.com/sheredom/json.h
How would you recommend adding it as a dependency?

module/input/Microphone/src/Microphone.cpp
c3144856: Overzealous checking on my part.  Changed to checking for -1 error condition and logs and error in that case.

module/input/Microphone/src/Microphone.cpp
c3144856: https://man7.org/linux/man-pages/man3/errno.3.html

"errno is thread-local" Should be ok according to this.

module/input/Microphone/src/Microphone.cpp
c3144856: Ok. I log the stderr message now.

module/input/Microphone/src/Microphone.cpp
c3144856: Changed to accept file paths now. Can be tested by typing "file /home/nubots/nuhear/audio/voice_001.wav" for example.

module/input/Microphone/src/Microphone.cpp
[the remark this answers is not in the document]
c3144856: Just to quiet down some overzealous warnings.

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,367 @@
...
+    bool success = false;
+    
+    pid_t pid = 0;
+    int stdout_pipe[2] = {};
+    int stderr_pipe[2] = {};
+    int stdin_pipe[2] = {};
+    posix_spawn_file_actions_t actions = {};
+    
+    char *path = getenv("PATH");
+    assert(path);
```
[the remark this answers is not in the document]
c3144856: Probably overkill on my part. Removed.

module/input/Microphone/src/Microphone.cpp
```diff
@@ -0,0 +1,367 @@
...
+    
+    
+    return true;
+}
+*/
+
+//handles: out parameter containing the stdin, stdout, stderr handles from the
+//spawned process
+//first char * in args array refers to the process name
+bool spawn_process(MicProcHandles *handles, char *const *args) {
```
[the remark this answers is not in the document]
c3144856: Ok I changed the out parameter to a reference and added a comment about the null termination of the args array

module/input/Microphone/src/Microphone.cpp
c3144856: Thanks, it's been changed to use NUClear::log and fmt::format now.

module/input/Microphone/src/Microphone.cpp
c3144856: I added another variable "actions_inited" denoted whether that structure had been initialized. Now we check that variable instead.

module/input/Microphone/src/Microphone.cpp
c3144856: https://man7.org/linux/man-pages/man3/posix_spawn_file_actions_addclose.3p.html
https://man7.org/linux/man-pages/man3/posix_spawn_file_actions_adddup2.3p.html

module/input/Microphone/src/Microphone.cpp
KipHamiltons: Same as we do with the catch header [here](https://github.com/NUbots/NUbots/blob/main/docker/Dockerfile#L220)

then it'll be in a folder such that you should be able to do `#include <json.h>`, I am pretty sure

module/input/Microphone/src/Microphone.cpp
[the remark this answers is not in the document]
c3144856: There's 100% less goto now Kip :-)

module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,391 @@
...
+            posix_spawn_file_actions_adddup2(&actions, stdout_pipe[1], STDOUT_FILENO);
+            posix_spawn_file_actions_adddup2(&actions, stderr_pipe[1], STDERR_FILENO);
+            posix_spawn_file_actions_adddup2(&actions, stdin_pipe[0], STDIN_FILENO);
+
+            posix_spawn_file_actions_addclose(&actions, stdout_pipe[1]);
+            posix_spawn_file_actions_addclose(&actions, stderr_pipe[1]);
+            posix_spawn_file_actions_addclose(&actions, stdin_pipe[0]);
+
+            if (posix_spawnp(&pid, (char const*) args[0], &actions, 0, (char* const*) args, envp) == 0) {
+                success = true;
```
tayloryoung6396, line 136: I know this technically works, but feels clunky for some reason, there's probably a better way to cleanup and return?

module/input/SpeechIntent/data/config/SpeechIntent.yaml
```diff
@@ -0,0 +1,3 @@
+# Controls the minimum log level that NUClear log will display
+log_level: DEBUG
+debug_mode: true
```
tayloryoung6396, line 3: I feel like "debug_mode" should just be using the nuclear log level debug

module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,391 @@
...
+                NUClear::log<NUClear::FATAL>(
+                    fmt::format("({}:{}) failed to read from spawned process stdout, errno = {}",
+                                __FILE__,
+                                __LINE__,
+                                errno));
+                return;
+            }
+            buffer[bytes_read] = 0;
+
+            if (enabled) {
```
tayloryoung6396, line 349: What is enabled?

module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,391 @@
...
+                char* intent_json = voice2json_recognize_intent(buffer);
+                if (!intent_json) {
+                    NUClear::log<NUClear::FATAL>(
+                        fmt::format("({}:{}) failed to recognize intent, errno = {}", __FILE__, __LINE__, errno));
+                }
+
+                std::unique_ptr<SpeechIntentMessage> intent = parse_voice2json_json(intent_json, strlen(intent_json));
```
tayloryoung6396, line 356: If the intent failed do you really want to parse it?

module/input/Microphone/src/Microphone.cpp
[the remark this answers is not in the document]
c3144856: Ok, I removed some casts.

module/input/SpeechIntent/src/SpeechIntent.cpp
c3144856: That is a bool that controls the behaviour of whether or not the module should respond to input reading from voice2json. This is useful when running in "transcribe-stream" mode (continuously running). When enabled the module will attempt to parse the voice2json json and emit it as a SpeechIntentMessage. When disabled it won't emit anything. Perhaps it deserves a more descriptive name.

module/input/SpeechIntent/src/SpeechIntent.cpp
c3144856: What looks clunky to you? All 3 return "paths" - spawn process success, pipe creation failure and spawn process failure need to cleanup those resources ? How should we proceed?

module/input/SpeechIntent/data/config/SpeechIntent.yaml
c3144856: It's been removed in favour or a command line argument "file" for testing.

module/input/SpeechIntent/src/SpeechIntent.cpp
c3144856: Ok, it's been renamed to "output_enabled"

tayloryoung6396: I'd have to read the code closer, but why run the reaction if the pasing is not enabled? You can just disable the reaction

c3144856: For the moment we are contiguously running the voice2json process in either transcribe-stream or transcribe-wav mode. This implies that we have to read from the voice2sjon stdout pipe when data is available. If we fail to read from it then the pipe will continue to fill up indefinitely. We may choose to change this in the future so that you can selectively turn voice2json on or off.

module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+                on<Trigger<SpeechIntentMsg>>().then([this](const SpeechIntentMsg& msg) { print_intent(msg); });
+                recognize_wav(wav_filename);
+            }
+            else if (args.size() > 1 && args[1] == "cli") {
+                this->config.transcribe_mode = TRANSCRIBE_MODE_FILE;
+                init();
+
+                on<Trigger<SpeechIntentMsg>>().then([this](const SpeechIntentMsg& msg) { print_intent(msg); });
+                on<IO>(STDIN_FILENO, IO::READ).then([this] {
```
tayloryoung6396, line 373: Same with this, maybe some comments would clear things up, someone else might have a different opinion, but enabling this reaction if the specific use case for it is needed seems like a cleaner way to have this optionally included

module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,405 @@
...
+        });
+    }
+
+    SpeechIntent::~SpeechIntent() {
+        if (voice2json_proc.stdout)
+            close(voice2json_proc.stdout);
+        if (voice2json_proc.stderr)
+            close(voice2json_proc.stderr);
+        if (voice2json_proc.stdin)
+            close(voice2json_proc.stdin);
```
tayloryoung6396, line 401: I like brackets

module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,405 @@
...
+    void print_intent(const SpeechIntentMsg& msg) {
+        NUClear::log(fmt::format("text = {}", msg.text.data()));
+        NUClear::log(fmt::format("intent = {}", msg.intent.data()));
+        NUClear::log("slots");
+        for (Slot slot : msg.slots) {
+            NUClear::log(fmt::format("{} = {}", slot.name.data(), slot.value.data()));
+        }
+    }
+
+    void SpeechIntent::init() {
```
tayloryoung6396, line 259: Do we use init functions elsewhere in the code base? Not saying it's wrong, just seems against the grain

tayloryoung6396 (commented):
General comments in these files would go a long way to understanding

module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+    SpeechIntent::SpeechIntent(std::unique_ptr<NUClear::Environment> environment)
+        : Reactor(std::move(environment)), config{} {
+
+
+        on<Configuration>("SpeechIntent.yaml").then([this](const Configuration& cfg) {
+            this->log_level = cfg["log_level"].as<NUClear::LogLevel>();
+        });
+
+        on<Trigger<CommandLineArguments>>().then([this](const CommandLineArguments& args) {
```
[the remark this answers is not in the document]
c3144856: Our reasoning for splitting it out - We don't want that code to be re-run in the case of the configuration file being reloaded.

module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,405 @@
...
+                init();
+
+                on<Trigger<SpeechIntentMsg>>().then([this](const SpeechIntentMsg& msg) { print_intent(msg); });
+                recognize_wav(wav_filename);
+            }
+            else if (args.size() > 1 && args[1] == "cli") {
+                this->config.transcribe_mode = TRANSCRIBE_MODE_FILE;
+                init();
+
+                on<Trigger<SpeechIntentMsg>>().then([this](const SpeechIntentMsg& msg) { print_intent(msg); });
```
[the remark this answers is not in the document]
c3144856: Ok, I see the problem now. init() won't be called in the case where this module is used by another module.

module/input/SpeechIntent/src/SpeechIntent.cpp
c3144856: Done :)

module/input/SpeechIntent/src/SpeechIntent.cpp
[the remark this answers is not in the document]
tayloryoung6396: @Bidski Would this be better handled as this? If you want to make sure it doesn't run again if the config file is changed but want it to run straight away?
handle = on<config> {
Do things
handle.disable();
}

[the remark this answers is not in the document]
Bidski: Using command line arguments in this way seems a little pointless tbh. I would just move all of this logic into `on<Configuration>` and not use `CommandLineArguments` at all for this.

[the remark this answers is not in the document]
Mrwerdo: Why do you think it's pointless?

The reason why we're trying to access command line argument is so that we have some way of testing the module with different audio sources. In particular, in a container environment, we can't access the microphone, so the audio source is instead a wav file. Are you recommending that we hard code a path into the configuration file that points to a wav file?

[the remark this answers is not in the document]
Bidski: Its pointless because you can achieve all of the same functionality using the configuration file.
You could easily set up a `sound_source` entry in the config file, and if that represents a valid file (you can use `std::filesystem` for this) then you load that as an audio file, if `sound_source` is `mic` (or some other appropriate keyword) then you look for the appropriate hardware and get your audio data from there.

Also, are you aware of [this](https://stackoverflow.com/a/48795232/4795763)? Not sure if this would work in WSL.

Mrwerdo:
I'm closing this pull request because the builds are unable to succeed.

[The new pull request is located here](https://github.com/NUbots/NUbots/pull/737).

PR #695, opened 2021-09-07
Fix some edge case bugs in dockerise

CMurtagh-LGTM:
Whilst looking at running our code base with podman I ran into a couple bugs that could possibly happen even when running with docker.


tools/utility/dockerise/platform.py
```diff
@@ -25,7 +25,7 @@ def selected(repository):
 
         names = [
             tag.split(":")[-1]
-            for tag in img_info[0]["RepoTags"]
```
KipHamiltons, line 28: Double list comprehension makes my brain hurt. Could you please split the assignment of `names` into smaller blocks? I am having trouble _comprehending_ what this does at all. (I wouldn't class this as scope creep, because the replacement you have offered takes the complexity over a threshold where my brain hurts.)

tools/utility/dockerise/platform.py
```diff
@@ -38,7 +38,7 @@ def selected(repository):
         else:
             print("WARNING There are multiple platforms with the same image tag.")
             print("        The possible tags are [{}]".format(", ".join(names)))
-            platform = list(sorted(names))[0]
```
KipHamiltons, line 41: Very curious what prompted this change

tools/utility/dockerise/run.py
```diff
@@ -17,7 +17,11 @@
 
 def _is_docker():
     path = "/proc/self/cgroup"
-    return os.path.exists("/.dockerenv") or os.path.isfile(path) and any("docker" in line for line in open(path))
+    return (
+        os.path.exists("/.dockerenv")
+        or os.path.isfile(path)
```
KipHamiltons, line 22: Can you please add brackets around this condition?

reading this, it's not clear to me when:
- the first condition is false
- the second condition is true

what happens? 

If operator precedence of `or` and `and` in python is equivalent, then the third condition never matters - the `and` requires the second condition to be true, but if the second condition is true and the precedence is the same, then the `or` is true, and the whole statement is true, because short-circuit.
Please just make this more clear, either way.

KipHamiltons (changes requested):

tools/utility/dockerise/platform.py
TrentHouliston: Likely an indexed access, what is returned by sorted is an iterator, not a list.
So previously `platform[0]` would have been invalid

tools/utility/dockerise/platform.py
TrentHouliston: This is more of a filter than a double list comprehension.
But I agree with kip in this case, easier to add a new line that splits out your `/`

tools/utility/dockerise/platform.py
KipHamiltons: [the python docs](https://docs.python.org/3/library/functions.html#sorted) say that it returns a new list though?
Based on that, it seems to me that this cast to a list is simply just redundant.

CMurtagh-LGTM: I was getting an error when I wrote it, now I cannot reproduce the error :slightly_frowning_face:. It is redundant code though, so I'm going to leave it in the pr.

KipHamiltons (commented):
LGTM

KipHamiltons (approved):

PR #697, opened 2021-09-07
Fixes clang build

KipHamiltons:
I used clang12 (which isn't merged yet), which had some _fresh_ issues. There are some decisions made in here which could do with some review. I'll point those out with comments.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,8 @@ add_compile_options(-Wall -Wpedantic -Wextra)
 # This is needed everywhere so that we can catch segmentation faults
 add_compile_options(-fnon-call-exceptions)
 
+add_compile_options(-std=c++17)
+
 # Make NUClear use the system clock, as we are syncing it with ptp
 add_compile_definitions(NUCLEAR_CLOCK_TYPE=std::chrono::steady_clock)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,7 @@ add_compile_options(-Wall -Wpedantic -Wextra)
 # This is needed everywhere so that we can catch segmentation faults
 add_compile_options(-fnon-call-exceptions)
 
+# We use C++17
 add_compile_options(-std=c++17)
 
 # Make NUClear use the system clock, as we are syncing it with ptp
--- a/nuclear/message/include/message/conversion/proto_matrix.hpp
+++ b/nuclear/message/include/message/conversion/proto_matrix.hpp
@@ -1427,9 +1427,8 @@ namespace message::conversion {
         proto.mutable_v()->resize(vector.size());
 
         // Copy the data across
-        Eigen::Map<::message::conversion::math::cvec>(
-            const_cast<uint8_t*>(reinterpret_cast<const uint8_t*>(proto.mutable_v()->data())),
-            proto.v().size()) = vector;
+        Eigen::Map<::message::conversion::math::cvec>(reinterpret_cast<uint8_t*>(proto.mutable_v()->data()),
+                                                      proto.v().size()) = vector;
 
         return proto;
     }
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -88,6 +88,9 @@ add_compile_options(-Wall -Wpedantic -Wextra)
 # This is needed everywhere so that we can catch segmentation faults
 add_compile_options(-fnon-call-exceptions)
 
+# We use C++17
+add_compile_options(-std=c++17)
```
KipHamiltons, line 92: GCC defaults to C++17, but clang defaults to C++14. We use a few C++17 features a decent bit - e.g. std::filesystem. I think this is reasonable.

nuclear/message/include/message/conversion/proto_matrix.hpp
```diff
@@ -1427,8 +1427,9 @@ namespace message::conversion {
         proto.mutable_v()->resize(vector.size());
 
         // Copy the data across
-        Eigen::Map<::message::conversion::math::cvec>(reinterpret_cast<uint8_t*>(proto.mutable_v()->data()),
-                                                      proto.v().size()) = vector;
+        Eigen::Map<::message::conversion::math::cvec>(
+            const_cast<uint8_t*>(reinterpret_cast<const uint8_t*>(proto.mutable_v()->data())),
```
KipHamiltons, line 1431: This deserves scrutiny. @TrentHouliston is this undefined??

shared/utility/math/filter/ParticleFilter.hpp
```diff
@@ -98,7 +98,7 @@ namespace utility::math::filter {
          */
         ParticleFilter(const StateVec& mean,
                        const StateMat& covariance,
-                       const ResampleMethod& resample_method_ = ResampleMethod::RESIDUAL | ResampleMethod::SYSTEMATIC) {
+                       const int& resample_method_ = ResampleMethod::RESIDUAL | ResampleMethod::SYSTEMATIC) {
```
KipHamiltons, line 101: `RESIDUAL | SYSTEMATIC` isn't  an enum element. The default wouldn't bind with clang. If we want to do this differently, let me know.

KipHamiltons (commented):
The parts worth reviewing ;P

nuclear/message/include/message/conversion/proto_matrix.hpp
TrentHouliston: This feels dirty, pretty sure it's just the Eigen map type that should be const

TrentHouliston: Actually that won't work either, Why is protobuf returning a const for it's mutable_v data now? when did that change?

KipHamiltons: I am looking everywhere for a definition of `mutable_v`, but I can't find anything. This class confuses me hahaha

CMurtagh-LGTM: @KipHamiltons `mutable_v` is defined [here](https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#embeddedmessage) with foo being [v](https://github.com/NUbots/NUbots/blob/main/nuclear/message/proto/Vector.proto#L20).

shared/utility/math/filter/ParticleFilter.hpp
CMurtagh-LGTM: How so, what is the error message? 
Did you try
```suggestion
                       const int& resample_method_ = ResampleMethod(ResampleMethod::RESIDUAL | ResampleMethod::SYSTEMATIC)) {
```

nuclear/message/include/message/conversion/proto_matrix.hpp
KipHamiltons: Thanks for the links! They pointed me in the right direction. Given that it's only not compiling on clang, it seems to me that protobuf didn't change their interface, but the interaction with clang12 has some unknown idiosyncrasy. I'm going to try updating our protobuf version.

KipHamiltons: It builds without the const_cast on the the newest protobuf release. I'll submit a PR to update that, then revert these changes.

shared/utility/math/filter/ParticleFilter.hpp
KipHamiltons: Explicitly casting worked. Thanks for the suggestion :+1:

CMurtagh-LGTM (approved):
LGTM

shared/utility/math/filter/ParticleFilter.hpp
TrentHouliston: Explicit casting is wrong. I do see where c++ is warning about. It's saying that the value that comes out of bitwise oring the two is not an enum. We are using the enum as constant flags rather than as enums. The solution might be to make this enum just constants

PR #699, opened 2021-09-08
Add game functionality to webots tool

Y0sh1dk:
Adds `game` subcommand to webots tool.

Usage:
```
# Start single robot instance
./b webots run <role>

# Start a full game (localhost)
./b webots game -r <num of robots> <role>

# Start a full game (webots running remotely)
./b webots game -a <address of webots instance> -r <num of robots> <role>
```
NOTE: `pip install -r requirements.txt` will need to be ran to install honcho

UPDATE: CTRL+C Now works correctly, Overriding the method that honchos signal handler calls with exec_stop() so it now gets called on CTRL+C


JosephusPaye:
Code looks good, though I haven't tested it.

ysims:
Had a look. It doesn't connect with the default settings in NUWebots.

```bash
12:05:18 Robot_1 (red):  | Webots (╯°□°）╯︵ ┻━┻: Cannot connect to server: 127.0.0.1:1001
12:05:18 Robot_1 (red):  | Webots (╯°□°）╯︵ ┻━┻: Failed to connect to server.
``` 

I think it should be `127.0.1.1:10001` for the default set up.

KipHamiltons:
Did you resolve the `ctrl + c` issue @Y0sh1dk? If not, could you please elaborate so that we can get people to take a look?

Y0sh1dk:
@ysims Woops!, looks like I missed a `0` in the port number, have fixed that up and added logic for the port numbers based on. I don't currently have Webots setup to test, but this should work now.
```
"red": {
...
    "ports": [
      10001,
      10002,
      10003
    ]
  },
  "blue": {
...
    "ports": [
      10021,
      10022,
      10023
    ]
  }
```

Y0sh1dk:
@KipHamiltons PR description updated :+1:

KipHamiltons:
@TrentHouliston @Bidski @CMurtagh-LGTM @JosephusPaye any ideas on how to get `ctrl + c` working/propagating to each robot container? The signal handling seems to be getting in the way, and we couldn't find an elegant solution

Bidski:
> @TrentHouliston @Bidski @CMurtagh-LGTM @JosephusPaye any ideas on how to get `ctrl + c` working/propagating to each robot container? The signal handling seems to be getting in the way, and we couldn't find an elegant solution

Im gonna need more context

Bidski:
Oops :P

JosephusPaye:
> [...] any ideas on how to get `ctrl + c` working/propagating to each robot container? The signal handling seems to be getting in the way, and we couldn't find an elegant solution

What does it do at the moment? Kill the parent process but not each of the container processes?

KipHamiltons:
I think that summarises the behaviour pretty well. Basically, you run the game command, it spins up a container for each player, then you can control c the game script, but the containers are still running trying to play. They can be killed by calling docker stop or signalling the pids directly, but ideally we would be able to just control c and kill it all

KipHamiltons:
@Y0sh1dk tried making a signal handler in python, but the dep he uses defines its own handler, and there were some issues with doing it that way

Bidski:
Do we need honcho? Can we not just use python subprocess to spawn the extra processes? Then your custom signal handler will also work

Y0sh1dk:
@Bidski  We dont *need* it, But it does multiplex the outputs with pretty colors :)

Y0sh1dk:
**UPDATE:**  CTRL+C Now works correctly, Overriding the method that honchos signal handler calls with exec_stop() so it now gets called on CTRL+C

Bidski:
> @Bidski We dont _need_ it, But it does multiplex the outputs with pretty colors :)

I mean that is beautiful, but the cost of having that seems a little expensive. So I see three options.

1. Make a PR for honcho so that they will optionally call a custom sigint handler as part of their handler (I would suggest overriding it completely, but it looks like they have their own logic in there as well)
1. Fork honcho and add the functionality we desire (you have to do this to make a PR anyway)
1. Re-implement some of what they are doing so you get that nice multiplexing (won't be too difficult to achieve, but you are re-implementing what they have) (i.e. using subprocess)

The only other possibility is to not use Ctrl+C to kill things and instead implement an extra command to handle all of the killing (avoiding the need for signal handlers altogether, I am honestly not a big fan of how often programs want you to Ctrl+C to kill them).

NOTE: I would be careful about just overriding their `terminate` function in case they have specific logic that is being run to safely/nicely clean up the processes

Y0sh1dk:
> Make a PR for honcho so that they will optionally call a custom sigint handler as part of their handler (I would suggest overriding it completely, but it looks like they have their own logic in there as well)

I did think about this, but the project seems a bit dead. Forking, adding the functionality and using the fork instead seems like a good option.

> implement an extra command

I did have this as `./b webots stop`, but removed in this commit https://github.com/NUbots/NUbots/commit/83fa7089cf9c134a723c11db7ae9f75b4752727f once the signal handler worked. I can add it back if we want to provide both ways to kill the containers?

> NOTE: I would be careful about just overriding 

Good point, They have logic to only kill the processes only once when you spam CTRL+C. I have overridden the [_killall](https://github.com/nickstenning/honcho/blob/master/honcho/manager.py#L156) method instead which keeps this logic inplace.

Bidski:
Rather than searching for all containers that are spawned from a certain image can you use the `--name` option for `docker run` and then use those `name`s to kill the containers?

TrentHouliston:
> Rather than searching for all containers that are spawned from a certain image can you use the `--name` option for `docker run` and then use those `name`s to kill the containers?

That often caused me grief, since sometimes the containers don't die and then you can't ever start something with the same name again

Bidski:
> > Rather than searching for all containers that are spawned from a certain image can you use the `--name` option for `docker run` and then use those `name`s to kill the containers?
> 
> That often caused me grief, since sometimes the containers don't die and then you can't ever start something with the same name again

My concern with killing everything from a particular image is that you could kill something that you dont want to kill (I know its not very likely that you will have **_that_** many containers running on a single machine, but still)

Y0sh1dk:
> Rather than searching for all containers that are spawned from a certain image can you use the --name option for docker run and then use those names to kill the containers?

I did initially start doing it like this, But there might be situations where you are running multiple 'games' running from the one machine, if the containers were named you would have to keep track how the 'games' running in order to name the containers correctly. But then there is also still the issue of not being able to kill a single 'game' if multiple are running.

Bidski:
> > Rather than searching for all containers that are spawned from a certain image can you use the --name option for docker run and then use those names to kill the containers?
> 
> I did initially start doing it like this, But there might be situations where you are running multiple 'games' running from the one machine, if the containers were named you would have to keep track how the 'games' running in order to name the containers correctly. But then there is also still the issue of not being able to kill a single 'game' if multiple are running.

If you have multiple games running on one machine and you are killing all containers started from a particular image then when you kill one game you will kill all games?

Y0sh1dk:
> If you have multiple games running on one machine and you are killing all containers started from a particular image then when you kill one game you will kill all games?

Correct, It was the tradeoff I made for being able to start multiple games without keeping track of current running games somehow.

Thinking about it, I can probably name the containers something like `<random string>_Robotx`, keeping track of the string so you can only stop the containers started for that game. I will give that a go now.

Bidski:
> > If you have multiple games running on one machine and you are killing all containers started from a particular image then when you kill one game you will kill all games?
> 
> Correct, It was the tradeoff I made for being able to start multiple games without keeping track of current running games somehow.
> 
> Thinking about it, I can probably name the containers something like `<random string>_Robotx`, keeping track of the string so you can only stop the containers started for that game. I will give that a go now.

Thats pretty much what I was thinking, but just using `<random_string>`

tools/webots.py
```diff
@@ -63,6 +72,11 @@ def register(command):
     run_subcommand = subparsers.add_parser("run", help="Run the simulation docker image for use with Webots")
     run_subcommand.add_argument("role", help="The role to run")
 
+    game_subcommand = subparsers.add_parser("game", help="Run a full game")
+    game_subcommand.add_argument("role", action="store", help="The role to run")
+    game_subcommand.add_argument("-r", "--robots", action="store", dest="num_of_robots", required=True)
+    game_subcommand.add_argument("-a", "--sim-address", action="store", default="127.0.0.1", dest="sim_address")
```
Bidski, line 78: Add help messages

tools/webots.py
```diff
@@ -174,31 +188,92 @@ def exec_build(target, roles, clean, jobs):
...
+    process_manager = Manager()
+
+    # Override honcho.manager.terminate() method, this is called by its signal handler on CTRL+C
+    process_manager._killall = exec_stop
+
+    # Add all robot run commands to process_manager
+    for i in range(1, int(num_of_robots) + 1):
+
+        robot_color, port_num = (
+            ("red", 10000 + i) if i <= int(num_of_robots) / 2 else ("blue", 10020 + i - int(int(num_of_robots) / 2))
```
Bidski, line 218: However if you set `type=int` in `game_subcommand.add_argument("-r", "--robots", action="store", dest="num_of_robots", required=True)` then you should be confident in `num_of_robots` being an `int` and so the cast should be unnecessary.
https://docs.python.org/3/library/argparse.html#type
```suggestion
            ("red", 10000 + i) if i <= int(num_of_robots // 2) else ("blue", 10020 + i - int(num_of_robots // 2)
```

tools/webots.py
```diff
@@ -3,22 +3,31 @@
...
 
+random.seed(datetime.now())
+
 # The docker image details for Robocup
 ROBOCUP_IMAGE_NAME = "robocup-vhsc-nubots"  # Provided by the TC and shouldn't be changed
 ROBOCUP_IMAGE_TAG = "robocup2021"  # Submitted in our team_config.json, shouldn't be changed here unless changed there
 ROBOCUP_IMAGE_REGISTRY = "079967072104.dkr.ecr.us-east-2.amazonaws.com/robocup-vhsc-nubots"  # Provided by the TC
 
+# Generate 4 digit random int to act as a unique identifier for this game
+GAME_IDENTIFIER = str(random.randint(1000, 9999))
```
Bidski, line 29: How about something [like this](https://stackoverflow.com/a/27302466/4795763) instead? Might be overkill?

Bidski (changes requested):

tools/webots.py
Y0sh1dk: Nice find!

tools/webots.py
Y0sh1dk: I think that is likely overkill for now, If we do ever run into issues we can definietly add something like that though.

Bidski (changes requested):

tools/webots.py
```diff
@@ -63,6 +72,26 @@ def register(command):
...
+        required=True,
+        help="The number of robot instances to start",
+    )
+    game_subcommand.add_argument(
+        "-a",
+        "--sim-address",
+        action="store",
+        default="127.0.0.1",
+        dest="sim_address",
+        help="Address that robot instances use to connect to Webots",
```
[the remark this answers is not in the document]
Y0sh1dk: Yeah, so just defaults to 127.0.0.1 when running locally, but can give it whatever address the webots server is running on.

Bidski (approved):

PR #706, opened 2021-09-11
Refactors ResampleMethod to prevent invalid configurations

KipHamiltons:
### Resample Method Refactor

This changes the `ResampleMethod` enum to only include the disjoint methods, with a bool to indicate if residual sampling should be used. This prevents invalid configurations, allowing us to simplify the logic and remove the error handling code.

The doc-tags of a lot of `ParticleFilter` have become stale/outdated, but I will update them in a follow-up PR which is almost done on my local machine


Bidski:
This seems quite excessive. What was clangs complaint?

KipHamiltons:
I disagree that it's excessive. I think it makes all of it much easier to understand and harder to misuse, both of which are important properties. The clang error is solveable, but it seems absurd to use an enum when we are trying to use combinations of the settings, which aren't represented as members of the enum. The Boolean logic is also unnecessarily unreadable.

shared/utility/math/filter/ParticleFilter.hpp
```diff
@@ -92,19 +94,13 @@ namespace utility::math::filter {
          * filter.
          * @param covariance The covariance of the multivariate normal distribution. Used for sampling the initial
          * particles in the filter.
-         * @param resample_method_ The method to use when resampling particles during the time update step.
-         * @param residual_method_ When using ResampleMethod::RESIDUAL for the resample method, this method will be used
-         * when the residual method needs to resample residual particles.
+         * @param resample_method_ The method to use when resampling particles during the time update step
```
Bidski, line 97: You are missing one

shared/utility/math/filter/ParticleFilter.hpp
```diff
@@ -127,12 +123,9 @@ namespace utility::math::filter {
...
          */
         ParticleFilter(const std::vector<std::pair<StateVec, StateMat>>& hypotheses,
-                       const ResampleMethod& resample_method_ = ResampleMethod::RESIDUAL | ResampleMethod::SYSTEMATIC) {
-
-            if (!check_resample_method(resample_method_)) {
-                throw std::runtime_error("Invalid setting for resampling method");
-            }
-            resample_method = resample_method_;
+                       const ResampleMethod& resample_method_ = ResampleMethod::SYSTEMATIC,
+                       const bool& use_residual_resampling_   = true)
```
Bidski, line 127: You need to update the comment block for this function too

Bidski (changes requested):

shared/utility/math/filter/ParticleFilter.hpp
KipHamiltons: I was avoiding changing the block comments in this PR, seeing as I am going to submit a PR updating all of them more thoroughly immediately after merging this one, but I'll fix these :shrug:

CMurtagh-LGTM (approved):
I think the code is more simple this way :stuck_out_tongue:

Bidski (approved):
LGTM

PR #707, opened 2021-09-12
Forces CI_BUILD variable off because we aren't ready for Werror/clang-tidy

KipHamiltons:
### Disabling the Effects of the CI_BUILD variable

Nerfing this variable allows us to "enable" it on CI, with a "one-line" fix to enable it by removing the `set` call.

Addresses issues with #353, regarding changes unrelated to the PR.


TrentHouliston (approved):
Sounds good

PR #709, opened 2021-09-13
NUsight CI fix for missed files when linting in pipeline

KipHamiltons:
### NUsight Formatting CI Fix

ATM when we run `yarn eslint[:fix]`, it only runs on files matching `src/**/*.ts{,x}` This restriction means that some of our typescript files aren't being covered by the checks, such as `build_scripts/build.ts`. 

This PR rectifies that by matching against all TS files, and fixes the two errors that the linter had with the previously uncovered files.


BrendanAnnable (approved):
LGTM

PR #711, opened 2021-09-13
Rename master->main

Y0sh1dk:
Update references to main due to renaming of master branch.

`./b target` was telling me my docker build context had changed but I hadn't touched it!, then noticed it was `diff`ing with master and not main :thinking:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,11 +5,11 @@ env:
   DOCKER_BUILDKIT: 1
 
 steps:
-  # Push up new images to DockerHub when they get to master
+  # Push up new images to DockerHub when they get to main
   # Makes development faster as you can just pull the images rather than building them yourself
-  # If you make modifications to the Dockerfile it should resume from the common point in master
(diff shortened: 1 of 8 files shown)


KipHamiltons:
Almost :stuck_out_tongue: You missed the build badges on the README. They'll need to be updated after GHA is approved, but we haven't done that yet :sweat_smile:

Y0sh1dk:
I decided to leave them because github actions should be in soon (hopfully :sweat_smile: ), But there you go, updated the references in there too.

KipHamiltons (approved):
Thank you sir

JosephusPaye:
The badge URLs might need editing beyond just `master` -> `main` (e.g. the numeric identifier might be different for the different branch).

Also, Buildkite was still building `master` as your main branch, but I've updated that.

Y0sh1dk:
>The badge URLs might need editing beyond just master -> main (e.g. the numeric identifier might be different for the different branch).

Just noticed this, I dont have access to the pipeline settings in buildkite where you grab the url from, are you able to update it @JosephusPaye or @KipHamiltons ?

Issue #705, opened 2021-09-10
Goal::MeasurementType only ever having the value of CENTRE and RobotModel.hpp returning Zero matrix


abihall:
So, following talk on issue #450 and #315, it is found that the Goal::MeasurementType only ever takes the value of CENTRE and never that of LEFT_NORMAL or RIGHT_NORMAL. 

As a result, half of RobotModel.hpp will not be reached and most of the functions defined will return the Zero matrix since Goal::MeasurementType doesn't take on any value but CENTRE.

abihall:
In RobotModel.hpp, both getSquarePostCamSpaceNormal and getCylindricalPostCamSpaceNormal will return the Zero Matrix always. I can't see anywhere where these are used anyway. What is the point of having them there?

KipHamiltons:
We don't use those functions, nor do we have plans to use them. Their invalidity for the only measurement type we actually use compounds things to the point where they're properly dead and should definitely be removed.

Given that we don't have a consensus on the `MeasurementType` enum and only use one measurement type in the code, I think it should be removed too. We should update where it's used so it's clear exactly what the measurement means too, because without the enum, the extra context is valuable.

If you made PRs to remove those, I would approve them :stuck_out_tongue_winking_eye: (you should link #450 to the PR which removes the enum, so it closes once it's merged)

PR #712, opened 2021-09-13
Remove unused functions from RobotModel

abihall:
**Removes unused RobotModel Functions**

This PR removes the unused functions found in RobotModel.cpp which include `getCylindricalPostCamSpaceNormal` and `getSquarePostCamSpaceNormal`. These currently return the Zero vector due to `Goal::MeasurementType` enumeration and are never used. 

This is related to issue #705


KipHamiltons:
I am very keen to remove these. They're properly dead. They are not called and even if they were, they don't apply to the only measurement type we actually use. On top of that, no one can on the team can confidently say they understand the internals because I spent months trying to get an insight when I was Eigenifying it because the logic was _invalid_ during its arma iteration (it would throw an exception if it was ever run, which it wasn't). I ended up piecing together what I thought the intentions of these functions was (after sitting down with @tayloryoung6396) and trying to approximate that with Eigen, but that means that we should have absolutely zero confidence about their correctness. If they had tests, we could have some, but they don't.

Basically, it's unused and almost certainly buggy code that no one fully understands. I won't approve without either @TrentHouliston or @Bidski 's blessing, but IMO this is a clear cut case.

TrentHouliston (commented):
I don't mind these being removed.
For what they are supposed to do, they are supposed to work out what the normal vector should be from your camera of each side of the goal post, for both cylindrical goals and square goals. THey went along with the other measurement types when we tried to use the normals as measurements (didn't work so well)

KipHamiltons (approved):
LGTM!

Issue #450, opened 2021-05-01
We don't know what the message::vision::Goal::MeasurementType enum means


KipHamiltons:
As mentioned in #315 [here](https://github.com/NUbots/NUbots/pull/315#discussion_r499075484), we don't know what the goal measurement types actually mean/do. We need to figure out what they do, then document that field in `Goal.proto`, and make sure that the use of this message is consistent with that finding.

KipHamiltons:
This was resolved by #713. We now only have two types of measurement, and we have documented what they mean. Closing

PR #713, opened 2021-09-13
Remove the measurement types that aren't unknown and centre

abihall:
**Remove the MeasurementType enumeration options**

This PR removes the `Goal::MeasurementType` enumeration since it always take on the value of `CENTRE`. The `MeasurementType` was used once in `GoalDetector`. These changes have been made following discussion in issues #705 and #450.

This PR can only be merged following the merge of PR #712


TrentHouliston:
Don't remove the measurement type, despite the fact that it's not used right now it very likely could be used in the future and it is best to make the ball and goal measurements match.

You can probably remove some of the enums here, but the point of this is to have different independent ways of measuring the goals. For example, using the width of the post, or projection to the bottom etc. For the ball as well width and projection-based measurements go in the type.

abihall:
> Don't remove the measurement type, despite the fact that it's not used right now it very likely could be used in the future and it is best to make the ball and goal measurements match.
> 
> You can probably remove some of the enums here, but the point of this is to have different independent ways of measuring the goals. For example, using the width of the post, or projection to the bottom etc. For the ball as well width and projection-based measurements go in the type.

I feel like having a unused enumeration in the proto file is not ideal, if different measurementTypes are needed in the future, it would not take much effort to implement them again and with correct understanding of what they actually mean.

TrentHouliston:
You don't want to end up in the situation where each object type has its own way of doing things, e.g. ball has a measurement type and goal doesn't etc. That's just adding difficulty when people want to learn how to use them.
Also if someone now works on localisation without having this enum, they'll write code as if it doesn't exist.
Then adding it back in the future will be far more difficult because it wasn't accounted for.

On the other hand, if it does exist with a single value, at least those developing localisation will know it needs to exist and handle it.

You want the way these to protos to be used to match
https://github.com/NUbots/NUbots/blob/main/shared/message/vision/Ball.proto
https://github.com/NUbots/NUbots/blob/main/shared/message/vision/Goal.proto

And it wouldn't be an unused enum, it'd just be an enum with a single option, which I would comment in this PR so people know what it means.

KipHamiltons:
If we want to keep this enum, I would like what each of the fields means documented. There's not a consensus on what the fields mean (#450), and if the enum is going to be useful, it should be clear.

shared/message/vision/Goal.proto
```diff
@@ -37,14 +37,10 @@ message Goal {
...
-    /// TODO(Vision team): Work out what these mean. See issue #450
     enum MeasurementType {
-        UNKNOWN_MEASUREMENT = 0;
-        CENTRE              = 1;
-        LEFT_NORMAL         = 2;
-        RIGHT_NORMAL        = 3;
-        TOP_NORMAL          = 4;
-        BASE_NORMAL         = 5;
+        UNKNOWN     = 0;
+        CENTRE      = 1; //distance from camera to bottom center of post
```
KipHamiltons, line 42: Just to make it match the style of the others - I know a guy who spent a while on them :laughing: 
```suggestion
        /// The distance is from the camera to bottom centre of post
        CENTRE      = 1;
```

KipHamiltons (approved):
LGTM

PR #716, opened 2021-09-17
Updates the VisualMesh module README

ysims:
This PR fills in the README for the VisualMesh module.

### Reviewers, Note

This is just what I got from reading the module, so 100% needs looking over. I also think some of it is a bit wordy, so someone who isn't familiar with this module should check if this makes sense to them.

### DIFF
--- a/module/vision/VisualMesh/README.md
+++ b/module/vision/VisualMesh/README.md
@@ -11,7 +11,15 @@ This module requires on the following configuration files to be set:
 - `data/config/networks` should include a `yaml` file with Visual Mesh network configuration, which can be obtained from tools in the [Visual Mesh](https://github.com/Fastcode/VisualMesh) repository after training. A guide on updating this can be found on [NUbook](https://nubook.nubots.net/guides/main/maintaining-subsystems#vision).
 - `data/config/VisualMesh.yaml` should include information on what cameras and network configuration to use.
 
-Once these are set appropriately, this module will trigger on any `Image`s emitted in the system and run the Visual Mesh as per the configuration.
+Once these are set appropriately, this module will trigger on any `message::input::Image`s emitted in the system.
+
+The module uses the `./visualmesh/VisualMeshRunner` to run the Visual Mesh.
+
+`VisualMeshRunner` gets a `LoadedModel` from `./visualmesh/load_model`, which uses the network configuration specified in `./data/config/VisualMesh.yaml` to create a network model through the `VisualMesh` library. This `LoadedModel` is the Visual Mesh model that the `Image` will be run on, and is only updated if the configuration changes.
+
+This `LoadedModel` is then used in the `VisualMeshRunner` anytime an `Image` needs to be processed. The `VisualMeshRunner` takes the `Image` and `Hcw` (world to camera homogeneous transformation), gets the lens and height data from these and passes it through to the `VisualMesh` itself to get the results. The `VisualMeshRunner` then returns the results and the module emits those results as a `message::vision::VisualMesh` message. This will rerun with a new `Image` and `Hcw` every time a new `Image` is received.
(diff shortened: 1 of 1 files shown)


module/vision/VisualMesh/README.md
```diff
@@ -3,12 +3,32 @@ VisualMesh
...
+- `data/config/VisualMesh.yaml` should include information on what cameras and network configuration to use.
+
+Once these are set appropriately, this module will trigger on any `message::input::Image`s emitted in the system.
+
+The module uses the `./visualmesh/VisualMeshRunner` to run the Visual Mesh.
+
+`VisualMeshRunner` gets a `LoadedModel` from `./visualmesh/load_model`, which uses the network configuration specified in `./data/config/VisualMesh.yaml` to create a network model through the `VisualMesh` library. This `LoadedModel` is the Visual Mesh model that the `Image` will be run on, and is only updated if the configuration changes.
+
+This `LoadedModel` is then used in the `VisualMeshRunner` anytime an `Image` needs to be processed. The `VisualMeshRunner` takes the `Image` and `Hcw` (world to camera homogeneous transformation), gets the lens and height data from these and passes it through to the `VisualMesh` itself to get the results. The `VisualMeshRunner` then returns the results and the module emits those results as a `message::vision::VisualMesh` message. This will rerun with a new `Image` and `Hcw` every time a new `Image` is received.
```
KipHamiltons, line 20: > This will rerun with a new `Image` and `Hcw` every time a new `Image` is received.

this makes it sound like the two `Image`s mentioned are distinct? I think this could be clearer

BryceTuppurainen, line 20: I suggest adding a link to the NUbook Mathematics page over the (world to camera homogenous transformation) section. Just found that the resources there were quite useful when I was getting up to speed for behaviour:

```suggestion
This `LoadedModel` is then used in the `VisualMeshRunner` anytime an `Image` needs to be processed. The `VisualMeshRunner` takes the `Image` and `Hcw` [(world to camera homogenous transformation)](https://nubook.nubots.net/system/foundations/mathematics#homogeneous-transformations)
, gets the lens and height data from these and passes it through to the `VisualMesh` itself to get the results. The `VisualMeshRunner` then returns the results and the module emits those results as a `message::vision::VisualMesh` message. This will rerun with a new `Image` and `Hcw` every time a new `Image` is received.
```

module/vision/VisualMesh/README.md
```diff
@@ -3,12 +3,32 @@ VisualMesh
...
+
+Note that this module does not receive a `message::input::Sensors` message. It uses the `Hcw` matrix from the `Image` message, as this will capture the robot's orientation at the time the `Image` is first processed to reduce error.
+
+## Consumes
+
+- `message::input::Image` the image to process
 
 ## Emits
 
+- `message::vision::VisualMesh` the mesh results for the given image
```
BryceTuppurainen, line 30: Specifying what these mesh results represent at a high level may be useful in a short one or two liner. Maybe:

the mesh results for the given image stored as a `data-structure listed here` of rays

Just to be a little more verbose and provide a clearer idea of what is being emit which can then be used to reason out the why in other modules.

Reading through VisualMesh.cpp and VisualMesh.proto it appears that msg is a class storing the object-to-camera transform, an ID, camera to world transform along with the image timestamp, and a series of rays referring to sample points in the mesh.

ysims: I've updated it, let me know if it's good

KipHamiltons (approved):
LGTM

BryceTuppurainen (approved):
LGTM

Bidski (approved):
Seems like a rudimentary overview, but seems roughly correct

PR #718, opened 2021-09-21
Fix the NUsight docker container

JosephusPaye:
Currently, the [`generate_messages.js`](https://github.com/NUbots/NUbots/blob/main/nusight2/build_scripts/generate_messages.js) script can't find the proto files in the docker container, since only the `nusight2`folder is mounted. This mounts the top-level NUbots folder as well, which fixes the issue.

The change to mono repo broke the script and we didn't notice.

### DIFF
--- a/nusight2/docker-compose.yml
+++ b/nusight2/docker-compose.yml
@@ -3,9 +3,9 @@ services:
   nusight:
     image: node:14
     volumes:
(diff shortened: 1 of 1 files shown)


Y0sh1dk (approved):
LGTM

TrentHouliston (approved):

PR #720, opened 2021-09-22
Removes Unused and Unnecessary PortAudio Dep

KipHamiltons:
### Removes PortAudio

This dependency may have been previously required by eSpeak. However, it has been superseded by `pcaudio`, which isn't required by the version of eSpeak we use.


KipHamiltons:
This is wrong. see here https://github.com/espeak-ng/espeak-ng/blame/467f3f565f9f8d6d1de62b854e3c2b21a52fd8a2/docs/guide.md#L42

PR #723, opened 2021-09-25
Add clearing of filtered_sensors after each loop

abihall:
**Add clearing of filtered_sensors after each loop in sensor filter**
filtered sensors was not cleared after each loop through the sensors when resetting the sensor filter. This makes sure it is cleared after using it to calculate forward kinematics and averages on a iteration

### DIFF
--- a/module/input/SensorFilter/src/SensorFilter.cpp
+++ b/module/input/SensorFilter/src/SensorFilter.cpp
@@ -238,6 +238,8 @@ namespace module::input {
 
                         // Accumulator CoM readings
                         rMFt += calculateCentreOfMass(model, filtered_sensors->Htx).head<3>() + rTFt;
+
+                        filtered_sensors->servo.clear();
                     }
 
                     // Average all accumulated readings
--- a/module/input/SensorFilter/src/SensorFilter.cpp
+++ b/module/input/SensorFilter/src/SensorFilter.cpp
@@ -210,34 +210,34 @@ namespace module::input {
                         for (uint32_t id = 0; id < 20; ++id) {
                             auto& original = utility::platform::getRawServo(id, *s);
                             // Add the sensor values to the system properly
-                            filtered_sensors->servo.push_back({0,
-                                                               id,
-                                                               original.torque_enabled,
-                                                               original.p_gain,
-                                                               original.i_gain,
-                                                               original.d_gain,
-                                                               original.goal_position,
-                                                               original.moving_speed,
-                                                               original.present_position,
-                                                               original.present_speed,
-                                                               original.load,
-                                                               original.voltage,
-                                                               static_cast<float>(original.temperature)});
+                            filtered_sensors.servo.push_back({0,
+                                                              id,
+                                                              original.torque_enabled,
+                                                              original.p_gain,
+                                                              original.i_gain,
+                                                              original.d_gain,
+                                                              original.goal_position,
+                                                              original.moving_speed,
+                                                              original.present_position,
+                                                              original.present_speed,
+                                                              original.load,
+                                                              original.voltage,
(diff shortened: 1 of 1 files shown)


Bidski (dismissed):
Seems important

Bidski:
The alternative would be to move the declaration of `filtered_sensors` inside the for loop since that is the only place it is used

KipHamiltons:
I suppose it becomes a question of whether it's worth keeping that in the scope it's used vs incurring a heap allocation every iteration

tayloryoung6396 (commented):
I think if you're not using any of the information between loops you're better clearing the entire variable or redeclaring it each loop, stops people making stupid mistakes in the future

KipHamiltons:
> I think if you're not using any of the information between loops you're better clearing the entire variable or redeclaring it each loop, stops people making stupid mistakes in the future

Good point. That does stupid-proof it a bit :laughing:

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -196,16 +196,17 @@ namespace module::input {
...
-                    Eigen::Vector3d gyro  = Eigen::Vector3d::Zero();
-                    Eigen::Vector3d rMFt  = Eigen::Vector3d::Zero();
-                    auto filtered_sensors = std::make_unique<Sensors>();
+                    Eigen::Vector3d acc  = Eigen::Vector3d::Zero();
+                    Eigen::Vector3d gyro = Eigen::Vector3d::Zero();
+                    Eigen::Vector3d rMFt = Eigen::Vector3d::Zero();
 
                     for (const auto& s : sensors) {
+
+                        auto filtered_sensors = std::make_unique<Sensors>();
```
TrentHouliston, line 205: No need for this to be a unique_ptr, just make it a stack variable

TrentHouliston (commented):
Another comment about this, you have to be careful about using the sensors like this. Sometimes you'll get errors on a servo in a packet and since you're using raw sensors directly, you'll end up with some undefined values in your IK.
This could make some very bad things happen to your averages

KipHamiltons (dismissed):
LGTM. Can follow up with error handling for the servos :+1:

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -196,48 +196,48 @@ namespace module::input {
...
-                    Eigen::Vector3d acc   = Eigen::Vector3d::Zero();
-                    Eigen::Vector3d gyro  = Eigen::Vector3d::Zero();
-                    Eigen::Vector3d rMFt  = Eigen::Vector3d::Zero();
-                    auto filtered_sensors = std::make_unique<Sensors>();
+                    Eigen::Vector3d acc  = Eigen::Vector3d::Zero();
+                    Eigen::Vector3d gyro = Eigen::Vector3d::Zero();
+                    Eigen::Vector3d rMFt = Eigen::Vector3d::Zero();
 
                     for (const auto& s : sensors) {
+                        auto filtered_sensors = Sensors();
```
TrentHouliston, line 204:
```suggestion
                        Sensors filtered_sensors;
```

TrentHouliston (changes requested):

module/input/SensorFilter/src/SensorFilter.cpp
KipHamiltons: I mean, at least call the ctor explicitly :p 
```suggestion
                        Sensors filtered_sensors{};
```

abihall: difference being?

TrentHouliston (approved):

PR #726, opened 2021-09-25
Added Servo Error check and separate function for servo error string message

abihall:
**Added Servo Error check and separate function for servo error string message**

When going through each servo joint for a sensor, the error message of the servo was not considered before using it to calculate forward kinematics which (may?) be a problem. 

Also, adding a separate function for generating the servo error string for logging as already done for error on the platform and errors on the FSR.


tayloryoung6396:
I feel like the servo error strings already exist? But I could be wrong

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -209,20 +238,28 @@ namespace module::input {
...
-                                                              original.i_gain,
-                                                              original.d_gain,
-                                                              original.goal_position,
-                                                              original.moving_speed,
-                                                              original.present_position,
-                                                              original.present_speed,
-                                                              original.load,
-                                                              original.voltage,
-                                                              static_cast<float>(original.temperature)});
+                            auto& error    = original.error_flags;
```
KipHamiltons, line 241:
```suggestion
                            const auto& error    = original.error_flags;
```

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -209,20 +238,28 @@ namespace module::input {
...
+                                                                  original.i_gain,
+                                                                  original.d_gain,
+                                                                  original.goal_position,
+                                                                  original.moving_speed,
+                                                                  original.present_position,
+                                                                  original.present_speed,
+                                                                  original.load,
+                                                                  original.voltage,
+                                                                  static_cast<float>(original.temperature)});
+                            }
```
KipHamiltons, line 262: I think this might have an issue - we accumulate then divide by the size. If we have errors on the servos, we are dividing by a number larger than the number of things accumulated, and so it won't be the average.

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -87,6 +86,36 @@ namespace module::input {
         return s.str();
     }
```
KipHamiltons, line 88: It feels like we should remove the old one? Unless I am misunderstanding, they do basically the exact same thing

KipHamiltons:
Thanks for following this up

KipHamiltons:
> I feel like the servo error strings already exist? But I could be wrong

How do you mean sorry @tayloryoung6396?

module/input/SensorFilter/src/SensorFilter.cpp
abihall: Ah, I was going to change that but forgot, thanks

tayloryoung6396:
> > I feel like the servo error strings already exist? But I could be wrong
> 
> How do you mean sorry @tayloryoung6396?

Specifically the error messages, but also reading in my phone so hard to see things

module/input/SensorFilter/src/SensorFilter.cpp
abihall: Actually no, i don't think we use the amount of servos we have to calculate averages we only use the number of sensor messages we have

tayloryoung6396:
> > > I feel like the servo error strings already exist? But I could be wrong
> > 
> > 
> > How do you mean sorry @tayloryoung6396?
> 
> Specifically the error messages, but also reading in my phone so hard to see things

Like isn't there an identical function just above?

module/input/SensorFilter/src/SensorFilter.cpp
tayloryoung6396: 100% use the old one...

abihall: I was going to do the one above but they need to take in different params since additional information is printed for the servos, which is why it probably wasn't implemented before for servo errors

tayloryoung6396: Copying the same function seems like badness

tayloryoung6396: I might not understand it.. but it looks like they take the same arguments just in different ways..

abihall: Okay I will try make them the same

abihall: I don't think I can merge them into one since they will pass different param types since they print different datatypes.

tayloryoung6396: I don't really see why we're making this a function.. but anyway.. if you really want to, I would modify the original errorString function to take the error value (the cm740 and servos return the same error packet afaik). And then just pass the function the servo id as the string
`                            if (error != RawSensors::Error::OK) {
                                NUClear::log<NUClear::WARN>(
                                    makeErrorString("Servo " + std::to_string(id + 1) + " ("
                                                        + std::to_string(static_cast<ServoID>(id)) + "):",
                                                    original,
                                                    error));
                            }`

abihall: Passing original still won't work either, since `original` and `input` are different types. It doesn't really need to be a function it was just to make it easier to follow and broken up a bit.

KipHamiltons: I think it’s pretty important that we decompose sensor filter into some smaller functions if we get chances like this. Atm it’s a huge monolith which is a marathon to try to understand. Decomposition into functions is a big help for situations like that.

regarding the issue of the different types, I get the problem now. I am wondering if we could make a string out of the servo name or id then use the old function which just took the string and flags perhaps? `shared/utility/input` has functions for getting the string name of a servo iirc - maybe we could use that?

tayloryoung6396: I agree with the decomposition, i think if that's the goal it should be done as a whole, I don't find it too complex having dealt with it for a long time. 

I like these kinds of "utility" functions to be general and not too repeating, i find having multiple functions to print the same set of error messages hard to swallow

abihall: I just made a new issue about breaking down sensor filter (#729) and will get onto doing that while trying to find potential problems. Not sure what to do about this current decomposition.

KipHamiltons:
Lgtm. 

Does this fit the bill @tayloryoung6396? We aren't repeating anything by doing it like this, and we get the benefit of still reporting the associated servo information when the servo has an error.

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -54,14 +54,13 @@ namespace module::input {
     using utility::nusight::graph;
     using utility::support::Expression;
 
-    std::string makeErrorString(const std::string& src, uint errorCode) {
+    [[nodiscard]] std::string makeErrorString(const std::string& src, uint errorCode) {
         std::stringstream s;
 
         s << "Error on ";
         s << src;
         s << ":";
```
Bidski, line 62:
```suggestion
        s << "Error on " << src << ":";
```

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -87,6 +86,14 @@ namespace module::input {
         return s.str();
     }
 
+    [[nodiscard]] std::string servoErrorString(uint32_t id, RawSensors::Servo servo, uint errorCode) {
+        std::string servoName(static_cast<ServoID>(id));
+        std::stringstream s;
+        s << "Voltage: " << servo.voltage << " Position: " << servo.present_position
+          << " Temperature: " << servo.temperature << " Load: " << servo.load;
+        return makeErrorString(servoName, errorCode).append(s.str());
```
Bidski, line 94:
```suggestion
        std::string servoName(static_cast<ServoID>(id));
        std::stringstream s;
        s << makeErrorString(servoName, errorCode);
        s << "Voltage: " << servo.voltage << " Position: " << servo.present_position
          << " Temperature: " << servo.temperature << " Load: " << servo.load;
        return s.str();
```

Bidski (changes requested):

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -54,13 +54,10 @@ namespace module::input {
...
+    [[nodiscard]] std::string makeErrorString(const std::string& src, uint errorCode) {
         std::stringstream s;
 
-        s << "Error on ";
-        s << src;
-        s << ":";
-
+        s << "Error on " << src << ":";
 
         if (errorCode & RawSensors::Error::INPUT_VOLTAGE) {
```
Bidski, line 62: I think you should change the checks to this sort of format
```suggestion
        if (errorCode & RawSensors::Error::INPUT_VOLTAGE == RawSensors::Error::INPUT_VOLTAGE) {
```
or this if you prefer
```suggestion
        if (errorCode & RawSensors::Error::INPUT_VOLTAGE != 0) {
```

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -54,13 +54,10 @@ namespace module::input {
     using utility::nusight::graph;
     using utility::support::Expression;
 
-    std::string makeErrorString(const std::string& src, uint errorCode) {
+    [[nodiscard]] std::string makeErrorString(const std::string& src, uint errorCode) {
```
Bidski, line 57:
```suggestion
    [[nodiscard]] std::string makeErrorString(const std::string& src, const uint& errorCode) {
```

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -87,6 +84,15 @@ namespace module::input {
         return s.str();
     }
 
+    [[nodiscard]] std::string servoErrorString(uint32_t id, RawSensors::Servo servo, uint errorCode) {
```
Bidski, line 87:
```suggestion
    [[nodiscard]] std::string servoErrorString(uint32_t id, RawSensors::Servo servo, const uint& errorCode) {
```

Bidski (changes requested):

module/input/SensorFilter/src/SensorFilter.cpp
Bidski: Dont forget the rest of them ☺ (also might be a good idea to verify that it works)

abihall: Do you mean for all of those if statements with similar things to do the same? this change has added a warning so I will still have to clear that up as well

abihall: Could we merge this PR and I will make another PR for these changes and other small changes like these. Could you also check the servo error check as well (lines 217 - 239)

KipHamiltons: I think this would be better in a follow up. This seems a little scope-creep-y to me, given that these lines weren't changed and this wasn't the intention of the PR

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -87,6 +84,15 @@ namespace module::input {
         return s.str();
     }
 
+    [[nodiscard]] std::string servoErrorString(uint32_t id, RawSensors::Servo servo, const uint& errorCode) {
+        std::string servoName(static_cast<ServoID>(id));
+        std::stringstream s;
+        s << makeErrorString(servoName, errorCode);
```
KipHamiltons, line 90: Should we delineate where the string from `makeErrorString` ends and this info starts more clearly? I am just thinking I would probably appreciate a clear border between the part that's tell me what the problem is and the diagnostic info added by this function

abihall: I think it is pretty clear as it is right now

abihall: I misunderstood what you were saying and I have made a change for what you said :)

KipHamiltons (dismissed):
LGTM. Can follow up with the new style bitwise-op checking in the if conditions.

Bidski:
> LGTM. Can follow up with the new style bitwise-op checking in the if conditions.

Its not exactly a style choice @KipHamiltons . its about using ints in a bool context and how thats not, strictly speaking, a wise choice

KipHamiltons:
> > LGTM. Can follow up with the new style bitwise-op checking in the if conditions.
> 
> Its not exactly a style choice @KipHamiltons . its about using ints in a bool context and how thats not, strictly speaking, a wise choice

I wasn't being quite that strict with my language :stuck_out_tongue_closed_eyes: anyway, I am of the opinion that should be in a follow-up regardless of what we call it

Bidski:
> > > LGTM. Can follow up with the new style bitwise-op checking in the if conditions.
> > 
> > Its not exactly a style choice @KipHamiltons . its about using ints in a bool context and how thats not, strictly speaking, a wise choice
> 
> I wasn't being quite that strict with my language :stuck_out_tongue_closed_eyes: anyway, I am of the opinion that should be in a follow-up regardless of what we call it

Or you could just deal with it now ....

tayloryoung6396 (approved):
lgtm

KipHamiltons:
@Bidski Can you please re-review

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -206,21 +212,29 @@ namespace module::input {
...
-                                                              original.load,
-                                                              original.voltage,
-                                                              static_cast<float>(original.temperature)});
+                            auto& original    = utility::platform::getRawServo(id, *s);
+                            const auto& error = original.error_flags;
+
+                            // Check for an error on the servo and report it
+                            if (error != RawSensors::Error::OK) {
+                                NUClear::log<NUClear::WARN>(servoErrorString(id, original, error));
+                            }
```
Bidski, line 221: I think you will need to push the bad servo packet in this case as well, pretty sure [this function](https://github.com/NUbots/NUbots/blob/c16974f0ad9b4b10545b56d24fff63b6f7d6e569/module/input/SensorFilter/src/SensorFilter.cpp#L241) is expecting a full set of servo readings. Something similar to what we are [doing here](https://github.com/NUbots/NUbots/blob/c16974f0ad9b4b10545b56d24fff63b6f7d6e569/module/input/SensorFilter/src/SensorFilter.cpp#L387-L430).

Bidski (changes requested):

ysims:
Closing this as the PR is out of date with our current SensorFilter and the ideas in the PR are currently in main.

Issue #396, opened 2021-03-24
Modules need unit tests


KipHamiltons:
Currently, only a handful of utilities have unit tests. We need modules to have unit tests too, if only to know when we break things.
The progression of this problem would probably go something like this:
1. A proof of concept module unit test
2. A fully featured set of unit tests for a module
3. A guide for making unit tests for modules on NUbook
4. Unit tests for all modules
5. Profit

In terms of building a proof of concept, currently _the best lead I have_ is to go off [Nuclear's API tests](https://github.com/Fastcode/NUClear/blob/master/tests/api/ReactionHandle.cpp), which set up a powerplant manually with one thread, then install a module manually. This removes a lot of the headache of multithreading which I can imagine would be annoying if you sent a bunch of test cases to a normal multithreaded powerplant to return asynchronously. I am happy to be wrong in this case - this is just what I am thinking could work.

Once we have a proof of concept, it should just be a matter of devising test cases for a module, and seeing it through. This will allow us to see what problems come up at that scale, which makes the guide we will write more complete.

KipHamiltons:
Rather than making a NUClear extension for testing reaction callbacks, I think we should move towards using NUClear reactions as wrappers for functions which we can test like normal functions. For example, instead of 
```cpp
on<Trigger<MessageName>>().then([this](const MessageName& msg) {
    // <lots of nasty, difficult to test logic, without calls to other functions>
    emit(response1);
    // <more nasty logic>
    emit(responseN);
});
```
we use the much easier to test and reason about:
```cpp
on<Trigger<MessageName>>().then([this](const MessageName& msg) {
    nice_function_to_handle_msg(msg);
    emit(response1);
    next_function(args);
    // <a function between each emit...>
    emit(responseN);
});
```
where `nice_function_to_handle_msg`, `next_function`, and all other functions between `emit`'s (or other calls to the powerplant) are testable units.
These will be testable without constructing a powerplant (making the Reactor's `powerplant` pointer null should work I think), because we won't be emitting anything in those functions. Logging will also be impossible, so we'll have to set `log_level` high enough that it never tries to log anything (this is hacky, I know... It also fails if the module logs at the highest log level).
We should have some way of configuring without the reaction as well, because we won't have a powerplant at all if this is to work how I am envisioning it, which (I think) means `on<Configuration>` will be a no-go.

PR #727, opened 2021-09-26
Adds Module/Reaction Testing API

KipHamiltons:
### Module Testing API

This PR defines an API for testing modules with as little boilerplate as possible. All that has to be done to test modules/reactions is to define a reactor to grab the emissions that you want to check. The API has facilities for installing module dependencies and for handling timeout conditions. This is one of the big steps on the path for #396.

This PR defines the API. #728 shows it in action.

### Reviewers, Note

I am keen for feedback on the default timeout period. I chose 500 ms, because I want the default to be fairly short. If people want longer tests, they can specify that manually.

### TODO:

- [ ] Add NUbook page for this feature
- [x] Add tests for this API, if that's feasible


shared/utility/module_test_utils/ModuleTester.hpp
```diff
@@ -0,0 +1,121 @@
...
+namespace utility::module_test {
+
+    /**
+     * @brief Class for running discrete PowerPlants to test Reactors and Reactions
+     *
+     * @tparam ModuleToTest The module which is to be tested
+     * @tparam timeout_ticks  The number of ticks of the TimeoutTimeUnit before the test times out and fails
+     * @tparam TimeoutTimeUnit The unit of time which the timeout_ticks are
+     */
+    template <typename ModuleToTest, int timeout_ticks = 500, typename TimeoutTimeUnit = std::chrono::milliseconds>
```
ysims, line 38: Do we want to let people chose their own timeout time unit? I'm just thinking that consistency in time units might be good across the tests. The only thing I can think of is if we're testing to make sure the module runs quickly and we want more precision than milliseconds, though I'm not sure that's likely?

shared/utility/module_test_utils/TestLogHandler.hpp
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * @brief Handler for logging during tests so we can pass Catch the messages too
+     */
+    class TestLogHandler : public NUClear::Reactor {
+    public:
+        /// @brief Registers the log handling reaction
+        explicit TestLogHandler(std::unique_ptr<NUClear::Environment> environment);
+
+    private:
+        /// @brief Used to restrict the log handling reaction so that only one message can be processed at once
```
ysims, line 37:
```suggestion
        /// @brief Used to restrict the log handling reaction so that only one message can be processed at a time
```

ysims (commented):
This looks cool. Makes sense to me. Pretty tidy as well.

shared/utility/module_test_utils/ModuleTester.hpp
```diff
@@ -0,0 +1,121 @@
...
+            INFO("Starting PowerPlant");
+            plant.start();
+        }
+
+    private:
+        /// @brief The plant which the test runs on
+        NUClear::PowerPlant plant;
+
+        /// @brief Creates a config object to instantiate plant with
+        NUClear::PowerPlant::Configuration gen_config(const int& num_threads) {
```
CMurtagh-LGTM, line 85:
```suggestion
        NUClear::PowerPlant::Configuration generate_config(const int& num_threads) {
```
gen is ambiguous

shared/utility/module_test_utils/ModuleTester.hpp
TrentHouliston: Without a good reason to do otherwise, I think it's better to either make things have units or use SI units. And since we can't use seconds here, proving the ability to control the units (and in a way documenting what it is if you don't use the units) is good

shared/utility/module_test_utils/TestLogHandler.hpp
```diff
@@ -0,0 +1,43 @@
...
+#ifndef UTILITY_MODULETESTUTILS_TESTLOGHANDLER
+#define UTILITY_MODULETESTUTILS_TESTLOGHANDLER
+
+#include <memory>
+#include <nuclear>
+
+namespace utility::module_test {
+
+    /**
+     * @brief Handler for logging during tests so we can pass Catch the messages too
```
miikyla, line 29: Handler for logging during tests so we can pass 'Catch' messages to? So we can pass the messages to? 
pass Catch the messages too????????

miikyla (commented):
Commenting is solid

shared/utility/module_test_utils/TestLogHandler.hpp
KipHamiltons: [catch](https://github.com/catchorg/Catch2) is the unit testing framework we use. So I am saying that we log the messages using catch logging macros. I could make that more clear though?

miikyla: Ok so after reading that you're wanting Catch to receive the messages as well? Defs reword since it's a bit confusing to read

shared/utility/module_test_utils/TestLogHandler.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+        // Where this message came from
+        std::string source{};
+
+        // If we know where this log message came from, we display that
+        if (message.task != nullptr) {
+            // Get our reactor name
+            std::string reactor = message.task->identifier[1];
+
+            // Strip to the last semicolon if we have one
```
Bidski, line 48: *cough* colon *cough*

Bidski (changes requested):

miikyla (commented):
Makes sense

ysims (approved):
This looks ok. I'd suggest leaving the example PR up until something using this goes into `main` so people know how to use it. Documentation would be good too.

PR #728, opened 2021-09-26
Adds Proof of Concept Module Test for New Testing API

KipHamiltons:
### Module Testing API Proof of Concept

This adds a super-basic test for `KinematicsConfiguration`. The test isn't really testing the module so much as showing off the new API introduced in #727.

### DIFF
--- a/module/motion/KinematicsConfiguration/tests/KinematicsConfigurationTest.cpp
+++ b/module/motion/KinematicsConfiguration/tests/KinematicsConfigurationTest.cpp
@@ -14,14 +14,68 @@
  * You should have received a copy of the GNU General Public License
  * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
  *
- * Copyright 2015 NUbots <nubots@nubots.net>
+ * Copyright 2021 NUbots <nubots@nubots.net>
  */
 
-// Uncomment this line when other test files are added
-//#define CATCH_CONFIG_MAIN  // This tells Catch to provide a main() - only do this in one cpp file
-//#include <catch.hpp>
+#define CATCH_CONFIG_MAIN
+#include <catch.hpp>
+#include <memory>
+#include <nuclear>
 
-// Remove this line when test files are added
-int main() {
-    return 0;
+#include "KinematicsConfiguration.hpp"
+
+#include "extension/FileWatcher/src/FileWatcher.hpp"
+
+#include "message/motion/KinematicsModel.hpp"
+
+#include "utility/module_test_utils/ModuleTester.hpp"
+#include "utility/strutil/ansi.hpp"
+
+namespace {
+    using message::motion::KinematicsModel;
+
+    std::unique_ptr<KinematicsModel> saved_model = nullptr;
+
+    class TestReactor : public NUClear::Reactor {
+    public:
+        TestReactor(std::unique_ptr<NUClear::Environment> environment) : Reactor(std::move(environment)) {
+
+            // Trigger on the KinematicsModel message
+            on<Trigger<KinematicsModel>>().then([this](const KinematicsModel& model) {
+                log<NUClear::INFO>("Received a KinematicsModel message. Test is now over.");
+                saved_model = std::make_unique<KinematicsModel>(model);
+
+                // When we receive the KinematicsModel message we can shutdown
+                powerplant.shutdown();
+            });
+        }
+    };
+}  // namespace
+
+TEST_CASE("Testing the Kinematics Configuration module", "[module][motion][KinematicsConfiguration]") {
+
+    using module::extension::FileWatcher;
+    using module::motion::KinematicsConfiguration;
+    using utility::support::ModuleTester;
+
+    static constexpr int NUM_THREADS = 2;
+
+    // Test is for KinematicsConfiguration
+    ModuleTester<KinematicsConfiguration> tester(NUM_THREADS);
+
+    // Configuration tasks depend on FileWatcher
+    tester.install<FileWatcher>("FileWatcher");
+
+    // Finally, we install the emission "intercepter" module, which saves the emissions for us
+    tester.install<TestReactor>("TestReactor");
+
+    tester.run();
+
+    // Require that a model was saved
+    REQUIRE(saved_model != nullptr);
+
+    // TODO(Devops&QA/Motion): Make this test actually check the config values as found in the file, rather than
+    //                         hardcoding this value which is subject to change
+    // Now check values in model to ensure correctness
+    REQUIRE(saved_model->head.INTERPUPILLARY_DISTANCE == 0.068f);
 }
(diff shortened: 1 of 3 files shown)


ysims (commented):
Looks neat and easy enough to use 👍

miikyla (commented):
Looks like code 👍

module/motion/KinematicsConfiguration/tests/KinematicsConfigurationTest.cpp
```diff
@@ -14,14 +14,71 @@
  * You should have received a copy of the GNU General Public License
  * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
  *
- * Copyright 2015 NUbots <nubots@nubots.net>
+ * Copyright 2021 NUbots <nubots@nubots.net>
```
abihall, line 17: Should we be changing these for all files we change from now on?

KipHamiltons: I dunno. Given that this file was basically empty, it seems appropriate, but I have no idea about the legal stuff and licence things. I have just been doing when it seemed reasonable 😂

module/motion/KinematicsConfiguration/tests/KinematicsConfigurationTest.cpp
```diff
@@ -14,14 +14,71 @@
...
+    tester.install<TestReactor>("TestReactor");
+
+    tester.run();
+
+    // We needed to have saved a model. If we didn't, we failed the test
+    if (saved_model != nullptr) {
+        // TODO(Devops&QA/Motion): Make this test actually check the config values as found in the file, rather than
+        //                         hardcoding this value which is subject to change
+        // Now check values in model to ensure correctness
+        REQUIRE(saved_model->head.INTERPUPILLARY_DISTANCE == 0.068f);
```
abihall, line 79: Is this the only value we want to check?

module/motion/KinematicsConfiguration/tests/KinematicsConfigurationTest.cpp
```diff
@@ -14,14 +14,71 @@
...
+    };
+}  // namespace
+
+TEST_CASE("Testing the Kinematics Configuration module", "[module][motion][KinematicsConfiguration]") {
+
+    using module::extension::FileWatcher;
+    using module::motion::KinematicsConfiguration;
+    using utility::module_test::ModuleTester;
+
+    static constexpr int NUM_THREADS = 2;
```
abihall, line 61: why two threads? what are the threads doing?

nuclear/module/NUClearModule.cmake
```diff
@@ -23,7 +23,7 @@ function(NUCLEAR_MODULE)
   # Parse our input arguments
   set(options, "")
   set(oneValueArgs "LANGUAGE")
-  set(multiValueArgs "LIBRARIES" "SOURCES" "DATA_FILES")
+  set(multiValueArgs "LIBRARIES" "SOURCES" "DATA_FILES" "TEST_MODULE_DEPENDENCIES")
```
Bidski, line 26: I would argue that changes to this file belong in the other PR as test reactors wont work without this and its the other PR that sets up the functionality to allow test reactors to be a thing

Bidski (changes requested):

.github/workflows/nubots.yaml
```diff
@@ -96,7 +96,7 @@ jobs:
         run: ./b build
 
       - name: Test the code
-        run: ./b tests run
+        run: for i in {1..20}; do ./b tests run Kine; done
```
abihall, line 99: What is this I am confused

KipHamiltons: This is me trying to debug the segfaults on the tests which I can't replicate on my machine :sweat:

abihall: Ahh okay that sucks 😓 can I help in anyway?

KipHamiltons: If you can replicate the segfault on your machine, that would be a huge help! My machine doesn't segfault, but adding this line [here](https://github.com/NUbots/NUbots/pull/728/commits/db3f4ed34d7a40f7d2394b819259f5374d7357e8) seems to introduce it. I don't know why though. Fixing the cause is the goal

KipHamiltons: Looks like it was a reference issue which has been fixed by explicit `std::move` in the `unique_ptr` construction.

module/motion/KinematicsConfiguration/tests/KinematicsConfigurationTest.cpp
KipHamiltons: I didn't institute this change, so I am not 100% about the reasoning (this test was made possible thanks to @Bidski's work he sent me), but I am thinking it's probably because we want to be testing the kinematics configuration in a multithreaded context, which is closer to the actual use case.

In terms of what they're doing, this variable is passed to the `ModuleTester` constructor which sets up the NUClear PowerPlant with that many threads for the test.

module/motion/KinematicsConfiguration/tests/KinematicsConfigurationTest.cpp
KipHamiltons: Well, not really. I don't really want to check any of the values like this, since this test would fail if the config changed, which is very bad, because it's very brittle - this is just a proof of concept. I plan to make this test proper before actually merging it - for example, by reading the config file directly, then comparing the values to the ones we read.

nuclear/module/NUClearModule.cmake
KipHamiltons: Excellent point. I have moved these changes up.

.github/workflows/nubots.yaml
KipHamiltons: I have stopped using `unique_ptr` at all because it seemed like that problem wasn't actually fixed by that change. This change seems to have fixed things... _here's hoping_

miikyla (commented):
seg fault gone?

module/motion/KinematicsConfiguration/tests/KinematicsConfigurationTest.cpp
JosephusPaye: Speaking of updating things, I'm pretty sure that email address doesn't work anymore.

module/motion/KinematicsConfiguration/tests/KinematicsConfigurationTest.cpp
```diff
@@ -14,14 +14,66 @@
...
```
KipHamiltons, line 61:
```suggestion
    static constexpr int NUM_THREADS = 1;
```

CMurtagh-LGTM:
@KipHamiltons It seems that the CI passed, does that mean that #736 is fixed?

KipHamiltons:
@CMurtagh-LGTM i guess???? Seems good

KipHamiltons:
The test passed! That means that the API is basically done. I will add a couple of tests for the API, but this test is a model for using the API. 
This test is brittle and ugly and dumb because it just checks one value which is hardcoded, but that doesn't matter for now :sweat_smile: 
Thank you @CMurtagh-LGTM for fixing filewatcher

KipHamiltons:
BTW the failed CI is a segfault in director tests I haven't touched. This stuff passes 😄

PR #733, opened 2021-10-02
Dockerhub Cleanup Fix

KipHamiltons:
### DockerHub Clean-up Workflow Fix
 
Basically, our PR tags weren't getting deleted. It wasn't immediately obvious, because the HTTP status codes being wrong doesn't make `curl` error. This adds a method (which should fail if done incorrectly) to delete tags properly.

Details about GHA and our image stuff like this should probably make their way into NUbook at some point. I might make an issue


.github/workflows/scripts/remove_dockerhub_tag.sh
```diff
@@ -0,0 +1,48 @@
...
+login_json() {
+cat <<EOF
+{
+  "username": "$DOCKERHUB_USERNAME",
+  "password": "$DOCKERHUB_PASSWORD"
+}
+EOF
+}
+
+TOKEN=$(curl -s -H "Content-Type: application/json" -X POST -d "$(login_json)" "https://hub.docker.com/v2/users/login/" | jq -r .token)
```
Y0sh1dk, line 35: I don't think `jq` will be installed

.github/workflows/dockerhub-cleanup.yaml
```diff
@@ -17,10 +17,20 @@ jobs:
...
+    # We need `jq` for the script, so we run in a container with it installed
+    container:
+      image: dwdraju:alpine-curl-jq
+
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      - name: Delete image
+      # Checks-out the repo under $GITHUB_WORKSPACE so we can access the delete script
+      - name: Checkout Code
+        uses: actions/checkout@v2
```
Y0sh1dk, line 28: I'm not super familiar with GHA, but will this action run inside the `alpine-curl-jq`? It wont have `git` installed if so.

Can we make this workflow trigger on push's too this branch to test?

KipHamiltons: crap. Good point. Let's see what happens

.github/workflows/dockerhub-cleanup.yaml
```diff
@@ -17,10 +17,20 @@ jobs:
     # The type of runner that the job will run on
     runs-on: ubuntu-20.04
 
+    # We need `jq` for the script, so we run in a container with it installed
+    container:
+      image: dwdraju:alpine-curl-jq
```
Y0sh1dk, line 22:
```suggestion
      image: dwdraju/alpine-curl-jq
```

.github/workflows/dockerhub-cleanup.yaml
KipHamiltons: Looks like the checkout isn't an issue. I have screwed up the `TAG` env var though

KipHamiltons:
[Looks like we have liftoff](https://github.com/NUbots/NUbots/pull/733/checks?check_run_id=3774783747)

Should be good to go

.github/workflows/dockerhub-cleanup.yaml
```diff
@@ -17,10 +17,20 @@ jobs:
...
+      image: dwdraju/alpine-curl-jq
+
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      - name: Delete image
+      # Checks-out the repo under $GITHUB_WORKSPACE so we can access the delete script
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      # Run out script which calls the dockerhub API to delete the PR's image
```
KipHamiltons, line 30:
```suggestion
      # Run our script which calls the dockerhub API to delete the PR's image
```

.github/workflows/dockerhub-cleanup.yaml
```diff
@@ -17,10 +17,20 @@ jobs:
...
+
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      - name: Delete image
+      # Checks-out the repo under $GITHUB_WORKSPACE so we can access the delete script
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      # Run our script which calls the dockerhub API to delete the PR's image
+      - name: Delete DockerHub image
```
Y0sh1dk, line 31:
```suggestion
      # Run our script which calls the dockerhub API to delete the PR's tag
      - name: Delete DockerHub image tag
```

Y0sh1dk (approved):
Nice!, LGTM

PR #734, opened 2021-10-02
The next round of clang-tidy fixes

KipHamiltons:
One day, we will integrate it into CI, without any more PRs to go through. One day...


module/behaviour/strategy/SoccerStrategy/src/SoccerStrategy.hpp
```diff
@@ -84,7 +84,7 @@ namespace module::behaviour::strategy {
         NUClear::clock::time_point ballSearchStartTime;
         NUClear::clock::time_point goalLastMeasured;
         void initialLocalisationReset();
-        void penaltyShootoutLocalisationReset(const message::support::FieldDescription& fieldDescription);
+        void penaltyShootoutLocalisationReset(const message::support::FieldDescription& fd);
```
CMurtagh-LGTM, line 87: `fd` is usually a file descriptor, I think the full name is better here.

module/platform/darwin/HardwareIO/src/darwin/Darwin.cpp
```diff
@@ -78,7 +77,9 @@ namespace Darwin {
 
         // Set servos to be enabled if they are not simulated
         for (size_t i = 0; i < config["servos"].config.size(); ++i) {
-            enabledServoIds[i] = !config["servos"][i]["simulated"].as<bool>();
+            enabledServoIds[i] =
+                static_cast<__gnu_cxx::__alloc_traits<class std::allocator<unsigned char>, unsigned char>::value_type>(
+                    !config["servos"][i]["simulated"].as<bool>());
```
CMurtagh-LGTM, line 82: This needs some justification.

module/platform/darwin/HardwareIO/src/Convert.hpp
```diff
@@ -80,30 +80,30 @@ namespace module::platform::darwin {
...
-        static float torqueLimit(const uint16_t value);
-        static uint16_t torqueLimitInverse(const float value);
+        static float torqueLimit(uint16_t value);
+        static uint16_t torqueLimitInverse(float value);
 
-        static float servoLoad(const uint8_t id, const uint16_t value);
+        static float servoLoad(uint8_t id, uint16_t value);
 
-        static float temperature(const uint8_t value);
+        static float temperature(uint8_t value);
```
CMurtagh-LGTM, line 106: Is it more performant to pass by value rather than const reference?

module/platform/darwin/HardwareIO/src/darwin/Darwin.cpp
```diff
@@ -129,10 +130,11 @@ namespace Darwin {
...
         // Do a self test so that we can move all the sensors that are currently failing to the end of the list
         std::vector<std::pair<uint8_t, bool>> sensors = selfTest();
-        std::sort(
-            std::begin(sensors),
-            std::end(sensors),
-            [](const std::pair<uint8_t, bool>& a, const std::pair<uint8_t, bool>& b) { return a.second > b.second; });
+        std::sort(std::begin(sensors),
+                  std::end(sensors),
+                  [](const std::pair<uint8_t, bool>& a, const std::pair<uint8_t, bool>& b) {
+                      return static_cast<int>(a.second) > static_cast<int>(b.second);
```
CMurtagh-LGTM, line 136: What is the reason on not using the `unit8_t` comparison

shared/extension/Script.hpp
```diff
@@ -82,29 +82,27 @@ namespace extension {
...
             , config()
-            , frames(frames) {}
+            , frames(std::move(frames)) {}
 
-        Script(const std::string& fileName,
-               const std::string& hostname,
-               const std::string& platform,
+        Script(std::string fileName,
+               std::string hostname,
+               std::string platform,
```
CMurtagh-LGTM, line 95: Should this be const reference?

shared/utility/motion/Balance.hpp
```diff
@@ -62,9 +62,9 @@ namespace utility::motion {
         NUClear::clock::time_point lastBalanceTime;
 
     public:
-        Balancer() : lastErrorQuaternion(), lastBalanceTime() {}
+        Balancer() = default;
         void configure(const YAML::Node& config);
-        void balance(const message::motion::KinematicsModel& hip,
+        void balance(const message::motion::KinematicsModel& model,
```
CMurtagh-LGTM, line 67: Are there other Kinematics Models other than a hip one? If so, don't rename this.

CMurtagh-LGTM (changes requested):
:P

module/behaviour/strategy/SoccerStrategy/src/SoccerStrategy.hpp
KipHamiltons: Good point.

module/platform/darwin/HardwareIO/src/darwin/Darwin.cpp
KipHamiltons: I am not sure why I didn't give this more scrutiny, as you say. I think this might be a bug. I'll post on slack.

module/platform/darwin/HardwareIO/src/darwin/Darwin.cpp
KipHamiltons: looking at this closer, it seems to be doing a `greater than` on the bool, which is implicitly cast to integer, which clang-tidy doesn't like. This seems like it should be a `std::partition` instead.

module/platform/darwin/HardwareIO/src/Convert.hpp
KipHamiltons: I think clang-tidy takes issue with this because the `const` applies here to a value (and therefore is irrelevant to the signature), so having it at all is misleading. I should change these to be `const` references though, to match the rest of the codebase.

shared/extension/Script.hpp
KipHamiltons: Probably

shared/utility/motion/Balance.hpp
KipHamiltons: `model` is more appropriate. Not sure why it was 'hip' in the first place.

tayloryoung6396: imo this isn't used/probably will be replaced if it ever is.. but `model` makes more sense than hip

tayloryoung6396: it would have been hip because our we don't really do kinematics for the arms and is definitely something that the balance engine wouldn't have used

module/platform/darwin/HardwareIO/src/Convert.hpp
CMurtagh-LGTM: yes

ysims (approved):
Can't see anything bad.

CMurtagh-LGTM (approved):
LGTM

PR #737, opened 2021-10-05
Make it technically possible to respond to voice commands

mrwerdo:
### Make it technically possible to respond to voice commands

This pull request adds:
- Speech Recognition
- Intention Recognition

A message looks something like:
```c++
struct SpeechIntentMsg {
    string text;
    string intent;
    vector<Slot> slots;
    float confidence;
}

struct Slot {
    string name;
    string value;
}
```

A slot represents a substitution made during sentence parsing. For example, consider the following grammar:
```
I saw a (red | green | blue){colour} car
```

Given the sentence `I saw a green car`, there would be a `SpeechIntentMsg` produced once the sentence was spoken where the slots vector contained one slot with `name=colour`, and `value=green`.

I know this slots language does not make sense, but we use it here for consistency with [voice2json's terminology](https://voice2json.org/sentences).

The system uses [arecord](https://linux.die.net/man/1/arecord) underneath the hood to find the microphone.

### Pre-PR Checklist:

Have you
- [x] Added/Updated tests for your changes, including regression tests for bug fixes
- [x] Added/modified [documentation directives](https://nubook.nubots.net/guides/general/documentation#extra-formatting) in relevant code
- [x] Added relevant labels to the PR

The module is being tested by hand at the moment.

### Reviewers, Note

This pull request replaces the [other pull request here with the same name](https://github.com/NUbots/NUbots/pull/693) due the integration tests being unable to complete.

To verify the code is able to perform intention recognition:
1. Create a wav file in the repository: `espeak -w output.wav "bring me an apple from the bench"`
2. Run the module passing the full path: `./b run speechintent -- file /home/nubots/NUbots/output.wav`

The module will translate the sentence and then perform intention recognition based on the grammar in `docker/voice2json/sentences.ini`.

### Things left to do

- [x] Install [json.h](https://github.com/sheredom/json.h) in `/usr/local/include/` instead of as a user header in the `SpeechIntent` module.
- [ ] Sync the following directories during the installation:
    - `/home/nubots/.local/share/voice2json/en-us_kaldi-zamia`
    - `/home/nubots/voice2json`
    - `/opt/kaldi/`


module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,425 @@
...
+    SpeechIntent::SpeechIntent(std::unique_ptr<NUClear::Environment> environment) : Reactor(std::move(environment)) {
+
+        on<Configuration>("SpeechIntent.yaml").then([this](const Configuration& cfg) {
+            this->log_level = cfg["log_level"].as<NUClear::LogLevel>();
+        });
+
+        // Choose between 3 modes based on command line arguments. The "cli" and "file" arguments are for testing
+        // purposes. In the case where no cmd line arguments are passed the module is initialized in
+        // "TRANSCRIBE_MODE_STREAM" mode.
+        on<Trigger<CommandLineArguments>>().then([this](const CommandLineArguments& args) {
```
Mrwerdo, line 365: @Bidski, @tayloryoung6396, @c3144856 This was the section that we were discussing in the previous pull request.

Mrwerdo: > Its pointless because you can achieve all of the same functionality using the configuration file. You could easily set up a `sound_source` entry in the config file, and if that represents a valid file (you can use `std::filesystem` for this) then you load that as an audio file, if `sound_source` is `mic` (or some other appropriate keyword) then you look for the appropriate hardware and get your audio data from there.
> 

The perspective that I was thinking about this from was that hard coding the path to the file means that you'd need to edit the file in order to change which sentences is processed by the module, and that could accidentally make its way into version control.

I'm happy to change this, though I think it will make using it a little bit more awkward when verifying the command works.

> Also, are you aware of [this](https://stackoverflow.com/a/48795232/4795763)? Not sure if this would work in WSL.

Yes. `--device` doesn't work on macOS or Windows because docker desktop runs inside a virtualised environment which doesn't have access to the device drivers. [See here](https://github.com/docker/for-mac/issues/900).

docker/Dockerfile
```diff
@@ -273,6 +273,9 @@ RUN mkdir -p /usr/local/include/zstr \
     && wget https://raw.githubusercontent.com/mateidavid/zstr/946ab409380e90e4a46c88c2da4f645a7ffaa51d/src/strict_fstream.hpp -O /usr/local/include/strict_fstream.hpp \
     && wget https://raw.githubusercontent.com/mateidavid/zstr/946ab409380e90e4a46c88c2da4f645a7ffaa51d/src/zstr_make_unique_polyfill.h -O /usr/local/include/zstr_make_unique_polyfill.h
 
+# json.h
+RUN install-header-from-source https://raw.githubusercontent.com/sheredom/json.h/adf662d166f1890f456b1c86a6d923dbfe6c7397/json.h
+
```
Bidski, line 278: Is this needed as a dependency for the stuff below?

module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,425 @@
+#include "SpeechIntent.hpp"
+
+#include <alsa/asoundlib.h>
+#include <fmt/format.h>
+#include <json.h>
```
Bidski, line 5: JSON is a subset of YAML, just use `yaml-cpp`

module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,425 @@
...
+        return success;
+    }
+
+    // python -m voice2json --base-directory /home/nubots/voice2json/ -p en transcribe-stream
+    bool voice2json_transcribe_stream(SpawnedProcess& process) {
+        const char* args[] = {"python",
+                              "-m",
+                              "voice2json",
+                              "--base-directory",
+                              "/home/nubots/voice2json/",
```
Bidski, line 140: This should probably be a config value

module/input/SpeechIntent/src/SpeechIntent.cpp
```diff
@@ -0,0 +1,425 @@
...
+    // when voice2json is called using the "transcribe-wav --input-size" flags
+    // it excepts the first line to be the file size (as ascii)
+    // followed by the wav file after the newline
+    bool write_audio_to_file(int fd, char* filename) {
+        ssize_t write_res = 0;
+        if (access(filename, F_OK) != 0) {
+            return false;
+        }
+
+        FILE* file = fopen(filename, "rb");
```
Bidski, line 219: Try to avoid using the C standard library, C++ file ops should always be preferred

Bidski (changes requested):

module/input/SpeechIntent/src/SpeechIntent.cpp
Bidski: > The perspective that I was thinking about this from was that hard coding the path to the file means that you'd need to edit the file in order to change which sentences is processed by the module, and that could accidentally make its way into version control.

This isn't really much of an issue. The per-robot/per-binary config system will probably be your friend here. Given that we will never be able to use the microphone in docker (at least not in a cross-platform way) so "hard coding" a path in the `docker` "robot" config file is not a horrible idea.

I would make a data folder `module/input/SpeechIntent/data/audio` and stick some test WAV files in there, your "hard coded" path would then just be `audio/test1.wav`.

Also consider unit tests will not have access to a microphone either, so if you set things up properly you will be keeping some data files in the repo anyway.

docker/Dockerfile
```diff
@@ -10,7 +10,7 @@
 ### The date in the Server command must either match, or be more recent than,
 ### the date in the FROM command
 ### Date format is ISO year/month/day
-FROM archlinux:base-20210829.0.32635
+FROM archlinux:base-20210829.0.32635 as base
```
KipHamiltons, line 13: What's the selling point of changing the dockerfile like this? It seems to me that it makes it a bit harder to follow, but not terribly so. Perhaps enough to counter the somewhat faster dockerfile builds when actively working with this library, but I am not sold either way.

KipHamiltons (changes requested):
Not quite a complete review. I apologise for _some_ of the pedantry :P

docker/PKGBUILD/kaldi.sh
```diff
@@ -0,0 +1,81 @@
+# Maintainer: Jingbei Li <i@jingbei.lli>
```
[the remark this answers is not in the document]
Mrwerdo: These files are copied from AUR packages. (If you search for their names you can find them. I do have a list somewhere too.) Most PKGBUILD files needed to be modified in order for them to compile. I thought of a few possibilities:
1. Submitting patches for these AUR packages so they actually build correctly.
2. Run `sed` to correct the problems.
3. Store the files in the repository and copy them in to build.

I went with option three because it was clearer how the package was being built. You can see the script that is building the dependency and it keeps the docker file cleaner.

The packages need to be updated whenever it stops working. More to the point, I didn't substantially modify the PKGBUILD files so that you could run `diff` to see the changes between the latest AUR version our version. That makes upgrading slightly easier.

docker/voice2json/readme.md
```diff
@@ -0,0 +1,36 @@
...
+python -m voice2json --profile en recognize-intent |                    # text to intent
+jq                                                                      # pretty print
+```
+
+The sentences.ini file has been created for targeting the general purpose command set and will need to be modified for the enhanced general purpose command set.
+
+Mount your docker container so that it can access your nubots folder...
+
+```bash
+docker run -it --rm --mount type=bind,source=/Users/realideasman/Developer/NUbots/,target=/home/nubots/NUbots nuhear:latest
```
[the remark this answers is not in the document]
Mrwerdo: Lol, now you know where I keep all my secret projects...

module/input/SpeechIntent/src/SpeechIntent.hpp
```diff
@@ -0,0 +1,168 @@
...
+
+        /**
+         * A number between 0 and 1.
+         * @see https://voice2json.org/formats.html#intents
+         * @remark You probably don't want to make any assumptions about this value. Either use empirical evidence
+         * or consult voice2json's documentation for a precise interpretation of this value.
+         */
+        float confidence;
+
+        void add_slot(std::string name, std::string value) {
```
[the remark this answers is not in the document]
Mrwerdo: I'm pretty sure this should have been deleted. It doesn't make sense to include it in the interface definition (who's gonna add slots after the message has been emitted?)

module/input/SpeechIntent/src/SpeechIntent.hpp
```diff
@@ -0,0 +1,168 @@
...
+            Slot slot = {name, value};
+            slots.push_back(slot);
+        }
+    };
+
+    struct ProcessSpawnResources {
+        int stdout_pipe[2]                 = {};
```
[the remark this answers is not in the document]
Mrwerdo: I'd suggest using more meaningful names instead, ones that correspond with the data that can be read from them.

docker/Dockerfile
Mrwerdo: I had it setup that way because I ways building multiple parallel dependencies. At the moment now the case is less compelling because Kaldi and OpenFST are the only ones that take a while to compile and Kaldi depends on OpenFST.

I've been thinking about adjusting the voice2json installation section to follow the conventions established in the docker file however I fear that I won't have enough time at the moment to improve this and I feel that the improvement is readably. Specifically, I was thinking of adding a section to the `install-from-source` script that can handle `makepkg` dependencies and then installing them in the toolchains sections.

Mrwerdo:
This branch takes longer than 6 hours to compile. It was part way through OpenFST which takes about 40 minutes. Kaldi comes afterwards and also takes about 40 minutes.

I'm kind of the opinion at the moment that we need to build these packages separately and then use Pacman (or some other tool) to manage the dependencies in the docker container and robot. I've been trying to install multiple packages at the same time, but couldn't until after one was built. If they were separated I'd be able to work on the at the same time.

tayloryoung6396:
> This branch takes longer than 6 hours to compile. It was part way through OpenFST which takes about 40 minutes. Kaldi comes afterwards and also takes about 40 minutes.
> 
> I'm kind of the opinion at the moment that we need to build these packages separately and then use Pacman (or some other tool) to manage the dependencies in the docker container and robot. I've been trying to install multiple packages at the same time, but couldn't until after one was built. If they were separated I'd be able to work on the at the same time.

@KipHamiltons @Bidski

ysims:
Superceded by #966

Issue #735, opened 2021-10-02
Webots full game with NUbots vs NUbots has the same Team ID


ysims:
When playing a full game with NUbots vs NUbots

```bash
./b webots game -r 6 webotsrobocup
```

both teams of robots have the same team ID. This is problematic because game controller messages are filtered by team, and if both have the same ID then we can have the scenario where team red robot 1 gets penalised and team blue robot 1 will think it's penalised and freeze when it shouldn't.

I think a good solution would be to allow the team ID to be passed as an environment variable that then adjusts the variable for that robot. Then for the full game stuff maybe just increment our current number.... that is, have one be 12 and the other 13. It doesn't matter so much what they are as long as it's documented, then when running a full NUbots vs NUbots game the appropriate team IDs can be put into the `game.json` file.

PR #741, opened 2021-10-06
De koeyer/docker team

Y0sh1dk:
Should fix https://github.com/NUbots/NUbots/issues/735

 - `team_id` is now updated in `GlobalConfig.yaml` on container startup using the env var passed too it.
 
  - Update Dockerfile naming convention


tools/webots.py
```diff
@@ -178,7 +182,7 @@ def exec_build(target, roles, clean, jobs):
 
     # Build the image!
     exit_code = subprocess.run(
-        ["docker", "build", "-t", f"{ROBOCUP_IMAGE_NAME}:{ROBOCUP_IMAGE_TAG}", "-f", "./nugus_sim.Dockerfile", "."]
+        ["docker", "build", "-t", f"{ROBOCUP_IMAGE_NAME}:{ROBOCUP_IMAGE_TAG}", "-f", "./Dockerfile.nugus_sim", "."]
```
KipHamiltons, line 185: Isn't this wrong? we have a file called `nugus_sim.Dockerfile` and we don't have a file named `Dockerfile.nugus_sim`?

tools/webots.py
```diff
@@ -229,8 +233,10 @@ def exec_run(role, num_of_robots=1, sim_address="127.0.0.1"):
     # Add all robot run commands to process_manager
     for i in range(1, num_of_robots + 1):
 
-        robot_color, port_num = (
-            ("red", 10000 + i) if i <= num_of_robots // 2 else ("blue", 10020 + i - num_of_robots // 2)
+        team_id, robot_color, port_num = (
+            (NUBOTS_TEAM_ID, "red", 10000 + i)
+            if i <= num_of_robots // 2
+            else (ALT_TEAM_ID, "blue", 10020 + i - num_of_robots // 2)
```
KipHamiltons, line 239: This hurts my brain a little. Could you please split up this logic so it's easier to follow? (I know you're just working with the same thing which was there, but I think what was there was probably a bit over the top anyway :P )

Y0sh1dk: :cry:

KipHamiltons: :brain:  <<  large :weary:

tools/webots.py
```diff
@@ -25,6 +25,10 @@
 ROBOCUP_IMAGE_TAG = "robocup2021"  # Submitted in our team_config.json, shouldn't be changed here unless changed there
 ROBOCUP_IMAGE_REGISTRY = "079967072104.dkr.ecr.us-east-2.amazonaws.com/robocup-vhsc-nubots"  # Provided by the TC
 
+# Webots team details
+NUBOTS_TEAM_ID = 12
```
KipHamiltons, line 29: I am curious what the significance of this is

ysims: This is our official team ID from RoboCup 2021, specified in the GameController repo.
13 is just the number after 12 :stuck_out_tongue:

KipHamiltons: I think it'd be worth adding that as a comment tbh. I feel like I could see someone in a couple of years having a lot of trouble trying to work out that :sweat_smile:

Y0sh1dk: We get assigned a different team id each robocup afaik. I got `12` from what is currently set in [GlobalConfig.yaml](https://github.com/NUbots/NUbots/blob/main/module/support/configuration/GlobalConfig/data/config/GlobalConfig.yaml). It doesnt matter what the alt team id is, as long as it is different

Y0sh1dk: Added a relevant comment and removed the alt team ID in favour of just incrementing the NUbots team ID for the other team ID.

KipHamiltons (approved):
LGTM! (if it works :P)

KipHamiltons:
Tested playing a full game on my machine and it works. very nice!

JosephusPaye:
Yeah there are tradeoffs, but I think `x.Dockerfile` is probably the way to go. This [StackOverflow thread](https://stackoverflow.com/questions/26077543/how-to-name-dockerfiles) has some takes on both sides.

PR #757, opened 2021-11-09
Disable/enable the walk on getup

ysims:
When the get up happens, the walk engine just keeps running (although the servo commands aren't used). 
I think we should disable the walk engine while getting up, and reenable it when the get up is killed. This PR makes that change. 

While I don't see any affect on the QuinticWalk, this is useful for resetting the Static Walk when the get up is finished. 
I'm wondering if this could be achieve a different way if there's method to execute code when a module gains/loses control, or if that's just a Director thing. 
Regardless though it should add the benefit of improving performance from not running the walk when unnecessary?

Not 100% sure on this solution to the problem.

### DIFF
--- a/module/behaviour/skills/Getup/README.md
+++ b/module/behaviour/skills/Getup/README.md
@@ -20,6 +20,8 @@ Include this module to allow the robot to get up after it has fallen over.
 - `extension::ExecuteScriptByName` to run getup and stand scripts
 - `message::motion::ExecuteGetup` instigates a getup
 - `message::motion::KillGetup` stops a getup
+- `message::motion::DisableWalkEngineCommand` stops the walk engine while getting up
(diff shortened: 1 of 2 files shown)


tayloryoung6396:
I would probably do it the same way the getup is handled. A trigger on the kill that enables the handle for the walk engine. The director will solve this

ysims:
> I would probably do it the same way the getup is handled. A trigger on the kill that enables the handle for the walk engine. The director will solve this

Ah right yeah I was hoping for that sort of functionality but wasn't sure how to do it, but yeah the getup uses that method. I'll change this PR to do that for the QuinticWalk.

ysims:
Ye nah this isn't even needed it's already done in SimpleWalkPathPlanner :laughing:

PR #759, opened 2021-11-23
Adds in a dynamic foot and torso controller and a static walk

ysims:
This PR adds a static walk engine with a dynamic foot and torso controller underneath.

**This PR is a draft - the walk doesn't work correctly yet!** 
I have attached a video below of the current state of the walk, where it's meant to lean the torso over the right foot. It appears to lean over too much, which doesn't seem to be a CoM problem because even if I make it move the torso straight over the support foot it seems to happen. It also seems to keep rotating its feet back and forth in the y-axis so it's rocking itself back and forth, maybe either odometry issues or overshooting the rotation. 

https://user-images.githubusercontent.com/35280100/143004416-2f236370-f175-4e36-bf63-42b81b50276f.mp4

I'm hoping that more eyes might help see if there are any maths issues or even just other general issues that I can fix up for when this will become a proper PR. 

You can run this in Webots using `webotsrobocup`.

There are a lot of files in this PR, main ones are 
- StaticWalk.cpp
- MovementController.cpp
- FootController.cpp
- TorsoController.cpp

## To Do

General things to do:
- [ ] Update NUbook
- [ ] Tests
- [x] Module READMEs
- [ ] Added [documentation directives](https://nubook.nubots.net/guides/general/documentation#extra-formatting)

Main things to do:
- [ ] Stable lean
- [ ] Stable step
- [ ] Working walk!

## Further Work

There are also a lot of extensions to this project that need to be done, but the goal right now is to get it walking forward correctly (without falling over...) and extensions can be a follow up PR. If you'd like to make suggestions for further things, feel free to. Current known things are
- Collisions with the feet when rotating and strafing.
- Adding a better walk engine such as reinforcement learning or ZMP approach rather than the static walk. 
- Minimise movement of the head during the walk.


tayloryoung6396:
One thing future me would like that may be useful to have in the back of you mind when developing the torso controller. I want the walk engine to try and keep the cameras parallel to the visual mesh observation plane. This will assist in vision related things, measurements, motion blur etc

ysims:
> One thing future me would like that may be useful to have in the back of you mind when developing the torso controller. I want the walk engine to try and keep the cameras parallel to the visual mesh observation plane. This will assist in vision related things, measurements, motion blur etc

Yeah, that would be useful for vision. The torso controller and foot controller work with any given rotation/translation for flexibility, so that sort of thing would go in the walk engine on top. 
The StaticWalk does something close to this atm. It tries to keep the torso and support foot level with the ground (ground space rotation). As long as head behaviour keeps the head in a good position, it should be pretty close to achieving that.

ysims:
:/

Issue #745, opened 2021-10-07
RoboCup Docker Image GitHub Actions Workflow


KipHamiltons:
Having a dedicated github actions workflow for pushing the robocup docker images would be nice. It has enough idiosyncrasies to be separate from the normal image push workflow, so this would involve making a new workflow for it.

PR #762, opened 2021-12-02
Slack message if the docker image was changed

### LINKED ISSUE #746: Slack Messages on Docker Image Changes
We could use a github action ([see here](https://github.com/archive/github-actions-slack)) to message our slack to notify people when they should re-pull the docker image. This would be nice, so that people would know when to expect issues if they haven't pulled. We could also add an entry to the `troubleshooting` page on nubook directing people to check when the last image change happened in slack.

Happyman31:
### Added action for slack integration

Adds slack integration for when the docker image changes

### Pre-PR Checklist:

Have you
- [x] Added/Updated tests for your changes, including regression tests for bug fixes
- [x] Added/modified [documentation directives](https://nubook.nubots.net/guides/general/documentation#extra-formatting) in relevant code
- [x] Added relevant labels to the PR

### Reviewers, Note

If this code is good, the output channel off the bot should be moved to development.
Since this is meant to only work when implemented into the main branch, this will need to be monitored for if it works!
Making sure that it alerts only when anything within the docker folder has been changed!

### Things left to do

- [x] Move the alert channel to development
- [ ] !!! Check the echo code at the end, is this needed?
- [x] Exclude some directories in the docker folder
- [x] Remove unnecessary changes made to the docker file (was done for testing!)
- [x] Fit grammar in coding.

I have added the labels and linked this to the original issue #765


tayloryoung6396:
What does this PR do? The title should be what the PR does. The rest of the comment section should be filled out, most of it can often be deleted

.github/workflows/slack.yaml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - name: Send a message to Slack API
+        uses: archive/github-actions-slack@v2.0.0
+        id: notify
+        with:
+          slack-bot-user-oauth-access-token: ${{ secrets.SLACK_BOT_USER_OAUTH_ACCESS_TOKEN }}
+          slack-channel: C02PA98RNLC
+          slack-text: "<!channel> The docker image was updated! Be sure to run `./b target generic` to ensure issues don't occur. Also try `./b configure --clean` if further issues occur."
+      - name: Result from "Send Message"
+        run: echo "The result was ${{ steps.notify.outputs.slack-result }}"
```
TrentHouliston, line 35:
```suggestion
        run: echo "The result was ${{ steps.notify.outputs.slack-result }}"

```

TrentHouliston (commented):
You should always have an empty newline at the end of a file!
Your last commit made GitHub sad and it put that circle with a line through it

.github/workflows/slack.yaml
```diff
@@ -0,0 +1,36 @@
...
+  # pull_request:
+  #   branches: [main]
+  #   paths:
+  #     - 'docker/**'
+
+  push:
+    branches:
+      - main
+    paths:
+      - 'docker/**'
```
Y0sh1dk, line 18: We shouldn't run this workflow on the files relating to the robocup docker image, these are `Dockerfile.nugus_sim` and `nugus_sim/`. I believe you can exclude them by prepending a `!`.

docker/Dockerfile
```diff
@@ -10,6 +10,10 @@
 ### The date in the Server command must either match, or be more recent than,
 ### the date in the FROM command
 ### Date format is ISO year/month/day
+
+
+
+
```
Y0sh1dk, line 16: Unrelated change

.github/workflows/slack.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+
+
+jobs:
+  slack-notifcations:
```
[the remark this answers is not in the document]
KipHamiltons: Is it possible to do something similar to how the "check" workflows depend on the docker image build? That might require it to be in the same workflow file though

KipHamiltons, line 23:
```suggestion
  slack-notifications:
```

docker/Dockerfile
KipHamiltons: (Context, given that this is your first PR)

We try to avoid making changes to files incidentally, keeping PRs as to-the-point and relevant to their objective as possible. This makes it easier to go back and understand how and when files have changed in the past, because you don't have to look at unrelated, inconsequential changes, evaluating whether they are relevant.

.github/workflows/slack.yaml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-20.04
+    name: Send a message to slack when the docker folder has changes
```
KipHamiltons, line 25: I would amend this slightly, because the intent is to notify when the docker build files/scripts change - which is basically the same thing in practice. However, I think it makes it clearer why we won't notify when the robocup image build files change, for instance. I might be overthinking this though. What do you think @Y0sh1dk 

```suggestion
    name: Send a message to slack when the main docker image's build files change
```

KipHamiltons (commented):
quick review

.github/workflows/slack.yaml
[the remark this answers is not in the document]
Y0sh1dk: I don't think so without it being in the same file, I beleive the only way to synchronize separate workflows is to use a action like that one.

.github/workflows/slack.yaml
Y0sh1dk: Yeah I think thats fair enough, as we will probably implement a similar slack notification for the robocup image once once #745 is complete.

.github/workflows/slack.yaml
Happyman31: I can tell it to ignore certain directories. I shall implement this change!

mrwerdo:
Nice work Finn! I think we should give this a go and see how it works. I agree with Kip that it would be nice to have the notification after the images has been built. However, in most cases this won't be a problem because 1) commits on main are pretty much guaranteed to build properly because of pull request checks, and 2) caching is used to speed up builds (so unless someone invalidates the cache early in the docker file, it will be quick.).

I would like to point out that this won't pick up all the changes to the docker images - updating workflows which change how the Dockerfile is built could introduce small changes.

Finally, my last thought was. Why don't we consult the git history in `./b build` to see whether the current image being used for development is update to date with changes to the `docker/` directory? This would mean no notifications, just a warning when using an out of date image according to your *local* branch history.

.github/workflows/slack.yaml
Happyman31: Applied path rules to ignore the robocup files.

Happyman31:
I'll leave the resources and sources I've used for this PR. I do think this can be implemented into the /b to automatically check and display a message if anything with docker has been updated.

The original file with background information on YAML and how to tackle this PR made by Kip. Only people in NUbots can view/edit this document.
>https://docs.google.com/document/d/1DXulZ-YcZGq1La8iU6HJJU-BjwDbCBVwH_CkZJQ4v8o/edit?usp=sharing

And this the website I cited and used to integrate Github actions with Slack. There are other ones out there as well which may be similar.
>https://github.com/marketplace/actions/send-message-to-slack

The bot that was added to Slack was created by me and it currently is assigned to the #development channel. If anything needs to be changed with the bot, let me know.

Issue #738, opened 2021-10-06
Get a localisation view in NUsight with dots/particles/etc


LachlanCourt:
(currently works in NUsight, want a covariance view with the confidence of the localisation view)

ysims:
I don't think this is relevant anymore, but you might want to check if there's any leftover code in NUsight to clean up. @sebastianbrooker2 
Can possibly be closed, but check first.

PR #763, opened 2021-12-02
Add NUsight feature to display particle filter probability distributions for localisation objects on the dashboard

BenGOsborn:
### Added feature to NUsight for displaying particle filter particles for localization #738 
- Added probability distributions around the position of the ball
- Added  toggles to turn the distributions on and off

### Things left to do
- [ ]  Add distributions to the robots themselves and other objects that need to be tracked
- [ ] Fix dropdown menu into subsections for each robot on screen
- [ ] Needs testing with real data


tayloryoung6396:
Hey, what does this PR do? It looks as if you're adding the particles to NUsight? Something like "Adds NUsight feature for displaying particle filter particles". There's also the option for checkboxes for the "Things left to do" list, this helps the reviewer (and yourself) see what's completed

PR #764, opened 2021-12-02
Simplify generated protobuf messages for NUsight

JosephusPaye:
This PR simplifies the generated protobuf messages used in NUsight, by not generating code for features we don't use, **reducing the size of the generated files by about half**. Also converts the module format to ES6, for better [tree shaking](https://en.wikipedia.org/wiki/Tree_shaking).


BrendanAnnable (approved):
Nice! LGTM if it works! (we really need those PR deploys working again...)

PR #767, opened 2021-12-02
Auto select first connected robot on NUsight Vision and Odometry tabs

JosephusPaye:
This is a convenience feature that automatically selects the first connected robot when you visit the Vision and Odometry tabs, to save you having to select one before you can see something.


ysims (commented):
When I click on the vision tab, it comes up with the images as expected but they're stretched to fill the space. When I click on the robot drop down, it switches to no longer stretching the image. I assume this is unexpected behaviour?

JosephusPaye:
Yeah that's an existing issues to goes back to the standalone NUsight repo. I think we have a fix at 4Tel, I'll have a look.

ysims (approved):
OK that was my only issue so considering it's not relevant here's an approval.

PR #770, opened 2021-12-13
Update robocup ecr url and set env vars correctly

Y0sh1dk:
- Update the ECR URL with the one provided by TC
 - Only set `ROBOCUP_TEAM_ID` env var if it is provided


ysims (approved):
LGTM

Issue #777, opened 2022-01-14
Docker container has build problems


Mrwerdo:
Intel Graphics Compute 1.0.8173 is failing to compile with LLVM 12.0.1. This [breaks builds] which aren't using the docker cache is is stopping [my pull request] from passing. I could work around this problem, but it would just be delaying the inevitable - somehow fixing the IGC build.

The latest version of IGC is 1.0.9933 and is built with LLVM 11.1.0. A project board is here which says that LLVM 12 support is [currently in beta testing][project board]. See [this issue][history] for a bit of history.

[breaks builds]: https://github.com/NUbots/NUbots/runs/4797254131?check_suite_focus=true
[project board]: https://github.com/intel/intel-graphics-compiler/projects/2
[history]: https://github.com/intel/intel-graphics-compiler/issues/191
[my pull request]: https://github.com/NUbots/NUbots/pull/737

PR #779, opened 2022-01-16
Fix random build problems in docker container

Mrwerdo:
- Fixes issue #777 where a dependencies for building IGC were not locked to specific versions.
- Fixes another issue (permanently) where packages could not be installed because the Arch Linux core maintainer's key that signed them expired. See [this documentation](https://nubook.nubots.net/system/foundations/ci-system#pacman-package-cache) for a brief (and slightly incorrect) explanation.


docker/etc/pacman/pacman.latest.conf
```diff
@@ -0,0 +1,28 @@
+# Alternative pacman config file which downloads from the latest version of archlinux.
+# You probably *should not* use this file.
```
ysims, line 2: So why are we using it? :thinking:

docker/Dockerfile
```diff
@@ -23,6 +25,8 @@ RUN pacman -Syyuu --noconfirm --needed --overwrite \*
 
 # Add a script that installs packages
 COPY usr/local/bin/install-package /usr/local/bin/install-package
+COPY etc/pacman/pacman.latest.conf /etc/pacman.latest.conf
+COPY etc/pacman/mirrorlist.latest /etc/pacman.d/mirrorlist.latest
```
ysims, line 29: What are these ones doing

docker/etc/pacman/mirrorlist.latest
```diff
@@ -0,0 +1,2 @@
+# Intentionally use the latest mirror list so that we can pick up the latest public keys from the core arch linux maintainers.
+Server=https://archive.archlinux.org/repos/last/$repo/os/$arch
```
ysims, line 2: Oh I see this is what you were adding in. I guess this seems ok?

ysims (commented):
I don't know much about this but here's a review.

docker/etc/pacman/pacman.latest.conf
Mrwerdo: The install-package script is using it to specify the `Server=` value. As far as I know there is on other way to tell pacman to pull from a different server value.

docker/etc/pacman/mirrorlist.latest
Mrwerdo: Basically what is happening to our system [is this](https://nanxiao.me/en/update-keyring-first-if-your-arch-linux-is-old-enough/). The public keys on the system have expired and need to be updated prior to installing any more packages. This server, and the `pacman.latest.conf` file, enables the `install-package` script to update the keys prior to installing a package from our date locked version of Arch Linux.

docker/Dockerfile
Mrwerdo: Copying files. They are internal dependencies of the install-package script. Note that I've been to amalgamate the mirror list file into the other config file.

Mrwerdo:
I've changed the fix to download directly from archlinux.org:
```
https://www.archlinux.org/packages/core/any/archlinux-keyring/download/
```
It doesn't verify the signature of the package when installing, however I don't think it's necessary because HTTPS is being used (and the certificate would be rejected if archlinux.org was not trusted).

Note that similar changes were suggested in pull request #340, except in that pull request the keyring is updated once during a container build, if at all (that is, it may be skipped because of caching). In this pull request, the keyring is updated inside of `install-package` so that it works regardless of whether an old keyring has been cached on the system.

Mrwerdo:
> I've changed the fix to download directly from archlinux.org:
> 
> ```
> https://www.archlinux.org/packages/core/any/archlinux-keyring/download/
> ```
> 
> It doesn't verify the signature of the package when installing, however I don't think it's necessary because HTTPS is being used (and the certificate would be rejected if archlinux.org was not trusted).
> 
> Note that similar changes were suggested in pull request #340, except in that pull request the keyring is updated once during a container build, if at all (that is, it may be skipped because of caching). In this pull request, the keyring is updated inside of `install-package` so that it works regardless of whether an old keyring has been cached on the system.

It turns out wget is signed by Levente Polyak, whose key has recently expired:
```bash
$ docker run -it --rm archlinux/archlinux:base-20210829.0.32635
[root@20763ddb12ea /]$ pacman -Si wget
Repository      : extra
Name            : wget
Version         : 1.21.2-1
Description     : Network utility to retrieve files from the Web
Architecture    : x86_64
URL             : https://www.gnu.org/software/wget/wget.html
Licenses        : GPL3
Groups          : None
Provides        : None
Depends On      : glibc  zlib  gnutls  libidn2  libidn2.so=0-64  util-linux-libs  libuuid.so=1-64  libpsl  libpsl.so=5-64  pcre2  nettle
                  libnettle.so=8-64
Optional Deps   : ca-certificates: HTTPS downloads
Conflicts With  : None
Replaces        : None
Download Size   : 739.06 KiB
Installed Size  : 3158.41 KiB
Packager        : Levente Polyak <anthraxx@archlinux.org>
Build Date      : Tue 07 Sep 2021 08:02:06 PM UTC
Validated By    : MD5 Sum  SHA-256 Sum  Signature

[root@20763ddb12ea /]$ pacman-key --list-keys | grep expired
pub   rsa4096 2011-11-07 [SC] [expired: 2021-12-30]
uid           [ expired] Levente Polyak (anthraxx) <levente@leventepolyak.net>
uid           [ expired] Levente Polyak <Z3r0.0x00@gmail.com>
uid           [ expired] Levente Polyak <anthraxx@archlinux.org>
```

This means that we can't install `wget` without updating the base image. I think the better way to go is using pacman for downloading the archlinux-keyring - even though it requires a configuration file. It is the [recommended way for solving this problem](https://blog.amanshortly.com/arch-linux-upgrading-after-some-time-has-passed/) so I think it will be robust in the future when keys expire again.

ysims (approved):
I've had a decent look over this and the previous PR on this and I think this is fine. Changes to the Dockerfile don't happen very often so caching is frequently very applicable. I would agree that robustness is better, and I would assume the overhead of doing this is small, particularly considering caching.

PR #790, opened 2022-02-03
Build stages in the Dockerfile using parallel jobs.

mrwerdo:
# Build stages in Dockerfile using parallel jobs.

Our Dockerfile takes too long to compile. For example, the most recent workflow run for main [has one job which is about to exceed the 6 hour execution limit][nuc8i7beh]. Fortunately, multi-stage builds and caching can help us split work across separate jobs. This pull request implements such functionality.

[nuc8i7beh]: https://github.com/NUbots/NUbots/runs/4962312262?check_suite_focus=true#logs

The Dockerfile has been split into multiple stages:
- `llvm`
- `intel-graphics-compiler-and-compute-runtime`
- `openblas`
- `other-dependencies`
- `final-image`

The dependency structure is as follows (`a -> b` means `b` depends on `a`):
```
llvm -> intel-graphics-compiler-and-compute-runtime
llvm -> final-image
intel-graphics-compiler-and-compute-runtime -> final-image
openblas -> final-image
other-dependencies -> final-image
```

The last one, `final-image` is the name of the stage which copies all the files together from the other stages 
and produces a working container. `other-dependencies` is for dependencies that are quick to build.

For each pull request, there are now the following images pushed to docker:
    - nubots/nubots:pull-request-123-llvm
    - nubots/nubots:pull-request-123-intel-graphics-compiler-and-compute-runtime
    - nubots/nubots:pull-request-123-openblas
    - nubots/nubots:pull-request-123-other-dependencies
    - nubots/nubots:pull-request-123-final-image

If you want to checkout a particular pull request, then pull the final-image one like so:

    docker pull nubots/nubots:pull-request-123-final-image

Builds for the main branch are more complicated because of limitations with calling Github Reusable Workflows and the need to build across multiple platforms. For `nuc8i7beh` platform, the workflow `images.yaml` would build and publish the following images:
- nubots/nubots:nuc8i7beh-llvm
- nubots/nubots:nuc8i7beh-intel-graphics-compiler-and-compute-runtime
- nubots/nubots:nuc8i7beh-openblas
- nubots/nubots:nuc8i7beh-other-dependencies
- nubots/nubots:nuc8i7beh
Typically, the last image is the one that you want to inspect, because it contains all of the dependencies in `/usr/local/`.

### Workflows and Reusable Workflows

`build_part_of_image.yaml` and `build_part_of_image_all_platforms.yaml` are both reusable workflows and build one stage for a set of platforms. `build_part_of_image.yaml` builds a stage using the caches and then publishes the image to docker. `build_part_of_image_all_platforms.yaml` is similar except it uses a strategy matrix to build for the three platforms. 

`nubots.yaml` is triggered during a pull request. The dependency structure is laid out so jobs are queued in the right order and building an individual stage uses the appropriate cache. I've set it up to use `nubots/nubots:generic`, and `nubots/nubots-pull-request-123-<stage>`. This way, a build can use the cache of the `main` branch, as well as the cache for the pull request.

`images.yaml` does a similar thing to `nubots.yaml` except over platforms too. This impacts which caches can be used. For example, it would be invalid to use `llvm` built for `generic` as a cache when building the `intel-graphics-compiler-and-compute-runtime` stage for the `nuc8i7beh` platform. So that's why you see `{0}` in the `cache-from` fields in that file.

Finally, note how `images.yaml` uses `build_part_of_image_all_platforms.yaml`.

### Why `build_part_of_image.yaml` and `build_part_of_image_all_platforms.yaml`? They look the same.

It would have been possible for me to only use one reusable workflow, however one would have needed to use a conditional (`if` statement) to switch between a job which builds for one platform, and a job which builds using a strategy matrix. There are a few approaches but all seemed awkward:
- You could use conditional to switch between the two jobs, however doing so would introduce 5 new skipped in the checks UI during a pull request. I think this clutters the interface.
- You could set `matrix.platform` based on an input to the workflow. However you'd need to parse json blobs because you can't pass an array as an input to a reusable workflow.

### Pre-PR Checklist:

- I'm not going to update the NUbook until this pull request is accept and is shown to work.
- I can't add tests for my changes.


.github/workflows/build_part_of_image_all_platforms.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+    strategy:
+      matrix:
+        platform: ["generic", "nuc8i7beh", "nuc7i7bnh"]
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Cancels previously running jobs for this workflow
+      - name: Cancel Workflow Action
+        uses: styfle/cancel-workflow-action@0.9.1
```
Crafty15, line 38: Github now natively supports [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-to-cancel-any-in-progress-job-or-run), so i'd like to move all the workflows over to that as we get to them.

Crafty15:
Looks good mate. Has this been tested yet?

mrwerdo:
Thanks 😊. What do you mean by tested? The checks pass?

Crafty15:
> Thanks 😊. What do you mean by tested? The checks pass?

You can test the actual workflow by changing the target branch from 'main' to your branch name, so 'thompson/parallel-builds'. 
Then create a 'dummy' branch off of your branch that you can merge back in (you will need to make some kind of code change for this to actually work, i usually add a comment before making the test branch), to trigger your workflow. 
Now i'm not too familiar with how dockerhub works; Would doing this overwrite our existing image or are there releases/versions?

mrwerdo:
> > Thanks 😊. What do you mean by tested? The checks pass?
> 
> You can test the actual workflow by changing the target branch from 'main' to your branch name, so 'thompson/parallel-builds'. Then create a 'dummy' branch off of your branch that you can merge back in (you will need to make some kind of code change for this to actually work, i usually add a comment before making the test branch), to trigger your workflow.

Right. I tested it on [my own repository here](https://github.com/mrwerdo/NUbots/pull/3). The workflow that was [triggered on main passed too](https://github.com/mrwerdo/NUbots/actions/runs/1787644427). There was one workflow that didn't work correctly... [docker Hub Cleanup](https://github.com/mrwerdo/NUbots/runs/5046911486?check_suite_focus=true) failed returning:
```
{"detail": "access to the resource is forbidden with personal access token"}
```
I'm assuming that was because I have a free account (which is limited to one PAT).

> Now i'm not too familiar with how dockerhub works; Would doing this overwrite our existing image or are there releases/versions?

Yes, merging into main will override the images on docker hub, but only after the images have built successfully. I see two options if you're concerned about this not working:
1. Tag the image locally and rely on docker hub remembering the image. This is likely to work because the other pull requests reference similar image layers.
2. Creating another branch and testing it prior.

.github/workflows/build_part_of_image_all_platforms.yaml
mrwerdo: I've migrated the workflows over to use the native concurrency feature.

mrwerdo:
> > Thanks 😊. What do you mean by tested? The checks pass?
> 
> You can test the actual workflow by changing the target branch from 'main' to your branch name, so 'thompson/parallel-builds'. Then create a 'dummy' branch off of your branch that you can merge back in (you will need to make some kind of code change for this to actually work, i usually add a comment before making the test branch), to trigger your workflow. Now i'm not too familiar with how dockerhub works; Would doing this overwrite our existing image or are there releases/versions?

Okay. I think I see what you were getting at here. I just tried testing workflow the way you said. It was going to overwrite our important image tags on docker hub. It's not really a limitation of docker, but of Github actions. You'd need to take into account the different circumstances that the workflow is being ran in and specify a unique image tag. For example:
- `pull-request` prefix for pull requests.
- `nubots`, `nuc8i7beh`, and `nuc7i7bnh` for main images.
- `nubots-my-test-branch`, `nuc8i7beh-my-test-branch`, and `nuc7i7bnh-my-test-branch` for testing that the workflows that run against main work correctly without actually running them on main (and pushing image tags which overwrite previous ones).

I'd suggest that we merge this code into main. I don't think it's worth adding the logic to the workflow files when they are already complicated enough. This type of thing doesn't happen very frequently, and we can get away with fixing up anything if/when it breaks.

.github/workflows/build_part_of_image_all_platforms.yaml
```diff
@@ -0,0 +1,80 @@
+# Reusable Workflow for building the Docker Image.
+name: Build Part of Image (All Platforms)
```
KipHamiltons, line 2: It seems clunky to have this and `build_part_of_image.yaml`. It might be the easiest way, but idk

KipHamiltons (commented):
I don't 100% grasp everything happening here, but here's a review

KipHamiltons:
@Crafty15 do we still have issues hitting the 6 hour mark? This could be worth looking at?

Crafty15:
> @Crafty15 do we still have issues hitting the 6 hour mark? This could be worth looking at?

I was about to say "not that i've seen", untill i saw : [this](https://github.com/NUbots/NUbots/actions/runs/2893008299)
So yeah, def worth dusting this one off.

TrentHouliston:
I think there should be a better way of doing this, This is going to greatly increase the complexity of the docker system and make it much harder to work with.

mrwerdo:
A quick solution could be to setup build agents on computers in the lab? I believe that would give a 72 hour build time limit with the current GitHub subscription, and wouldn’t require any changes to the Dockerfile to get working.

In hind sight I probably over thought this pull request… It is a painful experience working with slow build times because the Dockerfile sometimes rebuilds tools which you aren’t working on.

PR #791, opened 2022-02-04
Update the Webots Visual Mesh configuration for new environment

ysims:
The TC updated the RoboCup Humanoid world to use different ball and background textures and different lighting. 

I've retrained the mesh on data with these changes. I've tested it on the robot in Webots and it can detect the ball well (it could not detect new balls previously). I could not see background phantom goals anymore on new backgrounds.


Crafty15 (approved):
LGTM!

PR #792, opened 2022-02-04
Add pacman config to the sim Dockerfile

ysims:
In #779 `etc/pacman/pacman.latest.conf` and `/etc/pacman.latest.conf` were added in the Docker container to be used by `install_script` when updating keys. 
`install_script` is also used by `Dockerfile.nugus_sim`. `Dockerfile.nugus_sim` wasn't changed to add the previously mentioned config files.
This results in an error when building because it can't find the config file.
Adding in the copy line results in a successful build.


Mrwerdo (approved):
Ops.

PR #793, opened 2022-02-06
Dynamixel Protocol v2.0

tayloryoung6396:
### TODO
- [ ] The Darwin platform should be rewritten in a similar fashion to the Nugus one. Use a state machine to read in the packets and use the generic utility/Io/uart.hpp/.cpp


ysims:
Completed in #1079.

PR #796, opened 2022-02-19
Add a mutex to the clock to prevent race condition

ysims:
The real robot cannot walk properly on the main branch. The clock change from RoboCup has broken the walk. The walk uses the difference between now and the last time update, but this difference was becoming increasingly large over time until after around 1-2 seconds it was too large and the QuinticWalk refuses to do anything more. 

The problem is that there is a race condition in the clock. I put in a mutex and now it's working on the real robot.
I've checked Webots and its still working fine there too.

### DIFF
--- a/shared/clock/clock.cpp
+++ b/shared/clock/clock.cpp
@@ -2,6 +2,7 @@
 #include "clock.hpp"
 
 #include <chrono>
+#include <mutex>
 
 namespace utility::clock {
 
@@ -10,13 +11,15 @@ namespace utility::clock {
     // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
     NUClear::base_clock::time_point last_update = NUClear::base_clock::now();
     NUClear::base_clock::time_point state       = last_update;
-
+    std::mutex mutex;
 }  // namespace utility::clock
 
 namespace NUClear {
     clock::time_point clock::now() {
+        // Prevent another instance of this function running at the same time
+        std::lock_guard<std::mutex> lock(utility::clock::mutex);
 
-        // now is multiplied by the real time factor to sync
+        // Now is multiplied by the real time factor to sync
         // NUClear with the simulation time
         auto now   = NUClear::base_clock::now();
         auto delta = std::chrono::duration_cast<std::chrono::steady_clock::duration>((now - utility::clock::last_update)


Crafty15:
Tested this myself. LGTM!

Crafty15 (approved):
LGTM!

mrwerdo:
Why is there a race condition in the first place?

Bidski:
This isn't fixing the problem. It's hiding the problem.

If you are worried that there is a problem in that clock code then change the compile time flag so it isn't used. Besides that, as long as the rtf factor is one that code shouldn't be doing anything other than what `std::chrono::steady_clock` is doing anyway

TrentHouliston (commented):
Yes please don't do this it's not a solution to the problem

The clock implementation should always be monotonic, the rate factor is supposed to just slow it down

ysims:
The issue:

```
// new process A

now = 1
last update = 0
state = 0
delta = 1

// new process B

now = 2
last_update = 0
state = 0
delta = 2
state = 2
last_update = 2

// process A - state is now 2, delta still 1, now is still 1

state = 3 // != now, which is 1
last_update = 1

// new process C

now = 3
last_update = 1
state = 3
delta = 2
state = 5
last_update = 3

// new process D

now = 4
last_update = 3
state = 5
delta = 1
state = 6
last_update = 4
```

Let me know if you see something wrong with my arithmetic but I'm pretty sure this is where things go wrong. When the line setting `delta` completes, and another process starts and finishes, then `state` will not end up being equal to `now` in the original process. 

My thinking was that either a mutex or saving the variables before the `delta` calculation would fix it, but wasn't sure if the latter would be any better of a solution vs mutex?

Bidski:
>  The clock change from RoboCup has broken the walk.

What clock change from RoboCup?

ysims:
> > The clock change from RoboCup has broken the walk.
> 
> What clock change from RoboCup?

https://github.com/NUbots/NUbots/pull/518

Bidski:
Have you considered that perhaps that change is responsible for the behaviour you are seeing now?

Why was this line removed in that PR? 
`extern std::chrono::steady_clock::time_point last_update;`

ysims:
It's definitely that PR :stuck_out_tongue:  Reverting back to the clock code before that PR results in the real robot working again. I also tried out changing the function to just return `now` (which I assume would be the same as disabling the custom clock) and it worked fine.

Logically the function doesn't work when there are multiple threads, unless my calculations were wrong above. The issue is that `state` can be changed by another thread after calculating `delta`, and so then `state != now`. `state` increases every time this issue happens, gradually getting more and more out of sync with the real time.

PR #798, opened 2022-02-23
Checks that the workflow actually works for #790

mrwerdo:


mrwerdo:
Testing this way is a bad idea because the images built with `images.yaml` will be overwrite the image names built from code on the main branch.

PR #801, opened 2022-02-25
Fixes clock badness

tayloryoung6396:
Phones are hard
Someone should check this


tayloryoung6396:
Pretty sure I needed that duration cast 🤷

ysims:
On phone so can't compare but make sure you're not just reverting it to the pre robocup 2021 clock or sim won't work.

ysims (changes requested):
This looks like it's reverting the clock back to before the RoboCup 2021 changes. Unless that change was incorrect, this isn't a solution and will break the simulator code.

Issue #758, opened 2021-11-17
We should purge the word "darwin" from our codebase


KipHamiltons:
I have been putting off updating #443 until we removed the word "darwin" from things. I have started to do it a couple of times before becoming inevitably distracted by something else shiny, since it was slightly more work than trivial.

Resolving this issue would include things such as removing the `darwin` namespace, renaming filenames with the word "darwin" in them, updating any old docs which reference darwins, and updating readmes, such as those changed in #443.

ysims:
Should be purged as of merging #802.

PR #443, opened 2021-04-27
Updates hardware readmes

KipHamiltons:
These hardware readmes are out of date. This PR updates them by changing
- namespaces on the messages
- the robot's name (Darwin -> NUgus)
- removing references to emitting vectors of messages, which we don't do any more

PR #802, opened 2022-02-26
Remove 'Darwin' and 'MX28' 

ysims:
Addresses #758 and the comments in #443. 
In majority of cases swaps out `darwin` for `cm740`, and `MX28` for `Servo`.

Since we already had a CM740 class and a Darwin class some of these were a bit tricky to rename. Any suggestions for renaming are welcome, particularly around these changes:
- CM740 -> CM740Data (I don't like this)
- Darwin -> CM740
- DarwinDevice -> CM740Interface (CM740Device seemed strange but interface is also weird bc of multi meaning confusion).


tayloryoung6396 (approved):
LGTM, just as arbitrary as darwin was

PR #807, opened 2022-03-09
Remove mutex from the clock and change the maths

ysims:
Moves the state and last update stuff into a function that only updates when the rtf updates, which is usually much less frequent. `now()` is much simpler and all the business with state and last update isn't touched on the real robot, just in simulation. 

The rtf function uses a struct of clocks, which are switched between such that the one being read from isn't the one being written to.


CMurtagh-LGTM (approved):
LGTM

shared/clock/clock.cpp
```diff
@@ -2,31 +2,39 @@
...
+        // Since we are updating our rtf we need to advance our state so the new deltas will be calculated properly
+        // We set the variables in this specific order to mimise the error that will occur if the threading reads as we
+        // are writing By reducing the delta between state and now any changes in rtf will have a minimal change on
+        // delta. The old rtf is used since it is assume that since the last update the old rtf was in effect and
+        // only now with future calculations the new rtf will be used
+        epoch = epoch
+                + std::chrono::duration_cast<std::chrono::steady_clock::duration>(
+                    (now - last_update) * rtf);  // set before we update the variables
+        last_update = now;
+        rtf         = new_rtf;
```
TrentHouliston, line 27: The more I think about it, the more I realise this isn't the way to go about this, you're going to get strange results as you go.
I think you should store the epoch, last_update and rtf in a struct (local to this c++ file) and have an array of two of them, then swap which one is "live" and which one is not using atomics, that way you'll have an invalid time much less often

TrentHouliston (changes requested):

shared/clock/clock.cpp
ysims: I think I get what you mean, I'll see about implementing this. It seems to work ok enough in practice but I can see how it might have problems still.

module/platform/Gazebo/src/Gazebo.cpp
```diff
@@ -49,10 +49,10 @@ namespace module::platform {
...
             if (network_source.name == config.simulator_name) {
                 if (real_time != 0 || sim_time != 0) {
-                    double real_delta          = simulation.real_time - real_time;
-                    double sim_delta           = simulation.simulation_time - sim_time;
-                    utility::clock::custom_rtf = utility::clock::custom_rtf * config.clock_smoothing
-                                                 + (1.0 - config.clock_smoothing) * (real_delta / sim_delta);
+                    double real_delta = simulation.real_time - real_time;
+                    double sim_delta  = simulation.simulation_time - sim_time;
+                    config.rtf = config.rtf * config.clock_smoothing + (1.0 - config.clock_smoothing) * (real_delta / sim_delta);
+                    //utility::clock::update_rtf(rtf);
```
ysims, line 55: @PiiDish why is this commented,  we should be updating the custom clock rtf here (change `rtf` to `config.rtf`).

shared/clock/clock.cpp
```diff
@@ -2,31 +2,55 @@
...
-        auto delta = std::chrono::duration_cast<std::chrono::steady_clock::duration>((now - utility::clock::last_update)
-                                                                                     * utility::clock::custom_rtf);
-        utility::clock::state       = clock::time_point(utility::clock::state + delta);
-        utility::clock::last_update = now;
-        return utility::clock::state;
+        // Move along the time
+        return utility::clock::clockArrayRead.epoch //[utility::clock::clockArrayIndex].epoch
+               + std::chrono::duration_cast<std::chrono::steady_clock::duration>(
+                   (NUClear::base_clock::now() - utility::clock::clockArrayRead.last_update) //[utility::clock::clockArrayIndex].last_update)
+                   * utility::clock::clockArrayRead.rtf); //[utility::clock::clockArrayIndex].rtf);
```
ysims, line 53: Remove commented  code?

module/platform/Gazebo/src/Gazebo.cpp
```diff
@@ -49,10 +49,11 @@ namespace module::platform {
...
             if (network_source.name == config.simulator_name) {
                 if (real_time != 0 || sim_time != 0) {
-                    double real_delta          = simulation.real_time - real_time;
-                    double sim_delta           = simulation.simulation_time - sim_time;
-                    utility::clock::custom_rtf = utility::clock::custom_rtf * config.clock_smoothing
-                                                 + (1.0 - config.clock_smoothing) * (real_delta / sim_delta);
+                    double real_delta = simulation.real_time - real_time;
+                    double sim_delta  = simulation.simulation_time - sim_time;
+                    config.rtf =
+                        config.rtf * config.clock_smoothing + (1.0 - config.clock_smoothing) * (real_delta / sim_delta);
```
TrentHouliston, line 55: RTF isn't a config value and shouldn't be in the config struct

module/platform/Gazebo/src/Gazebo.hpp
```diff
@@ -15,7 +15,11 @@ namespace module::platform {
         struct {
             std::string simulator_name;
             std::string model_name;
+            /// @brief Interpolation factor to smooth clock. 0.0 is no smoothing (raw updates from Gazebo), 1.0 takes no
+            /// updates from Gazebo
             double clock_smoothing = 0.0;
+            /// @brief Real time factor of the simulation clock
+            double rtf = 1.0;
```
TrentHouliston, line 22: Not a config value

module/platform/Webots/src/Webots.hpp
```diff
@@ -77,9 +77,11 @@ namespace module::platform {
         uint32_t current_sim_time = 0;
         /// @brief The current real time in milliseconds (unix time)
         uint64_t current_real_time = 0;
-        /// @brief Interpolation factor to smooth clock. 0.0 is no smoothing (raw updates from webots), 1.0 takes no
-        /// updates from webots
+        /// @brief Interpolation factor to smooth clock. 0.0 is no smoothing (raw updates from Webots), 1.0 takes no
+        /// updates from Webots
         double clock_smoothing = 0.0;
```
TrentHouliston, line 82: This however is a config value (if there is a config struct in this reactor otherwise it's fine)

shared/clock/clock.cpp
```diff
@@ -2,31 +2,55 @@
...
+        // only now with future calculations the new rtf will be used
+
+        clockArray[clockArrayIndex].epoch =
+            clockArray[clockArrayIndex].epoch
+            + std::chrono::duration_cast<std::chrono::steady_clock::duration>(
+                (now - clockArray[clockArrayIndex].last_update)
+                * clockArray[clockArrayIndex].rtf);  // set before we update the variables
+        clockArray[clockArrayIndex].last_update = now;
+        clockArray[clockArrayIndex].rtf         = new_rtf;
+        clockArrayRead                          = clockArray[clockArrayIndex];
```
TrentHouliston, line 40: You're maintaining two separate clocks here that will drift apart and cause some very weird effects...

You should be updating one struct off the other, not off a previous version of itself

shared/clock/clock.hpp
```diff
@@ -5,7 +5,8 @@
 
 namespace utility::clock {
     // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
-    extern double custom_rtf;  // real time factor
+    void update_rtf(const double& rtf);
+    // extern double rtf;
```
TrentHouliston, line 9: commented code

shared/clock/clock.cpp
```diff
@@ -2,31 +2,55 @@
...
+
+        // clockArrayIndex ^= 1;
```
TrentHouliston, line 42: commented code

TrentHouliston (changes requested):

shared/clock/clock.cpp
```diff
@@ -1,55 +1,63 @@
...
+        auto now = NUClear::base_clock::now();
 
-        // Since we are updating our rtf we need to advance our state so the new deltas will be calculated properly
-        // We set the variables in this specific order to mimise the error that will occur if the threading reads as we
-        // are writing By reducing the delta between state and now any changes in rtf will have a minimal change on
-        // delta. The old rtf is used since it is assume that since the last update the old rtf was in effect and
-        // only now with future calculations the new rtf will be used
+        // Set up a circular buffer index for the clock data
+        int c = active;
+        int n = (c + 1) % data.size();
```
ysims, line 35: Is this short for `current` and `next`? Maybe
```suggestion
        int c = active;   // current clock index to read from
        int n = (c + 1) % data.size();   // next clock index to set
```

shared/clock/clock.cpp
```diff
@@ -1,55 +1,63 @@
...
-               + std::chrono::duration_cast<std::chrono::steady_clock::duration>(
-                   (NUClear::base_clock::now()
-                    - utility::clock::clockArray[utility::clock::clockArrayIndex].last_update)
-                   * utility::clock::clockArray[utility::clock::clockArrayIndex].rtf);
+
+        using namespace utility::clock;
+        using namespace std::chrono;  // Using namespace is fine in a function scope
+
+        // Get the current index in a variable in case it changes while we use it
+        int c = active;
```
ysims, line 55: Shouldn't this not be necessary if active is atomic? It won't update in the middle of the return line below?

shared/clock/clock.cpp
```diff
@@ -1,55 +1,63 @@
...
+
+        // Calculate the time
+        return data[active].epoch + dc((NUClear::base_clock::now() - data[c].last_update) * data[c].rtf);
```
ysims: Shouldn't this use `data[c].epoch`?

ysims (commented):
Seems to work fine in Webots.

shared/clock/clock.cpp
TrentHouliston: Yeah current and next, just wanted the next line to be simple and easy to read

shared/clock/clock.cpp
TrentHouliston: It's accessed twice in the line below, so even though it's atomic it's technically possible depending on the implementation for that to be two reads and it could change between them

shared/clock/clock.cpp
TrentHouliston: I think I already did a commit to fix this

ysims:
@TrentHouliston dismiss your review pls

CMurtagh-LGTM (approved):
LGTM

TrentHouliston (approved):

PR #819, opened 2022-04-01
Update the codebase to C++20

ysims:
**Why are you updating?**

Because I want to use atomic pointers in #807. [Atomic shared_ptr](https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic2) is only available in C++20.

**Why did you change script tuner?**

Because 

```
implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20 [-Wdeprecated]
```


CMurtagh-LGTM (approved):
LGTM

PR #821, opened 2022-04-02
Remove NUsight Dependabot

ysims:
Dependabot currently spams PRs to update NUsight dependencies. This turns off Dependabot for NUsight dependencies, but keeps GitHub Actions. 

Reasons for turning it off
- Clutters the PRs list since there are a lot of dependencies
- We don't do anything with the PRs


CMurtagh-LGTM (approved):
Yes

PR #827, opened 2022-05-09
Fix for vision on the real robot: update calculation for k before used in Visual Mesh

ysims:
Issue was that the real robot was constantly giving no mesh points on screen error, no matter what angle it was at (pitch and roll work ok on the real robot). 

Making this change fixed it so it'll draw the mesh on the ground.

- [ ] Figure out why this works and if it's the right fix
- [ ] Test in webots


TrentHouliston (approved):
Indeed should be divide here (or multiply by the reciprocal)

PR #828, opened 2022-05-25
Simplify head behaviour

TheCoolKidsTable:
This PR simplifies head behaviour

Adds working functionality (tested in WeBots) for:

- Tracking ball if visible
- Executing a predefined search sequence to look for ball if lost

Changes:

- Refactors head behaviour
- Adds ability to enable/disable goal angle smoothing in head command

Things that can be done to improve it:

- Optimize a search sequence and test on real robot
- Currently, when a ball is detected it is causing some noisy spikes in ball position, this is pretty much taken care of by the smoother in HeadController, however its possible these spikes are caused by something that could be potentially avoided completely
- Clamping of requested angles in HeadController could be updated such that the max/min head pitch is a function of the current head yaw position


module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
```diff
@@ -62,66 +54,30 @@ namespace module::behaviour::skills {
     using utility::support::Expression;
 
     inline Eigen::Vector2d screenAngularFromObjectDirection(const Eigen::Vector3d& v) {
+        // Returns {yaw, pitch}
```
ysims, line 57: This would be better as doc directives, info here https://nubook.nubots.net/guides/general/code-conventions#documentation
same for any other functions

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
```diff
@@ -132,546 +88,58 @@ namespace module::behaviour::skills {
...
+
+        on<Trigger<Sensors>, Single, Sync<HeadBehaviourSoccer>>().then(
+            "Head Behaviour Main Loop",
+            [this](const Sensors& sensors) {
+                // Get the time since the ball was last seen
+                float timeSinceBallLastMeasured =
+                    std::chrono::duration_cast<std::chrono::duration<float>>(NUClear::clock::now() - ballLastMeasured)
+                        .count();
+                // Only look for ball if not getting up
+                if (!isGettingUp) {
```
ysims, line 107: you tested this get up check?

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
```diff
@@ -132,546 +88,58 @@ namespace module::behaviour::skills {
...
+                        Eigen::Vector2d angles               = screenAngularFromObjectDirection(rBCc);
+                        std::unique_ptr<HeadCommand> command = std::make_unique<HeadCommand>();
+                        command->yaw                         = angles[0];
+                        command->pitch                       = angles[1];
+                        command->robot_space                 = true;
+                        command->smooth                      = true;
+                        emit(std::move(command));
+                    }
+                    else {
+                        // We haven't seen the ball in a while, lets look around
```
ysims, line 119: Maybe have something more specific, like
'Balll hasn't been seen in a while. Look around using search positions.'

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
```diff
@@ -132,546 +88,58 @@ namespace module::behaviour::skills {
...
+                        float timeSinceLastSearchMoved = std::chrono::duration_cast<std::chrono::duration<float>>(
+                                                             NUClear::clock::now() - searchLastMoved)
+                                                             .count();
+
+                        if (timeSinceLastSearchMoved > fixation_time_ms / 1000) {
```
ysims, line 124: Maybe have a comment here explaining what it's doing. Like
'Robot will move through the search positions, and linger for `fixation_time_ms`. Once  `fixation_time_ms` time has passed, send a new head command for the next position in the list of search_positions.'

module/motion/HeadController/src/HeadController.cpp
```diff
@@ -76,76 +76,64 @@ namespace module::motion {
...
                 float yaw   = 0;
                 for (auto& angle : goalAnglesList) {
                     if (angle.first == ServoID::HEAD_PITCH) {
-                        angle.second = std::fmin(std::fmax(angle.second, min_pitch), max_pitch);
+                        angle.second = utility::math::clamp(float(min_pitch), angle.second, float(max_pitch));
                         pitch        = angle.second;
                     }
                     else if (angle.first == ServoID::HEAD_YAW) {
-                        angle.second = std::fmin(std::fmax(angle.second, min_yaw), max_yaw);
+                        angle.second = utility::math::clamp(float(min_yaw), angle.second, float(max_yaw));
```
ysims, line 129:
```suggestion
                        angle.second = utility::math::clamp(min_yaw, angle.second, max_yaw);
```

module/motion/HeadController/src/HeadController.hpp
```diff
@@ -49,6 +49,7 @@ namespace module::motion {
         Eigen::Vector2f currentAngles;
```
ysims, line 49: I can't comment on the line, but just above here there's a load of `static constexpr const char*`'s that don't seem to be used anymore?

ysims (changes requested):
Test on robot: should do this sometime, we'll be tuning vision on the real robot tomorrow so that might be an opportunity to test this.
Noisy spikes in ball position: is a result of using vision balls. Once we have ball localisation working, should switch to that and it shouldn't be a problem anymore.
That pitch/yaw clamping function should be added as an issue or project somewhere, it'd be good for a newer person.
This removes a lot of code. Are there aspects of the old code that we may want to record somewhere, as a summary for when we add in more functionality to this module?

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
TheCoolKidsTable: Yep, it works

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
```diff
@@ -62,66 +54,30 @@ namespace module::behaviour::skills {
...
-                    }
-                }
+                fixation_time_ms  = config["fixation_time_ms"].as<float>();
+                max_yaw           = config["max_yaw"].as<float>();
+                min_yaw           = config["min_yaw"].as<float>();
+                max_pitch         = config["max_pitch"].as<float>();
+                min_pitch         = config["min_pitch"].as<float>();
+
+                // Create vector of search positions
+                search_positions.push_back(Eigen::Vector2d(min_yaw, max_pitch));
```
LachlanCourt, line 79: I wonder if these positions would be better off coming from the config file too?

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.hpp
```diff
@@ -41,113 +41,33 @@ namespace module::behaviour::skills {
...
-
-        bool orientationHasChanged(const message::input::Sensors& sensors);
-
-
         // CONFIG - loaded elsewhere
-        float max_yaw   = 0.0f;
-        float min_yaw   = 0.0f;
-        float max_pitch = 0.0f;
-        float min_pitch = 0.0f;
+        float max_yaw           = 0.0f;
```
LachlanCourt: If these are loaded from the config what is the purpose of giving them values here? Is this in case the config does not exist?

module/motion/HeadController/src/HeadController.hpp
```diff
@@ -33,7 +33,7 @@ namespace module::motion {
     class HeadController : public NUClear::Reactor {
     private:
         const size_t id;
-        double min_yaw, max_yaw, min_pitch, max_pitch, head_motor_gain, head_motor_torque, p_gain;
+        double min_yaw, max_yaw, min_pitch, max_pitch, head_motor_gain, head_motor_torque, smoothing_factor;
```
LachlanCourt: These values are floats everywhere else, what's the reason for them being doubles here?

TheCoolKidsTable: No reason, will update.

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
ysims: I was thinking this, but in this case you'd then have two places to remember to change one value (min/max values and config search positions) so I'm not sure it's worth it. This was is also more readable so you know where it's going to by just reading this.

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.hpp
ysims: Always initialise your variables! Clang tidy will probably complain.

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
TheCoolKidsTable: Yeah, we could load the values directly into the search positions from config with the current search path

module/motion/KinematicsConfiguration/data/config/KinematicsConfiguration.yaml
```diff
@@ -26,7 +26,7 @@ head:
   # head movement limits
   limits: # pitch limit is a function of yaw ... need to calculate this properly
     yaw: [-pi/4, pi/4] # [min_yaw, max_yaw]
-    pitch: [0, pi/2] # [min_pitch, max_pitch]
+    pitch: [-pi/2, pi/8] # [min_pitch, max_pitch]
```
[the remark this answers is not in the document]
TheCoolKidsTable: Nah, I was just playing around with values. Still uses positive pitch is up convention for now. We should find some good limits on the real robot and update

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.hpp
LachlanCourt: Yes mum 😅

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
LachlanCourt: I am just wondering if we'll want a different head behaviour for goalie in which case we could just give it a different config file, but this is probably fine for now

LachlanCourt: I just discovered that yaml does self referencing, how much do you guys hate this or is it worth a shot?
```
min_yaw:  &min_yaw 0.7
max_yaw: &max_yaw 0.7
min_pitch: &min_pitch -0.1
max_pitch: &max_pitch 0.1
search_positions: [[*min_yaw , *max_pitch ], [*min_yaw , *min_pitch], [0, *min_pitch],  ...... ]
```

TheCoolKidsTable: Is it possible to then initialize a std::vector<Eigen::Vector2d(x,y)> directly from the yaml? or would you have to load each position into the vector individually?

LachlanCourt: Not super famiar with the std::vector library sorry, but I think it should be possible? Something like this though I would imagine?
```
search_positions = config["search_positions"].as<std::vector<Eigen::Vector2d>>()
```

TrentHouliston: The yaml library should be smart enough to do that

TheCoolKidsTable: Awesome, will update.

ysims:
Other things
- Remove the nugus1 specific config file. Shouldn't need per robot config, but if we do it'll probably be different to whatever the old stuff had
- Update the readme in the HeadBehaviourSoccer module. Check the HeadController readme too.

module/behaviour/skills/HeadBehaviourSoccer/README.md
```diff
@@ -7,15 +7,14 @@ Controls the state of the head
 
 ## Usage
 
-
+Include this module to allow the robot to focus on a ball or search for a ball if it is not visble
```
ysims, line 10:
```suggestion
Include this module to allow the robot to focus on a ball or search for a ball if it is not visible
```

TheCoolKidsTable: Spelling is hard :'(

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
```diff
@@ -49,8 +45,6 @@ namespace module::behaviour::skills {
 
     using utility::input::ServoID;
```
ysims, line 46: Is this used? And the accompanying include.

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.cpp
```diff
@@ -24,23 +24,19 @@
```
ysims, line 24: Can't commend on it, but above (line 22), probably don't need `<string>`?

module/behaviour/skills/HeadBehaviourSoccer/src/HeadBehaviourSoccer.hpp
```diff
@@ -23,18 +23,8 @@
 #include <nuclear>
 #include <set>
```
ysims, line 24: Do we need this? Or the ones below, `message/motion/HeadCommand` and `message/vision/Ball`

module/motion/HeadController/README.md
```diff
@@ -7,15 +7,19 @@ Controls the motion of the head
...
+The final goal angle is clamped based on the maximum/minium pitch and yaw of the head servos before being sent to servo controller
 
+## Consumes
 
+* `message::motion::HeadCommand` command containing desired angles, smoothing flag and reference frame information
 
 ## Emits
 
-
+*`message::behaviour::ServoCommands` instructs servo controller to move head to desired angle
```
ysims, line 23:
```suggestion
* `message::behaviour::ServoCommands` instructs servo controller to move head to desired angle
```

module/behaviour/skills/HeadBehaviourSoccer/README.md
```diff
@@ -7,15 +7,14 @@ Controls the state of the head
...
-
+* `message::vision::Balls` uses vision ball to calculate angles to rotate head
 
 ## Emits
 
-
+* `message::motion::HeadCommand` instructs the head to move to desired angle
 
 ## Dependencies
-
```
ysims, line 21: Eigen?

module/motion/HeadController/src/HeadController.cpp
```diff
@@ -60,7 +60,7 @@ namespace module::motion {
         , max_pitch(0.0)
```
ysims, line 60: Not this line, but lines above:
Check if some of these are needed or not. Might not need `utility/math/comparison`.

module/motion/HeadController/README.md
```diff
@@ -7,15 +7,19 @@ Controls the motion of the head
...
 
+- If command requests smoothing (`command->smooth = true`), then the goal angle emitted to is smoothed using exponential filter, tunable using `smoothing_factor`
+- If command request is in world space (`command->goalRobotSpace = false`) the goal angle is converted from world space to robot space
 
-## Consumes
+The final goal angle is clamped based on the maximum/minium pitch and yaw of the head servos before being sent to servo controller
 
+## Consumes
 
+* `message::motion::HeadCommand` command containing desired angles, smoothing flag and reference frame information
```
ysims, line 20: Also consumes KinematicsModel

module/motion/HeadController/README.md
```diff
@@ -7,15 +7,19 @@ Controls the motion of the head
...
 
 ## Emits
 
-
+*`message::behaviour::ServoCommands` instructs servo controller to move head to desired angle
```
ysims, line 24: it emits a headcommand too (at the beginning), probably explain this.
emits a nusight graph, probably mention this
it also emits a registeraction, mention this (it registers the module to be allowed to move the servos, and it specifically only registers for the head and gives itself a high priority of 30 (more is higher priority). if it loses priority of the head, it disables the main reaction loop and if it gets back priority it enables the reaction again).

module/motion/HeadController/README.md
```diff
@@ -7,15 +7,19 @@ Controls the motion of the head
...
 
 ## Dependencies
-
```
ysims, line 21: Eigen, IK, some maths stuff too

module/motion/HeadController/README.md
```diff
@@ -7,15 +7,28 @@ Controls the motion of the head
...
+* `message::motion::HeadCommand` command containing desired angles, smoothing flag and reference frame information
+* `message::input::Sensors` contains information to map desired angles into robot frame from world frame
+* `message::motion::KinematicsModel` contains maximum and minimum values for head pitch/yaw
 
 ## Emits
 
-
+* `message::motion::HeadCommand` instructs servo controller to move head to initial position when configuring
+* `message::behaviour::ServoCommands` instructs servo controller to move head to desired goal angle
+* `utility::behaviour::RegisterAction` it registers the module to be allowed to move the servos, and it specifically only registers for the head and gives itself a high priority of 30 (more is higher priority). if it loses priority of the head, it disables the main reaction loop and if it gets back priority it enables the reaction again).
```
ysims, line 27:
```suggestion
* `utility::behaviour::RegisterAction` registers the module so it is allowed to move the servos. It registers for access to the head with a priority of 30 (a higher number is a higher priority). If it loses priority of the head, it disables the main reaction loop and if it gets back priority it enables the reaction again.
```
I'm still not sure if I like my wording of this but this is probably a bit nicer.

LachlanCourt (dismissed):

ysims:
@TheCoolKidsTable can probably merge this now?

ysims (approved):
LGTM

PR #830, opened 2022-05-29
Document Ball Detector

ysims:
Main goal of this type of PR is to fill in readmes and ensure code is commented and makes sense.
This PR does this for the BallDetector module

I've gone through all of this now and have some things I've commented on below. Input appreciated.


module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,61 +118,79 @@ namespace module::vision {
...
+                    // From the link, the formula with arcsin is used since a ball is a spherical object
+                    // The variables are:
+                    //      delta: arccos(radius)
+                    //      d_act: field.ball_radius
+                    //      D: distance
+                    // Rearranging the equation gives
+                    //      distance = field.ball_radius / (2 * sin(arccos(radius)/2))
+                    // Using sin(arccos(x)) = sqrt(1 - x^2)
+                    //      distance = field.ball_radius / ( 2 * sqrt( 1 - (radius/2)^2 ) )
+                    // !!
```
ysims, line 177: My calculation didn't give me what is in the code?

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,126 +118,155 @@ namespace module::vision {
...
                     // 1) Do this in world space, not camera space!
                     // https://en.wikipedia.org/wiki/Line%E2%80%93plane_intersection#Algebraic_form
                     // Plane normal = (0, 0, 1)
                     // Point in plane = (0, 0, field.ball_radius)
                     // Line direction = axis
-                    // Point on line = camera = Hcw.topRightCorner<3, 1>()
+                    // Point on line = camera = Hwc.translation.z() = rCWw.z()
                     Eigen::Affine3f Hcw(horizon.Hcw.cast<float>());
+                    // Since the plane normal zeros out x and y, only consider z
+                    // why is this not (ball_radius - rCWw.z()) / axis.z()?
```
ysims, line 263: The calculation for `d` below doesn't seem to match the equation in the link?

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -276,20 +307,25 @@ namespace module::vision {
...
                 }
 
+                // Emit the Balls message, even if no balls were found
                 emit(std::move(balls));
             });
     }
 }  // namespace module::vision
 
 
+// Can we delete the below code?
```
ysims, line 328: What's the go with this code, can it be removed? Should I make an issue about adding a check in? Or is the degree of fit check above a similar thing?

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,126 +118,155 @@ namespace module::vision {
...
                     Ball b;
 
                     // Average the cluster to get the cones axis
                     Eigen::Vector3f axis = Eigen::Vector3f::Zero();
+                    // Add up all the unit vectors of each point (camera to point in world space) in the cluster
                     for (const auto& idx : cluster) {
                         axis += rays.col(idx);
                     }
-                    axis /= cluster.size();
+                    axis /= cluster.size();  // can this line be skipped if it's being normalised in the next line?
```
ysims, line 149: ?

module/vision/BallDetector/src/BallDetector.cpp
ysims: The existing calculation seems to be okay when testing in Webots, but I'm not sure why. What I think it should be is almost twice the size.

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -73,24 +73,23 @@ namespace module::vision {
                 std::vector<int> indices(horizon.mesh->indices.size());
                 std::iota(indices.begin(), indices.end(), 0);
 
-                // Partition the indices such that we only have the ball points that dont have ball surrounding them
+                // Partition the indices such that the ball points that dont have ball surrounding them are removed,
+                // and then resize the vector to remove those points
```
TrentHouliston, line 77: This comment has been made wrong.
We don't get the bulk of the ball points (points where the ball is surrounded by ball) we get the edges (ball points that have points around them that are not ball)

We keep the edge points and remove everything else

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,126 +118,155 @@ namespace module::vision {
...
-                // and react based on that last balls message.
+                // Balls is still emitted even if there are no balls, to indicate to other modules that no balls are
+                // currently visible
                 if (clusters.empty()) {
                     log<NUClear::DEBUG>("Found no balls.");
                     emit(std::move(balls));
                     return;
                 }
 
+                // Reserve the memory prematurely for efficiency
```
TrentHouliston, line 132:
```suggestion
                // Reserve the memory in advance for efficiency
```

module/vision/BallDetector/src/BallDetector.cpp
TrentHouliston: Yeah it can be skipped if it's being normalised

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,126 +118,155 @@ namespace module::vision {
...
+                balls->timestamp = horizon.timestamp;  // time when the image was taken
+                balls->Hcw       = horizon.Hcw;        // world to camera transform at the time the image was taken
 
+                // Check each cluster to see if it's a valid ball
                 for (auto& cluster : clusters) {
                     Ball b;
 
                     // Average the cluster to get the cones axis
                     Eigen::Vector3f axis = Eigen::Vector3f::Zero();
+                    // Add up all the unit vectors of each point (camera to point in world space) in the cluster
```
TrentHouliston, line 145: As we kept the edge points we are adding up all the edge points to hopefully find an average in the middle.

This part of the code was intended to be fixed up at some point in the future to instead try to find a central axis vector of a cone. Currently this does weird things when a ball is partially obscured when it could be done better

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,126 +118,155 @@ namespace module::vision {
...
-                    // Should we use the average distance instead?
                     float radius = 1.0f;
                     for (const auto& idx : cluster) {
                         const Eigen::Vector3f& ray(rays.col(idx));
-                        if (axis.dot(ray) < radius) {
-                            radius = axis.dot(ray);
-                        }
+                        // arccos of the dot product gives the angle between the rays
+                        // the smaller arccos is, the bigger the angle - if between 0 and pi
+                        radius = axis.dot(ray) < radius ? axis.dot(ray) : radius;
```
TrentHouliston, line 158: Ideally, this part of the code would use the cone radius once a cone was fit to the edges.

Also, this comment describes what is happening but not why. 
A good comment will explain the things you can't get by reading the code.
Making a comment that just describes what code is doing is bad because if the code changes in the future and not the comment it'll get very confusing and reading the code can give you the information you need.

In this case, we are trying to find the largest ball radius that is possible based on the edge points we have seen from the central axis.
The idea is that the cone that contains all the ball points from the centre of the ball is most likely to be the true radius of the ball since you don't have any points that are outside of that cone.
The way the visual mesh works is that you'll have points on the ball but not outside the ball, so the largest distance between the centre of the ball and a point that is ball is hopefully as close to the edge as possible.

If you did want to improve this part as well, you can look at the neighbours in the visual mesh for each of these points (which are not ball) and you know those are beyond the edge, so between the points that are ball and not ball you have a bound of where the edge of the ball lives

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,126 +118,155 @@ namespace module::vision {
...
                     }
 
-                    // Ball cam space info
-                    b.cone.axis   = horizon.Hcw.topLeftCorner<3, 3>().cast<float>() * axis;
-                    b.cone.radius = radius;
+                    // Set cone information for the ball
+                    // The rays are in world space, multiply by Rcw to get the axis in camera space
+                    b.cone.axis = horizon.Hcw.topLeftCorner<3, 3>().cast<float>()
+                                  * axis;    // Hcw is not an affine transform type (cannot use linear())
+                    b.cone.radius = radius;  // arccos(radius) is the angle between the furthest vectors
```
TrentHouliston, line 165: Just make it an Affine3d if you need Affine3d

`Eigen::Affine3d(horizon.Hcw).cast<float>().rotation() * axis`

module/vision/BallDetector/src/BallDetector.cpp
TrentHouliston: It's the radius vs diameter that you're having trouble with is my guess.
d_act is `2r` not `r`
Therefore in your rearranged math the 2s cancel giving what is in the code below

image

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,126 +118,155 @@ namespace module::vision {
...
+                    // From the link, the formula with arcsin is used since a ball is a spherical object
+                    // The variables are:
+                    //      delta: arccos(radius)
+                    //      d_act: field.ball_radius
+                    //      D: distance
+                    // Rearranging the equation gives
+                    //      distance = field.ball_radius / (2 * sin(arccos(radius)/2))
+                    // Using sin(arccos(x)) = sqrt(1 - x^2)
+                    //      distance = field.ball_radius / ( 2 * sqrt( 1 - (radius/2)^2 ) )
+                    // !!
```
TrentHouliston, line 177:
```suggestion
                    // The variables are:
                    //      delta: arccos(radius)
                    //      d_act: 2 * field.ball_radius
                    //      D: distance
                    // Rearranging the equation gives
                    //      distance = (2 * field.ball_radius) / (2 * sin(arccos(radius)/2))
                    // Using sin(arccos(x)) = sqrt(1 - x^2)
                    //      distance = field.ball_radius / sqrt(1 - (radius/2)^2
                    // !!
```

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,126 +118,155 @@ namespace module::vision {
...
                     b.measurements.back().srBCc      = cartesianToReciprocalSpherical(b.cone.axis * distance);
                     b.measurements.back().covariance = config.ball_angular_cov.asDiagonal();
 
-                    // Angular positions from the camera
+                    // Angular positions from the camera, doesn't consider depth - x is removed and unit vector is used
                     b.screen_angular = cartesianToSpherical(axis).tail<2>();
-                    b.angular_size   = Eigen::Vector2f::Constant(std::acos(radius));
+                    // acos(radius) is the angular diameter - ie the radius from the perspective of the camera
+                    // used for both directions since it's a ball
+                    b.angular_size = Eigen::Vector2f::Constant(std::acos(radius));
```
TrentHouliston, line 193: We really should remove these from the message, the message should contain the minimal information required to reconstruct. This information is redundant to radius and axis

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,126 +118,155 @@ namespace module::vision {
...
+                    // DISCARD IF THE DISAGREEMENT BETWEEN THE ANGULAR AND PROJECTION BASED DISTANCES ARE TOO LARGE
+                    // Intersect cone axis vector with a plane midway through the ball with normal vector (0, 0, 1)
                     // 1) Do this in world space, not camera space!
                     // https://en.wikipedia.org/wiki/Line%E2%80%93plane_intersection#Algebraic_form
                     // Plane normal = (0, 0, 1)
                     // Point in plane = (0, 0, field.ball_radius)
                     // Line direction = axis
-                    // Point on line = camera = Hcw.topRightCorner<3, 1>()
+                    // Point on line = camera = Hwc.translation.z() = rCWw.z()
                     Eigen::Affine3f Hcw(horizon.Hcw.cast<float>());
```
TrentHouliston, line 261: This has been used a few times, maybe just make an Affine3f right up the top of the loop once and reuse it

module/vision/BallDetector/src/BallDetector.cpp
TrentHouliston: This might be wrong, really need to use more of the naming convention I think...

`Hwc.translation.z()` would be the height of the camera above the ground (positive value)
Subtracting the ball radius would give the height from the centre of the ball to the camera

`axis.z()` should hopefully be in world space (should probably rename it `uACw`) which will be negative, so abs it and divide to get distance along the ground.
That said isn't the distance supposed to be from the camera? And is the width based distance from the camera? Either way we need to name these variables better

module/vision/BallDetector/src/BallDetector.cpp
TrentHouliston: This code was intended to be used rather than averaging the edge points to get the centre and radius but according to @Bidski it didn't work so well.

That said averaging has it's problems too so it's possible this could still be used and improved to work properly.

shared/message/vision/Ball.proto
```diff
@@ -39,6 +39,7 @@ message Ball {
         fmat3 covariance = 3;
     }
     /// Cone describing the ball from the camera's perspective
+    /// See Geometry.proto
```
TrentHouliston, line 42: Rather than using a cone object, I think it might be better to just have an axis and radius directly in here. The original idea was to have a common cone object so we could use it for math functions that worked with cones. But that never happened so now we just have badly named variables for cones

TrentHouliston (changes requested):

module/vision/BallDetector/src/BallDetector.cpp
ysims: Yeah lol I wrote this wrong, too many negatives for my brain apparently

ysims:
```suggestion
                // Partition the indices such that the ball points that have ball points surrounding them are removed,
                // and then resize the vector to remove those points
```

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,61 +118,79 @@ namespace module::vision {
...
                     return;
                 }
 
+                // Reserve the memory prematurely for efficiency
                 balls->balls.reserve(clusters.size());
 
-                balls->id        = horizon.id;
-                balls->timestamp = horizon.timestamp;
-                balls->Hcw       = horizon.Hcw;
+                balls->id        = horizon.id;         // subsumption id
```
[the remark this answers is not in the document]
ysims: Haha true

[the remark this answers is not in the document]
ysims:
```suggestion
                balls->id        = horizon.id;         // camera id
```

module/vision/BallDetector/src/BallDetector.cpp
ysims: Even with that, I don't get what is below, I get
$$ D = \frac{R}{(sin(\frac{arccos(r)}{2}))} $$

Doing it out in full:

$$ \delta = 2 \arcsin \left( \frac{d_{act}}{2D} \right) $$

Let $\delta = \arccos(r)$, where $r$ is `radius` in the code
Let $d_{act} = 2R$, where $R$ is `field.ball_radius`
and $D$ is the distance we're trying to find

$$ \arccos(r) = 2 \arcsin ( \frac{2R}{2D} ) $$

$$ \arccos(r) = 2 \arcsin ( \frac{R}{D} ) $$

$$ \frac{\arccos(r)}{2} = \arcsin ( \frac{R}{D} ) $$

$$ \sin \left( \frac{\arccos(r)}{2} \right) = \frac{R}{D} $$

$$ D = \frac{R}{\sin \left( \frac{\arccos(r)}{2} \right)} $$

module/vision/BallDetector/src/BallDetector.cpp
ysims:
```suggestion
                    // Add up all the unit vectors of each point (camera to point in world space) in the cluster to find an average vector, which represents the central cone axis
```

ysims: Note to self: add an issue for this

module/vision/BallDetector/src/BallDetector.cpp
ysims: Hmmm I guess this is easier to read

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -117,126 +118,155 @@ namespace module::vision {
...
+                        // the smaller arccos is, the bigger the angle - if between 0 and pi
+                        radius = axis.dot(ray) < radius ? axis.dot(ray) : radius;
                     }
 
-                    // Ball cam space info
-                    b.cone.axis   = horizon.Hcw.topLeftCorner<3, 3>().cast<float>() * axis;
-                    b.cone.radius = radius;
+                    // Set cone information for the ball
+                    // The rays are in world space, multiply by Rcw to get the axis in camera space
+                    b.cone.axis = horizon.Hcw.topLeftCorner<3, 3>().cast<float>()
```
ysims, line 163:
```suggestion
                    b.cone.axis = Eigen::Affine3d(horizon.Hcw).cast<float>().rotation()
```

module/vision/BallDetector/src/BallDetector.cpp
TrentHouliston: Ah also your second radius is off by 2 as well, The `radius` that you have is the dot product from the axis of the cone to a point on its edge, what the angular diameter of this equation wants is the entire diameter of the shape, this is twice the angle from the edge to the centre of the cone. So 2*acos(r) which cancels with your /2

module/vision/BallDetector/src/BallDetector.cpp
ysims: Once we have a benchmarking test set up in Webots, I might have a go at improving this given what you've said. Then I can see if any changes really improves it. 

I get what this means more or less now, but how would you describe this radius? Since it's not the true radius, since it comes from the unit vectors. Is it the approximate radius of the ball on the image itself? Or is there a better way to describe it?

shared/message/vision/Ball.proto
ysims: While we're here, why is there a repeated Measurement? It seems to only add in one Measurement in BallDetector, seems it shouldn't be repeated and the Measurement components could be pulled out into here. Type doesn't seem to be used either.

module/vision/BallDetector/src/BallDetector.cpp
ysims: I'll have a look at this too then. There's NUbots and NUWebots branches for vision webots tests for ball accuracy so once that's done can properly improve and test if it's an improvement.

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -106,137 +108,167 @@ namespace module::vision {
...
+                    // Given a flat X-Y plane that intersects through the middle of the ball, find the point where
+                    // the central axis vector (uBCw) intersects with this place. Get the distance from this point.
+                    // https://en.wikipedia.org/wiki/Line%E2%80%93plane_intersection#Algebraic_form
+                    // Plane normal = (0, 0, 1)
+                    // Point on plane = (0, 0, field.ball_radius) // SHOULD THIS BE (0,0, BALL RADIUS + rWCw)
+                    // Line direction = uBCw
+                    // Point on line = camera = Hwc.translation.z() = rCWw.z() // SHOULD THIS BE (0,0,0)
+                    // Since the plane normal zeros out x and y, only consider z
+                    const float d = (Hcw.inverse().translation().z() - field.ball_radius) / std::abs(uBCw.z());
+                    const Eigen::Vector3f srBCc     = uBCw * d;  // used later, so save it in a variable
```
ysims, line 208: @TrentHouliston 
I've updated the variable names to hopefully match conventions.
It seems the angular based distance is determined using world space but centered on the camera (like using uXCw type vectors)
But projection based seems to be going straight from world. My thoughts on what projection based should be is that 
- the line is ofc uBCw (central axis vector), 
- the point on the line is (0,0,0) since the line is on the camera and we're measuring from the camera (in world space)
- the point on the plane is like (ball_radius + rWCw) because from the camera, go down to the world and then up to the ball radius

Given this, the equation becomes $d = \frac{p_0.z()}{l.z()} = \frac{rBCw.z()}{\text{uBCw.z()}}$
What is in the code is pretty much the same, since it's $\frac{rCBw.z()}{abs(uBCw.z()}$
the way in the code makes the top thing positive and abs the bottom, but since those negatives will cancel anyway it might make more sense to go with the straight derivation.

module/vision/BallDetector/README.md
```diff
@@ -1,13 +1,40 @@
...
-============
+# BallDetector
 
 ## Description
 
+The BallDetector module performs post-processing on the output of the VisualMesh to determine where balls are in the image, if any. If there are no balls, it will emit an empty `Balls` message.
+
+The BallDetector first tries to create balls, and then removes candidate balls if they fail to pass a set of checks.
+
+Balls are created by clustering neighbouring Visual Mesh points that have a high enough confidence that they are a ball point.
```
Bidski, line 9: Only ball edge points are clustered

module/vision/BallDetector/README.md
```diff
@@ -1,13 +1,40 @@
...
+
+Each cluster is checked to see if
+
+1. It is below the green horizon
+2. It has a high enough degree of fit
```
Bidski, line 14: Degree of fit to a ... circle?

module/vision/BallDetector/README.md
```diff
@@ -1,13 +1,40 @@
...
+Zero or many balls could be detected and emitted in a Balls message.
 
 ## Usage
 
+Add this module to get information about balls, given the required messages are present.
+
+## Consumes
+
+- `message::vision::GreenHorizon` which contains points that represent the green horizon, and the Visual Mesh output. The Visual Mesh output is used to create the ball clusters and the green horizon is used to remove balls above the horizon.
+- `message::support::FieldDescription` which contains parameters that define the structure and size of the field. This is used in the checks mentioned above, to determine if the ball candidate is likely to be a ball.
```
Bidski, line 32: Also contains information about the real world size of the ball

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -106,137 +108,168 @@ namespace module::vision {
...
+                    // the central axis vector (uBCw) intersects with this place. Get the distance from this point.
+                    // https://en.wikipedia.org/wiki/Line%E2%80%93plane_intersection#Algebraic_form
+                    //      Plane normal: (0, 0, 1)
+                    //      Point on plane: (0, 0, rWCw.z() + field.ball_radius)
+                    //      Line direction: uBCw
+                    //      Point on line: (0, 0, 0)
+                    // Since the plane normal zeros out x and y, only consider z
+                    // rWCw = -Hcw.inverse().translation().z()
+                    const float d                   = (field.ball_radius - Hcw.inverse().translation().z()) / uBCw.z();
+                    const Eigen::Vector3f srBCc     = uBCw * d;  // used later, so save it in a variable
```
ysims, line 209: I don't get why this is `srBCc`. `uBCw` is just a normal unit vector to the ball, then scaled by `d`, the distance, to get `rBCw`?
And below, the norm would just be `d` right? Why make two variables for the same thing?

module/vision/BallDetector/src/BallDetector.cpp
ysims: Gonna resolve this here since the code has changed a bit and I've made a new comment thread on that.

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -106,137 +108,167 @@ namespace module::vision {
...
-                    axis /= cluster.size();
-                    axis.normalize();
+                    uBCw.normalize();  // get cone axis as a unit vector
 
-                    // Find the ray with the greatest distance from the axis
-                    // Should we use the average distance instead?
+                    // FIND UNIT RADIUS OF BALL
+                    // Find the ray (uPCw) with the greatest distance from the central axis (uBCw) to then determine the
+                    // largest radius possible from the edge points available. The radius is the upper bound on the ball
+                    // radius if it were a metre away. arccos(radius) is the angle between the central ball
```
Bidski, line 163: I dont think the ball being a meter away has anything to do with it. This is just an upper bound on the ball radius

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -106,137 +108,167 @@ namespace module::vision {
...
                     float radius = 1.0f;
                     for (const auto& idx : cluster) {
-                        const Eigen::Vector3f& ray(rays.col(idx));
-                        if (axis.dot(ray) < radius) {
-                            radius = axis.dot(ray);
-                        }
+                        // Unit vector from the camera to the ball edge, in world space
+                        const Eigen::Vector3f& uECw(uPCw.col(idx));
+                        // Choose this vector if it's further than previously seen vectors, to get the largest angle
+                        radius = uBCw.dot(uECw) < radius ? uBCw.dot(uECw) : radius;
```
Bidski, line 170: You might want to say that we are checking `<` here because `radius` is actually `cos(theta)` (the `cos` of the angle, `theta`, between the axis and the ball edge) and larger `theta` results in smaller `cos(theta)`. This logic can become non-intuitive when you are talking about finding the "largest" angle

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -106,137 +108,167 @@ namespace module::vision {
...
                     }
 
-                    // Ball cam space info
-                    b.cone.axis   = horizon.Hcw.topLeftCorner<3, 3>().cast<float>() * axis;
-                    b.cone.radius = radius;
+                    // The vectors are in world space, multiply by Rcw to get the central axis in camera space
+                    b.uBCc   = Hcw.rotation() * uBCw;
+                    b.radius = radius;  // arccos(radius) is the angle between the central axis and edge of the ball
```
Bidski, line 175: Probably better to refer to `radius` being equal to `cos(theta)`, where `theta` is the angle between the central axis and edge of the ball

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -106,137 +108,167 @@ namespace module::vision {
...
+                    //      distance = field.ball_radius / sqrt(1 - radius^2)
                     float distance = field.ball_radius / std::sqrt(1.0f - radius * radius);
+                    // Convert uBCc into a position vector (rBCc) and then convert it into Spherical Reciprocal
+                    // Coordinates (1/distance, phi, theta)
+                    b.srBCc              = cartesianToReciprocalSpherical(b.uBCc * distance);
+                    b.screen_angular_BCw = cartesianToSpherical(uBCw).tail<2>();  // no depth (x-coordinate)
+
+                    // Projection-based distance
+                    // Given a flat X-Y plane that intersects through the middle of the ball, find the point where
+                    // the central axis vector (uBCw) intersects with this place. Get the distance from this point.
```
Bidski, line 200:
```suggestion
                    // the central axis vector (uBCw) intersects with this plane. Get the distance from the camera to this point.
```

module/vision/BallDetector/src/BallDetector.cpp
Bidski: Pretty sure `srBCc` should just be `rBCc` here since it is in cartesian coordinates not spherical coordinates.

Although your method results in a very similar equation I think the logic on getting there is a little more convoluted. The way that is currently implemented, the plane is the field (with a slightly higher z-value so it passes through the ball) and the camera is above is the field. I feel like this is easier to visualise.

module/vision/BallDetector/src/BallDetector.hpp
```diff
@@ -27,6 +27,7 @@ namespace module::vision {
 
     class BallDetector : public NUClear::Reactor {
     private:
+        // See BallDetector.yaml for info on configuration variables and their values
```
Bidski, line 30: If we are using doxygen-like documentation system you should duplicate the definitions of these config quantities here so that they show up in the docs (unless doxygen can parse yaml files???)

shared/message/vision/Ball.proto
```diff
@@ -23,29 +23,22 @@ package message.vision;
...
+    fvec3 srBCc = 2;
+    /// Unit radius of the ball from the camera in world space
+    /// Cosine of the angle from the centre of the ball to the edge of the ball
+    float radius = 2;
     /// Angular positions from the camera
-    fvec2 screen_angular = 3;
-
-    fvec2 angular_size = 4;
-    /// Colour of the ball for NUsight. Changes depending on the perceived ball location
+    fvec2 screen_angular_BCw = 4;
```
Bidski, line 37: Really not a fan of this name ... but also dont have any suggestions for improving it

Bidski (changes requested):

module/vision/BallDetector/src/BallDetector.cpp
Bidski: It should be `rBCc`, not `srBCc` ... not sure why we made the extra variable though, maybe a remnant of some old code that got deleted.

shared/message/vision/Ball.proto
Bidski: It is repeated because localisation should, theoretically, handle multiple measurements for each object. The idea is that you measure the distance to the ball using multiple, distinct, methods to help improve the localisation accuracy. We never ended up implementing any other measurement methods for the ball, but left in the ability to handle multiple measurements in case someone ever came up with one.

This multiple measurement thing should be done for all vision objects (balls, goals, field lines, center circle, obstacles_ wherever possible. @TrentHouliston might be able to explain this better

ysims: If it's something we should, do, then I'll see about updating this to use both projection and angular based methods? And revert removing it

module/vision/BallDetector/src/BallDetector.cpp
ysims: I think this comes down to personal preference. The original stuff seemed more convoluted to me.
Maybe I'm not understanding the original one well enough, but it seems like it doesn't quite following the spaces correctly and then has to do an abs() to fix up the signs.
The way I thought about it was going from the formula and making sure all the spaces were correct throughout. 
In the end it is the same thing though. I'm just not sure how to nicely comment the original stuff to show how you get it mathematically.

module/vision/BallDetector/src/BallDetector.cpp
ysims: How is it an upper bound on the ball radius? If it's a unit vector, then projecting it out further than a unit (one metre) would give a larger radius? It has to be an upper bound from a particular perspective, since we know in reality the radius is `field.ball_radius`?

shared/message/vision/Ball.proto
Bidski: Im pretty sure that projection isn't distinct from angular (I think the two measurements are derived from each other or something?). @TrentHouliston knows more about this than I do, but there was a reason we didnt add that measurement

module/vision/BallDetector/src/BallDetector.cpp
Bidski: Everything is in world space, so thats not an issue. Its just a question of where your origin is, is it at camera height or ball height? And it really doesn't matter which one you choose, so it becomes a question of which version is easier to understand. 

You are projecting a ray from the camera to the "ground" plane so, to me, having the plane at "ground" level makes more sense. Then, the camera being above the plane follows from that. The alternative is your suggestion, where the ground is below the camera, but as I said either option is equally correct.

shared/message/vision/Ball.proto
TrentHouliston: Projection-based and width based are distinct, I don't know why they were removed. My guess is someone was annoyed that the projection-based was bad because odometry was bad.

module/vision/BallDetector/src/BallDetector.cpp
Bidski: Because its not actually a radius that we are measuring, its (`cos` of) the angle between the cone axis and the ball edge. Its an upper bound because we are looking for the largest such angle. We could just as feasibly take the minimum angle (the lower bound) or the mean value. Literally the next thing we do is use the angular distance formula to figure out what distance the ball has to be away from the camera for the "radius" we calculate to make sense because, as you said, the actual radius is fixed to the radius of the ball.

module/vision/BallDetector/README.md
TrentHouliston: cone?

shared/message/vision/Ball.proto
TrentHouliston: My suggestion was deleting it :P

module/vision/BallDetector/README.md
Bidski: I mean we are fitting a cone, but this check in particular is really measuring how circular it is

module/vision/BallDetector/src/BallDetector.cpp
Bidski: I think at one point we called this the `gradient` of the cone (I think it was in the message we called it this). Which is also technically correct, but the quantity we are actually measuring isn't actually the gradient of the cone either.

ysims: The naming of it seems not great to me. Angular radius seems like a better name.

Bidski: Its also a longer name and more annoying to type :stuck_out_tongue: just comment it lol

shared/message/vision/Ball.proto
```diff
@@ -23,30 +23,23 @@ package message.vision;
...
-        fvec3 srBCc      = 2;  // Spherical Reciprocal Coordinates (1/distance, phi, theta)
-        fmat3 covariance = 3;
-    }
-    /// Cone describing the ball from the camera's perspective
-    Cone                 cone         = 1;
-    repeated Measurement measurements = 2;
+    /// Central axis unit vector of the ball from the camera in camera space
+    fvec3 uBCc = 1;
+    // Spherical Reciprocal Coordinates (1/distance, phi, theta)
+    fvec3 srBCc = 2;
```
ysims, line 32: Is there a particular reason we want to store it in spherical reciprocal coordinates (and from the camera) rather than something like `rBCw` (or `rBCc`)?

Bidski: The distance measurement scales better in reciprocal space, cant remember why we want spherical. @TrentHouliston ?

TrentHouliston: The reason is that it's the domain of the covariance.
So when our measurements have a covariance value that corresponds to the spherical coordinates. Because our error is in the angular position, and reciprocal distance gives an ok approximation of making the error curve in the distance look like a normal distribution. (normally the further you go the greater your error is which isn't very normal)

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -259,10 +263,10 @@ namespace module::vision {
 
                     // DISCARD IF THE DISTANCE FROM THE BALL TO THE ROBOT IS TOO CLOSE
                     // Prevents the robot itself being classed as a ball, ie its arms/hands
-                    if (distance < config.minimum_ball_distance) {
+                    if (angular_distance < config.minimum_ball_distance) {
```
Bidski, line 266: Might be better to move the projection/angular distance throwout before this one, otherwise you would need to check that both angular and projection distances are both further away than `config.minimum_ball_distance`.

module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -271,12 +275,15 @@ namespace module::vision {
                     }
 
                     // Discard this ball if projection_distance and distance are too far apart
-                    if ((std::abs(projection_distance - distance) / max_distance) > config.distance_disagreement) {
+                    const float max_distance = std::max(projection_distance, angular_distance);
+
+                    if ((std::abs(projection_distance - angular_distance) / max_distance)
+                        > config.distance_disagreement) {
```
Bidski, line 281: @TrentHouliston do we even need this throwout if we are passing both distance measurements to localisation? Wouldnt it be better to let localisation deal with the discrepency?

Bidski (changes requested):

shared/message/vision/Ball.proto
ysims: I deleted it...

module/vision/BallDetector/src/BallDetector.cpp
ysims: I assumed this was a throw out because it implies an assumption made didn't really work out, eg ball radius, and so it's probably not a ball. When I was tuning the other day with a new mesh, this was a useful throwout to get rid of some bad detections.

TrentHouliston: Localisation could do more with it, but in this case it's when it gets to the "this isn't a ball is it" threshold. Like if it saw a ball on a poster on the wall or something

TrentHouliston (approved):
This is definitely better than it was, which is good!

Bidski (approved):
I concur with @TrentHouliston, definitely much better than it was

Bidski (approved):
Sure, but if you dismiss my review again you wont get another one :stuck_out_tongue:

PR #833, opened 2022-06-02
Add in platform-specific configuration and set the hostname in docker to 'webots' if the role contains the word 'webots'

ysims:
- Updates the Configuration extension to allow platform-specific configuration files
- Update the docker run command to use 'webots' as the hostname if the user is using `./b run` and their binary contains the word `webots`
- Updates the QuinticWalk config - in Webots, the default config was used and there was no way to run with the config in the already-existing QuinticWalk `webots` config directory. With these changes, the Webots platform should run with the config file in the QuinticWalk `webots` config directory. Updates this config to the same values as the default config so the webots robot will use the right values.
- Added two platform config files for the visualmesh module, one for the nugus and one for webots. The default file only contains the log level, since each platform should implement it's own camera config values. Having the real robot config as the default does not work, since the merging method for config files means that the webots file will contain config values for the real Left and Right camera.

This is tested and it seems to work fine.

[NUbook PR here.](https://github.com/NUbots/NUbook/pull/223)

### DIFF
--- a/module/input/SensorFilter/data/config/webots/SensorFilter.yaml
+++ b/module/input/SensorFilter/data/config/webots/SensorFilter.yaml
@@ -1,10 +1,5 @@
-log_level: ERROR
-
-buttons:
-  debounce_threshold: 7
-
 foot_down:
-  method: Z_HEIGHT
+  method: FSR
   known_methods:
     - name: FSR # Literal FSR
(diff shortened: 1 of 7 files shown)


LachlanCourt (approved):
LGTM

PR #835, opened 2022-06-04
Always restore tty settings

Bidski:
This PR prevents your terminal from going insane after running a `./b` command by
- overriding signal more signal handlers
- resetting TTY (terminal) settings when a signal is caught or when the script ends


ysims (approved):
Fixes my issue, ty

PR #836, opened 2022-06-04
Change naming convention of webots roles

TheCoolKidsTable:
This PR simply changes the names of webots roles to have the following consistent naming convention:
- `webots`
- `webots_keyboard`
- `webots_robocup`
- `webots_scripttuner`


ysims:
Looks good, though if we're making these snake cake I feel like we should use that convention for all roles. ie `camera_test` (or even `test_camera` if we want ordering to put test roles next to each other)

ysims (commented):
If it's proper snake case, we should have underscores between each word. module names can stick together though. eg filter_playback, look_at_ball, firmware_installer

LachlanCourt (changes requested):
This seems fine to me, please update NUbook and make associated PR though before merging this in - lots of the role names are mentioned throughout

TrentHouliston (commented):
I think swapping the order this PR has proposed would be good

so `webots_keyboardwalk` over `keyboardwalk_webots`

that way all the webots roles will sit together and typing the name will be easy (w tab rolename)

ysims:
I don't know if there's much benefit to having them grouped together by role name, since you're unlikely to be turning on all of the roles in that grouping, whereas you're likely to be turning on all the webots roles.

TrentHouliston:
But wouldn't you rather all the webots sit together in the configure list? That way if you're not working on webots you can just easily disable all the webots roles.

And tabbing is just in a shell with autocomplete, so if you type something and hit tab it'll complete the webots_ part and then you can just type the role name

ysims (commented):
Get a NUbook PR up as per @LachlanCourt's suggestion and then we can merge this

LachlanCourt (approved):
LGTM when the NUbook PR is merged :)

TrentHouliston:
What would be even better... Folders!

so like `webots/keyboardwalk.role` making a `webots/keyboardwalk` module

Would require a small change to the `nuclear/roles/CMakeLists.txt` and maybe `nuclear/roles/NUClearRole.cmake`

ysims:
> What would be even better... Folders!
> 
> so like `webots/keyboardwalk.role` making a `webots/keyboardwalk` module
> 
> Would require a small change to the `nuclear/roles/CMakeLists.txt` and maybe `nuclear/roles/NUClearRole.cmake`

I like this, I'll make an issue. Although this PR I think is still ok in the meantime, since our role names are over the place.

PR #837, opened 2022-06-04
Switch around nugus2 and nugus3 Camera config files after nuc swap

ysims:
The nucs were swapped around in the nugus' earlier this year. Camera config needs to be swapped around to the right nugus'. They have already had serial numbers checked and calibration performed.


TheCoolKidsTable (approved):
LGTM!

PR #839, opened 2022-06-06
Adjusts offsets for metal leg robot

Crafty15:
Right leg was slightly forward compared to the left when zombied. 
Double and triple checked that the horns were mounted in the correct position and servos were calibrated and came to the conclusion that the legs may have errors from manufacturing.
This adjusts the offsets in the HardwareIO.yaml file to compensate for these errors.


ysims:
This should be in a nugus1 config folder, otherwise all the robots will use this config.

LachlanCourt (approved):
LGTM

PR #840, opened 2022-06-06
Adds a stable walk for the real NUgus robot

Crafty15:
- Reduced frequency by 0.75
- Increased double support ratio by 0.60
- All other changes from main were already on the robot, but the two above were what increased stability.
- Also increased trunk height to straighten out legs.


ysims:
This should be a change specifically for nugus1 (ie make a nugus1 folder)

LachlanCourt (approved):
LGTM

ysims:
If it works for all robots, it can be the default config file

Crafty15:
> If it works for all robots, it can be the default config file

Works for at least 2 of them. Do we want to keep the sub folder?

ysims:
> > If it works for all robots, it can be the default config file
> 
> Works for at least 2 of them. Do we want to keep the sub folder?

No point if it works better than the current default config for the robots. Small tweaks to it can be done in sub folders, but that can come later. Move it back to the default and then lets merge this

ysims (approved):
LGTM

PR #841, opened 2022-06-10
Update getup to use filtered sensors

TheCoolKidsTable:
This PR updates the getup to use filtered sensors and increases the number of last sensors used in the average.

Fixes:

- The getup being executed while walking due to noisy raw sensors
- The getup being executed before the robot hits the ground (by increasing the number of last sensors msgs used)

Potential improvements:

- Use the actual time since fallen to execute getup rather then the average of last X msg's

### DIFF
--- a/module/behaviour/skills/Getup/src/Getup.cpp
+++ b/module/behaviour/skills/Getup/src/Getup.cpp
@@ -26,8 +26,8 @@
 #include "extension/Script.hpp"
 
 #include "message/behaviour/ServoCommand.hpp"
+#include "message/input/Sensors.hpp"
 #include "message/motion/GetupCommand.hpp"
-#include "message/platform/RawSensors.hpp"
 
 #include "utility/behaviour/Action.hpp"
 #include "utility/input/LimbID.hpp"
@@ -37,9 +37,9 @@ namespace module::behaviour::skills {
     using extension::Configuration;
     using extension::ExecuteScriptByName;
 
+    using message::input::Sensors;
     using message::motion::ExecuteGetup;
     using message::motion::KillGetup;
-    using message::platform::RawSensors;
 
     using utility::behaviour::ActionPriorities;
     using utility::behaviour::RegisterAction;
@@ -63,9 +63,9 @@ namespace module::behaviour::skills {
             EXECUTION_PRIORITY = config["EXECUTION_PRIORITY"].as<float>();
         });
 
-        on<Last<1000, Trigger<RawSensors>>, Single>().then(
+        on<Last<250, Trigger<Sensors>>, Single>().then(
             "Getup Fallen Check",
-            [this](const std::list<std::shared_ptr<const RawSensors>>& sensors) {
+            [this](const std::list<std::shared_ptr<const Sensors>>& sensors) {
                 Eigen::Vector3d acc_reading = Eigen::Vector3d::Zero();
 
                 for (const auto& s : sensors) {
--- a/module/behaviour/skills/Getup/src/Getup.cpp
+++ b/module/behaviour/skills/Getup/src/Getup.cpp
@@ -63,7 +63,7 @@ namespace module::behaviour::skills {
             EXECUTION_PRIORITY = config["EXECUTION_PRIORITY"].as<float>();
         });
 
-        on<Last<20, Trigger<RawSensors>>, Single>().then(
+        on<Last<1000, Trigger<RawSensors>>, Single>().then(
             "Getup Fallen Check",
             [this](const std::list<std::shared_ptr<const RawSensors>>& sensors) {
                 Eigen::Vector3d acc_reading = Eigen::Vector3d::Zero();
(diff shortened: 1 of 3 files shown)


ysims:
Have you checked if the old odometry method works? Since pitch and roll are fine, it would be much better to use odometry than raw sensors.

TheCoolKidsTable:
Good point, I for some reason assumed the sensors.accelerometer was filtered...

Will investigate with Htw.

module/behaviour/skills/Getup/src/Getup.cpp
```diff
@@ -61,24 +61,33 @@ namespace module::behaviour::skills {
             // load priorities for the getup
             GETUP_PRIORITY     = config["GETUP_PRIORITY"].as<float>();
             EXECUTION_PRIORITY = config["EXECUTION_PRIORITY"].as<float>();
+            log<NUClear::WARN>("CONFIGURED");
```
ysims, line 64: Remove log

module/behaviour/skills/Getup/src/Getup.cpp
```diff
@@ -61,24 +61,33 @@ namespace module::behaviour::skills {
...
         });
 
-        on<Last<20, Trigger<RawSensors>>, Single>().then(
+        on<Last<250, Trigger<Sensors>>, Single>().then(
```
ysims, line 67: You should only need one since the sensors are already filtered?

ysims (commented):
We only really need the for loop for filtering raw sensors. Odometry should already be filtered, we shouldn't think we've fallen over until we really have, and if that isn't the case we should tune Odometry. Ideally we get up as soon as we detect we've fallen, since we don't have long before being deemed incapable and we want our robots back in the game asap.

TheCoolKidsTable:
I think we want a small delay after it detects it's fallen before it executes getup, just enough so that it doesn't try to start before it's even hit the ground and to let it settle into a stable state.

We could potentially increase the fallen angle to help with this as well.

ysims:
> I think we want a small delay after it detects it's fallen before it executes getup, just enough so that it doesn't try to start before it's even hit the ground and to let it settle into a stable state.
> 
> We could potentially increase the fallen angle to help with this as well.

Yeah fallen angle would help, but also the get up scripts usually have a long initial frame. If we had an angular velocity measurement for the robot we could use that.

module/behaviour/skills/Getup/src/Getup.cpp
BryceTuppurainen:
```cpp
// Average of the last 250 sensors measurments
for (const auto& s : sensors) {
    Hwt = s->Htw.inverse().matrix().cast<double>();
    uZTw += Hwt.rotation().block(0, 2, 3, 1);
    uXTw += Hwt.rotation().block(0, 0, 3, 1);
}
uZTw = uZTw.normalized()
```

Seems that there's an average being taken of the filtered measurements again here?

@TheCoolKidsTable I agree I'm not sure why we're doing so for filtered measurements? Is this to eliminate/dampen the effect of outlier measurements?

TheCoolKidsTable: Yeah the idea is now more if the angle has exceeded FALLEN_ANGLE for the last X msg's, then start the getup.

The main problem I was trying to solve was the getup being executed while it was walking in real life. Will need to investigate in the lab if current odometry solves this already.

LachlanCourt (dismissed):
If this works then seems fine, haven't tested myself

module/behaviour/skills/Getup/src/Getup.cpp
TheCoolKidsTable: I also just noticed an issue in sim where if its walking and has delay it doesn't stop trying to walk while on ground.

I will remove the last 250 msg's component and test

ysims: Yeah 250 messages would cause a delay of almost three seconds before it detects the getup. We get sensors 90 times a second.

ysims:
The recent change looks much better. How does it go in webots?

TheCoolKidsTable:
Works fine in webots.

BryceTuppurainen (approved):
lgtm, we'll want to do plenty of testing on the actual bots to make sure that it is working correctly but after removing this averaging this makes sense :+1:

module/behaviour/skills/Getup/src/Getup.cpp
```diff
@@ -61,24 +61,33 @@ namespace module::behaviour::skills {
...
             "Getup Fallen Check",
```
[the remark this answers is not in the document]
BryceTuppurainen: I think you're looking at an old version of the code @BryceTuppurainen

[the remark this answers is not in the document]
ysims: I think you're looking at an old version of the code @BryceTuppurainen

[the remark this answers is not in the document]
BryceTuppurainen: Yes, this has become outdated now, I created this 5-days ago initially before Tom fixed it. I'll resolve this conversation

ysims:
Have you tested on the robot? Also probably fix up your PR description.

ysims (approved):
LGTM

PR #842, opened 2022-06-11
Fix kick script executing when intializing

TheCoolKidsTable:
It seems an ExecuteKick message is being emitted when the kick script is registering action and since the current trigger in kick script will execute a kick regardless of if its kickCommand is valid, a kick is executed.

This PR makes for follwoing changes:

- Adds check for valid limb id and type in `kickCommand` which fixes the kick script being executed
- Adds check to ensure the robot is not currently getting up
- Lowers modules priority to ensure getup starts if the robot falls during a kick


module/behaviour/skills/KickScript/src/KickScript.cpp
```diff
@@ -65,24 +65,37 @@ namespace module::behaviour::skills {
...
+                log<NUClear::WARN>("Kick command LimbID is unknown");
+                return;
+            }
+
             // Execute the penalty kick if the type is PENALTY
             if (kickCommand.type == KickCommandType::PENALTY) {
                 emit(std::make_unique<ExecuteScriptByName>(id, std::vector<std::string>({"KickPenalty.yaml"})));
             }
-            else {
+            // Execute the penalty kick if the type is PENALTY
```
Bidski, line 81: Copy-pasting is bad

Bidski (changes requested):
This method should be fine, just fix up your comments

Bidski:
Sorry @TheCoolKidsTable,  I should have been more explicit. That comment (along with the other one) says _what_ the code says. Comments should say _why_ the code does what it does, what does the code _mean_?

ysims (commented):
It looks like you're not really using the subsumption system here? We shouldn't have to check if the robot is getting up or kicking currently if we use subsumption. It looks like it would break in some cases.

BryceTuppurainen (approved):
All the changes makes sense to me, can't see anything I'd critique here :+1: (That hasn't already been mentioned)

PR #843, opened 2022-06-12
Update the Director DSL to solve a premature execution flaw

TrentHouliston:
This PR updates the DSL for the Director a little to avoid some confusion and fix the only remaining flaw I was aware of.

Adds the `Needs` keyword which will allow groups of providers to work together in a group, For example, the Walk may `Needs` the LeftLegIK and RightLegIK and won't run unless it is able to get both of them and transitively the servos involved.

Previously there was a flaw where the walk would run as it could get the IK modules, but those IK modules couldn't get the servos. Because the tasks were emitted dynamically the Director couldn't know this until it was already too late and the walk had been run. Now by providing a need keyword we can define these relationships explicitly and ensure it does not happen.

Also, changes how entering works to be Start/Stop and adds a Uses keyword to in the future provide information about what can be run.

Eventually, I want to add a function to `Uses` objects called `uses.why()` which will explain why something isn't running to help with debugging.


module/extension/Director/src/director/remove_task.cpp
```diff
@@ -36,25 +37,31 @@ namespace module::extension {
...
-                //      when/causing.
-                //  maybe the leaving without a causing needs it's own DSL keyword, (might be leaving and current
-                //      leaving needs renaming)
+                // Run the Stop reactions for this provider group since it is no longer running
+                for (auto& provider : group.providers) {
+                    if (provider->classification == ProviderClassification::STOP) {
+                        auto task = provider->reaction->get_task();
+                        if (task) {
+                            task->run(std::move(task));
+                        }
```
ysims, line 54: So getting the task and running it is how to run a provider?
Also does this imply that multiple STOP Providers might exist for one ProviderGroup?

module/extension/Director/src/director/remove_task.cpp
```diff
@@ -36,25 +37,31 @@ namespace module::extension {
...
+                    }
+                }
 
+                // Remove any subtasks this group had recursively
                 for (const auto& t : group.subtasks) {
```
ysims, line 59: Does removing the tasks still let the STOP Provider/s run?

module/extension/Director/src/provider/ProviderGroup.hpp
```diff
@@ -40,8 +40,6 @@ namespace module::extension::provider {
         enum State {
             /// This Provider group currently isn't doing anything
             IDLE,
-            /// This Provider group is currently in an "Entering" state
-            ENTERING,
```
ysims, line 44: Can't be in an entering state?

shared/extension/Behaviour.hpp
```diff
@@ -79,7 +79,7 @@ namespace extension::behaviour {
 
 
     /**
-     * This type is used as a base extension type for the different Provider DSL keywords (Provides, Entering, Leaving)
+     * This type is used as a base extension type for the different Provider DSL keywords (Start Stop, Provide, Leave)
```
ysims, line 82:
```suggestion
     * This type is used as a base extension type for the different Provider DSL keywords (Start, Stop, Provide, Leave)
```

shared/extension/Behaviour.hpp
```diff
@@ -243,81 +249,124 @@ namespace extension::behaviour {
...
-     * This is a special task that should be emitted when a Provider finishes the task it was given.
-     * When this is emitted the director will re-execute the Provider which caused this task to run.
+     * Adds a Uses object to the callback to allow access to information from the context of this provider.
+     *
+     * The uses object provides information about what would happen if you were to emit a specific task from this
+     * provider as well as what has happened in previous calls.
+     *
+     * It has two main uses
+     *   It allows you to see what would happen if you emitted a task (run, block proxy, etc)
+     *   It allows you to see what has happened in previous calls, e.g. are we re-running because that was emitted done
```
ysims, line 291:
```suggestion
     *   It allows you to see what has happened in previous calls, e.g. are we re-running because what was emitted is done
```
I'm not sure if this is right but the sentence didn't make sense.

shared/extension/behaviour/commands.hpp
```diff
@@ -35,36 +35,38 @@ namespace extension::behaviour::commands {
...
-        /// LEAVING Providers are executed as a module is relinquishing control
-        LEAVING,
+        /// Provide providers are typical providers that run when given a task
+        PROVIDE,
+        /// Leave providers are able to be forced to run when a higher priority task needs a causing it makes
+        LEAVE,
+        /// Start providers run once when a provider group first gets a task
+        START,
+        /// Stop providers run once there are no tasks running for any provider of this group
+        STOP
```
ysims, line 45: Are Leaving and Stop similar, but Stop is like Start in that it runs once and doesn't Cause anything, and then Leaving will keep running until it's happy and its Cause is satisfied?

module/extension/Director/src/Director.cpp
```diff
@@ -125,18 +134,44 @@ namespace module::extension {
...
+
+    void Director::add_needs(const NeedsExpression& needs) {
+        auto it = providers.find(needs.reaction->id);
+
+        if (it != providers.end()) {
+            auto provider = it->second;
+
+            // Can't add causing to a Start or Stop
+            if (provider->classification == ProviderClassification::STOP
+                || provider->classification == ProviderClassification::START) {
```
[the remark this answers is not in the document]
ysims: And what happens if a Provider needs to run a Leave but it doesn't have its Needs? Will it just wait?

module/extension/Director/src/director/remove_task.cpp
TrentHouliston: Getting a task and running it or pushing it to the thread pool is how you run a NUClear reaction. This particular one is me running it directly rather than handing it to the thread pool. Since the providers are NUClear reactions I'm running it here.

module/extension/Director/src/provider/ProviderGroup.hpp
TrentHouliston: Entering was previously done so that it could run some warmup tasks before doing the normal providers. However, I realised while trying to code it there is no difference between an entering and a normal provider so it makes no sense to have entering.

The role of entering is now just handled by having a normal provider that does what you want and the ordering is now handled by the fact that the system will try to run the versions of the provider in the order that you declare them.

so if you have a normal provider for walk that needed standing it couldn't run, but if you had a second provider declared after that that would cause standing it'll do the warmup automatically without a need for a specific entering element

shared/extension/Behaviour.hpp
TrentHouliston: You're right it didn't!

shared/extension/behaviour/commands.hpp
TrentHouliston: Leave is a bad name for what it has turned into. If you have a better name for it I'll happily change it.

As an example of why Leaving is a thing

You have a walk command with two normal `Provide` commands one is a normal walk and the other has a `Causing` for `Standing` that stops the walk and stands still.

You have a second module for `Look` that only works when the robot is in the `Standing` state that has a lower challenge priority than walk. It can still force the walk to start using the `Provide` statement that would stop the walk and make it stand. Which is not ideal.

Leaving providers are only able to be bullied into that state by a module with higher challenge priority. So if the look command were a higher priority it would be able to force it into a `Leave` provider. But if it was lower priority it would only be able to force it into other `Provide` statements.

So what you should do is to make the walk provider that makes the walk stop a `Leave` provider. That way it can only be forced into the stop state by another module that wants something the walk engine can do that would go against the main task (walking). So in a way it's "Leaving" although it might never actually stop running walk commands. Look might finish and the walk will go back to normal without ever "Leaving"

 and the other just walks for example a `Walk` has a `Provide` that causes `Standing` and one that just walks then another module with lower priority that is running (say a look module wants to be standing still) can only run when

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
TrentHouliston: Stop and Start are sort of in their own world. They don't really act as providers like the others. Start just run one time when a provider gets a task and had none previously, and stop runs once when there was a task and now there isn't one.

So think in walk you'd do resets in one of them, but most modules probably won't even use them unless they have some state that can't adapt to time jumps

module/extension/Director/src/director/remove_task.cpp
TrentHouliston: Yeah, stop providers will just get the last thing they ran with

module/extension/Director/src/provider/ProviderGroup.hpp
ysims: Ah fair enough

ysims:
You going to rename `Leave` in this PR?

shared/extension/Behaviour.hpp
```diff
@@ -79,7 +79,7 @@ namespace extension::behaviour {
 
 
     /**
-     * This type is used as a base extension type for the different Provider DSL keywords (Provides, Entering, Leaving)
+     * This type is used as a base extension type for the different Provider DSL keywords (Start, Stop, Provide, Leave)
```
ysims, line 82:
```suggestion
     * This type is used as a base extension type for the different Provider DSL keywords (Start, Stop, Provide)
```

ysims (commented):
Will look at this more in depth later (on my phone atm)

module/extension/Director/src/provider/ProviderGroup.hpp
```diff
@@ -40,15 +40,13 @@ namespace module::extension::provider {
...
             /// This Provider group is currently in a "Normal" state
             NORMAL,
-            /// This Provider group is currently in a "Leaving" state
-            LEAVING,
+            /// This Provider group is currently being Pushed by another provider group
+            PUSHED,
```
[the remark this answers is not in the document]
TrentHouliston: Yeah a provider can still be pushed by a higher priority provider

module/extension/Director/src/provider/ProviderGroup.hpp
```diff
@@ -40,15 +40,13 @@ namespace module::extension::provider {
...
             /// This Provider group is not running because it has requirements that are not met, in order to meet
             /// these requirements it is taking control of another currently running Provider group
-            PROXYING
+            PUSHING
```
ysims, line 49: When are these set in the code?

TrentHouliston: The algorithm that's not yet completed!

ysims: Fair enough haha

ysims (approved):
Yeah sure

PR #844, opened 2022-06-12
Ignore CI builds for things that haven't changed

TrentHouliston:
CI was taking too long so I added in code so it'll only build parts that have changed


Crafty15 (approved):
LGTM

PR #845, opened 2022-06-14
Stop extra kick in webots_keyboardwalk and add in a kick timeout

ysims:
Alternative to #842.

- Adds Stand module to `webots_keyboardwalk.role` inline with other motion roles
- Makes kick priority lower than get up priority
- Uses shared pointer to check kick command exists
- Adds a timeout so that the kick isn't executed too long after a kick command (since we probably are no longer necessarily in a good position to kick)
- Fill in module readme
- Add document directives

### DIFF
--- a/module/behaviour/skills/KickScript/data/config/KickScript.yaml
+++ b/module/behaviour/skills/KickScript/data/config/KickScript.yaml
@@ -1,3 +1,4 @@
 log_level: WARN
 
 "KICK_PRIORITY": 50
+message_timeout: 3000
--- a/module/behaviour/skills/KickScript/data/config/KickScript.yaml
+++ b/module/behaviour/skills/KickScript/data/config/KickScript.yaml
@@ -1,4 +1,3 @@
 log_level: WARN
 
 "KICK_PRIORITY": 50
-"EXECUTION_PRIORITY": 50
--- a/module/behaviour/skills/KickScript/data/config/KickScript.yaml
+++ b/module/behaviour/skills/KickScript/data/config/KickScript.yaml
@@ -1,4 +1,6 @@
 log_level: WARN
 
-"KICK_PRIORITY": 50
+# Priority set when wanting to kick
+kick_priority: 40
+# Time between kick command message and kicking before kick is discarded
 message_timeout: 3000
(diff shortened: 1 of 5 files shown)


module/behaviour/skills/KickScript/src/KickScript.cpp
```diff
@@ -48,27 +49,45 @@ namespace module::behaviour::skills {
...
+
+            // Don't kick if it's been a while since we got the KickScriptCommand
+            if (std::chrono::duration_cast<std::chrono::milliseconds>(NUClear::clock::now() - time_since_message)
+                    .count()
+                > message_timeout) {
+                updatePriority(0);
+                return;
+            }
+
             // auto direction = kickCommand.direction;
```
TheCoolKidsTable, line 86: Can probably remove this comment

module/behaviour/skills/KickScript/src/KickScript.cpp
```diff
@@ -48,27 +49,45 @@ namespace module::behaviour::skills {
...
-            KICK_PRIORITY      = config["KICK_PRIORITY"].as<float>();
-            EXECUTION_PRIORITY = config["EXECUTION_PRIORITY"].as<float>();
+            kick_priority   = config["kick_priority"].as<float>();
+            message_timeout = config["message_timeout"].as<Expression>();
         });
 
         on<Trigger<KickScriptCommand>>().then([this](const KickScriptCommand& cmd) {
-            kickCommand = cmd;
-            updatePriority(KICK_PRIORITY);
+            kickCommand        = std::make_shared<KickScriptCommand>(cmd);
```
TheCoolKidsTable, line 65: We should update var names to match convention while we are here:`kick_command`

module/behaviour/skills/KickScript/data/config/KickScript.yaml
```diff
@@ -1,4 +1,6 @@
 log_level: WARN
 
-"KICK_PRIORITY": 50
-"EXECUTION_PRIORITY": 50
+# Priority set when wanting to kick
+kick_priority: 40
+# Time between kick command message and kicking before kick is discarded
```
TheCoolKidsTable, line 5: Add units to comment: (in milliseconds)

TheCoolKidsTable (changes requested):
I tested in WeBots and it fixes all the issues I encountered previously. 

Other then very small changes suggested, LGTM!

TheCoolKidsTable:
Fix build issue and its good to go.

TheCoolKidsTable (approved):
LGTM!

PR #849, opened 2022-06-18
Add the algorithm to solve a director task execution

TrentHouliston:
This PR will add a function to the director system that will work out what provider options the director is able to use when executing a task.

The result will be a recursive data structure that stores all the options that the algorithm has at each step.
It will consist of requirements objects that are themselves solutions, including potentially pushing solutions where they will take control of other running tasks and influence them to run differently

The following step in the algorithm will be to take these solutions for each of the tasks emitted by a single provider and merging/pruning them into a single set of providers that can be executed to run the tasks.


module/extension/Director/src/provider/Provider.hpp
```diff
@@ -38,19 +38,14 @@ namespace module::extension::provider {
...
-                , current(std::move(current_))
-                , handle(std::move(handle_)) {}
-
+                          const bool& current_)
+                : type(type_), validator(std::move(validator_)), current(current_) {}
             /// The type of state that this condition is checking
             std::type_index type;
             /// Expression to determine if the passed state is valid
             std::function<bool(const int&)> validator;
             /// Function to get the current state
```
LachlanCourt, line 47: This isn't a function anymore

module/extension/Director/src/director/challenge_priority.cpp
```diff
@@ -75,7 +83,7 @@ namespace module::extension {
 
                 // Add this new parent we found and set t so we can loop up further
                 t = group.active_task;
```
LachlanCourt, line 85: Is this supposed to be overwriting the iterator variable here? It gives me bad juju but I guess the goal is to skip straight to the active task if possible?

module/extension/Director/src/Director.cpp
```diff
@@ -116,22 +116,25 @@ namespace module::extension {
...
+                    w->current = valid;
+                    emit(std::make_unique<StateUpdate>(id, w->type, state));
                 }
             });
-            handle.disable();
+            w->handle.disable();
 
             // Add it to the list of when conditions
-            provider->when.emplace_back(when.type, when.validator, when.current, handle);
+            provider->when.push_back(w);
```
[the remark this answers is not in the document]
CMurtagh-LGTM: In the current state of c++ there are enough optimisations that the choice is not as clear cut:
- `emplace_back` constructs a new object in the datastructure.
- `push_back` copies/moves an object into the datastructure.

Clearly if the object is not yet constructed `emplace_back` is the best option as we compare: construct then assign vs only construct.

Now if the object is already constructed, note for both we can either do a copy or a move. We consider how copy-assign, copy-construct, move-assign and move-construct work. Basically, efficiency wise they are very similar, and semantically wise `push_back` is preferred as it calls the assignment, which is what you expect for an already existing object.

module/extension/Director/src/director/challenge_priority.cpp
TrentHouliston: It is acting like an iterator variable, but really it's just a pointer (a shared_ptr in this case).
This part of the algorithm is just moving up the tree to find each direct ancestor

module/extension/Director/src/provider/Provider.hpp
TrentHouliston: True!

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
LachlanCourt: Fair enough! Thanks for the explanation 😊

module/extension/Director/src/Director.hpp
```diff
@@ -117,6 +117,103 @@ namespace module::extension {
...
+                enum State {
+                    /// This option is not blocked to us
+                    OK,
+                    /// This provider is blocked as the task given didn't have enough authority to run it
+                    BLOCKED_PRIORITY,
+                    /// This provider is blocked due to a when condition that cannot be met
+                    BLOCKED_WHEN,
+                    /// This provider is blocked as executing it would cause a loop of providers
+                    BLOCKED_LOOP,
+                };
```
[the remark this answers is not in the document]
TrentHouliston: Honestly, I'll be using `state != OK` most of the time. I put in these enums because there are differing reasons why it was blocked and when debugging it could help to be able to see the exact reason

ysims (approved):
This looks probably okay

PR #850, opened 2022-06-19
Stop Configuration running multiple times with the same data during installation phase of the program

ysims:
Fixes the issue where multiple runs of the Configuration reactor occurs with the same data at the beginning of the program. It adds a precondition function that checks if the reaction should run. The function includes checking if the file has been loaded in the initial stage (NO_OP) and if so and it's not the default config, then it doesn't run.

I'm not sure if using the `NO_OP` is a bit dodgy, and it might be better to see if there's a way to modify `FileWatcher` to set a particular event type for this.


Bidski:
If checking for `NO_OP` is the correct option then I would do it [here](https://github.com/NUbots/NUbots/blob/main/module/extension/FileWatcher/src/FileWatcher.cpp#L46-L64)  and dont call `reactor.powerplant.submit(std::move(task));` (not sure what else you need to not to do, @TrentHouliston )

ysims:
Yeah. My understanding was the FileWatcher is a generic module for seeing file events. While Configuration is the one that doesn't want to see repeat config at the beginning. But I can't really see how else to do it in Configuration if this `nullptr` thing won't work.

Bidski:
You are quite correct, `FileWatcher` is its own module and it shouldn't know anything about the inner workings of the `Configuration` extension, so doing the checks in `FileWatcher` is not a smart idea.

What you sort of want to do is send a list of files to `FileWatcher` and have it notify you if any one of those files gets modified. However, `libuv` only lets you trigger an event for 1 file at a time, so I feel like making this work would be complicated

CMurtagh-LGTM:
Why are we changing the `get` method, shouldn't we be adding overriding `precondition`?
```
bool precondition(threading::Reaction& reaction){
::extension::FileWatch watch = DSLProxy<::extension::FileWatch>::get<DSL>(r);
...
    if (watch.events == ::extension::FileWatch::Event::NO_OP && ((component == hostname) ||
        (component == binaryName) || (component == platform))) {
        return false;
    }
...
}
```

CMurtagh-LGTM:
As I remember precondition stops the reaction on false and proceeds on a true return.

ysims:
I'm not familiar with these different functions, but I'll have a go with the suggestion, ty @CMurtagh-LGTM

shared/extension/Configuration.hpp
```diff
@@ -346,67 +346,82 @@ namespace NUClear::dsl {
...
+
+            /// @brief Checks if the Configuration Reaction's parameters are valid
+            /// @details Parameters are not valid if the FileWatch object doesn't exist or it isn't a yaml file.
+            /// If this is the installation phase (denoted by the NO_OP FileWatch event), and this is not the default
+            /// configuration file, then the Reaction will also not run. This is to prevent Configuration Reactors
+            /// running multiple times during installation for one module.
+            /// @throws std::runtime_error if there is a YAML parsing error
+            /// @tparam DSL Magic NUClear type. Ignore for the purpose of understanding this function
+            /// @param t The associated Configuration Reaction
+            /// @return True if the file is valid, otherwise false.
```
CMurtagh-LGTM, line 397:
```suggestion
            /// @return False is the reaction is not to be run, otherwise true
```
Can return true or false if file is a `NO_OP`.

CMurtagh-LGTM (changes requested):
Looking good
However I don't like how they are now out of order; `precondition` should go above `get` and below `bind`.

CMurtagh-LGTM (dismissed):
LGTM

ysims:
Added some stuff for robustnuss, it won't check hostname, platform name or role name on stuff in the path before "config" (and including) and the actualy yaml file name.

CMurtagh-LGTM (approved):

PR #853, opened 2022-06-22
Update priority configuration and code conventions for our core behaviours 

Tom0Brien:
The goal of this PR is to make our current core behaviors consistent for setting and updating priority in subsumption system and follow code conventions.

Core behaviours modified are:
- `HeadController`
- `Getup`
- `KickScript`
- `SimpleWalkPathPlanner`
- `DirectWalkController`
- `Stand`
- `Nod`

The changes made for them include:
- Naming conventions for variables and functions
- Priority for each of the behaviors actions are stored in their config
- Move `emit<Scope::INITIALIZE>(std::make_unique<RegisterAction>` to same location in code
- Move initializing variables to header (except subsumption_id)
- Add comments in header files for variables

### DIFF
--- a/roles/robocup.role
+++ b/roles/robocup.role
@@ -31,6 +31,7 @@ nuclear_role(
   behaviour::skills::Getup
   behaviour::skills::KickScript
   behaviour::skills::HeadBehaviourSoccer
+  behaviour::skills::Stand
   behaviour::planning::KickPlanner
   behaviour::planning::SimpleWalkPathPlanner
   behaviour::strategy::SoccerStrategy


module/behaviour/planning/SimpleWalkPathPlanner/data/config/SimpleWalkPathPlanner.yaml
```diff
@@ -1,5 +1,8 @@
 log_level: WARN
 
+# Priority in subsumption system
+walk_path_planner_priority: 1
```
ysims, line 4: This looks like it'll be too low compared to the Stand module.

module/behaviour/skills/DirectWalkController/src/DirectWalkController.cpp
```diff
@@ -72,16 +79,20 @@ namespace module::behaviour::skills {
...
             }
             else {
-                emit(std::make_unique<ActionPriorities>(ActionPriorities{subsumptionId, {0, 0}}));
+                update_priority(0);
             }
         });
     }
+
+    void DirectWalkController::update_priority(const float& priority) {
+        emit(std::make_unique<ActionPriorities>(ActionPriorities{subsumption_id, {priority}}));
```
ysims, line 96: You're only setting one priority? There are two limb sets in this action. Check the others too.

ysims (changes requested):
You should check out the stand module as well. It's important so if we're not doing anything yet, the priority is on just standing. Nod would be good to verify for feedback on force playing.

module/behaviour/skills/DirectWalkController/src/DirectWalkController.cpp
```diff
@@ -72,16 +79,20 @@ namespace module::behaviour::skills {
...
             if (command.type == MotionCommand::Type::DIRECT_COMMAND) {
-                emit(std::make_unique<ActionPriorities>(ActionPriorities{subsumptionId, {26, 11}}));
-                emit(std::make_unique<WalkCommand>(subsumptionId, command.walk_command));
+                update_priority(cfg.direct_walk_priority);
+                emit(std::make_unique<WalkCommand>(subsumption_id, command.walk_command));
             }
             else if (command.type == MotionCommand::Type::STAND_STILL) {
-                emit(std::make_unique<ActionPriorities>(ActionPriorities{subsumptionId, {26, 11}}));
-                emit(std::make_unique<StopCommand>(subsumptionId));
+                update_priority(cfg.direct_walk_priority);
```
abihall, line 86: Should this be a different priority rather than the same as above if statement?

module/behaviour/skills/DirectWalkController/src/DirectWalkController.cpp
```diff
@@ -72,16 +79,20 @@ namespace module::behaviour::skills {
...
+                emit(std::make_unique<StopCommand>(subsumption_id));
             }
             else {
-                emit(std::make_unique<ActionPriorities>(ActionPriorities{subsumptionId, {0, 0}}));
+                update_priority(0);
```
abihall, line 90: Why is it that every other motion command other than direct or stand still is zero? Is there a situation that we will get another motion command other than these two and they need to be set to zero?

module/motion/HeadController/src/HeadController.cpp
```diff
@@ -52,81 +52,80 @@ namespace module::motion {
...
-                log_level = config["log_level"].as<NUClear::LogLevel>();
-
-                // Gains
-                head_motor_gain   = config["head_motors"]["gain"].as<double>();
-                head_motor_torque = config["head_motors"]["torque"].as<double>();
+                log_level                    = config["log_level"].as<NUClear::LogLevel>();
+                cfg.head_controller_priority = config["head_controller_priority"].as<double>();
+                cfg.head_motor_gain          = config["head_motors"]["gain"].as<double>();
+                cfg.head_motor_torque        = config["head_motors"]["torque"].as<double>();
+                cfg.smoothing_factor         = config["smoothing_factor"].as<float>();
```
abihall, line 64: Should some of the others be loaded here from config as well? e.g. min_yaw

module/behaviour/skills/DirectWalkController/src/DirectWalkController.cpp
ysims: If any other motion command is happening, DirectWalkController doesn't handle that and so shouldn't be doing anything (ie no priority on limbs).

module/behaviour/planning/SimpleWalkPathPlanner/data/config/SimpleWalkPathPlanner.yaml
ysims: oh yeah lol my bad should be fine. was on the bus so didn't go checking.

ysims:
If you don't think it's important to have different priorities for the arms and legs in some of these modules (I think it's mostly walk path planner and direct walk) then change it so it just gets one limb set like kick and get up. Otherwise give them two priorities and set up the yaml file to have eg ["priority"] ["arm"] and similar for leg.

LachlanCourt (approved):
LGTM

ysims (approved):
Sure

PR #855, opened 2022-06-24
Adds a KickRight script to be used with walking.

Crafty15:
First and last frames are matched (mostly, had to adjust final ankle pitch) to the stand script for smooth transitions between scripts.
The kick is pretty much a quick punt to minimise the chance of falling over when dribbling.
EDIT: Changed to KickRight.yaml, so this is now the main kick script, and we can make another for penalties.


ysims (approved):
LGTM

PR #857, opened 2022-06-24
Add unit tests to the Director and fix some bugs they exposed

TrentHouliston:
This PR adds many tests to the Director to ensure that when the algorithm is finished, we can be sure that it's working as we expect it to.
All the tests are currently flagged with `!mayfail`, so they won't impact if CI passes but will still get built (will test they compile properly)

This PR also fixes some errors that just using the Director DSL exposed

Tests to finish
- [x] simple test: test that emitting a task executes
- [ ] pass blocked task: test that one task is blocked other tasks can still use the resources it can't
- [x] priority steal: test that higher priority tasks can steal from lower priority tasks
- [x] start stop: test that the `Start` and `Stop` dsl words run when expected
- [x] triggered provider: test that you can trigger providers using other nuclear bind methods
- [x] needs blocking: test that if you need something and it's not available you're blocked
- [x] needs joining: test that if you need something and it becomes available later the task runs
- [x] needs priority loss: test that if a provider loses one of its dependent needs, it stops running all tasks and is blocked.
- [ ] needs when loss causing: check that if a provider with a needs has one of its needs lost to a when with a causing that it cancels all the other needs and runs that causing
- [ ] needs peeling: test that if you need something and someone else steals it, your entire needs tree gets kicked out
- [x] entire task: test that if you emit several non optional tasks, either all run or none do
- [x] optionals are optional: test that optional tasks don't need to run to run the rest
- [x] done task: test that emitting a `Done` task reruns your parent
- [x] when blocking: test that if a `When` condition isn't met the task is blocked
- [x] when causing: test that if `When` blocked and a `Causing` can provide it, it runs as expected
- [x] when causing ladder: test that when causing relationships can be applied recursively
- [ ] when causing ladder leaves: test that when a causing ladder happens, only leaves at the same level get executed
- [x] when causing multi: test that multiple when/causing relationships can run at once if they are orthogonal
- [ ] when causing stopped: test that when running a causing if the causing provider loses its task you reevaluate
- [ ] when causing new: test that if a previously inactive provider with an appropriate causing starts, the blocked provider will try to push it


module/extension/Director/tests/entire_task.cpp
```diff
@@ -0,0 +1,89 @@
...
+// Anonymous namespace to avoid name collisions
+namespace {
+
+    template <int i>
+    struct SimpleTask {
+        std::string msg;
+    };
+    struct ComplexTask {
+        std::string msg;
+    };
```
CMurtagh-LGTM, line 16: In the other files you have a constructor for these structs :thinking:

shared/extension/Behaviour.hpp
```diff
@@ -362,6 +363,36 @@ namespace extension::behaviour {
      */
     struct Done {};
 
+    /**
+     * A reactor subtype that can be used when making a behaviour reactor.
+     *
+     * It exposes the additional DSL words that are added by the Behaviour DSL so they can be used without the need for
+     * using statements
+     */
+    class BehaviourReactor : public NUClear::Reactor {
```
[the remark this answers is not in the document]
CMurtagh-LGTM: This is probably out of the scope of this PR though.

module/extension/Director/tests/optionals_are_optional.cpp
```diff
@@ -0,0 +1,96 @@
...
+// Anonymous namespace to avoid name collisions
+namespace {
+
+    template <int i>
+    struct SimpleTask {
+        std::string msg;
+    };
+    struct ComplexTask {
+        std::string msg;
+    };
```
CMurtagh-LGTM, line 16: You used a constructor in the other files.

module/extension/Director/tests/start_stop.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include "Director.hpp"
+#include "TestBase.hpp"
+
+// Anonymous namespace to avoid name collisions
+namespace {
+
+    template <int i>
+    struct Step {};
+
```
CMurtagh-LGTM, line 12:
```suggestion
```

module/extension/Director/tests/triggered_provider.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+#include "Director.hpp"
+#include "TestBase.hpp"
+
+// Anonymous namespace to avoid name collisions
+namespace {
+
+    template <int i>
+    struct Step {};
+
```
CMurtagh-LGTM, line 12:
```suggestion
```

module/extension/Director/tests/when_blocking.cpp
```diff
@@ -0,0 +1,91 @@
...
+                emit<Task>(std::make_unique<SimpleTask>());
+            });
+            on<Trigger<Step<4>>, Priority::LOW>().then([this] {
+                // This should make it run since it's now allowed
+                events.push_back("emitting allowed condition");
+                emit(std::make_unique<Condition>(Condition::ALLOW));
+            });
+            on<Trigger<Step<5>>, Priority::LOW>().then([this] {
+                // This task should run fine since it's allowed
+                events.push_back("emitting task again x2");
```
CMurtagh-LGTM, line 57:
```suggestion
                events.push_back("emitting task again _2");
```
`x2` implies that we emit the task twice here. I think you were trying to make the message unique and that `_2` would be better.

CMurtagh-LGTM (commented):
:stuck_out_tongue:

shared/extension/Behaviour.hpp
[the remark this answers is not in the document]
TrentHouliston: It would be another PR

The original idea was that there was a DSL that is the Behaviour DSL that's defined in this file. Then there is a module/algorithm called the Director that provides and implements.

That said, it's quite possible that the language in Behaviour.hpp has been muddied and should be cleared up to remove mentions of the Director (or just name the whole thing Director as you suggested)

module/extension/Director/tests/done_task.cpp
```diff
@@ -0,0 +1,64 @@
...
+        explicit TestReactor(std::unique_ptr<NUClear::Environment> environment)
+            : TestBase<TestReactor>(std::move(environment)) {
+
+            on<Provide<SubTask>>().then([this] {
+                events.push_back("subtask executed");
+                emit<Task>(std::make_unique<Done>());
+            });
+
+            on<Provide<SimpleTask>>().then([this] {
+                if (!executed) {
```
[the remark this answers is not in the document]
TrentHouliston: Looks like a hold over from an earlier version of the test before I had the shutdown on idle code.
I'll remove it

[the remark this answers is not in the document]
TrentHouliston: I should really update that variable so the test would work (if there was an algorithm)

module/extension/Director/tests/entire_task.cpp
TrentHouliston: Thought I had normalised a constructor on all of them, will fix

module/extension/Director/tests/when_blocking.cpp
TrentHouliston: True, I think `#2` helps even more signify it's number 2

CMurtagh-LGTM:
I'll give this another look when you mark it ready to review.

TrentHouliston:
The remainder of these tests are being moved into an issue for later so the parts of the Director can be finished first (as small details of those remaining tests may change)

module/extension/Director/tests/needs_blocking.cpp
```diff
@@ -0,0 +1,80 @@
...
+    // Check the events fired in order and only those events
+    REQUIRE(events
+            == std::vector<std::string>{
+                "requesting simple task",
+                "simple task",
+                "requesting low priority complex task",
+                "requesting high priority complex task",
+                "emitting tasks from complex: high priority complex task",
+                "high priority complex task",
+            });
```
ysims, line 79: Step<1> is emitted, which uses SimpleTask. "request simple task'
SimpleTask is run. "simple task"
Step<2> is emitted. "requesting low priority complex task"
at this point, Step<1> trigger has run and isn't needing to run again right? Why does it keep control of SimpleTask then?

ysims (commented):
partial

module/extension/Director/tests/simple_task.cpp
```diff
@@ -0,0 +1,53 @@
...
+            });
+
+            /**************
+             * TEST STEPS *
+             **************/
+            on<Trigger<Step<1>>, Priority::LOW>().then([this] {
+                events.push_back("emitting task");
+                emit<Task>(std::make_unique<SimpleTask>());
+            });
+            on<Startup>().then([this] {  //
```
ysims, line 31: Looks like you were gonna put in a comment?

module/extension/Director/tests/start_stop.cpp
```diff
@@ -0,0 +1,81 @@
...
+            on<Stop<SimpleTask>>().then([this] { events.push_back("stop"); });
+
+            on<Provide<Runner>>().then([this](const Runner& runner) {
+                // Emit a task to execute
+                if (runner.task) {
+                    events.push_back("running task");
+                    emit<Task>(std::make_unique<SimpleTask>());
+                }
+                else {
+                    events.push_back("removing task");
```
ysims, line 34: Does this stop because it runs but doesn't emit any tasks?

ysims (commented):
another partial

module/extension/Director/tests/simple_task.cpp
TrentHouliston: I'm just tricking the formatter so it doesn't make a mess

module/extension/Director/src/Director.cpp
```diff
@@ -211,9 +211,9 @@ namespace module::extension {
 
         // A task has arrived, either it's a root task so we send it off immediately, or we build up our pack for when
         // the Provider has finished executing
-        on<Trigger<DirectorTask>, Sync<DirectorTask>>().then(
+        on<Trigger<BehaviourTask>, Sync<BehaviourTask>>().then(
```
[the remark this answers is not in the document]
TrentHouliston: This is where the weird boundary is.
In theory, there is a Behaviour DSL, and the Director is an algorithm that operates with it.
So this is the director working with a BehaviourTask it has received from the DSL

module/extension/Director/tests/start_stop.cpp
TrentHouliston: When it emitted the task previously that task becomes this provider's "task pack" or pack of functions it wants to run.
The second time when it emits nothing that means it removes all tasks it didn't reemit.

So you can imagine for a provider, each time it runs it emits the tasks it needs.
If it fails to emit one, then the currently running tasks it didn't emit are ended.

module/extension/Director/tests/when_causing_ladder.cpp
```diff
@@ -0,0 +1,102 @@
...
+             **************/
+            on<Trigger<Step<1>>, Priority::LOW>().then([this] {
+                // Start up the helper
+                events.push_back("emitting helper task");
+                emit<Task>(std::make_unique<Helper>(), 10);
+            });
+            on<Trigger<Step<2>>, Priority::LOW>().then([this] {
+                // Emit the task
+                events.push_back("emitting task");
+                emit<Task>(std::make_unique<SimpleTask>());
```
ysims, line 72: shouldn't this have a priority >10? What is it if it's empty? or does that make it a root level task and so it has very high priority?

module/extension/Director/tests/when_causing_multi.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+            on<Provide<Subtask<1>>, When<Condition<1>, std::equal_to, Condition<1>::ALLOW>>().then([this] {  //
+                events.push_back("subtask 1 executed");
+            });
+            on<Provide<Subtask<2>>, When<Condition<2>, std::equal_to, Condition<2>::ALLOW>>().then([this] {  //
+                events.push_back("subtask 1 executed");
+            });
+
+
+            on<Provide<Helper<1>>>().then([this] { events.push_back("helper 1 waiting"); });
```
ysims, line 49: did you want to trick the formatter here too or is this one fine as is :stuck_out_tongue:

module/extension/Director/src/Director.cpp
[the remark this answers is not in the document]
ysims: So you mean Provide, Stop, Start, When, etc etc are Behaviour DSL?
I don't like it being called Behaviour though since that word is already used for something else. It may cause confusion and delay.
What about Action? That is a term used in the subsumption system which I assume is all disappearing once this is finished.

module/extension/Director/tests/when_causing_multi.cpp
ysims: (There's another one just below this for the second Helper too)

TrentHouliston: I don't mind this one being a single line

module/extension/Director/tests/when_causing_ladder.cpp
TrentHouliston: Good point, this wouldn't be able to trigger the ladder would it

module/extension/Director/tests/util/lcs.hpp
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * @brief Longest common subsequence algorithm that returns which elements from a and b form the common subsequence.
+     *
+     * This algorithm compares two lists and finds the longest subsequence you can make that is included in both of the
+     * lists. It returns a pair of bool vectors that indicate which elements are a part of the subsequence.
+     *
+     * @tparam T    the type of the elements to be compared
+     *
+     * @param a the first list to compare
+     * @param b the second list o compare
```
ysims, line 17:
```suggestion
     * @param b the second list to compare
```

ysims (dismissed):
Yeah should be good

ysims (approved):

PR #858, opened 2022-06-25
Adds HardwareIO for nugus3's metal legs

Crafty15:
Adjusted HardwareIO for nugus3's metal legs. Took out the minor difference needed to account for manufacturing faults in nugus1.


ysims (approved):
LGTM

PR #859, opened 2022-06-25
Kick the ball when it is within range and angle from VisionBalls

BryceTuppurainen:
### This branch allows SoccerStrategy to:

- Determine the torso distance and angle to the ball location from VisionBalls using the x and y components rBTt transform

- During normal_playing() if the distance and angle to the ball are suitable then create a KickCommand emit to kick with the RIGHT_LEG only

### Reviewers Please Note:

Currently this has the configured walk from main, although before I merge this I will be testing the values that I tuned last night on the bot to ensure that this is running appropriately for our game at 2pm

### Things left to do

- [x] Still going to do a little tidy-up before merge to ensure that this leaves main in a state that it can be worked on
- [x] Need to ensure that the distance to the ball is appropriate for all of the robots in case there are any failures with the metal legs robots


ysims:
Looks like you're still working on this right? Did you want to convert it to a draft PR?

module/behaviour/Controller/src/Controller.cpp
```diff
@@ -92,7 +92,9 @@ namespace module::behaviour {
             [this](const ActionPriorities& update) {
                 auto& request = requests[update.id];
 
-                // Find the largest priority
+                // TODO(BehaviourTeam): This can be converted to min_element
+                // ! in order to have the highest priority being
+                // ! 0 rather than the upper limit being extremely high Find the largest priority
```
Crafty15, line 97: My brain immediately negates anything behind an exclamation point in this sort of context. Are these there for a reason?

ysims, line 97: I would disagree, it's not really that big of a deal and we have the problem the other way around anyway. Get up/relax are at one end of the spectrum, stand is at the other end (priority epsilon or something).

module/behaviour/skills/HeadBehaviourSoccer/data/config/HeadBehaviourSoccer.yaml
```diff
@@ -1,22 +1,14 @@
 log_level: WARN
 
-# Search path positions
+# Search path positions # TODO(BehaviourTeam) Confirm that this is correct: (values in radians)
```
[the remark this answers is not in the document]
BryceTuppurainen: No worries, I've thrown in an updated todo item so that we can adjust this later on

module/behaviour/Controller/src/Controller.cpp
BryceTuppurainen: Screenshot from 2022-07-08 12-38-01

BryceTuppurainen:
Closing this PR as it has been merged into obrien/simple-ball-localisation and will be pulled into RoboCup 2022 from there instead

PR #860, opened 2022-06-28
Adds new HardwareIO for nugus2

Crafty15:
- Adjusted for metal legs
- Adjusts shoulder pitch so horn zero (Zombie) position is straight up (Same as n1 and n3)

### DIFF
--- a/module/platform/cm740/HardwareIO/data/config/nugus2/HardwareIO.yaml
+++ b/module/platform/cm740/HardwareIO/data/config/nugus2/HardwareIO.yaml
@@ -50,27 +50,27 @@ servos:
     simulated: false
   -  # [10] R_HIP_PITCH
     direction:  1
-    offset:  0.345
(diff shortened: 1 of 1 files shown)


ysims (commented):
If it's the same as n3, then we should probably just make this the default and only have a specific one for n1

Crafty15:
> If it's the same as n3, then we should probably just make this the default and only have a specific one for n1

Gonna have to double check what i've done here. Looks like i have the same IO for n1 and n2, with the minor differences in hip and knee values.

Crafty15:
> > If it's the same as n3, then we should probably just make this the default and only have a specific one for n1
> 
> Gonna have to double check what i've done here. Looks like i have the same IO for n1 and n2, with the minor differences in hip and knee values.

I checked them side by side and they are slightly different

ysims (approved):
lgtm

PR #862, opened 2022-06-30
Only include primary source files in Doxygen

CMurtagh-LGTM:
We currently include all source files in the codebase for Doxygen. I feel its better if we stick to only the files in `shared` and `module` so as not to include the more obscure types in the documentation.


ysims (commented):
Seems like a good idea. Roles would be nice to add as well, though I assume we'd need to set something up so that would work.

KipHamiltons:
I'd tend to agree. Seems reasonable to me too.

ysims (approved):
Sure

Issue #854, opened 2022-06-22
Folder capability for roles


ysims:
Add capability to put roles into folders.

eg `webots/keyboardwalk.role`

Would require a small change to the nuclear/roles/CMakeLists.txt and maybe nuclear/roles/NUClearRole.cmake

PR #866, opened 2022-06-30
Add capability to put roles in folders

ysims:
The purpose of this PR is to add the functionality to be able to put roles into subfolders within the `roles` directory. Likely will only have one extra level but I've tried to make it robust enough for however many levels the user wants. 

Configures, builds and runs binaries in Docker fine. Tested on the robot. Only annoying thing is that this PR makes a `webots` folder for binaries, but there is a binary called `webots`. So everyone's docker containers will need configure cleaning and the robot's binaries deleted.

- [x] Updated NUbook: NUbots/NUbook#243
- [x] Fix issues
- [x] Test on real robot

#854


nuclear/roles/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ foreach(rolefile ${rolefiles})
     # Roles may be within subfolders - get the subfolder and name
     file(RELATIVE_PATH role_folder ${role_source_dir} ${rolefile})
     cmake_path(REMOVE_EXTENSION role_folder)
+    string(REPLACE "/" "." role_folder ${role_folder})
```
TrentHouliston, line 28: You're replacing a dot here but in your regex you're not accepting `.`s
Actually, now that I think about it it's not accepting `_` either...

ysims: Yeah I was before but then I realised there were `_`s in the role names and it didn't seem to be affecting them. I guess I will see about adding them in.

ysims: I changed them from `/` to `.` because it seemed less bad. Having them be `_` means I might not be able to recover the path

nuclear/roles/CMakeLists.txt
```diff
@@ -22,14 +22,16 @@ foreach(rolefile ${rolefiles})
   # Check it is a .role file
   if(roleext STREQUAL .role)
 
-    # Get role name and add it to our list of roles
-    get_filename_component(role_name ${rolefile} NAME_WE)
+    # Roles may be within subfolders - get the subfolder and name
+    file(RELATIVE_PATH role_folder ${role_source_dir} ${rolefile})
```
KipHamiltons, line 26: Have you considered the case where a role is not in a subfolder? What about nested subfolders?
I feel like `role_folder` is slightly misleading as a variable name, since it's the relative path, which includes the role file's name, right? In the case where it's not in a subfolder, it wouldn't contain any folder names at all, if I'm not mistaken.

nuclear/roles/CMakeLists.txt
KipHamiltons: If I'm not mistaken, the reason it accepts those with `_` when `_` isn't in the regex is that `^ROLE_[a-zA-Z0-9]+` matches anything where `[a-zA-Z0-9]` matches one character. It's not matching the whole word in that case, it's just matching until the first `_`, which is still a match. Hence, unless the first character after `ROLE_` is not included in `[a-zA-Z0-9]`, it'll keep matching without modifications.

KipHamiltons (commented):
Partial 😅

tools/install.py
```diff
@@ -51,21 +51,45 @@ def run(target, local, user, config, toolchain, **kwargs):
...
 
+    # Build directory on the robot
     build_dir = b.binary_dir
 
-    cprint("Installing binaries to " + target_binaries_dir, "blue", attrs=["bold"])
-    files = glob.glob(os.path.join(build_dir, "bin", "*"))
-    subprocess.run(["rsync", "-avPl", "--checksum", "-e ssh"] + files + [target_binaries_dir])
+    # Traverse all binaries, including in subfolders
+    for dirpath, _, filenames in os.walk(os.path.join(build_dir, "bin")):
+        if not "lib" in dirpath:  # lib folder contains libraries, not binaries
```
KipHamiltons, line 71: It seems like we used to install libraries - how come this changed?

tools/install.py
```diff
@@ -51,21 +51,45 @@ def run(target, local, user, config, toolchain, **kwargs):
...
+            # Get the paths locally and on the robot for this binary folder
+            binary_subdirectory = os.path.normpath(dirpath).split(os.path.join(build_dir, "bin"))[1].strip("/")
```
KipHamiltons, line 73: this makes me think hard

tools/install.py
```diff
@@ -51,21 +51,45 @@ def run(target, local, user, config, toolchain, **kwargs):
 
     # Target location to install to
     if local:
+        target_name = target
```
KipHamiltons, line 54: this is almost certainly beyond the scope of this PR, but I think `target` is kind of confusing here, since that's a word we use for other things. It's `./b target generic`, then in the cmake scripts, a target is a thing we build, then here it's the location we're installing to. Idk. It's a bit to juggle

KipHamiltons (commented):
The rest :shipit:

nuclear/roles/CMakeLists.txt
Bidski: `role_path` might alleviate @KipHamiltons confusion?

tools/install.py
Bidski: Yea, you kind of need to install the libraries, these are important

ysims: I definitely got confused somewhere there. But this is now making me wonder why I have to put in all this stuff with making directories on the robot for the subfolders anyway, since I assume `lib` isn't there on initial robot set up? Maybe the error I was getting was actually because of the existing `webots` binary clashing with the new `webots` subfolder. Will test when I'm in the lab.

Bidski:
@ysims see #952 for a "review" and suggestions on how to improve this PR

nuclear/roles/CMakeLists.txt
```diff
@@ -2,67 +2,66 @@
...
-          CACHE INTERNAL "A list of the roles that are enabled to be built." FORCE
-      )
-    endif()
+  # Convert path to underscore separated folders
+  get_filename_component(role_path ${role_path} DIRECTORY)
+
+  # Join the role name with the role path again
+  if(role_path)
+    string(REPLACE "/" "_" role_folder ${role_path})
+    string(CONCAT role_folder ${role_folder} ";" ${role_name})
```
Bidski, line 40: Not sure `role_folder` is actually used anywhere? (either in this file or in `NUClearRole.cmake`)

nuclear/roles/NUClearRole.cmake
```diff
@@ -22,33 +21,29 @@ function(NUCLEAR_ROLE)
...
+  if(role_path)
+    set_property(TARGET ${role} PROPERTY FOLDER "roles/${role_path}/")
+  else()
+    set_property(TARGET ${role} PROPERTY FOLDER "roles/")
+  endif()
 
   # Store the used NUClear modules on the target as a property This can be used later in scripts to work out what
   # modules are used in the role
-  set_target_properties(${target_name} PROPERTIES NUCLEAR_MODULES "${role_modules}")
+  set_property(TARGET ${role} PROPERTY NUCLEAR_MODULES ${role_modules})
```
Bidski, line 46: You can also combine this line with the other `set_target_properties` above
```suggestion
  set_target_properties(${role} PROPERTIES NUCLEAR_MODULES ${role_modules})
```

Bidski (changes requested):
Not much point in me reviewing this any further since the majority of the changes here are my work

nuclear/roles/NUClearRole.cmake
```diff
@@ -24,23 +23,27 @@ function(NUCLEAR_ROLE)
...
 
-  # Set our output directory to be bin
-  set_target_properties(${role} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin/")
-
   # IDE folder
-  set_target_properties(${role} PROPERTIES FOLDER "roles/")
+  if(role_path)
+    set_target_properties(${role} PROPERTY FOLDER "roles/${role_path}/")
+  else()
+    set_target_properties(${role} PROPERTY FOLDER "roles/")
```
TrentHouliston, line 41: @Bidski led you astray
```suggestion
    set_target_properties(${role} PROPERTIES FOLDER "roles/${role_path}/")
  else()
    set_target_properties(${role} PROPERTIES FOLDER "roles/")
```

nuclear/roles/NUClearRole.cmake
```diff
@@ -24,23 +23,27 @@ function(NUCLEAR_ROLE)
...
+  endif()
 
   # Store the used NUClear modules on the target as a property This can be used later in scripts to work out what
   # modules are used in the role
-  set_target_properties(${role} PROPERTIES NUCLEAR_MODULES "${role_modules}")
+  set_target_properties(${role} PROPERTY NUCLEAR_MODULES ${role_modules})
```
TrentHouliston, line 46:
```suggestion
  set_target_properties(${role} PROPERTIES NUCLEAR_MODULES ${role_modules})
```

TrentHouliston (changes requested):

nuclear/roles/NUClearRole.cmake
Bidski: Oops 😛

ysims: I think that was me manually doing it when I was fixing up the other thing and missed that part 😅

TrentHouliston (approved):

Bidski (approved):

PR #872, opened 2022-07-02
Disher/fall relax

PiiDish:
### Pre-PR Checklist:

Have you
- [ ] Added/modified [documentation directives](https://nubook.nubots.net/guides/general/documentation#extra-formatting) in relevant code

### Reviewers, Note

Relax fall wortksa just. Readying for last practice match

### Things left to do
- [ ] Get it to relax before it hits ground


module/behaviour/skills/FallingRelax/data/config/FallingRelax.yaml
```diff
@@ -1,6 +1,8 @@
-{
-	"FALLING_ANGLE" : 0.8,
-	"FALLING_ACCELERATION" : 8,
-	"RECOVERY_ACCELERATION" : [9.5, 10.5],
-	"PRIORITY" : 90
-}
+# Common
+log_level: WARN
+relax_fall_priority: 5
+falling_angle: 0.65 #0.611
```
BryceTuppurainen, line 4: Stating that this value is in Radians in a comment (as well as falling_acceleration) would be useful for newer members trying to read through these config files.

module/behaviour/skills/FallingRelax/src/FallingRelax.cpp
```diff
@@ -1,3 +1,112 @@
...
+//  * GNU General Public License for more details.
+//  *
+//  * You should have received a copy of the GNU General Public License
+//  * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+//  *
+//  * Copyright 2013 NUbots <nubots@nubots.net>
+//  */
+
+
+// #include "FallingRelax.hpp"
```
BryceTuppurainen, line 21: Please delete artifacts such as this. They should be managed in the GitHub commit history, or on a branch/stashed that isn't in a PR so that we don't end up having commented out code on main

module/behaviour/skills/FallingRelax/src/FallingRelax.cpp
```diff
@@ -34,54 +144,63 @@
 namespace module::behaviour::skills {
 
     // internal only callback messages to start and stop our action
-    struct Falling {};
-    struct KillFalling {};
+    // struct Falling {};
```
BryceTuppurainen, line 147: This can be deleted if unused rather than commented out.

module/behaviour/skills/FallingRelax/src/FallingRelax.cpp
```diff
@@ -34,54 +144,63 @@
...
-                magnitude /= sensors.size();
+            if (!relaxed && !sensors.empty() && acos(Eigen::Vector3d::UnitZ().dot(uZTw)) > cfg.falling_angle) {
+
+                // We might be falling, check the accelerometer
+                double gravity = sensors.back()->accelerometer.norm();
 
-                if (magnitude < FALLING_ACCELERATION) {
-                    falling = true;
-                    updatePriority(PRIORITY);
+                if (gravity < cfg.falling_acceleration) {
```
BryceTuppurainen, line 198: Have we found that on the real robot this is sufficient to detect falling with the amount of noise that occurs on the IMU? If not, I would suggest breaking this out to a function which includes filtering/cutoffs appropriately so that we don't randomly collapse during play.

PR #874, opened 2022-07-03
Adds default kicks, new stand up front and removes old directories and scripts

Crafty15:
- Removed individual getup scripts as all robots now use the defaults.
- Added a new StandUpFront script that is a bit easier on the arms. Referred to as "The Banana".
- Added tuned left and right default kicks .
- Tuned the above kick scripts to suit n1. It seems to work a bit differently from the others, possibly due to having metal hips.
- Removed unneeded n2 and n3 script directories.

NOTE: The robots will need to have their directories cleaned of any left over old scripts


ysims (approved):
lgtm

PR #875, opened 2022-07-04
Update vision to use updated Visual Mesh that has opencl binary caching to reduce install time

ysims:
VisualMesh takes a few minutes to install on `main`, due to compiling OpenCL in the Visual Mesh engine. This PR makes the code compatibile with the change to the Visual Mesh that includes saving and loading precompiled OpenCL binaries. See FastCode/VisualMesh#11 for details. The changes here involve including a cache directory path to the engine call. 

Functionally, there should be around a 44 second wait on Visual Mesh installation the very first time this is run on each robot, and then unless the OpenCL binary is deleted it should only take a couple of seconds to install from then on. The binary is stored in `/home/nubots/vision_cache/` on each robot (can be changed in VisualMesh.yaml).

Todo:
- [x] Update the Dockerfile to use the right version of the Visual Mesh once the Visual Mesh PR is merged


module/vision/VisualMesh/src/visualmesh/VisualMeshRunner.cpp
```diff
@@ -50,6 +53,28 @@ namespace module::vision::visualmesh {
...
 
+        template <template <typename> class Engine, typename Scalar>
+        struct BuildEngine;
+
+        template <typename Scalar>
+        struct BuildEngine<::visualmesh::engine::opencl::Engine, Scalar> {
+            static std::shared_ptr<::visualmesh::engine::opencl::Engine<Scalar>> build(
+                ::visualmesh::NetworkStructure<float> net,
+                std::string cache) {
+                fs::create_directory(cache);  // create the cache folder if it doesn't exist
```
TrentHouliston, line 64: You probably want to make all of them if you decide to nest your directory beyond 1 level
```suggestion
                fs::create_directories(cache);  // create the cache folder if it doesn't exist
```

TrentHouliston (changes requested):

TrentHouliston (approved):

CMurtagh-LGTM (approved):

PR #299, opened 2020-04-18
Add a system configuration tool

TrentHouliston:
This tool can be used to store many of the system configuration files in the repository.
It will then apply them onto the robot when the system configuration role is run as a superuser.

- [x] Setup the networking configuration files
- [x] Setup the udev rules
- [x] Setup the systemd and enable them
- [x] Update the install instructions in NUbook

https://github.com/NUbots/NUbook/pull/245

PR #881, opened 2022-07-07
Lock pacman date

CMurtagh-LGTM:
Currently we lock the date in the dockerfile, but we don't when we install on the robot. This leads to the us building with `libasan.so.6` but the robot has `libasan.so.8`.

https://wiki.archlinux.org/title/Arch_Linux_Archive#How_to_restore_all_packages_to_a_specific_date

### DIFF
--- a/doc/ArchInstall/arch-chroot_install.sh
+++ b/doc/ArchInstall/arch-chroot_install.sh
@@ -31,6 +31,12 @@ cat << EOF > /etc/hosts
 127.0.1.1       ${HOSTNAME}
 EOF
 
+# Lock the pacman date
+# ↓↓↓ UPDATE MIRRORS HERE ↓↓↓
+cat > /etc/pacman.d/mirrorlist << EOF
+Server=https://archive.archlinux.org/repos/2021/08/29/\$repo/os/\$arch
+EOF
+
 #########
 # USERS #
 #########


TrentHouliston:
This PR https://github.com/NUbots/NUbots/pull/299
already would do this here
https://github.com/NUbots/NUbots/pull/299/files#diff-a91f875a6871edf665823f77cd53ce163b37db50bc4ab840e26e0ed32eec8dd1R5-R7

And in such a way that you don't have to keep reinstalling/updating from arch_chroot install

doc/ArchInstall/arch-chroot_install.sh
```diff
@@ -31,6 +31,12 @@ cat << EOF > /etc/hosts
 127.0.1.1       ${HOSTNAME}
 EOF
 
+# Lock the pacman date
+# ↓↓↓ UPDATE MIRRORS HERE ↓↓↓
+cat > /etc/pacman.d/mirrorlist << EOF
+Server=https://archive.archlinux.org/repos/2021/08/29/\$repo/os/\$arch
+EOF
```
Bidski, line 38: An alternative (and I am open to opinions here) is to create a `docker/etc/pacman.d/mirrorlist` file that contains the appropriate mirror information and then download that file from the repo. Having `UPDATE_MIRRORS_HERE` everywhere that the mirrors needs to be changed is asking for someone to miss one of them.

Bidski (changes requested):

CMurtagh-LGTM:
> This PR https://github.com/NUbots/NUbots/pull/299
already would do this here
https://github.com/NUbots/NUbots/pull/299/files#diff-a91f875a6871edf665823f77cd53ce163b37db50bc4ab840e26e0ed32eec8dd1R5-R7

> And in such a way that you don't have to keep reinstalling/updating from arch_chroot install

Yes, but until then this is a step forward.

TrentHouliston:
But it would only help if we reflash a robot since this is only done on install

CMurtagh-LGTM:
I'm planning on going through and manually updating the mirror list and running `pacman -Syyuu`.

Bidski:
I think you also need the LiveUSB mirrorlist in [this script](https://github.com/NUbots/NUbots/blob/6ae2d34010de1cc44547f174251c174bfbdf342b/doc/ArchInstall/arch_install.sh) before running `pacstrap` as well. If I am not mistaken, this should copy across the mirrorlist into the new installation. This should prevent the need for running `pacman -Syyuu` manually and will probably mean you dont need to update the mirrorlist the way you currently are in this PR.

The only thing that may throw a spanner in the works is [`reflector`](https://wiki.archlinux.org/title/Reflector). Might need to disable the reflector service??

CMurtagh-LGTM:
> I think you also need the LiveUSB mirrorlist in [this script](https://github.com/NUbots/NUbots/blob/6ae2d34010de1cc44547f174251c174bfbdf342b/doc/ArchInstall/arch_install.sh) before running `pacstrap` as well. If I am not mistaken, this should copy across the mirrorlist into the new installation. This should prevent the need for running `pacman -Syyuu` manually and will probably mean you dont need to update the mirrorlist the way you currently are in this PR. If you dont deal with the mirrorlist before you run `pacstrap` you will need to run `pacman -Syyuu` in the chroot script after updating the mirrorlist
> 
> The only thing that may throw a spanner in the works is [`reflector`](https://wiki.archlinux.org/title/Reflector). Might need to disable the reflector service??

I didn't think of that, this is going to be more difficult than I thought. I might just do it manually on the robot I need, then look at #299

CMurtagh-LGTM:
Its worse than I thought, `hwids` was replaced by `hwdata`. This means that a downgrade is impossible as pacman does not know that `hwids` was replaced by `hwdata` if we set the archive before the swap.

PR #882, opened 2022-07-07
Patch intel-compute-runtime to resolve a memory leak

CMurtagh-LGTM:
Apply the patch from https://github.com/intel/compute-runtime/pull/550 into the version we use.


KipHamiltons:
Have you considered `wget`ing the patch into a pipe into `patch`, such as we do in the IGC install script?

TrentHouliston (changes requested):
The patch file is no longer needed

TrentHouliston (approved):
Wait nevermind I thought the second file was the patch file, turns out it's just that we have a custom build script for this

CMurtagh-LGTM:
> Have you considered `wget`ing the patch into a pipe into `patch`, such as we do in the IGC install script?

I like this way better.

Issue #736, opened 2021-10-03
The FileWatcher destructor can segfault


KipHamiltons:
I have been butting heads with a segfault and I have tracked down where the segfault originates. I can't replicate the issue on my machine - it only happens during a particular test.

[This GHA run](https://github.com/NUbots/NUbots/pull/728/checks?check_run_id=3780850190#step:8:35) isolates the problem reasonably well. Single threaded, [after all the code in the test case](https://github.com/NUbots/NUbots/pull/728/files#diff-ba446b102ba87bc70b1dacd1ec0c8d21b1581e6bd1a5cf509ba45fbfeef14f03R78), we get inside the destructor for FW, then promptly segfault. [Somewhere in that first `for` loop](https://github.com/NUbots/NUbots/pull/728/files#diff-0e68052f9d92d448f711689136865870e93be8a480a5aaf25e826292fde4a430R296-R301), it's having the problem.

@CMurtagh-LGTM do you have experience with FileWatcher? Any ideas?

Also, if anyone can replicate this error, that would make it much much easier to debug - adding print statements to send to GHA to execute after building for 20 mins is less than ideal lol.

This currently blocks #728 and transitively #727, since I can't merge the test API in good faith when the only demo segfaults lmao

JosephusPaye:
Oooh so it was in a destructor

KipHamiltons:
Yessir. I have no idea why tho :(

CMurtagh-LGTM:
```
Start of filewatcher destructor
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
TestmotionKinematicsConfiguration is a Catch v2.11.1 host application.
Run with -? for options
-------------------------------------------------------------------------------
Testing the Kinematics Configuration module
-------------------------------------------------------------------------------
/__w/NUbots/NUbots/module/motion/KinematicsConfiguration/tests/KinematicsConfigurationTest.cpp:55
...............................................................................
/__w/NUbots/NUbots/module/motion/KinematicsConfiguration/tests/KinematicsConfigurationTest.cpp:55: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal
===============================================================================
test cases: 1 | 1 failed
assertions: 2 | 1 passed | 1 failed
```
[...]

PR #885, opened 2022-07-08
Close the event handles in the uv-loop in FileWatcher

CMurtagh-LGTM:
As libuv is a c library we need to call the constructor for the type explicitly.
Possibly fixes segfault #736
Importantly https://github.com/libuv/libuv/blob/v1.x/include/uv.h#L437 needs to be set so the polymorphism can work. One important instance of the polymorphism is calling `uv_close` with the reinterpret cast.


module/extension/FileWatcher/src/FileWatcher.cpp
```diff
@@ -236,6 +236,7 @@ namespace module::extension {
                 if (paths.find(dir) == paths.end()) {
                     auto& p        = paths[path];
                     p.handle       = std::make_unique<uv_fs_event_t>();
+                    uv_fs_event_init(loop.get(), p.handle);
```
TrentHouliston, line 239: Is there a delete that needs to be run that should be attached to the deleter?

TrentHouliston (changes requested):

module/extension/FileWatcher/src/FileWatcher.cpp
CMurtagh-LGTM: Its freed here https://github.com/NUbots/NUbots/blob/main/module/extension/FileWatcher/src/FileWatcher.cpp#L297

CMurtagh-LGTM: I could add it to the deleter.

module/extension/FileWatcher/src/FileWatcher.hpp
```diff
@@ -23,6 +23,20 @@
 
 namespace module::extension {
 
+
+    ///@brief Call the cleanup function and delete the ptr
+    ///@tparam must be be reinterpret castable into uv_handle_t
```
KipHamiltons, line 28:
```suggestion
    /// @tparam must be be reinterpret castable into uv_handle_t
```
The one above as well. Always reviewing the important things 😂😂

KipHamiltons (commented):
Stuff like this is impossible to review on my phone lol, but I did spot this 😝😝😝

module/extension/FileWatcher/src/FileWatcher.cpp
```diff
@@ -110,9 +110,9 @@ namespace module::extension {
     FileWatcher::FileWatcher(std::unique_ptr<NUClear::Environment> environment)
         : Reactor(std::move(environment))
         , loop(std::make_unique<uv_loop_t>())
-        , add_watch(std::make_unique<uv_async_t>())
```
TrentHouliston, line 113: Not a fan of new...
I think I'd rather either a custom wrapper struct for these, or shared_ptrs so you can set the deleter per object

module/extension/FileWatcher/src/FileWatcher.cpp
```diff
@@ -291,22 +292,20 @@ namespace module::extension {
     }
 
     FileWatcher::~FileWatcher() {
-        for (const auto& path : paths) {
+        // Remove all the handles to the uv loop will finish
+        for (auto& path : paths) {
             if (path.second.handle) {
-                uv_fs_event_stop(path.second.handle.get());
-                uv_close(reinterpret_cast<uv_handle_t*>(path.second.handle.get()), [](uv_handle_t* /* handle */) {});
+                path.second.handle.reset();
```
TrentHouliston, line 298: Do we need these anymore if we have the variables? they'll delete themselves anyway

CMurtagh-LGTM: As per the comment, they need to be freed for libuv to finish. We wait for libuv to finish in the deconstructor of FileWatcher.

CMurtagh-LGTM:
I've determined all the reworking of the Filewatcher is out of scope for this PR and that I'll just fix #736 here.

TrentHouliston (approved):
¯\_(ツ)_/¯

PR #887, opened 2022-07-12
Update our team id for RoboCup 2022

ysims:
It is 1, because we are number 1!!!

see here for team ids for this year:
https://github.com/RoboCup-Humanoid-TC/GameController/blob/master/resources/config/hl_kid/teams.cfg


BryceTuppurainen:
LGTM :smile:

Just need to remember to change the robot_id for each robot before we begin play.

PR #889, opened 2022-07-12
Simple Ball Localisation and Kick Behaviour

Tom0Brien:
### This PR adds the following:
- Move VisionBall trigger from behavior modules to its own module to allow filtering and additional throw outs
- This new module emits the distance, absolute yaw angle, time of measurement, rBTt and srBCc of the latest VisionBall message which is closest to our current estimate
- During normal_playing() if the distance and angle to the ball are suitable a KickCommand is emitted to kick with leg closest to the ball 
- TODO: Rotate in the direction of last seen ball


ysims:
I think this was added in #957. @Tom0Brien can you check and close if so.

PR #892, opened 2022-07-12
Temp robocup docker image

Crafty15:
plz help


docker/Dockerfile
```diff
@@ -27,6 +27,11 @@ RUN pacman -Syyuu --noconfirm --needed --overwrite \*
 COPY usr/local/bin/install-package /usr/local/bin/install-package
 COPY etc/pacman/pacman.latest.conf /etc/pacman.latest.conf
 
+# Update the pacman keyring
+RUN pacman -Sy archlinux-keyring --noconfirm
+RUN pacman-key --init
+RUN pacman -Su --noconfirm
+
```
CMurtagh-LGTM, line 34:
```suggestion
# Update the pacman keyring
RUN pacman-key --init
RUN pacman -Sy archlinux-keyring --noconfirm
RUN pacman -Su --noconfirm

```

.github/workflows/nubots.yaml
```diff
@@ -97,21 +97,21 @@ jobs:
...
           context: docker
           build-args: |
             platform=generic
             BUILDKIT_INLINE_CACHE=1
           push: true
           cache-from: |
-            type=registry,ref=nubots/nubots:pull-request-${{ github.event.number }}
-            type=registry,ref=nubots/nubots:generic
+            type=registry,ref=crafty15/nubots:pull-request-${{ github.event.number }}
+            type=registry,ref=nubots/crafty15/nubots:generic
```
mrwerdo, line 109: Remove the `nubots/` prefix. I think it should be `crafty15/nubots:generic`.

docker/etc/pacman/pacman.conf
```diff
@@ -0,0 +1,100 @@
...
+#UseSyslog
+#Color
+#TotalDownload
+# We cannot check disk space from within a chroot environment
+#CheckSpace
+#VerbosePkgLists
+
+# By default, pacman accepts packages signed by keys that its local keyring
+# trusts (see pacman-key and its man page), as well as unsigned packages.
+SigLevel    = Never DatabaseOptional
```
TrentHouliston, line 41:
```suggestion
SigLevel    = Never TrustAll
```

mrwerdo:
The problem is this:
1. On the Arch Linux website, [python-distlib](https://archlinux.org/packages/community/any/python-distlib/) version 0.3.4-1 was signed by Antonio Rojas on the 2022-01-04.
2. Our docker file is date locked to 2021-08-29, so the version of python-distlib in the docker image is always going to be 0.3.2-1. This was signed by Eli Schwartz.
3. The install-package command updates the keychain to trust the same signers as what Arch Linux uses. Unfortunately, Eli Schwartz’s certificate is no longer trusted because his certificate expired.

So we have to install packages by explicitly trusting Eli Schwartz’s certificate, or update the dependency.

@TrentHouliston, you bet met to it. 😜

PR #896, opened 2022-07-12
Disable PGP rather than trying to use a separate keychain

TrentHouliston:
This changes how PGP is handled so that rather than trying to update to the latest keychain before installing (which can fail sometimes) it just blindly installs packages without checking them.

Since we're date locked sometimes keys will expire and it won't install because of PGP errors. This gets rid of those errors by not checking.

The downside is there is an infinitesimal chance that if someone were able to compromise your package download they could add something malicious.

### Pre-PR Checklist:

Have you


docker/etc/pacman.conf
```diff
@@ -0,0 +1,29 @@
+[options]
+HoldPkg      = pacman glibc
+Architecture = auto
+TotalDownload
+Color
+ILoveCandy
+
+# Trust anything so we don't get signature errors for old packages while date locked
```
Bidski, line 8:
```suggestion
# Trust everything so we don't get signature errors for old packages while date locked
```

Bidski (changes requested):

miikyla (approved):
Looks good

Crafty15 (approved):
LGTM

PR #900, opened 2022-07-13
Fix cmake so we can set the CI build flag

CMurtagh-LGTM:
Don't add FORCE if there is no if statement protecting it.


ysims (approved):
LGTM

PR #908, opened 2022-07-25
Change bad name in cmake ToolchainLibraryFinder

CMurtagh-LGTM:
The current name is `VERSION_BINARY_ARGUMENTS` but there is no relation to the version of anything, I've changed it to `BINARY_ARGUMENTS` as I feel its a better representation of what its doing.

### DIFF
--- a/nuclear/cmake/Modules/ToolchainLibraryFinder.cmake
+++ b/nuclear/cmake/Modules/ToolchainLibraryFinder.cmake
@@ -11,7 +11,7 @@ function(ToolchainLibraryFinder)
       "PATH_SUFFIX"
       "BINARY"
       "VERSION_FILE"
-      "VERSION_BINARY_ARGUMENTS"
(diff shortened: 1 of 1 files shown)


TrentHouliston (commented):
The reason this was called version binary arguments was that it is supposed to be the arguments you pass to the binary so it prints out its version.

CMurtagh-LGTM:
but, its just the arguments that you pass to the binary?

TrentHouliston:
Yeah the arguments you pass to the binary specifically to get the version.
For ToolchainLibraryFinder for some libraries it can output a version number when it does configure, but to do that it needs a version.

I think you can get it from regexing a source file, or running a binary

CMurtagh-LGTM:
This is why we need better documentation :disappointed:.

ysims:
Please delete the branch too if this isn't needed @CMurtagh-LGTM

PR #909, opened 2022-07-25
Update the versions used in the Dockerfile

TrentHouliston:
The versions of libraries used in the codebase are getting older and we should make an effort to update them from time to time.
The longer we wait the harder it gets.

I have made some progress updating the versions (especially the tricky intel compute runtime ones) and figured I would share the results.
I'll update this as I get time but otherwise if others want to advance the version numbers and fix the builds hopefully we will get this done so we can date lock to a more recent date.


docker/Dockerfile
```diff
@@ -91,117 +95,125 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr
...
+    -DIGC_OPTION__LLD_MODE=Prebuilds \
+    -DIGC_OPTION__LLVM_PREFERRED_VERSION='14.0.6' \
     -DIGC_OPTION__LLVM_MODE=Prebuilds \
-    -DCMAKE_CXX_FLAGS="-Wno-deprecated-declarations" \
-    -DINSTALL_SPIRVDLL=0 \
+    -DIGC_OPTION__LINK_KHRONOS_SPIRV_TRANSLATOR=ON \
+    -DIGC_OPTION__USE_PREINSTALLED_SPRIV_HEADERS=ON \
+    -DIGC_OPTION__SPIRV_TOOLS_MODE=Prebuilds \
+    -DIGC_OPTION__SPIRV_TRANSLATOR_MODE=Prebuilds \
+    -DIGC_OPTION__VC_INTRINSICS_MODE=Prebuilds \
```
KipHamiltons, line 159: Good lord I hate intel open source

docker/Dockerfile
```diff
@@ -91,117 +95,125 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr
...
+RUN install-from-source https://github.com/KhronosGroup/SPIRV-Headers/archive/refs/tags/sdk-1.3.216.0.tar.gz
+RUN install-from-source https://github.com/KhronosGroup/SPIRV-Tools/archive/refs/tags/v2022.2.tar.gz \
+    -DSPIRV_WERROR=Off \
+    -DBUILD_SHARED_LIBS=ON \
+    -DSPIRV_TOOLS_BUILD_STATIC=OFF \
+    -DSPIRV-Headers_SOURCE_DIR=/usr/local
+RUN install-from-source https://download.savannah.nongnu.org/releases/libunwind/libunwind-1.6.2.tar.gz
+
+# We install ncurses rather than building it as it's used by many of the build tools and they don't like it when it changes
+RUN install-package ncurses
```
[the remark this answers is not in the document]
TrentHouliston: Put it where it's required, ncurses is only really needed at that point.
Could move it up if you want

docker/Dockerfile
```diff
@@ -76,8 +80,8 @@ RUN ln -s /usr/local/bin/install-from-source /usr/local/bin/install-header-from-
     && ln -s /usr/local/bin/install-from-source /usr/local/bin/install-from-source-with-patches
 
 # Generate toolchain files for generic
-COPY usr/local/toolchain/generate_generic_toolchain.py /usr/local/generate_toolchain.py
-RUN python /usr/local/generate_toolchain.py --prefix /usr
+RUN --mount=type=bind,source=usr/local/toolchain,target=/usr/local/toolchain \
```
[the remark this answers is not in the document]
TrentHouliston: It's the same thing, just a slightly different form!
Saves a layer too

docker/Dockerfile
TrentHouliston: Just needed a patch or two :P

CMurtagh-LGTM:
`ninja: error: 'nuclear/message/descriptor.pb: /home/nubots/NUbots/nuclear/message/proto/Vector.proto', needed by 'nuclear/message/protobuf/message/Geometry.pb.cc', missing and no known rule to make it`

docker/Dockerfile
```diff
@@ -91,117 +95,125 @@ RUN python /usr/local/generate_toolchain.py --prefix /usr
...
     --with-tcmalloc-pagesize=64 \
     --enable-minimal
 
 # Protobuf
-RUN install-package protobuf
-RUN install-autotools-from-source https://github.com/protocolbuffers/protobuf/releases/download/v3.17.3/protobuf-cpp-3.17.3.tar.gz \
+RUN install-package protobuf && \
+    install-autotools-from-source https://github.com/protocolbuffers/protobuf/releases/download/v21.4/protobuf-cpp-3.21.4.tar.gz \
     --with-zlib=/usr/local \
     --with-protoc=/usr/sbin/protoc
```
Bidski, line 212: With changes in `Neutron.cmake` to use DEPFILEs in `add_custom_command`s `ninja` will always think that the message protos are dirty and will always rebuild them.

The reason that `ninja` thinks they are dirty is because it treats the first specified output as the "primary" output and then gets upset that the primary output isnt the first output mentioned in the DEPFILE.

Unfortunately, `protobuf` uses and `std::unordered_map` to store output file names and so we are unable to rely on a consistent ordering of the outputs in the DEPFILE. This patch sorts the output files to give us that consistent order
```suggestion
RUN install-package protobuf && \
    install-from-source-with-patches cmake \
    https://github.com/protocolbuffers/protobuf/releases/download/v21.4/protobuf-cpp-3.21.4.tar.gz \
    https://gist.githubusercontent.com/TrentHouliston/f2e318ff0b96991a840778f10ac2dc63/raw/86bab8774fcb1f21b2bc8ed4fd35e9ecfddf3b5d/sort_manifest.patch \
    -- \
    -DWITH_PROTOC="/usr/sbin/protoc" \
    -Dprotobuf_BUILD_TESTS=OFF \
    -Dprotobuf_BUILD_LIBPROTOC=ON \
    -DBUILD_SHARED_LIBS=ON \
    -DZLIB_ROOT=${prefix}
```

Bidski (changes requested):

docker/Dockerfile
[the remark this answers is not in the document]
Bidski: Saving layers is important. There is a maximum layer limit of about 120 (IIRC), when you hit this limit your image will build successfully but the import of the image into docker will silently fail and it will leave you with the impression that everything finished successfully.

docker/Dockerfile
TrentHouliston: Suggesting I use my own patch!
I'll make it a proper PR then link to that commit for the patch

CMurtagh-LGTM:
I can't get valgrind to run on the robot on this version. After this PR is merged I'll have a go at another update.

https://bbs.archlinux.org/viewtopic.php?id=276422 :sob:

PR #911, opened 2022-07-29
Add configs specific to robocup for the system configurator.

CMurtagh-LGTM:
Add new files for robocup specific configs.
We can lower the number of the files for when we use robocup so they override the lab ones.

- [x] NUbook documentation https://github.com/NUbots/NUbook/pull/265
- [x] Test


TrentHouliston (commented):
For RoboCup you don't want to change the details of the bond network.
Leave the bonds the same so they're the same IP addresses as the lab for ethernet (that will still be part of the bond)
Instead change the details on the wifi interface to be the IP addresses you need for wifi (while keeping the lab ips for ethernet)

CMurtagh-LGTM:
Also fixes long boot time caused by the incorrect unit `wpa_supplicant@wlan0`.

module/tools/SystemConfiguration/data/system/default/etc/wpa_supplicant/wpa_supplicant-wlp58s0.conf
```diff
@@ -4,6 +4,12 @@ update_config=1
 fast_reauth=1
 ap_scan=1
 
+network={
+	ssid="NETGEAR89"
+	#psk="fancyriver224"
+	psk=6d1c66448797d8bcdb0ecca60f820d2d4be0e012eea7c3542388cfe0befcfa54
+    priority=3
+}
```
ysims, line 12: :monocle_face:

ysims: Won't this make the robots connect to the netgear router instead of epsilon-z normally?

CMurtagh-LGTM: Normally the router should be off, so it will fallback to the next priority :shrug:

ysims (approved):
Seems fine.

PR #914, opened 2022-07-31
Add an information source so the Director can communicate to the Behaviour DSL

TrentHouliston:
This adds an InformationSource class interface that lets the Behaviour DSL communicate with the Director so they can provide information about the task that is running

### DIFF
--- a/module/extension/Director/src/information/get_task_data.cpp
+++ b/module/extension/Director/src/information/get_task_data.cpp
@@ -14,7 +14,7 @@
  * You should have received a copy of the GNU General Public License
  * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
  *
- * Copyright 2021 NUbots <nubots@nubots.net>
+ * Copyright 2022 NUbots <nubots@nubots.net>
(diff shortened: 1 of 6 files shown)


module/extension/Director/src/information/get_task_data.cpp
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * The NUbots Codebase is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright 2021 NUbots <nubots@nubots.net>
```
ysims, line 17:
```suggestion
 * Copyright 2022 NUbots <nubots@nubots.net>
```

ysims (approved):

PR #915, opened 2022-07-31
Complete the basic director work without `Causing` dsl word

TrentHouliston:
Adds the code that is required to finish the director algorithm except for the Pushing part of the algorithm.

All features should work after this except for actions involving the Causing keyword

Note, this PR is being merged while not 100% working.
There is an unresolved issue regarding how Needs works with reserving tasks.
Currently if you emit a lower priority task and there is another task that needed it, it'll run the new lower priority task for a single cycle before replacing it with the higher priority other task that needed it.


CMakeLists.txt
```diff
@@ -92,8 +92,8 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS
     CACHE STRING "Enable/Disable output of compile commands during generation." FORCE
 )
 
-# Enable all warnings that we can
-add_compile_options(-Wall -Wpedantic -Wextra)
+# Enable warnings
+add_compile_options(-Wall -Wextra)
```
KipHamiltons, line 96: Ummmmmmmm

KipHamiltons (commented):
I glance and I see this ☹️

CMakeLists.txt
TrentHouliston: Apparently, the enum templates that I use are not allowed by pedantic and I don't have a specific rule I can turn off :(
Should work out what pedantic does and if any of them are actually important

KipHamiltons: Remove now pls 😇

module/extension/Director/src/director/challenge_priority.cpp
```diff
@@ -106,6 +107,7 @@ namespace module::extension {
         // This can happen when we are checking tasks with a `Needs` relationship against a task higher in the tree and
         // we already own that task. In that case we say that the challenger wins because we already own it and can
         // replace it.
+        // TLDR parent beats child
```
ysims, line 110: yikes

module/extension/Director/src/director/direct_priority.cpp
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * The NUbots Codebase is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright 2021 NUbots <nubots@nubots.net>
```
ysims, line 17:
```suggestion
 * Copyright 2022 NUbots <nubots@nubots.net>
```

docker/Dockerfile
```diff
@@ -249,7 +249,7 @@ RUN install-cmake-from-source https://github.com/libuv/libuv/archive/v1.42.0.tar
     -DBUILD_TESTING=OFF
 
 # NUClear!
-RUN install-from-source https://github.com/Fastcode/NUClear/archive/f4a60c7269e277a504635a603c2890fb1a70289d.tar.gz \
+RUN install-from-source https://github.com/Fastcode/NUClear/archive/304ff3b5e21b5bf7ea222f5cf1adca7622036f2a.tar.gz \
```
ysims, line 252: I can't find this commit on the repo (although it exists), what is the purpos (and also should we be trying to bring our NUClear version up to date?)

KipHamiltons: github says it's not on a branch 😳

module/extension/Director/src/Director.cpp
```diff
@@ -110,16 +118,23 @@ namespace module::extension {
...
                 throw std::runtime_error("You cannot use the 'When' DSL word with Start or Stop.");
             }
 
+            // If the initial read throws an exception then it starts as false
+            bool current = false;
+            try {
+                // This throws an exception when the state hasn't been emitted yet
+                current = when.validator(when.current());
+            }
+            catch (...) {
```
KipHamiltons, line 133: Empty catch everything feels a little dirty. Can we log the exception? is that worth doing?

module/extension/Director/src/Director.cpp
```diff
@@ -215,43 +229,89 @@ namespace module::extension {
...
         // the Provider has finished executing
-        on<Trigger<BehaviourTask>, Sync<BehaviourTask>>().then(
-            "Director Task",
-            [this](const std::shared_ptr<const BehaviourTask>& task) {
-                // Root level task, make the pack immediately and send it off to be executed as a root task
-                if (providers.count(task->requester_id) == 0) {
-                    emit(std::make_unique<TaskPack>(TaskPack({task})));
+        on<Trigger<BehaviourTask>, Sync<Director>>().then("Director Task", [this](const BehaviourTask& t) {
+            // Make our own mutable copy of the task
+            auto task = std::make_shared<BehaviourTask>(t);
```
KipHamiltons, line 234: something about this being a shared_ptr kind of feels awkward to me. two (probably oversimplistic, naive) things I think are:
- this is shared because the pack_builder maps to shared_ptr<behaviourtask>. Could the pack_builder own these? I think it could, because they're being transferred to being owned by the pack [here](https://github.com/NUbots/NUbots/blob/e834d2807364a2962d0bde8baae28ad93323fb9d/module/extension/Director/src/Director.cpp#L304). So I am thinking the pack_builder could map to unique_ptrs perhaps
- We're deref'ing the pointer a bunch here - would it be more efficient to create it on the stack then move/copy it to the heap when you add it to the pack_builder map? (I honestly don't know this one - just curious whether you think the derefs or copy/move would be more expensive)

I feel like this kind of unimportant, surface level stuff is the kind of stuff I'll be able to provide for this review haha

module/extension/Director/src/director/challenge_priority.cpp
```diff
@@ -58,23 +59,23 @@ namespace module::extension {
...
-        auto get_ancestor_priorities = [this](const std::shared_ptr<const BehaviourTask>& task) {
+        auto get_ancestor_priorities = [this](const std::shared_ptr<BehaviourTask>& task) {
             // We are our first ancestor
             std::vector<TaskPriority> ancestors;
 
             // The task might be a root task, in which case we won't have any provider
             // In that case we set the type to nullptr_t to indicate that we are a root task
             ancestors.emplace_back(
-                providers.count(task->requester_id) != 0 ? providers[task->requester_id]->type : typeid(nullptr_t),
+                providers.count(task->requester_id) != 0 ? providers.at(task->requester_id)->type : typeid(nullptr_t),
```
KipHamiltons, line 69: What's the rationale behind using `.at`? are we unconfident that the requester id is in providers? this seems odd

module/extension/Director/src/director/find_ok_solution.cpp
```diff
@@ -0,0 +1,103 @@
...
+                OkSolution(false, o.provider, std::move(o.requirements), std::move(o.used), std::move(blocking_groups));
+                return o;
+            }
+        }
+
+        // We couldn't find an OK solution, return blocked
+        return OkSolution(true, std::move(blocking_groups));
+    }
+
+    Director::OkSolution Director::find_ok_solution(const Solution::Option& option,
```
KipHamiltons, line 34: The name `OkSolution` is a meme

module/extension/Director/src/director/find_ok_solution.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+    std::vector<Director::OkSolution> Director::find_ok_solutions(const std::vector<Solution>& solutions) {
+
+        // Find each one individually but pass through the used types
+        std::vector<OkSolution> ok_solutions;
+        std::set<std::type_index> used_types;
+        for (const auto& solution : solutions) {
+
+            // Choose an option
+            ok_solutions.push_back(find_ok_solution(solution, used_types));
```
KipHamiltons, line 94: This is hurting my brain a bit. So we have local set of `used_types`, which accumulates the used_types from the ok_solutions, but then we pass that set into the find_ok_solution function for the next call? This seems very odd (but it might be deliberate).

KipHamiltons (commented):
I read some of the code 🚀

docker/Dockerfile
TrentHouliston: It's main, just one commit ahead on main from what you have. The bug I fixed doesn't influence anything else so updating isn't a matter of urgency (except for this)

ysims: https://github.com/Fastcode/NUClear/commit/b36b8731af9bde5742ece832d0a3c0f317ac6e6e
this is the next commit after on main?
If you got to it https://github.com/Fastcode/NUClear/commit/304ff3b5e21b5bf7ea222f5cf1adca7622036f2a
"This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."

TrentHouliston: Hmm, was supposed to include this PR, maybe I got the hash wrong https://github.com/Fastcode/NUClear/pull/57

TrentHouliston: Seems I picked a commit on that PR rather than the one on main after it merged

module/extension/Director/src/director/challenge_priority.cpp
TrentHouliston: I thought the same thing as I wrote the comment :P

module/extension/Director/src/Director.cpp
TrentHouliston: The purpose of this is that if you are trying to check a when condition and the thing you're checking doesn't exist you assume you can't run.

module/extension/Director/src/Director.cpp
TrentHouliston: It might be able to be made into a unique pointer, but it would influence a lot of places that would have to become raw pointers which I'm not a fan of... (e.g. watchers)

module/extension/Director/src/director/challenge_priority.cpp
TrentHouliston: In a map, if you use `operator[]` if the object doesn't exist it'll be created but when using `.at` it will just error if it doesn't exist.
For groups since there is no default constructor you can't even use `operator[]` and since I want to control when I am creating instances I use `.at`.

module/extension/Director/src/director/find_ok_solution.cpp
TrentHouliston: Think I should change it to `RunnableSolution`?
I wasn't super keen on it either...

module/extension/Director/src/director/find_ok_solution.cpp
TrentHouliston: So as you go down the options, you use types that can't be used again. However, when you are traversing down multiple options each option isn't going to be interfering with each other so they each get a copy of the current state.
When you've selected an option it gets passed down to the sibling solutions to limit them.

Yes it does hurt the brain a little :P

ysims (approved):
Seems good

module/extension/Director/src/Director.cpp
KipHamiltons: Raw pointers, or, references to the unique ptr's data? same same, but different - stealing the idea of using refs to unique ptr data from you 😅

module/extension/Director/src/director/find_ok_solution.cpp
KipHamiltons: I like that? Like, ok solution is _ok_ 😝 it just is a bit funny hahaha

module/extension/Director/src/director/find_ok_solution.cpp
KipHamiltons: Yep, my brain exploded

module/extension/Director/src/director/challenge_priority.cpp
```diff
@@ -58,23 +59,23 @@ namespace module::extension {
         }
 
         // Function to get the priorities of the ancestors of this task
-        auto get_ancestor_priorities = [this](const std::shared_ptr<const BehaviourTask>& task) {
+        auto get_ancestor_priorities = [this](const std::shared_ptr<BehaviourTask>& task) {
             // We are our first ancestor
             std::vector<TaskPriority> ancestors;
```
Bidski, line 64: Is this comment implying that this vector should be initialised with a single value?

module/extension/Director/src/director/challenge_priority.cpp
Bidski, line 69: Use `contains`

module/extension/Director/src/director/challenge_priority.cpp
```diff
@@ -58,23 +59,23 @@ namespace module::extension {
...
 
-            // Loop up through the providers until we reach a point where a task was emitted by a non provider
-            for (auto t = task; providers.count(t->requester_id) != 0;) {
+            // Loop up through the providers until we reach a point where a task was emitted by a root provider
+            for (auto t = task; providers.at(t->requester_id)->classification != Provider::Classification::ROOT;) {
 
                 // Get the provider that emitted this task, and from that the provider group
-                auto provider = providers[t->requester_id];
-                auto& group   = groups[provider->type];
+                auto provider = providers.at(t->requester_id);
```
Bidski, line 77: Or is there a reason we want to copy this?
```suggestion
                auto& provider = providers.at(t->requester_id);
```

module/extension/Director/src/director/challenge_priority.cpp
```diff
@@ -58,23 +59,23 @@ namespace module::extension {
...
+                auto& group   = provider->group;
 
                 // If there is no active task something has gone wrong with the algorithm
                 if (group.active_task == nullptr) {
```
Bidski, line 81: Is there any extra info we can add to this throw to make it more informative to the user? Or are we so broken at this point that there isnt any point?

module/extension/Director/src/director/direct_priority.cpp
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright 2022 NUbots <nubots@nubots.net>
+ */
+
+#include <algorithm>
+#include <tuple>
+#include <vector>
```
Bidski, line 22:
```suggestion
#include <memory>
```

module/extension/Director/src/director/find_ok_solution.cpp
Bidski: `Acceptable`? `Adequate`? `Usable`? `Valid`? `Mediocre`?

module/extension/Director/src/director/reevaluate_group.cpp
```diff
@@ -0,0 +1,92 @@
...
+        // Go through each of the watchers and try to re-run them and their siblings
+        for (const auto& t : watchers) {
+
+            // Check if we have the priority to override the currently active task
+            if (t != group.active_task && challenge_priority(group.active_task, t)) {
+                // Find the provider group that requested this task and reevaluate it
+                // Maybe it's blocked on something else and that's why it's not our active task
+                auto p  = providers.at(t->requester_id);
+                auto& g = p->group;
+                reevaluate_group(g);
```
Bidski, line 48: I assume there is a reason for not doing this?
```suggestion
                reevaluate_group(providers.at(t->requester_id)->group);
```

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -14,15 +14,244 @@
...
+        for (const auto& task : tasks) {
+            solutions.push_back(solve_task(task));
+        }
+
+        if (run_level >= RunLevel::OK) {
+            // Try to find an ok solution for the pack
+            auto ok_solutions = find_ok_solutions(solutions);
+
+            // Loop through the solutions and add watches for the things we want
+            for (int i = 0; i < int(tasks.size()); i++) {
```
Bidski, line 46:
```suggestion
            for (int i = 0; i < int(tasks.size()); ++i) {
```

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -14,15 +14,244 @@
...
+        for (const auto& t : pack.second) {
+            if (t->type == typeid(::extension::behaviour::Done)) {
+
+                auto parent_provider = providers.at(group.active_task->requester_id);
+                auto& parent_group   = parent_provider->group;
+
+                // If it's a root provider, then we just remove the task
+                if (parent_provider->classification == Provider::Classification::ROOT) {
+                    std::cout << "Removing " << parent_group.subtasks.front() << " from "
+                              << NUClear::util::demangle(parent_group.type.name()) << std::endl;
```
Bidski, line 159: `log`?

Bidski (changes requested):
Partial review for now

module/extension/Director/src/Director.cpp
TrentHouliston: There is one on line 128

module/extension/Director/src/Director.cpp
TrentHouliston: Raw pointers I think, or std::ref objects which I'm not a fan of... Because you can't have an `std::vector` of references

module/extension/Director/src/Director.hpp
```diff
@@ -183,38 +220,74 @@ namespace module::extension {
...
          *
-         * @param task      the task we are finding solutions for
+         * @param task      the task we are finding solutions for, also used for authority checks
          *
          * @return the set of possible solution options for this task
          */
-        Solution solve_task(const std::shared_ptr<const ::extension::behaviour::commands::BehaviourTask>& task);
+        Solution solve_task(const std::shared_ptr<BehaviourTask>& task);
+
+        struct OkSolution {
```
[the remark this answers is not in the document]
TrentHouliston: It sure is... I'll go add some

module/extension/Director/src/director/challenge_priority.cpp
TrentHouliston: It once upon a time was... and now this comment is wrong

module/extension/Director/src/director/challenge_priority.cpp
TrentHouliston: It's a shared pointer, I generally take those by copy even when I could take them by reference

module/extension/Director/src/Director.cpp
```diff
@@ -52,14 +53,22 @@ namespace module::extension {
 
     void Director::add_provider(const ProvideReaction& provide) {
 
-        auto& group = groups[provide.type];
+        // Create if it doesn't already exist
+        if (groups.count(provide.type) == 0) {
```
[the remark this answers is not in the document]
TrentHouliston: Replaced all `.count` with `.contains`

module/extension/Director/src/Director.cpp
TrentHouliston: As for your comment about dereferencing lots of dereferencing vs one is not really any more expensive. Any half decent optimiser will just do it once to get the memory and work with that memory from there

module/extension/Director/src/director/challenge_priority.cpp
TrentHouliston: If we reach this, it's not a user error it's a bug in the algorithm so I think this is fine as it gives you a place to start debugging

module/extension/Director/src/director/reevaluate_group.cpp
TrentHouliston: nope

module/extension/Director/src/director/run_task_pack.cpp
TrentHouliston: This shouldn't be here
I should add logs at some point though

module/extension/Director/src/Director.cpp
TrentHouliston: Wouldn't even matter if it weren't created here, since it's not declared as volatile the compiler can just assume the value stays the same the entire time. It actually can cause a class of bug when multithreading

module/extension/Director/src/provider/ProviderGroup.hpp
```diff
@@ -30,21 +31,70 @@ namespace module::extension::provider {
...
+        using TaskList = std::vector<std::shared_ptr<BehaviourTask>>;
+
+        ProviderGroup(const std::type_index& type_) : type(type_) {}
+
+        struct WatchHandle {
+            WatchHandle(const std::function<void()>& deleter_) : deleter(deleter_) {}
+
+            WatchHandle(WatchHandle&& other) {
+                deleter       = other.deleter;
+                other.deleter = std::function<void()>();
```
Bidski, line 44:
```suggestion
                deleter = std::exchange(other.deleter, std::function<void()>());
```

module/extension/Director/src/provider/ProviderGroup.hpp
```diff
@@ -30,21 +31,70 @@ namespace module::extension::provider {
...
+            }
+
+            WatchHandle& operator=(WatchHandle&& other) {
+                deleter       = other.deleter;
+                other.deleter = std::function<void()>();
+                return *this;
```
Bidski, line 50:
```suggestion
                if (this != &other) { 
                    deleter = std::exchange(other.deleter, std::function<void()>());
                }
                return *this;
```

module/extension/Director/src/Director.hpp
```diff
@@ -279,14 +285,73 @@ namespace module::extension {
...
+            /// We can run all our tasks on providers now
+            OK,
+            /// We can't run all our tasks on providers now, but we can push to fix our requirements
+            PUSH,
+            /// We can't run all our tasks on providers now, and we can't push to fix our requirements
+            BLOCKED
+        };
+
+        /**
+         * Tries to passed tasks in the pack, but only up to the passed run level.
```
Bidski, line 343: Want to try some english in this sentence?

Bidski (changes requested):
Another partial review

Bidski (dismissed):
Cant see anything too offensive

ysims (dismissed):
lets fix main!

ysims (approved):

PR #917, opened 2022-08-01
Remove Darwin Global Config and Update Default for 2022

ysims:
Removes the Darwin config, since it's not really needed.

Updates the default team ID to the team ID for 2022. This will be useful during practice games until next competition.


KipHamiltons (approved):
Seems fine to me 🤷‍♂️

PR #920, opened 2022-08-06
Removed right camera from visual mesh

PiiDish:
- Removing because of computational speed
- Remove right eye in visual mesh


ysims (approved):
should be good

TrentHouliston (commented):
Why not just reduce concurrency and keep both eyes?

TrentHouliston:
You're not going to get a frame rate increase by removing an eye. It's just all the frame rate is going to be on that eye rather than split over two

PR #921, opened 2022-08-06
Added MainThread to 'ON' statements

PiiDish:
Main Thread - <on.....


module/motion/QuinticWalk/src/QuinticWalk.cpp
```diff
@@ -124,7 +124,7 @@ namespace module::motion {
             }
         });
 
-        on<Configuration>("QuinticWalk.yaml").then([this](const Configuration& cfg) {
+        on<Configuration, MainThread>("QuinticWalk.yaml").then([this](const Configuration& cfg) {
```
Bidski, line 127: Adding `MainThread` to `on<Configuration>` should be a pointless and unneeded operation, pretty sure you don't need it for `on<Startup>` either

Bidski (changes requested):

TrentHouliston (commented):
You should only need this on the reactions that are time sensitive. So the update on the walk, and the hardware io ones that work with sensors. Maybe subsumption too. Basically the critical path. The rest of these will do nothing except make things slower that are important. 

If you overload the main thread with too much stuff you'll be back to the exact same problem as you started with

Bidski (approved):

ysims:
Not really needed atm, not having problems with walk. Also this is out of date in that we don't use this module anymore.

PR #924, opened 2022-08-09
Adds extra camera configs

Crafty15:
- Cameras kept developing a pink hue over time. Our best guess at the moment is that it's due to the constant auto white balancing getting confused by so much green. 
- Setting `BalanceWhiteAuto: Once` makes the cameras only adjust on start up. Having something neutral for them to look at while they adjust seems to work well. 
- Manual adjustment is also possible using more settings which are to be listed on a NUbook page.  
- Frame rate has been capped at 20fps as it was found to have the least amount of flickering from lab lighting.


Jesse27 (commented):
LGTM!

module/input/Camera/data/config/Cameras/Left.yaml
```diff
@@ -6,6 +6,16 @@ settings:
...
+  GammaEnable: true # Set to false to adjust gamma. Default = true
+  Gamma: 0.50
+  BalanceWhiteAuto: Once # Options: Off, Once, Continuous
+  BalanceRatioSelector: Red # Options: Red or Blue. Controls the balance ratio of the selected color relative to green. Used for white balancing.
+  BalanceRatio: 2.56
+  AutoExposureEVCompensation: 0 # Range from -3(darker) to 3(lighter) with a step of 1/3. Default = 0.
+  AcquisitionFrameRateEnable: true
+  AcquisitionFrameRate: 20.00
+  SaturationEnable: false
+  Saturation: 0
```
ysims, line 21: The majority of these seem to not be in use?

Crafty15: Yeah, i left them in in case we wanted to try manual tuning.

ysims: Maybe comment out the ones we don't use?

TrentHouliston: There are going to be hundreds of settings you don't use, that said it might be a good idea to use arv-tool to dump them all out so you can decide what's important

Crafty15: Yeah it would be good to document. I'll put that on the list.

module/input/Camera/data/config/Cameras/Left.yaml
```diff
@@ -6,6 +6,16 @@ settings:
...
+  GainAuto: Continuous # Options: Off, Once, Continuous
+  Gain: 15 # Range: 0.00 - 18.01'
+  GammaEnable: true # Set to false to adjust gamma. Default = true
+  Gamma: 0.50
+  BalanceWhiteAuto: Once # Options: Off, Once, Continuous
+  BalanceRatioSelector: Red # Options: Red or Blue. Controls the balance ratio of the selected color relative to green. Used for white balancing.
+  BalanceRatio: 2.56
+  AutoExposureEVCompensation: 0 # Range from -3(darker) to 3(lighter) with a step of 1/3. Default = 0.
+  AcquisitionFrameRateEnable: true
+  AcquisitionFrameRate: 20.00
```
ysims, line 19: Might be good to have a quick note here about this helping with flickering

TrentHouliston: What kind of flickering were you getting?

Crafty15: Was kind of pulsating from the fluros in the lab. One of our photography minded members suggested adjusting the frame rate to suit the grid frequency. I played around with it a bit and this seemed like a sweet spot.

ysims:
Would be good to have a link to the docs for the parameters unless there isn't a webpage for it? And make an issue in NUbook for it to be documented.

ysims (approved):
LGTM

PR #935, opened 2022-08-21
Remove 'enum' specifier from template declarations which caused warnings

KipHamiltons:
This PR gets rid of the `enum template` warnings by removing the `enum` specifier. :man_shrugging:


TrentHouliston (changes requested):
I refuse!
Plz remove -Wpedantic instead!

TrentHouliston:
Actually... I approve so long as this still enforces that you can only put a value from the enum in there

CMurtagh-LGTM:
I agree with Trent here, we shouldn't religiously stick with ISO c++ when we have the option for neater looking code.

KipHamiltons:
> Actually... I approve so long as this still enforces that you can only put a value from the enum in there

That's what it does ;) https://godbolt.org/z/d73rc3cdb

TrentHouliston (approved):

PR #936, opened 2022-08-22
Fix the video encoder

TrentHouliston:
It had a bug


tools/nbs/video.py
```diff
@@ -51,7 +50,7 @@ def packetise_stream(decoder):
 
     for packet in decoder:
         # Check for compressed images
-        if packet.type in ("message.output.CompressedImage", "message.input.Image"):
+        if packet.type.name in ("message.output.CompressedImage", "message.input.Image"):
```
ysims, line 53: This part doesn't work, I tried now with removing just `bytes_read` and it works. My issue last night was OOM, probably because my gpu was busy with other things.

ysims:
```suggestion
        if packet.type in ("message.output.CompressedImage", "message.input.Image"):
```

ysims (approved):
lgtm

PR #937, opened 2022-08-24
Remove wants to kick from walk path planner

ysims:
`robocup` on `main` is broken, the robot won't walk because SimpleWalkPathPlanner expects a `WantsToKick` message and it never gets one. It came from KickPlanner, but we don't use it anymore and it was removed from the role at some point.


CMurtagh-LGTM (approved):
LGTM

PR #938, opened 2022-09-01
Add react three fiber

BrendanAnnable:
The experimental beginnings of using react three fiber as an alternative way to express webgl scene graphs in a react-style declarative syntax.

Likely better in every way to our current framework, except perhaps performance (TBD). More exploration needed.

Next step is to try porting vision and then revaluate.


nusight2/src/client/components/storybook/fullscreen.tsx
```diff
@@ -9,7 +9,7 @@ export function fullscreen(story: () => JSX.Element) {
         // This is an inline <style> and not in `fullscreen.css` as that would make it apply globally
         // to all stories, due to the way we build and extract the styles from CSS files.
       }
-      <style>{'body { margin: 0; padding: 0; }'}</style>
+      <style>{'body { margin: 0 !important; padding: 0 !important; }'}</style>
```
JosephusPaye, line 12: This should be fixed in https://github.com/NUbots/NUbots/pull/942

BrendanAnnable:
Replaced by #1115

PR #942, opened 2022-09-05
Fix `fullscreen` decorator in NUsight storybook

JosephusPaye:
This PR fixes an issue with the fullscreen Storybook decorator that causes scrollbars to show unnecessarily.

The issue is borne out of Storybook adding extra padding to stories by default. This extra padding combined with the fullscreen decorator (which is set to be as large as the screen), causes the page to grow larger than the screen, resulting in scrollbars.

The fix is to the Storybook [`layout` parameter](https://storybook.js.org/docs/react/configure/story-layout) to remove the extra padding. The fullscreen wrapper's style is also inlined for simplicity.

<details>
<summary>Before</summary>

screenshot showing scrollbars on a fullscreen story

</details>

<details>
<summary>After</summary>

image

</details>


TrentHouliston (approved):

nusight2/src/client/components/storybook/fullscreen.tsx
```diff
@@ -1,16 +1,5 @@
...
-    <div className={style.fullscreen}>
-      {
-        // This is an inline <style> and not in `fullscreen.css` as that would make it apply globally
-        // to all stories, due to the way we build and extract the styles from CSS files.
-      }
-      <style>{'body { margin: 0; padding: 0; }'}</style>
-      {story()}
-    </div>
-  )
+  return <div style={{ width: '100vw', height: '100vh' }}>{story()}</div>
```
BrendanAnnable, line 4: Is there an example of a story which doesn't work without this decorator? Can we remove the decorator altogether?

JosephusPaye: Removing the `fullscreen` decorator breaks these stories (no image):

- vision objects
- vision grid_layout (both static and animated)

I also initially had trouble with the animated three stories (nugus and darwin model, three animated scene) - it would render into a small rectangle in the middle of the screen. Now it flashes like that for a moment and then goes fullscreen.

LachlanCourt (dismissed):

nusight2/src/client/components/storybook/fullscreen.tsx
JosephusPaye: Gonna merge now. Happy to follow up with improvements.

LachlanCourt (approved):

PR #943, opened 2022-09-05
Add a story for the CheckboxTree component in NUsight

JosephusPaye:
I think the CheckboxTree component was created before Storybook, and currently has no stories. This PR adds a story that demonstrates how it's used.

Also improves label styling to enable ellipsis truncation when the label text is longer than the available space.

### DIFF
--- a/nusight2/src/client/components/checkbox_tree/checkbox_tree.stories.tsx
+++ b/nusight2/src/client/components/checkbox_tree/checkbox_tree.stories.tsx
@@ -1,7 +1,6 @@
 import React from 'react'
 import { storiesOf } from '@storybook/react'
 import { action, computed, observable } from 'mobx'
-import { observer } from 'mobx-react'
 import { createTransformer } from 'mobx-utils'
 
 import { CheckedState, TreeModel, TreeNodeModel } from './model'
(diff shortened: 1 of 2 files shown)


TrentHouliston (dismissed):
Who can say no to good stories?

nusight2/src/client/components/checkbox_tree/checkbox_tree.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const Story = observer(() => (
+      <CheckboxTree
+        model={model}
+        onCheck={onNodeCheck}
+        onExpand={onNodeExpand}
+        renderLabel={renderLabel}
+      />
+    ))
+
+    return <Story />
```
BrendanAnnable, line 152: No observables are accessed in this observer component, so we can simplify

```suggestion
  return <CheckboxTree
      model={model}
      onCheck={onNodeCheck}
      onExpand={onNodeExpand}
      renderLabel={renderLabel}
    />
```

BrendanAnnable (approved):

TrentHouliston (approved):

PR #945, opened 2022-09-05
Switch to `createImageBitmap` for decoding JPEG images in NUsight

JosephusPaye:
This PR adds the ability to pass [`ImageBitmap`](https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmap) as a image texture for rendering, and switches vision to use [`createImageBitmap()`](https://developer.mozilla.org/en-US/docs/Web/API/createImageBitmap) when decoding JPEGs for rendering, instead of using a DOM `` element.

Additionally, this will allow us to move the decoding and rendering code into a webworker with an offscreen canvas (where we can't use an `` element), if necessary in the future.

Anecdotally, it seems a tiny bit faster at decoding, probably because using `createImageBitmap()` directly removes the overhead of going through the DOM.


nusight2/src/client/components/vision/image.ts
```diff
@@ -18,6 +18,11 @@ export interface ElementImage extends BaseImage {
   element: HTMLImageElement
 }
 
+export interface BitmapImage extends BaseImage {
+  type: 'bitmap'
+  bitmap: ImageBitmap
```
BrendanAnnable, line 23: Perhaps we should rename (in a separte PR?) `data` + `element` + `bitmap` to `image` so that we don't have to duplicate code e.g. needing to add basically a duplicate method `bitmapTexture` below.

BrendanAnnable (approved):
LGTM

nusight2/src/client/components/vision/image.ts
JosephusPaye: Done. Tried to merge all the `...Texture` methods but the data one needs to call `get()` on its image since the image is a computed. So I've merged the element and bitmap ones, keeping data separate.

TrentHouliston (dismissed):

nusight2/src/client/components/vision/image.ts
```diff
@@ -10,12 +10,17 @@ interface BaseImage {
...
+  image: IComputedValue<Uint8Array>
 }
 
 export interface ElementImage extends BaseImage {
   type: 'element'
-  element: HTMLImageElement
+  image: HTMLImageElement
+}
+
+export interface BitmapImage extends BaseImage {
```
BrendanAnnable, line 21: Would be good to collapse `ElementImage` and `BitmapImage` into one type if we can

BrendanAnnable (dismissed):

nusight2/src/client/components/vision/image.ts
JosephusPaye: I've unified them here: https://github.com/NUbots/NUbots/pull/945/commits/48885d43d9baf8a21328b04be62fda34bb1fb83c

`ElementOrBitmapName` was the best name I could come up with for the combined type. Happy for suggestions.

PR #949, opened 2022-09-14
Add Tailwind to NUsight

JosephusPaye:
This PR adds [Tailwind CSS](https://tailwindcss.com) to NUsight and Storybook.

If merged, other UI-level PRs that depend on Tailwind can follow, like the new nbs playback scrubber.


nusight2/src/client/components/app/style.module.css
```diff
@@ -1,11 +1,3 @@
...
-
-html,
-body {
-  margin: 0;
-  padding: 0;
-}
-
 .app {
   font: 14px 'Open Sans', Helvetica, Arial, sans-serif;
   min-height: 100vh;
```
JosephusPaye, line 3: This is for the `font-weight` line below, can't comment directly on that.

Would it make sense to remove that (in a separate PR if necessary)? The 300 font weight reduces the text contrast a fair bit, especially for small sized text. Another issue is that it's not in Storybook, so the components are developed with a normal 400 weight but get displayed at 300 in NUsight.

Screenshot of before and after

LachlanCourt (approved):
LGTM

PR #951, opened 2022-09-21
Adds left and right goalie dive scripts

Crafty15:
This was mainly for quick a demonstration but they turned out pretty good so i think they are worth keeping.
Fall is pretty soft as it does more of a fall than a dive. 
Also set gains to 0 for the final frame.


ysims (approved):
LGTM!

PR #953, opened 2022-09-22
Fix the extract images nbs tool

ysims:
The extract images tool doesn't debayer the image. This PR changes it so it does. The tool also outputs lens data as json. This has been changed so it works as-is with the Visual Mesh.
Edit: Further change has been made to change images in RGB format to BGR so OpenCV saves it properly. This is applicable for images from Webots.


requirements.txt
```diff
@@ -7,3 +7,4 @@ si-prefix
 numpy
 argcomplete
 honcho
+wand
```
Bidski, line 10: What is this for?

tools/nbs/extract_images.py
```diff
@@ -19,6 +26,83 @@ def register(command):
...
+        img = cv2.cvtColor(img, cv2.COLOR_BayerGB2RGB_VNG)
+
+    # Save the image
+    file_name = os.path.join(
+        output,
+        "{}_{:012d}.jpg".format(packet.msg.name, int(packet.msg.timestamp.seconds * 1e9 + packet.msg.timestamp.nanos)),
+    )
+    cv2.imwrite(file_name, img)
+    # Hacky! Image saved above has strange behaviour
+    Image(filename=file_name).convert("jpg").save(filename=file_name)
```
Bidski, line 53: This seems like a strange thing to do. What is wrong with the opencv saved image?

tools/nbs/extract_images.py
```diff
@@ -19,6 +26,83 @@ def register(command):
...
+    rCOc = np.array([-Hco.x.t, -Hco.y.t, -Hco.z.t])
+    Roc = np.array([[Hco.x.x, Hco.y.x, Hco.z.x], [Hco.x.y, Hco.y.y, Hco.z.y], [Hco.x.z, Hco.y.z, Hco.z.z]])
+    rCOo = np.matmul(Roc, rCOc).tolist()
+
+    Hoc = [
+        [Roc[0][0].item(), Roc[1][0].item(), Roc[2][0].item(), rCOo[0]],
+        [Roc[0][1].item(), Roc[1][1].item(), Roc[2][1].item(), rCOo[1]],
+        [Roc[0][2].item(), Roc[1][2].item(), Roc[2][2].item(), rCOo[2]],
+        [0, 0, 0, 1],
+    ]
```
Bidski, line 87: ?
```suggestion
    Hoc = np.linalg.inv(packet.msg.Hcw)
```

Bidski (changes requested):

tools/nbs/extract_images.py
ysims: The image it saves is somehow broken, nothing wants to read it (most things think the file is empty).

Bidski: Can you provide some examples?

ysims: I've just tried to get an image (from the same recording) and it's now fine :shrug:

tools/nbs/extract_images.py
```diff
@@ -4,10 +4,16 @@
 import os
 import types
 
+import cv2
```
LachlanCourt, line 7: Should add these to requirements

tools/nbs/extract_images.py
```diff
@@ -4,10 +4,16 @@
...
+import numpy as np
+import tensorflow as tf
```
CMurtagh-LGTM, line 9:
```suggestion
```
This isn't used.

tools/nbs/extract_images.py
```diff
@@ -19,6 +25,82 @@ def register(command):
     command.add_argument("--output", "-o", nargs="?", default=os.getcwd(), help="The folder to extract the images into")
 
 
+# Gets the image and debayers if needed, then saves the image
+def process_image(packet, output):
```
CMurtagh-LGTM, line 29:
```suggestion
def process_save_image(packet, output):
```
I think this describes the function better.

tools/nbs/extract_images.py
```diff
@@ -29,42 +111,7 @@ def run(files, output, **kwargs):
...
-                        "centre": [0, 0],
-                        "fov": packet.msg.lens.fov,
-                    },
-                },
-                f,
-                indent=4,
-                sort_keys=True,
-            )
+        # Get the image and save
+        process_image(packet, output)
```
ysims, line 115:
```suggestion
        process_save_image(packet, output)
```

Bidski (approved):
Not over-joyed that tensorflow is being used (transitively) in this, but I get that this was done in an older PR (you are a bad man @TrentHouliston), but otherwise this seems fine

PR #958, opened 2022-09-25
Changes default cmake settings from debug to release

ysims:
Most of the time we want to be in release. In debug, things are much slower. It gets annoying having to change it every time we `configure --clean`, and sometimes it is forgotten... :grimacing:


Crafty15 (changes requested):
TMLG

CMakeLists.txt
```diff
@@ -32,13 +32,13 @@ set(CI_BUILD
...
-  message(STATUS "Setting build type to 'Debug' as none was specified.")
+  message(STATUS "Setting build type to 'Release' as none was specified.")
   set(CMAKE_BUILD_TYPE
-      Debug
+      Release
       CACHE STRING "Choose the type of build." FORCE
   )
   # Set the possible values of build type for cmake-gui
-  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
+  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Release" "Debug" "MinSizeRel" "RelWithDebInfo")
```
Bidski, line 41: This change seems irrelevant to me as the ordering of these arguments are unimportant to the default value

Bidski (changes requested):

CMakeLists.txt
ysims:
```suggestion
  set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
```

Bidski (approved):
Also, adding `-DCMAKE_BUILD_TYPE=Release` to these lines might be an idea too? [https://github.com/NUbots/NUbots/blob/sims/default-release/tools/configure.py#L38-L43](https://github.com/NUbots/NUbots/blob/sims/default-release/tools/configure.py#L38-L43)

Otherwise this is fine (i.e. the above is only a suggestion and not a requirement)

PR #959, opened 2022-09-26
Tune walk

Tom0Brien:
This PR adds a tune for the walk on real robot.

Update - L.Craft
- Both robots seem to be reacting well to speed changes. So be aware that when using keyboard walk, a setting of 0.100 will barely move, and a setting of 2,000-ish will take off like a flash.
- Might not play to well if the start speed is too high when walking autonomously. 
- Works fine with both types of hips.


LachlanCourt (dismissed):

ysims (changes requested):
Can we double check these particularly on n3 since it seemed to not work the other day

Tom0Brien:
Yeah will need do.

Crafty15 (dismissed):
LGTM

ysims:
I am curious about the walk command - the walk engine does have an upper bound (walk command changes step length) so at some point making it higher will do nothing new and you'll get a warning from the walk engine saying something about xy being too big and it's truncating to the maximum step length. At what point does this happen?

ysims (approved):
lgtm

PR #966, opened 2022-10-14
Legge/wavreader module & role

thomaslegge98:
This pull request creates a wav reader module inside module/input. It will take in a test wav file and output the string of its data. The intention of this module is to be used in a future module named "SpeechRecognition". The speech rec module will receive the wavreader's data to carry out future speech recognition.


thomaslegge98 (commented):
Working Dockerfile

Bidski:
There are a lot of changes in this PR that should not exist in this PR

Bidski:
Also, if all you want to do is read the entire file into a vector, we have a [utility](https://github.com/NUbots/NUbots/blob/main/shared/utility/file/fileutil.cpp#L48-L68) for that

docker/Dockerfile
```diff
@@ -352,3 +352,9 @@ RUN chmod 600 /home/nubots/.ssh/id_rsa /home/nubots/.ssh/config
 RUN install -d -m 0755 /home/nubots/NUbots
 RUN install -d -m 0777 /home/nubots/build
 WORKDIR /home/nubots/NUbots
+
+#Speech Rec
+RUN sudo chmod a+rw /opt
+RUN curl -L https://raw.githubusercontent.com/NUbots/NUbots/legge/speech-rec-module-and-role/doc/ArchInstall/arch-speechrec.sh -o /opt/arch-speechrec.sh
+RUN sudo chmod a+rw /opt/arch-speechrec.sh
+RUN /bin/bash /opt/arch-speechrec.sh
```
Bidski, line 360: Modify your script so that it installs to `/usr/local` (rather than `/opt/kaldi`), there are examples of how we build and package things in [docker/usr/local/package](https://github.com/NUbots/NUbots/tree/main/docker/usr/local/package). 

If you install it into `/usr/local` then it will be installed on the robot when you run `./b install` (along with all of the usual optimisations we apply) and you wont need to run the script separately on the robot.

We also already build OpenBLAS for the robots [here](https://github.com/NUbots/NUbots/blob/164974f75c85a3176bcda40dce01d93ccb03bd05/docker/Dockerfile#L178-L181), you should prefer to use this over the one provided by `kaldi` if possible

Bidski (changes requested):

CMurtagh-LGTM:
There's a bunch of undefined reference errors during linking to names in `fst` and `kaldi`, I conjecture that we need a couple more find files.

model/README
```diff
@@ -0,0 +1,9 @@
+US English model for mobile Vosk applications
+
+Copyright 2020 Alpha Cephei Inc
```
CMurtagh-LGTM, line 3: What is the licence for this?

module/sound/SpeechRecognition/src/SpeechRecognition.cpp
```diff
@@ -0,0 +1,39 @@
...
+        : Reactor(std::move(environment)), config{} {
+
+        on<Configuration>("SpeechRecognition.yaml").then([this](const Configuration& cfg) {
+            // Use configuration here from file SpeechRecognition.yaml
+            this->log_level = cfg["log_level"].as<NUClear::LogLevel>();
+        });
+
+        // This reactor receives audiodata from emission
+        on<Trigger<AudioData>>().then([this](const AudioData& audioData) {
+            VoskModel* model           = vosk_model_new("/home/nubots/NUbots/model");
```
CMurtagh-LGTM, line 25: Move the model to somewhere in the data folder `module/sound/SpeechRecognition/data`.

Bidski: also make the path to the model a config value

.vscode/settings.json
```diff
@@ -23,7 +23,12 @@
     "yaml.format.printWidth": 120,
     "files.associations": {
         "*.role": "cmake",
-        ".Doxyfile": "properties"
+        ".Doxyfile": "properties",
+        "*.ipp": "cpp",
+        "string_view": "cpp",
+        "regex": "cpp",
+        "array": "cpp",
+        "string": "cpp"
```
ysims, line 31: Revert changes

model/ivector/online_cmvn.conf
```diff
@@ -0,0 +1 @@
+# configuration file for apply-cmvn-online, used in the script ../local/run_online_decoding.sh
```
ysims, line 1: Don't forget to move this into the `SpeechRecognition` data director. Do we need all of the files inside `model`, can we take just a subset?

module/sound/SpeechRecognition/README.md
```diff
@@ -0,0 +1,16 @@
+SpeechRecognition
+=================
+
+## Description
+This module takes in English speech as a wav file and converts it into text. It uses vosk-api to decipher the audio into text.
```
ysims, line 5:
```suggestion

This module takes in English speech as binary data and converts it into text. It uses vosk-api to decipher the audio into text.
```
Newline for consistency with other readmes. This module takes binary audio data, it isn't specific to the WavReader module.

module/sound/SpeechRecognition/README.md
```diff
@@ -0,0 +1,16 @@
+SpeechRecognition
+=================
```
ysims, line 2:
```suggestion
# SpeechRecognition
```

module/sound/SpeechRecognition/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Description
+This module takes in English speech as a wav file and converts it into text. It uses vosk-api to decipher the audio into text.
+
+## Usage
+
+## Consumes
+Uses an Audio Data message.
```
ysims, line 10:
```suggestion

- message::input::AudioData with the audio data to convert to text
```

module/sound/SpeechRecognition/src/SpeechRecognition.cpp
```diff
@@ -0,0 +1,39 @@
...
+            VoskRecognizer* recognizer = vosk_recognizer_new(model, 16000.0);
```
ysims, line 26: The sample rate shouldn't be hardcoded - it's probably something we should get from the AudioData message. And we should only call this line if the sample rate changes.

module/sound/SpeechRecognition/src/SpeechRecognition.cpp
ysims: And move the line into the on configuration, then make the model a class variable.

module/sound/SpeechRecognition/src/SpeechRecognition.cpp
```diff
@@ -0,0 +1,39 @@
...
+            int final                  = vosk_recognizer_accept_waveform(recognizer,
+                                                        reinterpret_cast<const char*>(audioData.WavData.data()),
+                                                        audioData.WavData.size());
+            log<NUClear::DEBUG>(audioData.WavData.size());
+            log<NUClear::DEBUG>(final);
+            log<NUClear::DEBUG>(vosk_recognizer_result(recognizer));
+            vosk_recognizer_free(recognizer);
```
ysims, line 33: Should be if we're making a new one and on Shutdown

module/sound/SpeechRecognition/src/SpeechRecognition.cpp
```diff
@@ -0,0 +1,39 @@
...
+            vosk_model_free(model);
```
ysims, line 34: on Shutdown, and if the config is updated and a new one is made.

module/sound/SpeechRecognition/src/SpeechRecognition.cpp
```diff
@@ -0,0 +1,39 @@
...
+        // This reactor receives audiodata from emission
+        on<Trigger<AudioData>>().then([this](const AudioData& audioData) {
+            VoskModel* model           = vosk_model_new("/home/nubots/NUbots/model");
+            VoskRecognizer* recognizer = vosk_recognizer_new(model, 16000.0);
+            int final                  = vosk_recognizer_accept_waveform(recognizer,
+                                                        reinterpret_cast<const char*>(audioData.WavData.data()),
+                                                        audioData.WavData.size());
+            log<NUClear::DEBUG>(audioData.WavData.size());
+            log<NUClear::DEBUG>(final);
+            log<NUClear::DEBUG>(vosk_recognizer_result(recognizer));
```
ysims, line 32: What is final? Can we just call the function without storing it in a variable, then I assume `vosk_recognizer_result(recognizer)` gives you the text? Looking at the way this works... we'll need to account for any race conditions.

shared/message/input/AudioData.proto
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+syntax = "proto3";
+
+package message.input;
+
+import "google/protobuf/timestamp.proto";
+
+message AudioData{
+        bytes WavData = 1;
```
ysims, line 27:
```suggestion
        bytes audio = 1;
```
or something similar. This is not specifically data from a wav file, it is any audio data, ie the message will also be used with a microphone module. should also add in sample rate.

ysims (changes requested):
It would be nice if we could come up with some interesting UI for the WavReader module, but I think it is good more as a way of testing it all works. I think we should move the implementation (should be more or less a copy and paste) into a test for the SpeechRecognition module. See the Director module's tests for an idea of how this works. You could have incrementally harder tests too, and slap on a `[!mayfail]` for any that don't work so we have a benchmark and something to improve towards. Longer text can be a test, since that seems to be a problem atm.

You should also make a message for the SpeechRecognition module to emit, with a string containing the speech.

module/sound/SpeechRecognition/src/SpeechRecognition.cpp
TrentHouliston: Better yet, this should be a unique or shared pointer with a deleter, that's the proper way to interact with C libraries like this

module/sound/SpeechRecognition/src/SpeechRecognition.cpp
TrentHouliston: Use shared/unique pointers with deleters and c++ itself will handle the cleanup

shared/message/input/AudioData.proto
TrentHouliston: Another thing is if you ever want to be able to use this in realtime you'll need to find a way to make this be chunked, If I'm reading this right currently this always just contains the entire WAV file

module/sound/SpeechRecognition/src/SpeechRecognition.cpp
TrentHouliston: Here make a shared/unique pointer with a deleter that uses the free command.

module/sound/SpeechRecognition/src/SpeechRecognition.cpp
TrentHouliston: Also try to avoid using the word `final` as a variable, it confuses a lot of syntax highlighters and is technically a keyword in c++ (although I think it's not relevant here)

shared/message/input/AudioData.proto
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright 2013 NUbots <nubots@nubots.net>
+ */
+
+syntax = "proto3";
+
+package message.input;
+
+import "google/protobuf/timestamp.proto";
+
+message AudioData{
```
TrentHouliston, line 26: Putting `Data` on a message is a tautology

PR #967, opened 2022-10-22
Add functionality so Director Providers can Idle

TrentHouliston:
This PR adds a new Idle word to the director.
This is so that it can choose to not update the subtasks it emitted last time.

The eventual intention for this is to work with the upcoming done information that will inform providers which of their subtasks emitted a "done" task.

Also fixes a few minor bugs that I noticed while coding this


ysims (approved):
seems fine

PR #968, opened 2022-10-24
Fix challenge priority function in Director and adds a test that targets the error

ysims:
`challenge_priority` function was broken, the ancestors were not right. This prevented a parent task from beating its child in the `challenge_priority` function. This PR adds a test for that case and fixes the function.


TrentHouliston (approved):
I approve, but I helped write it...

Bidski (approved):
Anything to help a parent beat its child

PR #969, opened 2022-10-24
Add extra information so Director Providers can know why they were executed

TrentHouliston:
Adds an optional RunInfo object that can be given alongside a provider that can give additional information about why it was run.
Currently only provides an enum that describes the reason the provider was executed.

### DIFF
--- a/module/extension/Director/src/Director.hpp
+++ b/module/extension/Director/src/Director.hpp
@@ -391,7 +391,7 @@ namespace module::extension {
          *
          * @param reason the reason to hold until the lock is destroyed
          *
-         * @return a lock object that will reset the current run reason when it is destroyed
+         * @return a lock object that will reset `current_run_reason` to its default when destroyed
          */
         RunReasonLock hold_run_reason(const ::extension::behaviour::RunInfo::RunReason& reason);
 
--- a/module/extension/Director/src/Director.hpp
+++ b/module/extension/Director/src/Director.hpp
@@ -321,11 +321,13 @@ namespace module::extension {
         /**
          * Runs the passed task on the passed provider.
          *
-         * @param task      the task we are running
-         * @param provider  the provider that we are running the task on
+         * @param task          the task we are running
+         * @param provider      the provider that we are running the task on
+         * @param run_reason    the reason that we are running this task
          */
         void run_task_on_provider(const std::shared_ptr<BehaviourTask>& task,
-                                  const std::shared_ptr<provider::Provider>& provider);
+                                  const std::shared_ptr<provider::Provider>& provider,
+                                  const ::extension::behaviour::RunInfo::RunReason& run_reason);
 
         /**
          * The level of solution that we can run at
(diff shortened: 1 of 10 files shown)


module/extension/Director/src/Director.hpp
```diff
@@ -373,20 +375,49 @@ namespace module::extension {
          */
         void run_task_pack(const TaskPack& pack);
 
+        /**
+         * A RunReasonLock instance will reset the current_run_reason back to OTHER_TRIGGER when it is destroyed.
+         *
+         * This is used so that the run reason can temporarily be set to something else, and then reset back to the
+         * default if there is an exception or other error.
+         */
+        using RunReasonLock = std::unique_ptr<void, std::function<void(void*)>>;
```
Bidski, line 384: A part of me wants you to replace this with a struct that is purpose built for this, but i dont know how big a part of me that is :stuck_out_tongue:

module/extension/Director/src/Director.hpp
```diff
@@ -373,20 +375,49 @@ namespace module::extension {
...
         std::shared_ptr<void> _get_task_data(const uint64_t& reaction_id) override;
 
+        /**
+         * Provides the RunInfo data via the InformationSource interface so it can be accessed
+         *
+         * @param reaction_id the provider reaction that is requesting its information.
+         *
+         * @return the information about why this task has been executed
+         */
+        ::extension::behaviour::RunInfo _get_run_info(const uint64_t& reaction_id) override;
```
Bidski, line 419: `_get_task_data`? `_get_run_info`? Have you been spending too much time in python?

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -174,7 +174,10 @@ namespace module::extension {
                     remove_task(task);
                 }
                 else {
-                    run_task_on_provider(parent_group.active_task, parent_group.active_provider);
+                    // TODO check somehow if this provider is equipped to handle done
```
Bidski, line 177:
```suggestion
                    // TODO(thouliston) check somehow if this provider is equipped to handle done
```

module/extension/Director/src/Director.hpp
```diff
@@ -370,20 +372,49 @@ namespace module::extension {
...
+         */
+        RunReasonLock hold_run_reason(const ::extension::behaviour::RunInfo::RunReason& reason);
+
     public:
         /// Called by the powerplant to build and setup the Director reactor.
         explicit Director(std::unique_ptr<NUClear::Environment> environment);
         virtual ~Director();
 
         /**
-         * @brief Provides the task data via the InformationSource interface so it can be accessed
```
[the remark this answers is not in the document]
TrentHouliston: It was inconsistent so I removed it everywhere

shared/extension/behaviour/InformationSource.hpp
```diff
@@ -23,6 +23,33 @@
 #include <memory>
 #include <typeindex>
 
+namespace extension::behaviour {
+
+    /**
+     * Provides information about how and why this provider was executed.
+     */
+    struct RunInfo {
```
Bidski, line 31: Is it worth making this a smart enum and building in string conversions so you don't need the `decode_reason` function in your test?

Bidski (changes requested):

shared/extension/behaviour/InformationSource.hpp
TrentHouliston: Seems a bit overkill just for a test

module/extension/Director/src/Director.hpp
TrentHouliston: It's because in the information source there are static functions of the same name and they can't collide

module/extension/Director/src/Director.hpp
TrentHouliston: A very small part

Bidski (approved):
LGTM

module/extension/Director/src/Director.hpp
```diff
@@ -373,20 +375,49 @@ namespace module::extension {
...
+
+        /**
+         * Holds a run reason as the current run reason for the current thread.
+         *
+         * It returns a lock object that acts as an RAII object that will reset the current run reason when it is
+         * destroyed.
+         *
+         * @param reason the reason to hold until the lock is destroyed
+         *
+         * @return a lock object that will reset the current run reason when it is destroyed
```
CMurtagh-LGTM, line 394:
```suggestion
         * @return a lock object that will reset the default run reason when it is destroyed
```
This is what it actually does.

CMurtagh-LGTM (changes requested):

module/extension/Director/src/Director.hpp
TrentHouliston: I meant `current run reason` in terms of the variable `current_run_reason` rather than what I was setting it to, but I can see that there is confusion in that

CMurtagh-LGTM:
```suggestion
         * @return a lock object that will reset the default current_run_reason when it is destroyed
```

CMurtagh-LGTM (approved):

PR #970, opened 2022-10-24
Use a recursive mutex in Director instead of Sync

ysims:
Sync isn't working to prevent multithreading issues, use a mutex. Use a recursive mutex because of `_get_task_data`, which also needs a mutex but can be called indirectly within Director reactions.


TrentHouliston (approved):
There might be better ways to fix this in the future.
But for now this works and that's better than what's there

PR #971, opened 2022-10-24
Add information in Director Uses word to know if a group is done

TrentHouliston:
Fixes up the Uses<X> word so that it provides information.
Adds the done state of the group to the Uses information so it can be used to determine when specific subtasks are done


ysims (approved):
LGTM

PR #973, opened 2022-10-27
Add modules that use the Director to control the servos

ysims:
Adds in Director modules for the lowest level of the system, the servos.

The **Servos** module has reactions that accept servo and limb Tasks and Provide the necessary functionality to complete those tasks. Each individual servo has its own Provider, so that only one servo Task can occur at a time. It is Done when the time is up. Limbs are wrappers for many servos, and will be Done when all their servos on the limb are Done.  

The **Kinematics** module Provides the functionality for leg and head IK Tasks. This will run the IK utility function to get joint angles and then makes a limb Task for the appropriate limb. 

- [ ] Updated NUbook if necessary

Has been tested with the QuinticWalk module (requires adding in emits for arms and legs, see branch before the commits that clean up the PR).


shared/message/motion/LimbsIK.proto
```diff
@@ -0,0 +1,56 @@
...
+import "message/motion/ServoCommand.proto";
+import "Matrix.proto";
+import "Vector.proto";
+
+/// Left leg motion information including target matrix, time and gain and torque for each servo
+message LeftLegIK {
+    /// When the command should execute
+    google.protobuf.Timestamp time = 1;
+    /// Target left foot position to torso
+    mat4 Htl = 2;
```
TrentHouliston, line 33: I think I'd name these `Htf`... maybe, I could be convinced this is better

shared/message/motion/Limbs.proto
```diff
@@ -0,0 +1,83 @@
...
+    enum ID {
+        RIGHT_HIP_YAW     = 0;
+        RIGHT_HIP_ROLL    = 1;
+        RIGHT_HIP_PITCH   = 2;
+        RIGHT_KNEE        = 3;
+        RIGHT_ANKLE_PITCH = 4;
+        RIGHT_ANKLE_ROLL  = 5;
+    }
+    /// Information for each servo in the right leg, in the order of ID
+    repeated ServoCommand servos = 1;
```
TrentHouliston, line 50: Not sure, but it might be worth making this a map rather than a vector.
Is there a situation where you wouldn't want to set all of them?

TrentHouliston (changes requested):

shared/message/motion/ServoCommand.proto
```diff
@@ -0,0 +1,41 @@
...
+syntax = "proto3";
+
+package message.motion;
+
+import "google/protobuf/timestamp.proto";
+
+message ServoState {
+    /// Proportional gain. In proportion to the servo's position error. Gain
+    /// values are in range 0-100.
+    float gain = 1;
```
TrentHouliston, line 29: Maybe make this a vec3 for PID gain (which I think we can still do right?)
It'll default to 0s in a vec3 so people can just set gain.x

shared/message/motion/Limbs.proto
```diff
@@ -0,0 +1,76 @@
...
+/// Enum for each servo in the head
+enum HeadID {
+    YAW   = 0;
+    PITCH = 1;
+}
+
+/// Head servo command information including time, gain, torque, position
+message Head {
+    /// Information for each servo in the head
+    map<HEadID, ServoCommand> servos = 1;
```
TrentHouliston: `HEadID` :P

module/motion/Servos/src/Servos.hpp
```diff
@@ -5,9 +5,37 @@
...
+    using message::input::Sensors;
+    using message::motion::ServoTarget;
+    using utility::input::ServoID;
 
     class Servos : public ::extension::behaviour::BehaviourReactor {
+    private:
+        /// @brief Creates a reaction that sends a given servo command as a servo target command for the platform module
+        /// to use
+        // TODO(ysims): add capability to be Done when the servo reaches the target position
+        template <typename Servo, ServoID ID>
```
TrentHouliston, line 23: Probably actually want `ServoID::Value` rather than the constexpr thing later

TrentHouliston (changes requested):

module/motion/Servos/src/Servos.cpp
```diff
@@ -197,26 +197,26 @@ namespace module::motion {
...
+        add_servo_provider<RightHipPitch, ServoID::Value::R_HIP_PITCH>();
+        add_servo_provider<LeftHipPitch, ServoID::Value::L_HIP_PITCH>();
+        add_servo_provider<RightKnee, ServoID::Value::R_KNEE>();
+        add_servo_provider<LeftKnee, ServoID::Value::L_KNEE>();
+        add_servo_provider<RightAnklePitch, ServoID::Value::R_ANKLE_PITCH>();
+        add_servo_provider<LeftAnklePitch, ServoID::Value::L_ANKLE_PITCH>();
+        add_servo_provider<RightAnkleRoll, ServoID::Value::R_ANKLE_ROLL>();
+        add_servo_provider<LeftAnkleRoll, ServoID::Value::L_ANKLE_ROLL>();
+        add_servo_provider<HeadYaw, ServoID::Value::HEAD_YAW>();
+        add_servo_provider<HeadPitch, ServoID::Value::HEAD_PITCH>();
```
TrentHouliston, line 103: Don't need value in this spot!

TrentHouliston (changes requested):

module/motion/Servos/src/Servos.hpp
```diff
@@ -0,0 +1,74 @@
...
+                if (NUClear::clock::now() >= servo.command.time) {
+                    emit<Task>(std::make_unique<Done>());
+                    return;
+                }
+                emit(std::make_unique<ServoTarget>(servo.command.time,
+                                                   ID,
+                                                   servo.command.position,
+                                                   servo.command.state.gain,
+                                                   servo.command.state.torque));
+            });
```
TrentHouliston, line 37: I think if we get a new task we should always run it regardless of if the time has passed

```suggestion
            on<Provide<Servo>, Trigger<Sensors>>().then([this](const Servo& servo, const RunInfo& info) {
                if (info.run_reason = NEW_TASK) {
                    emit(std::make_unique<ServoTarget>(servo.command.time,
                                                   ID,
                                                   servo.command.position,
                                                   servo.command.state.gain,
                                                   servo.command.state.torque));
                }
                // If the time to reach the position is over, then stop requesting the position
                else if (NUClear::clock::now() >= servo.command.time) {
                    emit<Task>(std::make_unique<Done>());
                }
            });
```

TrentHouliston (changes requested):

shared/message/motion/LimbsIK.proto
ysims: it is specifically the left/right

TrentHouliston (dismissed):

shared/message/motion/LimbsIK.proto
TrentHouliston: The only reason I'd think about changing it is if you ever made a template function that took a LeftLegIK or RightLegIK object you'd be able to take either if they were the same internally

shared/message/motion/ServoCommand.proto
```diff
@@ -0,0 +1,41 @@
...
+    /// Used to set the servo on or off. Typically either 0 (off) or 100 (on)
+    float torque = 2;
+}
+
+message ServoCommand {
```
CMurtagh-LGTM, line 34: Why are we adding a second message called `ServoCommand` should this be called something else?

CMurtagh-LGTM (changes requested):

shared/message/motion/ServoCommand.proto
TrentHouliston: One is for interacting with the Director and one is for interacting with Controller.
Eventually the controller one should be deleted and only this remain.

CMurtagh-LGTM: Comment this please.

TrentHouliston: Should the comment go here or in the controller one. Because eventually the comment will make no sense once the other message is removed

CMurtagh-LGTM: It should go on both, then the one you remove will tell you to remove the comment on the other.

CMurtagh-LGTM (approved):

PR #976, opened 2022-10-30
Update module generation for Director and consistency

ysims:
Updates the code generated with the `./b module generate` tool. 

- Includes Behaviour extension (Director) and extends the class from the Behaviour reactor instead of the normal reactor
- Updates naming of config parameter and struct to match what we've been doing recently for consistency
- Updates README with what we've been doing lately for consistency. Consumes is nice to have, particularly if we have emits already. Prettier automatically formats the heading to `#`, less annoying if we go with that.


CMurtagh-LGTM:
What if we want a non-behaviour reactor?

tools/module/generate.py
```diff
@@ -127,17 +130,18 @@ def generate_cpp(parts):
...
 
-        {className}::{className}(std::unique_ptr<NUClear::Environment> environment) : Reactor(std::move(environment)), config{{}} {{
+        {className}::{className}(std::unique_ptr<NUClear::Environment> environment) : BehaviourReactor(std::move(environment)) {{
 
-            on<Configuration>("{className}.yaml").then([this](const Configuration& cfg) {{
+            on<Configuration>("{className}.yaml").then([this](const Configuration& config) {{
                 // Use configuration here from file {className}.yaml
-                this->log_level = cfg["log_level"].as<NUClear::LogLevel>();
+                this->log_level = config["log_level"].as<NUClear::LogLevel>();
             }});
```
TrentHouliston, line 145: We used `cfg` rather than config here because often there is a config struct people use. And if both are named config it becomes trickier

TrentHouliston (changes requested):
I also agree with @CMurtagh-LGTM about a flag to choose what kind of reactor to generate

Like maybe `./b module generate --type behaviour path/to/Module`

tools/module/generate.py
ysims: I updated the generated struct to be `cfg` since that's what we've been doing for a while now.

tools/module/generate.py
```diff
@@ -96,13 +96,16 @@ def generate_header(parts):
...
 
-        class {className} : public NUClear::Reactor {{
+        class {className} : public ::extension::behaviour::BehaviourReactor {{
         private:
-            /// The configuration variables for this reactor
-            struct {{
-            }} config;
+            /// @brief Stores configuration values
+            struct Config {{
+                Config() = default;
```
TrentHouliston, line 107: This doesn't do anything afaik

CMurtagh-LGTM (approved):

tools/module/generate.py
```diff
@@ -62,11 +62,11 @@ def run(path, **kwargs):
         print("\t", os.path.join(src_path, "README.md"))
 
     with open(os.path.join(src_path, "{}.hpp".format(module_name)), "w") as output:
-        output.write(generate_header(parts))
+        output.write(generate_header(parts, kwargs["director"]))
```
TrentHouliston, line 65: Rather than using kwargs here, make it an argument to run and just use it

tools/module/generate.py
```diff
@@ -88,32 +88,61 @@ def generate_cmake(parts):
...
 
 
-def generate_header(parts):
-    template = textwrap.dedent(
-        """\
-        #ifndef {define}
-        #define {define}
+def generate_header(parts, director):
+    if director:
+        template = textwrap.dedent(
```
TrentHouliston, line 93: I'd make these just different constants like "normal_template" and "behaviour_template". Then the context for the if statement

TrentHouliston (changes requested):

TrentHouliston (approved):

PR #979, opened 2022-11-11
Update the formatting system

TrentHouliston:
This PR updates the formatting system so that both the check in CI and the formatting is done using the same python file.

It also enables formatting for more file types and fixes an issue where NUsight2 wasn't actually getting formatted by prettier.

Currently I haven't applied any of the formatting as now is the time if anyone wants to make any formatting rule changes.


.github/workflows/nubots.yaml
```diff
@@ -176,60 +155,4 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Check clang-format
```
Bidski, line 157:
```suggestion
       - name: Check formatting 
```

Bidski (changes requested):

Bidski (commented):
Consider me approving but not approving to allow people to argue about formatting

CMurtagh-LGTM (dismissed):

CMurtagh-LGTM:
I'm not the biggest fan of removing the ability to skip files that weren't changed, but I don't think I've personally needed it in a while, so we'll see how this goes.

LachlanCourt (approved):

PR #980, opened 2022-11-11
Update zlib using GitHub and remove the local cache

TrentHouliston:
This PR fixes the broken Zlib link that makes the docker image fail to build.
It uses a GitHub release instead of the download on the website as it won't go away. Hopefully avoiding this problem in the future.

I've also removed the other cache as the only way it would work is if everyone was running their own cache.
If you want a caching solution ask me or @Bidski and we can give you an option if you have a computer to run it on.


Bidski (approved):
LGTM

docker/Dockerfile
```diff
@@ -13,7 +13,7 @@
 FROM archlinux:base-20210829.0.32635
 
 # ↓↓↓ UPDATE MIRRORS HERE ↓↓↓
-RUN echo -e "Server=http://host.docker.internal:7878/\$repo/os/\$arch\nServer=https://archive.archlinux.org/repos/2021/08/29/\$repo/os/\$arch" | tee /etc/pacman.d/mirrorlist
```
CMurtagh-LGTM, line 16: Why are we removing the local cache?

CMurtagh-LGTM (changes requested):

docker/Dockerfile
KipHamiltons: The local cache only worked if you had the cache container running. That was only ever done on the iMacs when they were bamboo runners. They aren't that any more

CMurtagh-LGTM (approved):

docker/Dockerfile
CMurtagh-LGTM: :+1:

TrentHouliston: If you want another option and have a computer somewhere to run something on in the lab I can give you another caching solution

JosephusPaye:
Whoops I mistakenly clicked the wrong button and disabled automerge on the GitHub mobile app. Don't mind me.

PR #981, opened 2022-11-19
Add sequence servo director providers and update existing servo code for better usage

ysims:
Adds in sequence providers, which performs like frames of servo positions. Each part of the sequence will be emitted as servo task groups one after the other, and once every sequence is completed it will emit Done. 

There are also some changes to the servo and limb messages to make it easier to use, particularly with scripts.


module/motion/Kinematics/src/Kinematics.cpp
```diff
@@ -47,9 +48,13 @@ namespace module::motion {
...
                     calculateLegJoints<double>(kinematics_model, Eigen::Affine3d(leg_ik.Htl), LimbID::LEFT_LEG);
 
-                // The order of the servos in LeftLegIK and LeftLeg should be LeftLeg.ID
+                for (const auto& joint : joints) {
+                    servos->servos[joint.first] =
+                        ServoCommand(leg_ik.time, joint.second, leg_ik.servos.at(joint.first));
+                }
+
+                // The servos are mapped to ServoIDs
                 for (long unsigned int i = 0; i < joints.size(); i++) {
```
TrentHouliston, line 57: This is now an empty for loop!

shared/message/motion/Limbs.proto
```diff
@@ -22,55 +22,83 @@ package message.motion;
...
-
 /// Left leg servo command information including time, gain, torque, position
 message LeftLeg {
-    /// Information for each servo in the left leg, in order of LegID
-    repeated ServoCommand servos = 1;
+    /// Information for each servo in the left leg, with the key corresponding to ServoID
+    map<uint32, ServoCommand> servos = 1;
+}
+message LeftLegSequence {
+    repeated LeftLeg pack = 1;
```
TrentHouliston, line 31: I feel like there are better names than pack... maybe `sequence` or `frames`

module/motion/Servos/src/Servos.hpp
```diff
@@ -60,11 +60,57 @@ namespace module::motion {
...
+                                               group.servos.at(utility::motion::ServoMap<Elements>::value))),
                                            0)...);
                 });
         }
 
+        /// @brief Keeps track of which Group in the Sequence in add_sequence_provider should be emitted next
+        /// @tparam Sequence Used to have unique counts for each Sequence Provider.
+        template <typename Sequence>
+        struct Count {
+            long unsigned int count = 0;
```
TrentHouliston, line 72: Should make a note somewhere that there is the potential for bugs here.
Really need a token system so we know which script we are getting the done events for.
Or do validation ourself that they are actually done

module/motion/Servos/src/Servos.hpp
```diff
@@ -60,11 +60,57 @@ namespace module::motion {
...
+                    }
+                    // If this is a new task, run the first pack of servos and increment the counter
+                    else if (info.run_reason == RunInfo::RunReason::NEW_TASK) {
+                        emit<Task>(std::make_unique<Group>(sequence.pack[0]));
+                        emit<Scope::DIRECT>(std::make_unique<Count<Sequence>>(1));
+                    }
+                    // If the subtask is done, we are done if it is the last servo pack, otherwise use the count to
+                    // determine the current pack to emit
+                    else if (info.run_reason == RunInfo::RunReason::SUBTASK_DONE) {
+                        if (sequence.pack.size() < count.count) {
```
TrentHouliston, line 99: This reads backwards to me
shouldn't it be `count.count < sequence.pack.size()` emit the next frame else emit done?
More like a loop counter

TrentHouliston (changes requested):

TrentHouliston (approved):
Needs more 🧟 but LGTM

PR #982, opened 2022-11-19
Update NatNet to Version 4.0

AlanaNoonan:


module/input/NatNet/src/NatNet.cpp
```diff
@@ -130,10 +130,9 @@ namespace module::input {
         // Read frame number
         mocap->frame_number = ReadData<uint32_t>::read(ptr, version);
 
-        // Read the markersets
         mocap->marker_sets = ReadData<std::vector<MotionCapture::MarkerSet>>::read(ptr, version);
 
-        // Read the free floating markers
+        // Read the free floating markers but don't process them
```
ysims, line 135: It looks like the free markers are processed below?

module/input/NatNet/src/NatNet.cpp
```diff
@@ -130,10 +130,9 @@ namespace module::input {
         // Read frame number
         mocap->frame_number = ReadData<uint32_t>::read(ptr, version);
 
-        // Read the markersets
```
ysims, line 133: How come this comment is removed?

module/input/NatNet/src/NatNet.cpp
```diff
@@ -201,6 +200,8 @@ namespace module::input {
                 // Stop processing
                 return;
             }
+
+            // Should probably be saving the informaion of the frame here
```
ysims, line 204: what does this mean? it looks like frame information is being saved into a message above?

module/input/NatNet/src/NatNet.cpp
```diff
@@ -402,10 +403,15 @@ namespace module::input {
...
-
         // Get our packet
         const Packet& packet = *reinterpret_cast<const Packet*>(input.data());
 
+        // std::stringstream ss;
+        // for (const auto& c : input) {
+        //     ss << std::setw(2) << std::setfill('0') << std::hex << int(c);
+        // }
+        // NUClear::log(ss.str());
+
```
ysims, line 414: remove comented out code

module/input/NatNet/src/Parse.hpp
```diff
@@ -202,6 +204,9 @@ namespace module::input {
                 marker.model_solved       = false;
             }
 
+            float residual = ReadData<uint32_t>::read(ptr, version);
+
```
ysims, line 208:
```suggestion
```

roles/natnet.role
```diff
@@ -9,7 +9,8 @@ nuclear_role(
   support::configuration::GlobalConfig
   network::NUClearNet
   network::NetworkForwarder
-  platform::cm740::HardwareIO
+  platform::cm740::HardwareSimulator
```
ysims, line 12: revert change

shared/message/input/MotionCapture.proto
```diff
@@ -34,6 +34,7 @@ message MotionCapture {
         uint32 id       = 1;
         fvec3  position = 2;
         float  size     = 3;
+        uint32 modelId = 4;
```
ysims, line 37:
```suggestion
        uint32 model_id = 4;
```
[we use snake_case ](https://nubook.nubots.net/guides/general/code-conventions#page-content)

module/input/NatNet/src/Parse.hpp
```diff
@@ -246,6 +251,12 @@ namespace module::input {
             m.id       = ReadData<uint32_t>::read(ptr, version);
             m.parentId = ReadData<uint32_t>::read(ptr, version);
             m.offset   = ReadData<Eigen::Matrix<float, 3, 1>>::read(ptr, version);
+
+            uint32_t nMarkers = ReadData<uint32_t>::read(ptr, version);
+
+            ptr += nMarkers * 3 * sizeof(float);
+            ptr += nMarkers * sizeof(int);
+
```
ysims, line 259: should this be wrapped in a version if?

module/input/NatNet/src/Parse.hpp
```diff
@@ -202,6 +204,9 @@ namespace module::input {
                 marker.model_solved       = false;
             }
 
+            float residual = ReadData<uint32_t>::read(ptr, version);
```
ysims, line 207: should this be wrapped in a version if?

module/input/NatNet/src/Parse.hpp
```diff
@@ -154,9 +154,11 @@ namespace module::input {
             rigidBody.id       = ReadData<uint32_t>::read(ptr, version);
             rigidBody.position = ReadData<Eigen::Matrix<float, 3, 1>>::read(ptr, version);
             rigidBody.rotation = ReadData<Eigen::Matrix<float, 4, 1>>::read(ptr, version);
-            rigidBody.markers  = ReadData<std::vector<Marker>>::read(ptr, version);
 
             // Version specific information
+            if (version < 0x03000000) {
+                rigidBody.markers = ReadData<std::vector<Marker>>::read(ptr, version);
```
ysims, line 160: how are rigid body markers read now?

ysims (commented):
It would be good to double check that the version checks are right and that the change did occur in that exact version. I assume all the changes should be wrapped in version checks? There don't seem to be many with checks.

module/input/NatNet/src/NatNet.cpp
Dereferenced-Null: This is leftover from an in progress build, I'll remove it

module/input/NatNet/src/Parse.hpp
Dereferenced-Null: According to the documentation the marker positions have been removed as redundant as they can be derived from the position and orientation plus offset to optimize packet size

module/input/NatNet/src/NatNet.cpp
TrentHouliston: Replace commented code with a config option and ability to write the raw packets to files for future debugging

module/input/NatNet/src/NatNet.cpp
```diff
@@ -402,10 +422,15 @@ namespace module::input {
...
     void NatNet::process(const std::vector<char>& input) {
-
         // Get our packet
         const Packet& packet = *reinterpret_cast<const Packet*>(input.data());
 
+        // std::stringstream ss;
+        // for (const auto& c : input) {
+        //     ss << std::setw(2) << std::setfill('0') << std::hex << int(c);
+        // }
+        // NUClear::log(ss.str());
```
TrentHouliston, line 432: Something like this might work for dumping packets out (with an appropriate config value)
Note this code has not been tested!
You also may want to add a config value for where to dump the packets to so you don't fill up your folder

```suggestion
        if (config.dump_packets) {
            static int i = 0;
            std::ofstream output(std::format("natnet_{:06d}.bin", i++));
            output.write(input.data(), input.size());
        }
```

TrentHouliston (changes requested):
A few changes,
Also based on reading the source NatNet depacketizer there seem to be a lot of version checks that are not replicated in this code.

module/input/NatNet/src/NatNet.cpp
```diff
@@ -25,6 +25,7 @@
 
 #include "extension/Configuration.hpp"
 
+// TODO: Convert this to snake case
```
ysims, line 28: finish and remove todo comment?

module/input/NatNet/src/NatNet.cpp
```diff
@@ -164,23 +167,33 @@ namespace module::input {
...
+
         short params                  = ReadData<short>::read(ptr, version);
         mocap->recording              = (params & 0x01) == 0x01;
         mocap->tracked_models_changed = (params & 0x01) == 0x02;
 
         // TODO(HardwareTeam): there is an eod thing here
+        ReadData<uint32_t>::read(ptr, version);
 
         // Apply the model information we have to the objects
         for (auto& markerSet : mocap->marker_sets) {
```
ysims, line 199: `marker_set`?

module/input/NatNet/src/NatNet.hpp
```diff
@@ -58,7 +72,9 @@ namespace module::input {
             std::string name{};
             uint32_t id{0};
             uint32_t parentId{0};
-            Eigen::Vector3f offset = Eigen::Vector3f::Zero();
+            Eigen::Vector3f offset   = Eigen::Vector3f::Zero();
+            Eigen::Vector3f rotation = Eigen::Vector3f::Zero();
+            std::vector<std::string> markerNames;
```
ysims, line 77:
```suggestion
            std::vector<std::string> marker_names;
```

module/input/NatNet/src/NatNet.hpp
```diff
@@ -68,18 +84,48 @@ namespace module::input {
             std::map<uint32_t, RigidBodyModel> boneModels;
```
ysims, line 84:
```suggestion
            std::map<uint32_t, RigidBodyModel> bone_models;
```

module/input/NatNet/src/NatNet.hpp
```diff
@@ -68,18 +84,48 @@ namespace module::input {
...
+        };
+
+        struct DeviceModel {
+            DeviceModel() = default;
+            std::string name{};
+            uint32_t id{0};
+            std::string serialNo{};
+            uint32_t DeviceType{0};
+            uint32_t ChannelDataType{0};
+            uint32_t Channels{0};
```
ysims, line 108:
```suggestion
            uint32_t device_type{0};
            uint32_t channel_data_type{0};
            uint32_t channels{0};
```

module/input/NatNet/src/NatNet.hpp
```diff
@@ -68,18 +84,48 @@ namespace module::input {
...
+            Eigen::Vector3f position    = Eigen::Vector3f::Zero();
+            Eigen::Vector4f orientation = Eigen::Vector4f::Zero();
+        };
+
         // Models we are using
         std::map<std::string, MarkerSetModel> markerSetModels;
         std::map<uint32_t, RigidBodyModel> rigidBodyModels;
         std::map<uint32_t, SkeletonModel> skeletonModels;
+        std::map<uint32_t, ForcePlateModel> forcePlateModels;
+        std::map<uint32_t, DeviceModel> deviceModels;
```
ysims, line 123: snake_case

module/input/NatNet/data/config/NatNet.yaml
```diff
@@ -1,3 +1,4 @@
 multicast_address: 239.255.42.99
 command_port: 1510
 data_port: 1511
+dump_packets: true
```
ysims, line 4: I feel like it might be better to use the log level instead of having this here. Maybe log level trace will dump the packets?

TrentHouliston: I think it should stay separate, dumping packets uses up disk space which should be a separate decision.
Should have a config option here for where to dump them though

module/input/NatNet/src/NatNet.cpp
```diff
@@ -164,23 +167,33 @@ namespace module::input {
...
             mocap->natnet_timestamp = ReadData<double>::read(ptr, version);
         }
         else {
             mocap->natnet_timestamp = ReadData<float>::read(ptr, version);
         }
 
+        // High res timestamps
+        if (version >= 0x03000000) {
+            mocap->mid_exposure_timestamp  = ReadData<float>::read(ptr, version);
+            mocap->data_recieved_timestamp = ReadData<float>::read(ptr, version);
```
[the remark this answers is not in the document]
TrentHouliston: Ahhh, `i` before `e` except after `c`, the classic rule that has so many exceptions (except not this one)

module/input/NatNet/src/NatNet.cpp
```diff
@@ -164,23 +167,33 @@ namespace module::input {
...
+            mocap->transmit_timestamp      = ReadData<float>::read(ptr, version);
+        }
+
         short params                  = ReadData<short>::read(ptr, version);
         mocap->recording              = (params & 0x01) == 0x01;
         mocap->tracked_models_changed = (params & 0x01) == 0x02;
 
         // TODO(HardwareTeam): there is an eod thing here
```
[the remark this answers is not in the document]
TrentHouliston: Given that there is a new ReadData just below this that's not getting stored anywhere is it that?
I don't remember what eod is referring to

module/input/NatNet/src/NatNet.cpp
```diff
@@ -314,9 +328,9 @@ namespace module::input {
         // Our pointer as we move through the data
         const char* ptr = &packet.data;
 
-        uint32_t nModels = ReadData<uint32_t>::read(ptr, version);
+        uint32_t Models = ReadData<uint32_t>::read(ptr, version);
```
TrentHouliston, line 331: `n_models` not `Models` that makes it look like a class

module/input/NatNet/src/NatNet.cpp
```diff
@@ -340,25 +354,22 @@ namespace module::input {
...
+                } break;
 
-                //Camera
-                case 5:{
-                    ReadData<CameraModel>::read(ptr,version);
-                    break;
-                }
+                // Camera
+                case 5: {
+                    ReadData<CameraModel>::read(ptr, version);
```
TrentHouliston, line 371: Did we want to store this somewhere?

module/input/NatNet/src/NatNet.cpp
```diff
@@ -372,32 +383,32 @@ namespace module::input {
...
         std::string name(&packet.data);
-        const char* appVersion    = &packet.data + 256;
-        const char* natNetVersion = appVersion + 4;
+        const char* app_version     = &packet.data + 256;
+        const char* nat_net_version = app_version + 4;
 
         // Update our version number
-        version =
-            (natNetVersion[0] << 24) | (natNetVersion[1] << 16) | (natNetVersion[2] << 8) | (natNetVersion[3] << 0);
+        version = (nat_net_version[0] << 24) | (nat_net_version[1] << 16) | (nat_net_version[2] << 8)
```
TrentHouliston, line 390: is this just big to little endian swap? `ntohl(*reinterpret_cast<int32_t*>(nat_net_vesion))`

module/input/NatNet/src/NatNet.cpp
```diff
@@ -372,32 +383,32 @@ namespace module::input {
...
+                                      + (nat_net_version[2] == 0 ? "" : "." + std::to_string(nat_net_version[2]))
+                                      + (nat_net_version[3] == 0 ? "" : "." + std::to_string(nat_net_version[3]));
 
         // Make our remote into an IP
-        std::string strRemote = std::to_string((remote >> 24) & 0xFF) + "." + std::to_string((remote >> 16) & 0xFF)
-                                + "." + std::to_string((remote >> 8) & 0xFF) + "."
-                                + std::to_string((remote >> 0) & 0xFF);
+        std::string str_remote = std::to_string((remote >> 24) & 0xFF) + "." + std::to_string((remote >> 16) & 0xFF)
+                                 + "." + std::to_string((remote >> 8) & 0xFF) + "."
+                                 + std::to_string((remote >> 0) & 0xFF);
```
TrentHouliston, line 408: This is a job for fmt::format for sure

module/input/NatNet/src/NatNet.cpp
```diff
@@ -425,11 +436,12 @@ namespace module::input {
         // Get our packet
         const Packet& packet = *reinterpret_cast<const Packet*>(input.data());
 
-        // std::stringstream ss;
-        // for (const auto& c : input) {
-        //     ss << std::setw(2) << std::setfill('0') << std::hex << int(c);
-        // }
-        // NUClear::log(ss.str());
+        // This isn't working yet for some reason, no file is being created
```
TrentHouliston, line 439: This should be a todo so you don't forget it.
But also get it working
You might need a makedirs call here since if you're making this in a folder that doesn't exist it won't work

module/input/NatNet/src/NatNet.hpp
```diff
@@ -26,6 +26,20 @@
...
+        /// @brief Stores configuration values
+        struct Config {
+            Config() = default;
+            /// @brief Motive Multicast Address
+            std::string multicast_address = "";
+            /// @brief UDP Command port for communicating with motive
+            uint32_t command_port = 0;
+            /// @brief UDP Data port for communicating with motive
+            uint32_t data_port = 0;
+            /// @brief Dump packets configuraton settings
```
TrentHouliston, line 39: No need to say this is a configuration setting. It's in the config struct.
When documenting describe what something does, not what it is. You can see what it is because you're looking at it :P

module/input/NatNet/src/Parse.hpp
```diff
@@ -251,10 +251,22 @@ namespace module::input {
...
-            ptr += nMarkers * 3 * sizeof(float);
-            ptr += nMarkers * sizeof(int);
-
+            const char* ptr2  = ptr + (nMarkers * 3 * sizeof(float));
+            const char* ptr3  = ptr2 + (nMarkers * (sizeof(int)));
+
+            if (version >= 0x03000000) {
+                // Previously skipped information about markers, may be needed in the future
+                // Marker position
+                ReadData<Eigen::Matrix<float, 3, 1>>::read(ptr, version);
```
TrentHouliston, line 260: If you're reading the new data, you should store it somewhere

module/input/NatNet/src/Parse.hpp
```diff
@@ -279,68 +291,66 @@ namespace module::input {
...
-                m.corners = ReadData<Eigen::Matrix<uint32_t, 4, 3>>::read(ptr, version);
-                m.plateType = ReadData<uint32_t>::read(ptr,version);
-                m.channelType = ReadData<uint32_t>::read(ptr,version);
-                m.nChannels = ReadData<uint32_t>::read(ptr,version);
-                for (uint32_t n = 1; n < m.nChannels; n++){
-                    ReadData<std::string>::read(ptr,version);
+                m.corners           = ReadData<Eigen::Matrix<uint32_t, 4, 3>>::read(ptr, version);
+                m.plateType         = ReadData<uint32_t>::read(ptr, version);
+                m.channelType       = ReadData<uint32_t>::read(ptr, version);
+                m.Channels          = ReadData<uint32_t>::read(ptr, version);
```
TrentHouliston, line 309: `n_channels`

shared/message/input/MotionCapture.proto
```diff
@@ -99,16 +100,19 @@ message MotionCapture {
...
     /// https://en.wikipedia.org/wiki/SMPTE_timecode
-    uint32 timecode         = 3;
-    uint32 timecode_sub     = 4;
-    double natnet_timestamp = 5;
+    uint32 timecode                = 3;
+    uint32 timecode_sub            = 4;
+    double natnet_timestamp        = 5;
+    double mid_exposure_timestamp  = 6;
+    double data_recieved_timestamp = 7;
+    double transmit_timestamp      = 8;
```
TrentHouliston, line 108: Be aware that when you reorder the numbers of protocol buffers like this you break any previous nbs files you have recorded. If you're doing it in the future for something you have recordings of, you can still put the fields here, but you need to use the bigger numbers.

TrentHouliston (changes requested):

module/input/NatNet/data/config/NatNet.yaml
ysims: Oh this is saving to a file, not printing to the terminal. That's fine then.

module/input/NatNet/src/NatNet.cpp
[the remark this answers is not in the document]
Dereferenced-Null: There is a uint32_t value sitting at the end of the packet which based on the printed hex packets appears to be a constant to indicate the end of the packet. I made the new ReadData for completeness but realistically if we don't have any use for the End of Data symbol then I can get rid of this?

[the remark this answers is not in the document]
ysims: Sounds like it might be good to read and check if it is the expected eod symbol, and if it's not then throw an error saying that the packet wasn't read correctly.

module/input/NatNet/src/NatNet.cpp
TrentHouliston: Also, I think something like this will work for the IP address?

```cpp  
  std::array<char, INET_ADDRSTRLEN> str{};
  inet_ntop(AF_INET, &remote, str.data(), str.size());
  std::string IP = str.data();
```

module/input/NatNet/src/NatNet.cpp
```diff
@@ -379,17 +388,15 @@ namespace module::input {
...
         // Extract the information from the packet
         std::string name(&packet.data);
         const char* app_version     = &packet.data + 256;
         const char* nat_net_version = app_version + 4;
 
         // Update our version number
-        version = (nat_net_version[0] << 24) | (nat_net_version[1] << 16) | (nat_net_version[2] << 8)
-                  | (nat_net_version[3] << 0);
-
+        version = ntohl(*reinterpret_cast<const char*>(nat_net_version));
```
TrentHouliston, line 399: To use ntohl (network to host long) you need to use a 32 bit integer, so you'd need to do the cast to one. This will just cast it to a char and then try to do it for that (you'll probably miss most of your version number)

so `ntohl(*reinterpret_cast<const uint32_t*>(nat_net_version));`

TrentHouliston (changes requested):

module/input/NatNet/src/NatNet.cpp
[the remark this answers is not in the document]
TrentHouliston: End of data makes sense. I totally would have put a comment and ignored it when I wrote it the first time

module/input/NatNet/src/NatNet.cpp
```diff
@@ -307,37 +329,55 @@ namespace module::input {
...
+                // Device
+                case 4: {
+                    DeviceModel m       = ReadData<DeviceModel>::read(ptr, version);
+                    device_models[m.id] = m;
+                } break;
+
+                // Camera
+                case 5: {
+                    CameraModel m = ReadData<CameraModel>::read(ptr, version);
+                    camera_model  = m;
```
ysims, line 380: Is the camera model only ever going to appear once? It seems that data would be lost if not?

module/input/NatNet/src/NatNet.cpp
```diff
@@ -307,37 +329,57 @@ namespace module::input {
...
+                    DeviceModel m       = ReadData<DeviceModel>::read(ptr, version);
+                    device_models[m.id] = m;
+                } break;
+
+                // Camera
+                case 5: {
+                    static uint32_t n = 0;
+                    CameraModel m     = ReadData<CameraModel>::read(ptr, version);
+                    camera_models[n]  = m;
+                    n++;
```
ysims, line 382: Are the models reset anywhere each time a new packet is processed?

roles/natnet.role
ysims, line 12: This doesn't seem like a change we would want?

Dereferenced-Null: I've got to ignore this locally, using the simulator is easier for testing but I'll change it back

module/input/NatNet/src/NatNet.cpp
Dereferenced-Null: No, so the model definition packet is sent when requested by our natnet client, which in our case only occurs once on start-up or if there is an error that means the model definitions need to be reset, which I haven't seen occur during testing unless I coded something wrong, so it should happen very rarely.

module/input/NatNet/src/NatNet.cpp
```diff
@@ -307,37 +329,65 @@ namespace module::input {
...
+    void NatNet::process_model(const Packet& packet) {
+        // Remove old definitions
+        marker_set_models.clear();
+        rigid_body_models.clear();
+        skeleton_models.clear();
+        force_plate_models.clear();
+        device_models.clear();
+        camera_models.clear();
 
+        std::stringstream ss;
```
ysims, line 341: tbh this is scope creep-y, you can revert and make an issue instead (and ideally address the issue in your next PR :stuck_out_tongue: )

TrentHouliston: Is this variable even used?

Dereferenced-Null: No worries, I'll do that now.

I think it used to be used for storing some part of one of the models but its redundant now so I'll take it out.

ysims (approved):

module/input/NatNet/src/NatNet.cpp
TrentHouliston: When I was doing the original it also sent me the models without being asked if you updated them while you were listening

Dereferenced-Null: Yeah it still does that, it uses the model definitions to check if the information being read is correct and if not it asks for a new model definition packet. For the camera's specifically its a bit weird because there's a camera model but we never read information specifically about the cameras in the data packets so the camera model is stored and not used at the moment.

TrentHouliston (approved):

PR #986, opened 2022-11-23
Replace Affine with Isometry

TrentHouliston:
It turns out the thing that we thought Affine3d was doing is [actually handled by Isometry](https://eigen.tuxfamily.org/dox/classEigen_1_1Transform.html).
Affine3d assumes a general affine transform which may include things like scale/shear so when we invert or access rotation it does extra work. Isometry3d is doing what we expect where it does the transpose of the rotation part etc.
All of our homogenous transforms should be using something like Isometry3d over Affine3d.

### Pre-PR Checklist:

- [x] Updated NUbook if necessary https://github.com/NUbots/NUbook/pull/271

### DIFF
--- a/module/localisation/RobotParticleLocalisation/src/RobotModel.hpp
+++ b/module/localisation/RobotParticleLocalisation/src/RobotModel.hpp
@@ -84,11 +84,11 @@ namespace module::localisation {
                                                          const Eigen::Matrix<Scalar, 4, 4>& Hcw) {
 
             // Create a transform from the field state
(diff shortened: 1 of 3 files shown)


ysims (changes requested):
This should be tested first

ysims (approved):
seems to not break anything

KipHamiltons:
My whole life has been a lie

Bidski:
> My whole life has been a lie

Not your whole life. Just the more important parts 😛

PR #987, opened 2022-11-23
Updates default camera settings to compensate for motion blur

Crafty15:
Change of plan from having a set exposure and trying to compensate for darker images with gain. Now sets an auto exposure upper limit to allow for some auto adjustment, and compensates somewhat by setting the grey value. Seems to work a bit better.
- Turned auto exposure back on and set an upper limit of 3000.
- Sets frame rate to 100.
- Turned off auto exposure grey target value and set to 30.


ysims (approved):
lgtm

PR #988, opened 2022-12-01
Add live Webots vision error testing (goal detection)

BaileyJoe:
### Webots vision benchmarking for goal detection

We use groundtruth data from the webots simulator to determine error in the goal detectors output and do benchmarking.

- Outputs average error in detected goal post position in vector and scalar form.
- Tracks number of 'bad' posts determined by difference from a threshold set in the goaldetector yaml file.
- Outputs values including and excluding the bad goal posts.


module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -372,6 +375,134 @@ namespace module::vision {
...
+                                                        goal_error3f.x(),
+                                                        goal_error3f.y(),
+                                                        goal_error3f.z()));
+
+
+                        log<NUClear::DEBUG>(fmt::format("Total Error:{}", goal_error_bad));
+                        log<NUClear::DEBUG>(fmt::format("Total Error x:{}, y:{}, z:{}",
+                                                        goal_error3f_bad.x(),
+                                                        goal_error3f_bad.y(),
+                                                        goal_error3f_bad.z()));*/
```
ysims, line 489: remove?

roles/webots/keyboardwalk.role
```diff
@@ -15,6 +15,11 @@ nuclear_role(
   # Platform
   input::SensorFilter
   platform::Webots
+  # Vision
+  vision::VisualMesh
+  vision::GreenHorizonDetector
+  vision::GoalDetector
+  vision::BallDetector
```
ysims, line 22: probably don't want to add this here, we could look later at making a new role with all the parts to do the benchmarking and maybe a module that does fixed actions so the benchmarks should be the same each time.

ysims (commented):
Looking nice!

module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -372,6 +375,134 @@ namespace module::vision {
                         }
                     }
 
+                    auto calc_error = [&](const Eigen::Vector3f& post1, const Eigen::Vector3f& post2) {
+                        return std::sqrt((post1.x() - post2.x()) * (post1.x() - post2.x()))
+                               + ((post1.y() - post2.y()) * (post1.y() - post2.y()))
+                               + ((post1.z() - post2.z()) * (post1.z() - post2.z()));
+                    };
+
+                    if (horizon.vision_ground_truth.exists) {
```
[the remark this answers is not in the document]
ysims: This should all have comments throughout

PR #994, opened 2022-12-15
Changes Script extension to take a sequence of scripts

ysims:
For Director.

Changes the Script to take a vector of Scripts rather than one, and concats the scripts into one set of servo sequences.

Also some minor bug fixes.

Example of use:
`emit<Script>(std::make_unique<LimbsSequence>(),
                     std::vector<ScriptRequest>{{"KickLeft.yaml", 1.0}, {"KickRight.yaml", 1.0}});`

### DIFF
--- a/shared/extension/behaviour/Script.hpp
+++ b/shared/extension/behaviour/Script.hpp
@@ -21,6 +21,11 @@ namespace extension::behaviour {
     using message::motion::ServoState;
     using utility::input::ServoID;
 
+    struct ScriptRequest {
+        std::string name;
+        float duration_modifier{1.0};
+    };
+
     // One frame of a script
     struct Frame {
         /// @brief Servo targets for each active servo in the frame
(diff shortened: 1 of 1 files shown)


shared/extension/behaviour/Script.hpp
```diff
@@ -98,38 +107,40 @@ namespace extension::behaviour {
...
+        /// @param start When the first script should start executing. Default is now.
         static void emit(NUClear::PowerPlant& powerplant,
                          std::shared_ptr<Sequence> msg,
-                         const std::string& script,
-                         const NUClear::clock::time_point& start = NUClear::clock::now(),
-                         const float& duration_modifier          = 1.0) {
-            // Load the script to a vector of Frame structs
-            auto frames = load(script).as<std::vector<Frame>>();
-            // time will be incremented through each frame
+                         std::vector<ScriptRequest> scripts,
```
CMurtagh-LGTM, line 118: Will we ever want to run a single script, if so perhaps overload this function to one that takes a single script request.

CMurtagh-LGTM (changes requested):

CMurtagh-LGTM (approved):

PR #995, opened 2022-12-15
Add a Director-based KickScript module

ysims:
The KickScript provides for the Kick Task. It emits a servo sequence for the KickLeft or KickRight script through the Script extension, depending on the requested leg. As well as requesting the Kick script, it also adds the Stand script on either side, in a script sequence.


module/skills/KickScript/src/KickScript.cpp
```diff
@@ -0,0 +1,52 @@
...
+            }
+            // If it isn't a new task, don't do anything
+            if (info.run_reason != RunInfo::RunReason::NEW_TASK) {
+                emit<Task>(std::make_unique<Idle>());
+                return;
+            }
+
+            if (kick.leg == LimbID::RIGHT_LEG) {
+                emit<Script>(std::make_unique<LimbsSequence>(),
+                             std::vector<ScriptRequest>{{"Stand.yaml"}, {"KickRight.yaml"}, {"Stand.yaml"}});
```
TrentHouliston, line 43: I would put the scripts to run in the configuration so you can change them on demand at robocup
Which you will 100% want to do

module/skills/KickScript/src/KickScript.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "extension/Behaviour.hpp"
+#include "extension/Configuration.hpp"
+#include "extension/behaviour/Script.hpp"
+
+#include "message/actuation/Limbs.hpp"
+#include "message/skills/Kick.hpp"
+
+#include "utility/input/LimbID.hpp"
+
+namespace module::skills {
```
TrentHouliston, line 12: I think this should be `skill` not `skills` simply because every other level of behaviour is also singular

TrentHouliston (dismissed):

module/skills/KickScript/src/KickScript.cpp
TrentHouliston: In the past we've had to do things like swap out for a Power kick for a penalty shootout, or sometimes right kick worked but left didn't so we just set them to the same

ysims: I think planning should take care of that

CMurtagh-LGTM (approved):

PR #999, opened 2022-12-16
Remove undefined behaviour in Protobuf to Neutron conversions

Bidski:
Things the PR does go here

Previous message conversion implementation would pass a reference to uninitialised memory and would then set values on this uninitialised memory, finally returning a reference to the uninitialised object that was passed in. This PR aims to remove the potential for undefined behaviour to creep in

### Pre-PR Checklist:

Have you

- [X] Added/updated tests for your changes, including regression tests for bug fixes
- [X] Added/modified [documentation directives](https://nubook.nubots.net/guides/general/code-conventions#documentation) in relevant code
- [X] Added a descriptive title and relevant labels to the PR

### Reviewers, Note

The described undefined behaviour is non-deterministic in nature and it will not always happen. However, it has been experienced primarily in release builds. Generally the undefined behaviour manifests as invalid data in an otherwise valid message (e.g. when playing back a sensors message with DataPlayback Htw may consist of data on the order of `1e-310`, but if you view the same message contents via `./b nbs json` all data in all messages will be valid)


nuclear/cmake/Modules/GenerateNeutron.cmake
```diff
@@ -0,0 +1,170 @@
...
+  set(multiValueArgs "")
+  cmake_parse_arguments(NEUTRON "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+  # Locations to store each of the output components
+  set(pb_out "${CMAKE_CURRENT_BINARY_DIR}/protobuf")
+  set(nt_out "${CMAKE_CURRENT_BINARY_DIR}/neutron")
+  set(py_out "${CMAKE_CURRENT_BINARY_DIR}/python")
+
+  # Extract the components of the filename that we need
+  get_filename_component(file_we ${NEUTRON_PROTO} NAME_WE)
```
CMurtagh-LGTM, line 44: :confused: `we`? I think that this isn't a good name, it does not really describe its job clearly.

Bidski: file whatever? describes its purpose extremely well i think :stuck_out_tongue: 

Also given that the cmake option is called `NAME_WE` i think its hard to get more explicit?

CMurtagh-LGTM: I must not know enough about cmake :shrug:

TrentHouliston (approved):

PR #998, opened 2022-12-15
Add a GetUp provider that will use the appropriate getup script

TrentHouliston:
This PR adds a Behaviour skill that when executed as a task will determine using `sensors.Htw` which way the robot is lying and then execute an appropriate script.

PR #1001, opened 2022-12-16
Add a module for planning when to run the GetUp task

TrentHouliston:
Adds a planner to monitor when a getup should be run
Depends on #998


module/planning/GetUpPlanner/src/GetUpPlanner.cpp
```diff
@@ -0,0 +1,70 @@
...
+            // We have finished getting up so we can reset getting_up and decide if we want to get up again (we failed)
+            // If we don't this will make our task lapse and other providers can take over
+            if (getting_up && info.run_reason == RunInfo::SUBTASK_DONE) {
+                getting_up = false;
+            }
+
+            // If we have been at recovery levels for long enough we can trigger a getup
+            if (!getting_up && recovery_frames > cfg.count) {
+                // Emit a getup
+                emit<Task>(std::make_unique<GetUp>());
```
ysims, line 58:
```suggestion
                getting_up = true;
                // Emit a getup
                emit<Task>(std::make_unique<GetUp>());
```

ysims (approved):
Appears to work

PR #1003, opened 2022-12-16
Add a strategy module to manage falling and recovery

TrentHouliston:
This strategy module will run both GetUp and FallingRelax planners and ensure that GetUp correctly has higher priority than Relax.

It's recommended that when using this strategy it has higher priority than anything else in the system.


shared/message/strategy/FallRecovery.proto
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright 2022 NUbots <nubots@nubots.net>
+ */
+syntax = "proto3";
+
+package message.strategy;
+
+/**
+ * Emit a PlanGetUp Task in order to tell the GetUpPlanner to start monitoring the robot's state and get it up if it
+ * falls.
+ */
```
ysims, line 26: doesn't mention relaxing? names are wrong too. maybe don't make this comment reliant on other message names? Something like 'watches the robot's state and relaxes if it is falling over, and gets up when it has fallen' - or even more generic 'watches the robot's state and handles the recovery if the robot is falling/has fallen'.

ysims (approved):
TMLG

PR #1004, opened 2022-12-20
Clean InEKF code and integrate with SensorFilter

ysims:
NOTE this is not merging into main, it is merging into `obrien/InEKF`.

@Tom0Brien some cleaning and I got it integrated into SensorFilter (somewhat). 

**Changes**
- Cleaned up the four classes for InEKF (namespace, moved into their own folder, snake_case, some other things probably...)
- Cleaned up the test a lot, and added in a script that converts their measurements into a yaml file with all the stuff we need. Should be easier to read now. 
- Replaced the UKF in the SensorFilter module with the InEKF

**Current state** is that it will stand still fine, but when I start it walking the robot moves up in space, like it is ascending to its mothership. 

**Todos**
- The test doesn't have comparison values, just measurement values. If there aren't any real values from the original repo, we could just run the original repo and make sure the state is the same in our code and theirs. So we can be sure we aren't changing the maths. 
- Add a test with actual ground truth, probably just use whatever we have in the UKF test, or get new webots data
- More cleaning up for the InEKF code - in particular, consider if the current design is fine... maybe make RobotState and NoiseParams only used by the InEKF not whatever is using the InEKF as well... Better initialisation, no hardcoding, documentation directives, and in general just some comments throughout. 
- Sort out the `reset_filter` stuff in SensorFilter. I more or less ignored it.
- Tune the config values. Should set up an automatic system in webots.
- Fix the position error

**Things I'm Unsure About**
- The kinematics covariance. It's a 6 by 6 matrix. The test has values for it, but most of the elements are zero in it. I assume it's the noise in the kinematics? But I'm not sure how we would get that for each measurement. The best I can think of is to set it to always the same thing and have it as a config value. Atm I have set it to all zero in SensorFilter. 
- The spaces. At the moment I'm just assuming and flipping them to make things work. No idea if the state gives Htw or Hwt... 
- The feet order. I assume it goes right then left, based on the y position of the kinematics in the test data. Applicable to both kinematics and foot down info. Code for this should be cleaned up using eg body side enum or something so it's clear (once we figure it out).
- The test uses the previous gyro and acc measurements, not the current in each loop. I've set it to the current in SensorFilter but I'm not sure why the test is like that and if SensorFilter should do this as well.


ysims:
https://github.com/RossHartley/invariant-ekf/issues/9 addresses the kinematics covariance

ysims:
@Tom0Brien what do we want to do here - I think we should consolidate the branches so we just have the one. This branch was pretty much working, just the robot kept moving up/down. Would need to be brought up to date with main. Might be cleaner to merge it into yours then update with main.

ysims:
@Tom0Brien do we want to keep this up in case we want to use it at some point? Or get rid? If keep, I'd suggest doing as per my last comment (merge this into yours and bring yours up to date with main)

Tom0Brien:
> @Tom0Brien do we want to keep this up in case we want to use it at some point? Or get rid? If keep, I'd suggest doing as per my last comment (merge this into yours and bring yours up to date with main)

Can be closed, I have most up to date version of this on another branch, can come back to it later if we want to explore this idea again

PR #1005, opened 2022-12-30
Director Servos (Group) Update - ignore missing servos in map

ysims:
Fixes a bug that can occur when a servo group Task is made without all the servos in that group.
Mostly applicable to scripts. Makes it easier when running a sequence of scripts.
Ignores servos that don't exist in the message.

### DIFF
--- a/module/actuation/Servos/src/Servos.hpp
+++ b/module/actuation/Servos/src/Servos.hpp
@@ -39,7 +39,8 @@ namespace module::actuation {
         }
 
         /// @brief Creates a reaction that takes a servo wrapper task (eg LeftLeg) and emits a task for each servo.
-        /// Emits Done when all servo tasks are Done.
+        /// Emits Done when all servo tasks are Done. If individual servos in that wrapper task were not provided, then
+        /// it ignores them.
         /// @tparam Group is a servo wrapper task (eg LeftLeg)
         /// @tparam Elements is a template pack of Servos that Group uses
         template <typename Group, typename... Elements>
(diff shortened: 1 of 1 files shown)


module/actuation/Servos/src/Servos.hpp
```diff
@@ -58,10 +61,21 @@ namespace module::actuation {
...
+                        [&] {
+                            // Check if the Task includes this servo
+                            if (group.servos.contains(utility::actuation::ServoMap<Elements>::value)) {
+                                // Emit a Task for the servo if the map contains a value for it
+                                emit<Task>(std::make_unique<Elements>(
+                                    group.servos.at(utility::actuation::ServoMap<Elements>::value)));
+                            }
+                            else {  // if a servo was not filled in the map for this group, log it
+                                log<NUClear::DEBUG>("Requested a servo group Task but did not provide values for servo",
+                                                    utility::actuation::ServoMap<Elements>::value);
```
TrentHouliston, line 75: I'd drop this to TRACE, since it's probably going to show up a lot.
Also being that it's a smart enum is this printing the actual name rather than just a number?

TrentHouliston (dismissed):
Minor suggestion but approved otherwise

module/actuation/Servos/src/Servos.hpp
```diff
@@ -58,10 +61,21 @@ namespace module::actuation {
...
+                        // Runs an emit for each servo, if that servo was included in the map
+                        [&] {
+                            // Check if the Task includes this servo
+                            if (group.servos.contains(utility::actuation::ServoMap<Elements>::value)) {
+                                // Emit a Task for the servo if the map contains a value for it
+                                emit<Task>(std::make_unique<Elements>(
+                                    group.servos.at(utility::actuation::ServoMap<Elements>::value)));
+                            }
+                            else {  // if a servo was not filled in the map for this group, log it
+                                log<NUClear::DEBUG>("Requested a servo group Task but did not provide values for servo",
```
ysims, line 74:
```suggestion
                                log<NUClear::TRACE>("Requested a servo group Task but did not provide values for servo",
```

module/actuation/Servos/src/Servos.hpp
ysims: It's printing a number

TrentHouliston: In that case, you can change it to make it print out a string instead using the smart enums

TrentHouliston (approved):

PR #1006, opened 2022-12-30
Add in a Director-style kick planner module

ysims:
The logic is no different to what we have in SoccerStrategy currently. There is the addition of alignment with where we are kicking to, but it is currently disabled. There's also a config field to force a particular foot if one of the kicks is no good. A future idea could be that if we are forcing one foot, then wait until that foot in particular is aligned. 

Has been tested and works.

Minor fix to the KickScript module, config file is in lowercase because of a conflict with the other KickScript module.


ysims:
Can I get a review on this

module/planning/PlanKick/data/config/PlanKick.yaml
```diff
@@ -0,0 +1,18 @@
...
+# How long since the last ball measurement before ignoring it and assuming the ball is lost, in milliseconds
+timeout_threshold: 1000
+
+# Maximum distance of the ball required to kick the ball (in metres)
+ball_distance_threshold: 0.5
+# Maximum angle of the ball from the torso's x-axis required to kick the ball (in radians), both clockwise and anticlockwise
+ball_angle_threshold: 0.4
+
+# Whether to require aligning with the target
+align: false
```
TrentHouliston, line 13: I feel like this is redundant given you have a target angle threshold. Just set that threshold to pi and you have the same effect

module/planning/PlanKick/data/config/PlanKick.yaml
```diff
@@ -0,0 +1,18 @@
+# Controls the minimum log level that NUClear log will display
+log_level: DEBUG
+
+# How long since the last ball measurement before ignoring it and assuming the ball is lost, in milliseconds
+timeout_threshold: 1000
```
TrentHouliston, line 5: `ball_seen_timeout` might be a better name for this value

module/planning/PlanKick/src/PlanKick.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+                // Check if config specifies a leg
+                if (cfg.kick_leg == "left") {  // can only kick with left
+                    emit<Task>(std::make_unique<Kick>(LimbID::LEFT_LEG));
+                    return;
+                }
+                else if (cfg.kick_leg == "right") {  // can only kick with right
+                    emit<Task>(std::make_unique<Kick>(LimbID::RIGHT_LEG));
+                    return;
+                }
```
TrentHouliston, line 91: Should these be in the if statement above? Otherwise the if statement does nothing (you could just unconditionally set it to true and have no if)

module/planning/PlanKick/src/PlanKick.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+                // Else determine leg based on ball position
+                // If the ball is more to the left, then kick with the left leg
+                if (ball.rBTt.y() > 0.0) {
```
TrentHouliston, line 95: May as well combine these with the ones above and delete those
```suggestion
                if (cfg.kick_leg == "left" || ball.rBTt.y() > 0.0) {
```

module/skill/KickScript/src/KickScript.cpp
```diff
@@ -21,7 +21,7 @@ namespace module::skill {
     KickScript::KickScript(std::unique_ptr<NUClear::Environment> environment)
         : BehaviourReactor(std::move(environment)) {
 
-        on<Configuration>("KickScript.yaml").then([this](const Configuration& config) {
+        on<Configuration>("kickscript.yaml").then([this](const Configuration& config) {
```
TrentHouliston, line 24: Why has this been made lowercase?

shared/message/planning/KickTo.proto
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Copyright 2022 NUbots <nubots@nubots.net>
+ */
+syntax = "proto3";
+
+package message.planning;
+
+import "Vector.proto";
+
+message KickTo {
```
TrentHouliston, line 25: I know it's an extra transformation but I feel like it might be better to have this target in world space rather than torso space. The place you're kicking too will generally be a place in the world (won't rotate as the torso rotates and moves)

module/planning/PlanKick/src/PlanKick.hpp
```diff
@@ -0,0 +1,32 @@
...
+    class PlanKick : public ::extension::behaviour::BehaviourReactor {
+    private:
+        /// @brief Stores configuration values
+        struct Config {
+            float timeout_threshold       = 0.0;
+            float ball_distance_threshold = 0.0;
+            float ball_angle_threshold    = 0.0;
+            bool align                    = false;
+            float target_angle_threshold  = 0.0;
+            std::string kick_leg          = "";
```
TrentHouliston, line 19: Make this an enum (or smart enum I think we have a leg side enum?)

module/skill/KickScript/src/KickScript.cpp
ysims: because there is already a KickScript elsewhere

shared/message/planning/KickTo.proto
ysims: I don't think I can trust world space enough quite yet to do that

module/planning/PlanKick/src/PlanKick.cpp
ysims: This doesn't quite work, if it's more to the left but right has been forced in config, it'll kick with left still. We'd wants something like 
 ```suggestion
                if (cfg.kick_leg == "left" || (cfg.kick_leg == "auto" && ball.rBTt.y() > 0.0)) {
```

File: module/planning/PlanKick/src/PlanKick.hpp (hunk 7 above)
Comment (ysims) [reply module/planning/PlanKick/src/PlanKick.hpp:19]: Does yaml support enums?

File: shared/message/planning/KickTo.proto (hunk 6 above)
Comment (ysims) [reply shared/message/planning/KickTo.proto:25]: I still don't want world space here yet, none of our measurements are in world space at this point (ie goals). And this functionality is still in development. The aim should be to mostly move what we currently have in the codebase over. We don't really want to get to the end and have to rewrite it because we've gone overboard with changes.
Once we can align with vectors from the torso, then we can look at moving it to some other space.

File: module/skill/KickScript/src/KickScript.cpp (hunk 5 above)
Comment (TrentHouliston) [reply module/skill/KickScript/src/KickScript.cpp:24]: I think rename this module to `ScriptKick`, since it's implementing a scripted kick, not a kick script.

Comment (ysims) [reply module/skill/KickScript/src/KickScript.cpp:24]: I'll do this in another PR then, not here.

File: module/planning/PlanKick/src/PlanKick.cpp [hunk 8]
```diff
@@ -0,0 +1,93 @@
...
+                    emit<Task>(std::make_unique<Idle>());
+                    return;
+                }
+
+                // CHECK IF FACING POINT TO KICK TO
+                float align_angle = std::abs(std::atan2(kick.rPTt.y(), kick.rPTt.x()));
+
+                // Don't kick if we should align but we're not aligned to the target
+                if (align_angle > cfg.target_angle_threshold) {
+                    emit<Task>(std::make_unique<Idle>());
```
Crafty15, line 74: How do we plan on doing the goal alignment?

Crafty15 (commented):
LGTM.

module/planning/PlanKick/src/PlanKick.cpp
ysims: walk path planner. and you can make sure it works when we have the walk and anything else it needs. I've just put in rough things atm.

ysims:
Review pls

module/planning/PlanKick/src/PlanKick.cpp
```diff
@@ -0,0 +1,93 @@
...
+            // Use configuration here from file PlanKick.yaml
+            this->log_level             = config["log_level"].as<NUClear::LogLevel>();
+            cfg.ball_timeout_threshold  = config["timeout_threshold"].as<float>();
+            cfg.ball_distance_threshold = config["ball_distance_threshold"].as<float>();
+            cfg.ball_angle_threshold    = config["ball_angle_threshold"].as<float>();
+            cfg.target_angle_threshold  = config["target_angle_threshold"].as<float>();
+            cfg.kick_leg                = config["kick_leg"].as<std::string>();
+        });
+
+        on<Provide<KickTo>, Trigger<FilteredBall>, Needs<Kick>>().then(
```
TrentHouliston, line 36: The planner should not `Needs<Kick>` the idea would be that it wants to kick and tries by emitting a task. If it needs kick you can't even plan to kick unless the walk isn't running.

module/planning/PlanKick/src/PlanKick.cpp
```diff
@@ -0,0 +1,93 @@
...
+            [this](const KickTo& kick, const RunInfo& info, const FilteredBall& ball) {
+                // If the kicking subtask is done, then we are no longer kicking
+                if (info.run_reason == RunInfo::RunReason::SUBTASK_DONE) {
```
TrentHouliston, line 39: Im worried without needs kick this behaviour might not always fire (or at least it will need to be checked).

module/planning/PlanKick/src/PlanKick.cpp
```diff
@@ -0,0 +1,93 @@
...
+                    kicking = false;
+                }
+
+                // If we're already kicking, don't do anything
+                if (kicking) {
```
TrentHouliston, line 44: I'm not sure if we can, but we need a way so that if we were going to kick, and kick couldn't run, then the conditions to kick end before kick is possible the kick will be removed from the tasks queue

TrentHouliston (changes requested):
I think there are a few logic bugs that might make this not work properly when combined with other modules that use the body too.

module/planning/PlanKick/src/PlanKick.cpp
```diff
@@ -0,0 +1,90 @@
...
+            this->log_level             = config["log_level"].as<NUClear::LogLevel>();
+            cfg.ball_timeout_threshold  = config["timeout_threshold"].as<float>();
+            cfg.ball_distance_threshold = config["ball_distance_threshold"].as<float>();
+            cfg.ball_angle_threshold    = config["ball_angle_threshold"].as<float>();
+            cfg.target_angle_threshold  = config["target_angle_threshold"].as<float>();
+            cfg.kick_leg                = config["kick_leg"].as<std::string>();
+        });
+
+        on<Provide<KickTo>, Uses<Kick>, Trigger<FilteredBall>>().then(
+            [this](const KickTo& kick_to, const RunInfo& info, const Uses<Kick>& kick, const FilteredBall& ball) {
```
ysims, line 37:
```suggestion
            [this](const KickTo& kick_to, const Uses<Kick>& kick, const FilteredBall& ball) {
```

module/planning/PlanKick/src/PlanKick.cpp
```diff
@@ -0,0 +1,90 @@
...
+    using message::skill::Kick;
+    using utility::input::LimbID;
+    using utility::support::Expression;
+
+    PlanKick::PlanKick(std::unique_ptr<NUClear::Environment> environment) : BehaviourReactor(std::move(environment)) {
+
+        on<Configuration>("PlanKick.yaml").then([this](const Configuration& config) {
+            // Use configuration here from file PlanKick.yaml
+            this->log_level             = config["log_level"].as<NUClear::LogLevel>();
+            cfg.ball_timeout_threshold  = config["timeout_threshold"].as<float>();
```
ysims, line 29:
```suggestion
            cfg.ball_timeout_threshold  = config["ball_timeout_threshold"].as<float>();
```

ysims:
@TrentHouliston review pls

LachlanCourt (approved):
LGTM

module/planning/PlanKick/src/PlanKick.cpp
```diff
@@ -0,0 +1,103 @@
...
+            // If the kick conditions are not met, the function will have returned with no Tasks, ending the kick
+            // Otherwise, the kick conditions are met and we need to check if we are already kicking
+            // If we are already queued to kick, then only emit Idle to keep the previous Kick Task running
+            if (kick.run_state == GroupInfo::RunState::QUEUED) {
+                emit<Task>(std::make_unique<Idle>());
+                return;
+            }
+
+            // If the robot is not standing, make it stand before kicking
+            if (stability != Stability::STANDING) {
```
TrentHouliston, line 86: One day... pushing will get rid of this, I'm glad there is this workaround for now though
Manual pushing!

module/planning/PlanKick/src/PlanKick.hpp
```diff
@@ -0,0 +1,33 @@
...
+        struct Config {
+            float ball_timeout_threshold  = 0.0;
+            float ball_distance_threshold = 0.0;
+            float ball_angle_threshold    = 0.0;
+            float target_angle_threshold  = 0.0;
+            utility::input::LimbID kick_leg{};
+        } cfg;
+
+        /// @brief True if a Kick task has been requested, false otherwise. Resets to false when the Kick task is Done.
+        bool kicking = false;
```
TrentHouliston, line 24: Do we need this with the other PR you did for run state?

TrentHouliston (dismissed):
LGTM, but I think you were going to change modules like this to use the new Uses<> word

module/planning/PlanKick/src/PlanKick.hpp
ysims, line 24:
```suggestion
```

module/planning/PlanKick/src/PlanKick.cpp
```diff
@@ -0,0 +1,103 @@
...
+            // Use configuration here from file PlanKick.yaml
+            this->log_level             = config["log_level"].as<NUClear::LogLevel>();
+            cfg.ball_timeout_threshold  = config["ball_timeout_threshold"].as<float>();
+            cfg.ball_distance_threshold = config["ball_distance_threshold"].as<float>();
+            cfg.ball_angle_threshold    = config["ball_angle_threshold"].as<float>();
+            cfg.target_angle_threshold  = config["target_angle_threshold"].as<Expression>();
+            cfg.kick_leg                = config["kick_leg"].as<std::string>();
+        });
+
+        on<Provide<KickTo>, Uses<Kick>, Trigger<FilteredBall>, With<Stability>>().then([this](
```
TrentHouliston, line 40: Changed my mind... this should be `Trigger<Stability>`

TrentHouliston (changes requested):

ysims:
It does use the new Uses run state stuff, it's just the kicking variable was accidently left in.

module/planning/PlanKick/src/PlanKick.cpp
ysims, line 40:
```suggestion
        on<Provide<KickTo>, Uses<Kick>, Trigger<FilteredBall>, Trigger<Stability>>().then([this](
```
makes sense

TrentHouliston (approved):

TrentHouliston (approved):

PR #1007, opened 2023-01-01
Distribute the visual mesh runners across both cameras

ysims:
Rather than favouring the left camera. Will be useful when we get a fourth robot and have two robots with one camera, since it won't matter which side they are on.


Crafty15 (commented):
What is the outcome of doing this?

ysims:
Instead of running two threads on the left camera, run one thread on both. helpful if the robot does not have a left camera.

ysims:
#1039 replaces

PR #1009, opened 2023-01-09
Updates webots.py to search for roles in sub directories

Crafty15:
- webots and fake roles could no longer be found due be being moved to sub-directories.
- This requires that the full path to the role be used. For example: `./b webots build  roles/webots/robocup.role
`
- This [PR](https://github.com/NUbots/NUbook/pull/243) updates NUbook, but it needs updating for the full filepath


ysims (approved):
Lgtm

tools/webots.py
```diff
@@ -97,10 +97,12 @@ def register(command):
...
 def get_cmake_flags(roles_to_build):
-    roles_dir = os.path.join(b.project_dir, "roles")
-    roles_glob = os.path.join(roles_dir, "*.role")
 
-    available_roles = [Path(role_path).stem for role_path in glob.glob(roles_glob)]
+    available_roles = []
+    for dirpath, _, fnames in os.walk("roles"):  # iterate over the roles directory and all subdirectories
+        for f in fnames:  # iterate over all the files in the directory
+            if f.endswith(".role"):  # check that the file is a role file
+                available_roles.append(os.path.join(dirpath, f))  # add the role file to our list
```
CMurtagh-LGTM, line 105:
```suggestion
    for dirpath, _, fnames in os.walk("roles"):
        for f in fnames:
            if f.endswith(".role"):
                available_roles.append(os.path.join(dirpath, f))
```
Way too many comments.

CMurtagh-LGTM (changes requested):

tools/webots.py
TrentHouliston: All pretty self-documenting too!

Bidski: probably more simply achieved with a glob
```
from glob import glob

# ....

# Find all available roles
available_roles = glob(os.path.join("roles", "**", "*.role"), recursive=True)
```

Crafty15: Brilliant! Thanks

tools/webots.py
```diff
@@ -98,11 +98,8 @@ def register(command):
 
 def get_cmake_flags(roles_to_build):
 
-    available_roles = []
-    for dirpath, _, fnames in os.walk("roles"):  # iterate over the roles directory and all subdirectories
-        for f in fnames:  # iterate over all the files in the directory
-            if f.endswith(".role"):  # check that the file is a role file
-                available_roles.append(os.path.join(dirpath, f))  # add the role file to our list
+    # Find all available roles
+    available_roles = available_roles = glob(os.path.join("roles", "**", "*.role"), recursive=True)
```
CMurtagh-LGTM:
```suggestion
    available_roles = glob(os.path.join("roles", "**", "*.role"), recursive=True)
```
:stuck_out_tongue:

Also do we need to strip the `roles/` and `.role` from the filename https://github.com/NUbots/NUbots/pull/1010#discussion_r1068685321?

CMurtagh-LGTM (changes requested):

tools/webots.py
Bidski: Don't assume that the b script is always running from the project root directory, while that is commonly the case it doesn't have to be.

`available_roles = glob(os.path.join(b.project_dir, "roles", "**", "*.role"), recursive=True)`

Crafty15: `available_roles = available_roles = glob(os.path.join("roles", "**", "*.role"), recursive=True)`
I already fixed this.

Crafty15: > Don't assume that the b script is always running from the project root directory, while that is commonly the case it doesn't have to be.
> 
> `available_roles = glob(os.path.join(b.project_dir, "roles", "**", "*.role"), recursive=True)`

I'll add this when i can test it

Crafty15: 'twerks

CMurtagh-LGTM (approved):

PR #1014, opened 2023-02-01
Update webots camera FOV and Hpc

ysims:
Two problems with the Webots camera config
- FOV was taken from Webots, which is horizontal FOV, but the config is for diagonal FOV, so this has been updated
- Hpc looks like it was copied from a calibrated robot camera, in this PR it has been changed based on the model and given that the cameras in webots are positioned without any extra rotation relative to the last joint, the rotation is identity.

Original is

image

and with this update we have

image


Tom0Brien (approved):
LGTM!

PR #1020, opened 2023-02-12
Adds a Director-style Look skill

ysims:
This PR adds The Director Look skill, which is similar to the current HeadController module.

- Moved the angle bounds checking into the Kinematics module
- Made a Look message with a direction and smoothing bool
- Made a Look module that takes the direction and smooths if requested then emits a HeadIK Task

Has been tested with ball direction vector and it does follow the ball, functionality the same as the current stuff.

### DIFF
--- a/module/skill/Look/src/Look.hpp
+++ b/module/skill/Look/src/Look.hpp
@@ -1,0 +1,28 @@
+#ifndef MODULE_SKILL_LOOK_HPP
+#define MODULE_SKILL_LOOK_HPP
+
+#include <nuclear>
+
+#include "extension/Behaviour.hpp"
+
+namespace module::skill {
+
+    class Look : public ::extension::behaviour::BehaviourReactor {
+    private:
+        /// @brief Stores configuration values
+        struct Config {
+            float smoothing_factor = 0.0;
+            float head_gain        = 0.0;
+            float head_torque      = 0.0;
+        } cfg;
+
+        bool smooth = false;
+
+    public:
+        /// @brief Called by the powerplant to build and setup the Look reactor.
+        explicit Look(std::unique_ptr<NUClear::Environment> environment);
+    };
+
+}  // namespace module::skill
+
+#endif  // MODULE_SKILL_LOOK_HPP
--- a/module/skill/Look/src/Look.hpp
+++ b/module/skill/Look/src/Look.hpp
@@ -16,7 +16,8 @@ namespace module::skill {
             float head_torque      = 0.0;
         } cfg;
 
-        /// @brief Whether to smooth the head movements
+        /// @brief Store whether we are smoothing the head movements from the previous run, to help with transitioning
+        /// between smoothing and not smoothing. Smoothing is given by the message.
         bool smooth = false;
 
     public:
--- a/module/skill/Look/src/Look.hpp
+++ b/module/skill/Look/src/Look.hpp
@@ -16,6 +16,7 @@ namespace module::skill {
             float head_torque      = 0.0;
         } cfg;
 
+        /// @brief Whether to smooth the head movements
         bool smooth = false;
 
     public:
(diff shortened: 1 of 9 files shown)


PiiDish (approved):
LGTM

shared/message/skill/Look.proto
```diff
@@ -0,0 +1,30 @@
...
+ */
+syntax = "proto3";
+
+package message.skill;
+import "Vector.proto";
+
+/// A message telling the robot to look in the given direction
+message Look {
+    /// Vector from the camera to the point to look at, in camera space
+    vec3 rPCc = 1;
```
TrentHouliston, line 27: I would put this in world space instead of camera space, like rPCw.
The one that does things in camera space already exists as Head IK

shared/message/skill/Look.proto
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Copyright 2023 NUbots <nubots@nubots.net>
+ */
+syntax = "proto3";
+
+package message.skill;
+import "Vector.proto";
+
+/// A message telling the robot to look in the given direction
+message Look {
```
TrentHouliston, line 25: I think `LookAt` is more prescriptive as a task, and it would avoid name collisions in your module

module/skill/Look/src/Look.hpp
```diff
@@ -0,0 +1,29 @@
...
+    private:
+        /// @brief Stores configuration values
+        struct Config {
+            float smoothing_factor = 0.0;
+            float head_gain        = 0.0;
+            float head_torque      = 0.0;
+        } cfg;
+
+        /// @brief Whether to smooth the head movements
+        bool smooth = false;
```
TrentHouliston, line 20: Isn't this part of config?

ysims: No, but it seems my comment isn't clear enough so I'll fix that up. Whether smoothing is defined by the Look message. We store the last one so we know if we've switch between the two.

shared/message/skill/Look.proto
ysims: Kind of HeadIK but kind of not because this module also acts as a filter for smoothing the movement.

shared/message/skill/Look.proto
ysims: But we're not telling it what to look at, we're telling it where to look

shared/message/skill/Look.proto
TrentHouliston: Well at the very least, this can't be `rPCc`, because the camera space would move as you tried to align the target. it would need to be something like `rPCt`

ysims: I would say this is beyond the scope of the PR to fix the actual head controller implementation and instead an issue can be made. 
For anyone else reading this, for context: The current implementation doesn't work the way it is intended to work, and instead works as a derivative-style method. The input should be in torso space, but this causes issues with oscillation.

shared/utility/math/coordinates.hpp
```diff
@@ -81,6 +81,15 @@ namespace utility::math::coordinates {
...
+    /// @brief Converts the given vector from a cartesian direction vector to a 2d screen angular coordinates
+    /// @tparam T scalar template for the vectors
+    /// @param v The cartesian direction vector
+    /// @return The screen angular coordinates
+    template <typename T>
+    [[nodiscard]] inline Eigen::Matrix<T, 2, 1> screen_angular_from_object_direction(const Eigen::Matrix<T, 3, 1>& v) {
+        return {std::atan2(v.y(), v.x()), std::atan2(v.z(), v.x())};
+    }
+
 }  // namespace utility::math::coordinates
```
Crafty15, line 93: Maybe make an issue to write a unit test for this.

Crafty15 (approved):
LGTM.

PR #1024, opened 2023-02-16
Add VS Code debugging launchers for roles

Bidski:
This PR will generate a `.vscode/launch.json` file containing a launch configuration for each role that is currently being built that will allow those roles to be actively debugged inside of VS Code.

There are a couple of limitations to this

1. This will only work inside the devcontainer
2. Since the devcontainer currently does not expose any GPUs, no roles/modules that use a GPU can be debugged in this manner, you will have to fallback to the ./b run --gpus all --gdb role/to/debug method

### Pre-PR Checklist:

Have you

- [X] Added a descriptive title and relevant labels to the PR

### Reviewers, Note

Pull this branch and run `./b configure` followed by `./b build`. One of the build targets will generate `.vscode/launch.json` which will contain a launch configuration for all of the currently built roles.

Now switch to the `Run and Debug` tab in VS Code (or `Ctrl+Shift+D`) and you will be able to select a role to debug. You should also be able to add breakpoints in the editor (click just to the left of the line numbers and a red dot will appear).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,4 @@ tags
 *.sublime-workspace
 \.vscode/ipch
 .ccls-cache
+.vscode/launch.json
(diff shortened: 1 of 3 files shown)


ysims (approved):
cool

PR #1025, opened 2023-02-16
Updates the Look skill to use torso space and vector for smoothing

ysims:
The PR makes the Look skill somewhat more accurate.

- Uses rPCt instead of rPCc
- Smooths the vector, not the angle (no conversion to angles in the Look skill)

The changes result in

- The head looks directly at the ball, rather than before where it was not quite accurately looking at the ball particularly on the bounds of the head angles
- There is an oscillation issue around IK problems but the smoothing stops most of the movement

### DIFF
--- a/shared/message/skill/Look.proto
+++ b/shared/message/skill/Look.proto
@@ -23,7 +23,7 @@ import "Vector.proto";
 
 /// A message telling the robot to look in the given direction
 message Look {
-    /// Vector from the torso to the point to look at, in torso space
+    /// Vector from the camera to the point to look at, in torso space
     vec3 rPCt = 1;
     /// If true, smooth the goal angle using exponential smoothing
     bool smooth = 2;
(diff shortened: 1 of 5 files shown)


shared/message/skill/Look.proto
```diff
@@ -23,8 +23,8 @@ import "Vector.proto";
 
 /// A message telling the robot to look in the given direction
 message Look {
-    /// Vector from the camera to the point to look at, in camera space
-    vec3 rPCc = 1;
+    /// Vector from the torso to the point to look at, in torso space
+    vec3 rPCt = 1;
```
Tom0Brien, line 27: Comment should read "Vector from the camera to the point to look at, in torso space"

Tom0Brien: or rPTt if comment is correct

Tom0Brien (approved):
LGTM.

PR #1027, opened 2023-02-17
Adds a Director-style Look planner that can look for different features

ysims:
This PR adds in a Look planner. It can look for balls, goals and the field in general (for field lines).

The look will first determine if any of the requested features has been missing for too long, if it's been too long since a general field search, or if no features were requested. In these cases, the robot will conduct a search pattern.
If no search is required, the robot will fixate on the ball if ball search was requested. Otherwise, it will fixate on the goals unless goals were also not requested. If neither goals or ball were requested, then it performs the search pattern.

Some further changes

- The existing FilteredBall data was insufficient for calling the Look skill, and so rBCt has been included to the FilteredBall message
- The search pattern has been adjusted so that the robot can find balls at its feet (and further away)

The search pattern is conducted when a LookSearch Task is requested. This allows for easy transitions between search and non-search motions using the Start reaction to reset the index to 0.


module/skill/Look/src/Look.cpp
```diff
@@ -42,43 +41,30 @@ namespace module::skill {
             cfg.head_torque      = config["head_torque"].as<float>();
         });
 
-        on<Provide<LookTask>, Needs<HeadIK>, Trigger<Sensors>>().then(
-            [this](const LookTask& look, const Sensors& sensors) {
-                // Convert the look direction into angle requests
-                Eigen::Vector2d requested_angles(screen_angular_from_object_direction(look.rPCc.normalized()));
+        on<Provide<LookTask>, Needs<HeadIK>, Every<90, Per<std::chrono::seconds>>>().then([this](const LookTask& look) {
```
TrentHouliston, line 44: Why do you need the every? The only time anything should change it will be triggered

TrentHouliston: Oh right, smoothing.
In that case, instead of doing this did you want to Trigger<Sensors> and make it a closed loop?
Check the actual FK of head and interpolate it to the IK output

ysims: It's smoothed, so the targets for HeadIK change.

ysims: How did you even comment on this haha, this isn't in this PR :stuck_out_tongue:

TrentHouliston (approved):
lgtm

PR #1028, opened 2023-02-18
Move scripts from ScriptEngine to its own folder, for Director

ysims:
My current favourite solution to the Director scripts problem. 

Problem is that we don't have a centralised Scripts module (ie ScriptEngine) to dump all the scripts in with Director.

imo Director should allow us to call a script Task anywhere for whatever script (even a new one) without having to do more than including the script extension and emitting appropriately. 

The PR moves the ScriptEngine scripts into a folder (in utility, will be next to the scripts utility as in #1026), and adds a line in a cmake file that copies that folder into the build directory.


TrentHouliston (changes requested):
Good idea, but not the right place. I would instead put this in the libraries.cmake file that I think is in utility.

ysims:
> Good idea, but not the right place. I would instead put this in the libraries.cmake file that I think is in utility.

How does the location look now? The script folder will end up next to Script.hpp once the other script PR is in. I'm not sure if I've put the line in a good spot in the file or not though (if this is the right file).

TrentHouliston (approved):

PR #1030, opened 2023-02-21
Field Line Localisation

Tom0Brien:
# Overview
The following pull request adds a particle filter based localization method for estimating the robot's position and orientation (x, y, theta) in field space. The method relies solely on field line observations for measurement correction. The basic idea is to create n_particles from an initial distribution, which can be thought of as guesses for where the robot is. Next, the particle filter updates the guesses' position based on odometry input and checks how well the observations overlap with a known occupancy map of the field lines. We then resample the particles based on their weights, essentially discarding any particles which don't agree with the map. This process is ran iteratively, which results in all particles converging on a region which is most likely to be where the robot is.

This currently works great in Webots when the robot is stationary/moved using the grab tool or when the robot is walking while being fed ground truth odometry data from WeBots. However, when walking around with estimated odometry, I believe the current noise introduced is too large for the filter to handle.

The method can be broken down into multiple stages, visually illustrated in the diagram below:
image

### Intialization
At intilization, n particles are initialized around an initial guess of the state sampled from a multivariate normal distribution. A visual illustration of 20 particles being initialized around the centre of the field is shown below:

### Time Update
The time update updates all the particles current state based on odometry measurement, which currently uses just raw walk command as "odometry". We then add noise to all the particles (process noise) to account for uncertainty in the model. Adding noise also helps the filters performance.

### Measurement Update
The measurement update weights each particle based on how well the observations match with the field line occupancy map by converting each observation to an index in the map and retrieving the occupancy value. We then sum up all the occupancy values in a way such that maximum overlap results in a weight of 1. Additionally, we penalise the particle for being outside the map or having observations outside the map.

**TODO:** 
- [ ] Figure out how to implement new frame of reference which is on the field directly below robot, with x forward and y left of the robot, which will be referred to as as ground space. (Currently using world frame with yaw removed.)

### Resampling
Based on how well each particle is weighted determines how likely it is to be kept or discarded during a process known as resampling. 

The result of a few iterations of the filter after the measurement update and resampling is shown below:

## Field Line Map

An occupany grid of the field lines is used when weighting each particle in the measurement update. This map is generated dynamically based on the field line description on startup of the localisation module. An example of the field line map generated is shown below:

image

## Future Improvements
- Use goals for measurement update, could potentially help if at any point the robot falls or filter diverges.
- Methods for intialising and resetting filter, for example, when the robot is picked up and moved.
- Use actual odometry for time update
- Implement using ParticleFilter.hpp. The module is currently implemented standalone because I wanted to build everything from the ground up so I knew exactly what was happening. To do this I think all we'd need to do is change ParticleFilter.hpp slightly to accept custom measurement models for weighting particles.
- Fix localisation view in NUsight, it is currently displaying odometry and field dimensions are old


module/localisation/Localisation/src/Localisation.cpp
```diff
@@ -0,0 +1,432 @@
...
+                (fd.dimensions.border_strip_min_width + fd.dimensions.field_width / 2) / cfg.grid_size;
+            int centre_circle_r = (fd.dimensions.center_circle_diameter / 2) / cfg.grid_size;
+            fieldline_map.add_circle(centre_circle_x0, centre_circle_y0, centre_circle_r, line_width);
+
+            // Fill the surrounding cells close to the field lines with decreasing occupancy values
+            fieldline_map.fill_surrounding_cells(0.25 / cfg.grid_size);
+
+            // Save the map to a csv file
+            if (cfg.save_map) {
+                std::ofstream file("recordings/fieldline_map.csv");
```
ysims, line 176: it's not reeeally a recording - where is this relative to? NUbots or build?

module/localisation/Localisation/src/Localisation.cpp
```diff
@@ -0,0 +1,432 @@
...
+            }
+        }
+
+        // Ensure the weight is positive
+        weight = std::max(weight, 0.0);
+
+        return weight;
+    }
+
+    Eigen::Vector3d Localisation::compute_mean() {
```
ysims, line 356: I feel like a lot of these functions could be utilities.

module/localisation/Localisation/src/Localisation.cpp
```diff
@@ -0,0 +1,432 @@
...
+        const auto current_time = NUClear::clock::now();
+        const double dt         = duration_cast<duration<double>>(current_time - last_time_update_time).count();
+        last_time_update_time   = current_time;
+
+        // Update the particles using the walk command and the time since the last time update, with some tunable
+        // scaling
+        for (size_t i = 0; i < particles.size(); i++) {
+            double delta_x         = walk_command.x() * dt * cfg.scale_x;
+            double delta_y         = walk_command.y() * dt * cfg.scale_y;
+            double delta_theta     = walk_command.z() * dt * cfg.scale_theta;
```
ysims, line 388: hmm it'd be good to decouple the odometry walk command stuff from localisation. it isn't critical tho.

module/localisation/Localisation/src/Map.hpp
```diff
@@ -0,0 +1,165 @@
+#ifndef MODULE_LOCALISATION_MAP_HPP
```
ysims, line 1: break this up into hpp and cpp, it might also be good to put into utility

module/support/logging/DataLogging/data/config/DataLogging.yaml
```diff
@@ -8,5 +8,6 @@ output:
 # The below ones are just the most common and here as an example
 messages:
   message.output.CompressedImage: false
+  message.vision.FieldLines: false
```
ysims, line 11: revert

module/support/logging/DataPlayback/data/config/DataPlayback.yaml
```diff
@@ -16,5 +16,6 @@ buffer_time: 1500
 # The below ones are just the most common and here as an example
 messages:
   message.output.CompressedImage: false
+  message.vision.FieldLines: false
```
ysims, line 19: revert

module/vision/FieldLineDetector/README.md
```diff
@@ -0,0 +1,9 @@
+# FieldLineDetector
```
ysims, line 1: fill in

module/vision/FieldLineDetector/src/FieldLineDetector.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+        on<Configuration>("FieldLineDetector.yaml").then([this](const Configuration& cfg) {
+            // Use configuration here from file FieldLineDetector.yaml
+            this->log_level = cfg["log_level"].as<NUClear::LogLevel>();
+
+            this->config.confidence_threshold = cfg["confidence_threshold"].as<float>();
+            this->config.cluster_points       = cfg["cluster_points"].as<int>();
+        });
+
+        on<Trigger<GreenHorizon>, Buffer<2>>().then("Field Line Detector", [this](const GreenHorizon& horizon) {
```
ysims, line 38: how come we have the buffer?

module/vision/VisualMesh/data/config/docker/VisualMesh.yaml
```diff
@@ -0,0 +1,19 @@
...
+#   cache_directory: directory to save compiled opencl binaries
+#   spotlight:
+#     distance: the distance range for the spotlight mesh to be generated for
+#     radius: the search radius to do at that distance (NOT THE RADIUS OF THE OBJECT)
+
+cameras:
+  Left:
+    concurrent: 2
+    network: config/networks/RobocupNetwork.yaml
+    engine: opencl
```
ysims, line 14: hmmm if we're adding in a docker config it shouldn't be on the opencl engine. I'm not sure if it makes sense to have a docker config. if it's not nugus or webots, there shouldn't be camera data.

module/vision/VisualMesh/data/config/webots/VisualMesh.yaml
```diff
@@ -9,11 +9,11 @@
...
-    concurrent: 2
+    concurrent: 4
     network: config/networks/WebotsNetwork.yaml
     engine: cpu
     cache_directory: "" # unneeded for cpu engine
     classifier:
       height: [0.5, 1.0]
-      max_distance: 15
-      intersection_tolerance: 0.25
+      max_distance: 10
```
ysims, line 18: maybe 11? diagonal of the field is 10.82

roles/test/sensor.role
```diff
@@ -12,4 +12,5 @@ nuclear_role(
   platform::cm740::HardwareIO
   input::SensorFilter
   output::Overview
+  support::logging::DataPlayback
```
ysims, line 15: remove

roles/webots/localisation.role
```diff
@@ -0,0 +1,44 @@
+nuclear_role(
```
ysims, line 1: is this any different to webots/robocup?

ysims: I guess it's robocup and keyboardwalk?

shared/message/support/FieldDescription.proto
```diff
@@ -72,6 +72,10 @@ message FieldDescription {
         double center_circle_diameter = 16;
 
         double border_strip_min_width = 17;
+
+        double penalty_area_length = 18;
+
+        double penalty_area_width = 19;
     }
```
ysims, line 79: you don't need to do all of them, but probably add in documentation directives for the ones you added at least.

shared/message/vision/FieldLines.proto
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * The NUbots Codebase is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright 2013 NUbots <nubots@nubots.net>
```
ysims, line 17:
```suggestion
 * Copyright 2023 NUbots <nubots@nubots.net>
```

ysims (commented):
:bread:

TrentHouliston:
The purpose of world space (odometry space) is so that every single localisation after that (ball, field etc) doesn't have to constantly move all the particles in a time update.
All you need to do is estimate the position of the field in wold space and then let odometry work separately.

If odometry is so bad that it's unusable currently and the walk command is more reliable, then just make sensors odometry output the integration of the walk command until such a time that you can get a better odometry working.

module/localisation/Localisation/src/Localisation.cpp
```diff
@@ -0,0 +1,432 @@
...
+
+    using extension::Configuration;
+
+    using message::localisation::Field;
+    using message::motion::DisableWalkEngineCommand;
+    using message::motion::EnableWalkEngineCommand;
+    using message::motion::ExecuteGetup;
+    using message::motion::KillGetup;
+    using message::motion::StopCommand;
+    using message::motion::WalkCommand;
```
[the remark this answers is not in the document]
TrentHouliston: Use odometry from sensors as it's supposed to be used. And if it's not good enough then just make sensors output the integration of walk commands etc

roles/webots/localisation.role
Tom0Brien: It's just keyboardwalk with localisation on, so you can move the robot around while debugging/testing localisation

module/localisation/Localisation/src/Localisation.cpp
Tom0Brien: build, not sure where else to put it currently.

module/localisation/Localisation/src/Localisation.cpp
Tom0Brien: I agree, but at that point, we may as well move implementation over to use a general filter implementation within utility such as the current ParticleFilter.hpp, but I'd say thats out of scope of this PR atm.

module/vision/FieldLineDetector/src/FieldLineDetector.cpp
Tom0Brien: Not exactly sure, I imagine this module was first just duplicate of BallDetector which has the buffer. Goal detector has it as well.

Also in regards to this module, (Note: I haven't delved into what `partition_points` and `cluster_points` exactly do), from reading the comments their may be an easier way to extract field line points from mesh, since clustering isn't required here.

module/localisation/Localisation/src/Localisation.cpp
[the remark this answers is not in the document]
Tom0Brien: I think some further discussion on how we best go about integrating odometry is needed. I don't see clearly how to integrate it. My current thoughts are that we don't want to use our current odometry for more than z-height, roll, pitch which should be fairly accurate. (We then use this for projecting field line detection's onto the field plane).

module/vision/VisualMesh/data/config/webots/VisualMesh.yaml
```diff
@@ -9,11 +9,11 @@
 
 cameras:
   left_camera:
-    concurrent: 2
+    concurrent: 4
```
LachlanCourt, line 12: Was this just for testing or are we upping this going forward?

ysims:
@Tom0Brien can you make an issue with the future tasks as a tasklist? and anything brought up in reviews that you thing would be better in follow up PRs? we don't want this to be too complicated and then we can do smaller PRs from here improving the system. Also a NUbook page!!

module/input/SensorFilter/src/SensorFilter.hpp
```diff
@@ -44,6 +44,22 @@ namespace module::input {
...
         utility::math::filter::UKF<double, MotionModel> motionFilter{};
 
+        /// @brief Imu filter initial state [roll, pitch]
+        Eigen::Vector2d imu_state = Eigen::Vector2d::Zero();
+        /// @brief Imu filter initial covariance matrix
+        Eigen::Matrix2d P = Eigen::Matrix2d::Identity();
+        /// @brief Measurement noise  covariances
+        Eigen::Matrix2d R = 0.01 * Eigen::Matrix2d::Identity();
+        /// @brief Process noise covariance
+        Eigen::Matrix2d Q = 0.001 * Eigen::Matrix2d::Identity();
```
ysims, line 54: config values?

module/input/SensorFilter/src/SensorFilter.hpp
```diff
@@ -44,6 +44,22 @@ namespace module::input {
...
+
+        /// @brief Runs a Kalman filter for roll and pitch using accelerometer and gyroscope measurements
+        /// @param gyro Gyroscope measurements
+        /// @param acc Accelerometer measurements
+        /// @param dt Time since the last update
+        /// @return Updates imu_state and P with the new state and covariance
+        void run_imu_filter(Eigen::Vector3d gyro, Eigen::Vector3d acc, const double dt);
```
ysims, line 61: I would probably argue for this to be in a utility. Don't you have a generic simple kalman filter somewhere?

ysims (commented):
just got to be careful not to make SensorFilter more messy. I would argue making a new PR for updating SensorFilter

ysims (approved):
I think this is good enough to get in now, and then can do some smaller PRs to improve it.
@Tom0Brien have a look at the issue I put up, I went from the future work part on the main PR comment and some things from the review comments but can discuss if needed

PR #1031, opened 2023-03-01
Toggle Groups of Roles

### LINKED ISSUE #651: Flag for configuring only a set of roles

CMurtagh-LGTM:
Fixes #651

Adds flags to `configure` to set and unset groups of roles.


ysims:
I've had a play around with it - I did expect unset_roles to unset all of them. writing out individual roles might as well use `-i`. The main thing here is that it's annoying after `--clean` that everything is turned on. I'd like to be able to do `./b configure [--clean] --unset_roles --set_webots_roles` to turn on just the webots roles.

ysims:
@JesseWilliamson I've had a look at this and I think what I missed last time is that `--unset_all_roles` is a thing. 
If I do
`./b configure [--clean] --unset_all_roles --set_webots/_roles`
it works as I expect
we'll want to remove the trailing `/` on the folder names

tools/configure.py
```diff
@@ -1,11 +1,28 @@
...
 
 import b
 from utility.dockerise import run_on_docker
+from utility.roles import all_role_names, role_folders, dir_role_names
 from utility.shell import WrapPty
 
+ROLE_GROUPS = {
+    "all" : all_role_names(),
+    "cool" : ["robocup", "test/camera"],
+    "nice" : ["scripttuner", "natnet"],
```
ysims, line 14: these can be removed, assume it was test/debugging code

tools/utility/roles.py
```diff
@@ -0,0 +1,21 @@
...
+
+def all_role_names():
+    return dir_role_names("")
+
+def role_folders():
+    # Find all role folders
+    dnames = glob(os.path.join(b.project_dir, "roles", "**/"), recursive=True)
+    dnames.pop(0)
+    # Strip everything from file paths except role name and subdirectory in roles/
+    return [os.path.splitext(d.replace(os.path.join(b.project_dir, "roles", ""), ""))[0] for d in dnames]
```
ysims, line 21: in this part you'll need to remove the trailing `/`

TrentHouliston:
If you're going to set it up like this maybe use glob rules to set and unset too?

JesseWilliamson:
> If you're going to set it up like this maybe use glob rules to set and unset too?

Added. I'm curious why you said globbing rather than regex? I thought globbing was mostly just used for file names.
Also this kind of makes the other path-based arguments (--set_all_roles, --set_{dir}_roles) redundant, but I will leave them in so globbing isn't required.

tools/utility/roles.py
JesseWilliamson: I figured those were intentional to differentiate directories from custom groups like the cool_roles and nice_roles but I've removed them now

TrentHouliston:
Nobody wants to type out regex on the command line as arguments. And most of the time what you need can be represented with a glob. Another option is just to specify the roles you want to build directly as role file arguments. Then you can use the file system and shell to do globbing etc for you

ysims (approved):
Sweet! Looks good to me! Do you wanna put up a message on slack letting the team know about this new feature once it is merged. And will need an update to the build system page on NUbook

PR #1032, opened 2023-03-05
Add no task/running/queued information to Uses 

ysims:
This PR adds in information so a provider can find out if its task is already running or queued or not sent yet.

Pretty good for planners, since we don't want to go requesting a skill again when we already just requested it! ie Kick and Get Up planners.

Stops us having to keep track in the reactor if the kick or get up has already been requested.

Should add this to NUbots/NUbook#122.


module/extension/Director/tests/uses_state.cpp
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * The NUbots Codebase is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with the NUbots Codebase.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright 2022 NUbots <nubots@nubots.net>
```
ysims, line 17:
```suggestion
 * Copyright 2023 NUbots <nubots@nubots.net>
```

module/extension/Director/tests/uses_state_root.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+                    if (primary.run) {
+                        events.push_back("emitting subtask");
+                        emit<Task>(std::make_unique<SubTask>());
+                    }
+                    else {
+                        events.push_back("removing subtask");
+                        emit<Task>(std::unique_ptr<SubTask>(nullptr));
+                    }
+                });
```
TrentHouliston, line 68: To really test the root provider stuff properly, split this into 3x reactions triggered on separate messages

One that emits the task
One that emits the nullptr
And one that just reads the uses state

That way you can check that multiple different non provider reactions all behave as if they are a single "root" provider

module/extension/Director/src/information/get_group_info.cpp
```diff
@@ -23,10 +23,39 @@ namespace module::extension {
...
+
+        // If it is a root provider, get the root provider id
+        if (!providers.contains(reaction_id)) {
+            r_id = get_root_provider(root_type)->id;
+        }
+
         if (groups.contains(type)) {
-            return GroupInfo{groups.at(type).done};
+            // Check if the task is active
+            if (groups.at(type).active_task != nullptr) {
```
TrentHouliston, line 40: Pull the groups.at(type) out into a `auto& group =` since you use it a few times here

module/extension/Director/src/information/get_group_info.cpp
```diff
@@ -23,10 +23,39 @@ namespace module::extension {
...
+                // Check if the task is being run by this reaction
+                if (groups.at(type).active_task->requester_id == r_id) {
+                    return GroupInfo{GroupInfo::RunState::RUNNING, groups.at(type).done};
+                }
+            }
+
+            // Check if this task is in the reaction's task list
+            // If it is, then it is waiting to be run
+            if (providers.contains(r_id)) {
```
TrentHouliston, line 49: You know this is true from your earlier checks, you can probably make a `auto&` for it right after the root_type stuff

TrentHouliston (changes requested):
Looks pretty good, just a few things to cleanup, and a little more breadth in the root task unit test

TrentHouliston (approved):

PR #1035, opened 2023-03-12
Adds a strategy to stop walking and stand still

ysims:
This is useful for when we're penalised, in set or initial states and just don't want to move.

It checks if stability is standing and if not it emits a 0 walk command and if it is it emits stand.

### DIFF
--- a/module/strategy/StandStill/data/config/StandStill.yaml
+++ b/module/strategy/StandStill/data/config/StandStill.yaml
@@ -1,0 +1,2 @@
+# Controls the minimum log level that NUClear log will display
+log_level: DEBUG
--- a/module/strategy/StandStill/src/StandStill.cpp
+++ b/module/strategy/StandStill/src/StandStill.cpp
@@ -8,7 +8,7 @@
 #include "message/skill/Walk.hpp"
 #include "message/strategy/StandStill.hpp"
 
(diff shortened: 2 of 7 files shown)


LachlanCourt (dismissed):

module/strategy/StandStill/src/StandStill.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include "utility/skill/Script.hpp"
+
+namespace module::strategy {
+
+    using extension::Configuration;
+    using message::actuation::LimbsSequence;
+    using message::behaviour::state::Stability;
+    using message::skill::Walk;
+    using utility::motion::load_script;
```
ysims, line 19:
```suggestion
    using utility::skill::load_script;
```

module/strategy/StandStill/data/config/StandStill.yaml
```diff
@@ -0,0 +1,2 @@
+# Controls the minimum log level that NUClear log will display
+log_level: DEBUG
```
ysims, line 2:
```suggestion
log_level: INFO
```

ysims:
This should be good to review now

LachlanCourt (approved):
LGTM

Tom0Brien (approved):

PR #1036, opened 2023-03-12
Updates and Fixes for existing Director modules

ysims:
Some existing Director modules have issues. I uncovered them while working on getting an end to end system (robocup striker) working. 

- The FallingRelaxPlanner is not very compatible with Webots. The script to run for relaxing is set in config and is the Stand script for Webots and the Relax script for the real robot. Webots cannot run the relax script - it causes the robot to break in Webots. The tuning values were also somewhat premature and this fixes them so we don't fall over at the start of the program.
- Some logs were very spammy - i.e. FallingRelaxPlanner's falling log and Script utility's script loading log.
- The script utility was in motion instead of skill. It has been moved to skill.
- The PlanLook emits were not emitted as Tasks.
- The PlanLook was only running once and the Optional FilteredBall doesn't trigger the reaction, so an Every was added to update with any new FilteredBall info
- The Relax scripts weren't in our usual yaml format - they have been updated


shared/utility/skill/scripts/docker/Relax.yaml
```diff
@@ -1,28 +1,82 @@
...
-        { "id": "R_ANKLE_ROLL", "position": 0, "gain": .NaN, "torque": 0 },
-        { "id": "L_ANKLE_ROLL", "position": 0, "gain": .NaN, "torque": 0 },
-      ],
-  },
-]
+- duration: 25
+  targets:
+    - id: R_SHOULDER_PITCH
+      position: 0
+      gain: .NaN
```
LachlanCourt, line 5: Is the gain supposed to be NaN here? I would have thought it should be 0 too? Or is this a divide by zero thing?

LachlanCourt (dismissed):

shared/utility/skill/scripts/docker/Relax.yaml
ysims: pretty sure it's what it needs to be to relax the servos - not sure if 0 does the same thing but I can test

module/planning/FallingRelaxPlanner/src/FallingRelaxPlanner.cpp
```diff
@@ -93,21 +94,22 @@ namespace module::planning {
                 // We are falling
                 if (falling) {
                     // We are falling! Relax the limbs!
-                    log<NUClear::INFO>("Falling:",
-                                       "Gyroscope Magnitude:",
-                                       gyro_mag_state == State::FALLING    ? "FALLING"
-                                       : gyro_mag_state == State::UNSTABLE ? "UNSTABLE"
+                    log<NUClear::TRACE>("Falling:",
```
TrentHouliston, line 97: I think this is more of a debug line. Keep it as debug and change the config to be info or higher by default. Debug should be log lines that help you debug behaviour. Info lines should be for people. Trace should be for debugging code

TrentHouliston (commented):
I think some of the lines you made trace should still be debug

module/planning/FallingRelaxPlanner/src/FallingRelaxPlanner.cpp
ysims: you mean should be changed to debug, they were both originally info

TrentHouliston: Yes that

LachlanCourt (dismissed):
LGTM

TrentHouliston (approved):

TrentHouliston (approved):

PR #1041, opened 2023-03-13
Default modules to using INFO log level

ysims:
Probably makes more sense.

From @TrentHouliston 
> Trace should be if you're debugging code, so things that are only relevant if you would need to change a line of code to fix it
> Debug should be to debug behaviour, so a configuration change could make it work properly, or you just need to understand what it's doing
> Info is things that you would normally want to know when a module is running
> Warning is when something has gone wrong but the system can still perform it's intended function
> Error is when something has broken and the module cannot do it's intended behaviour (this time)
> Fatal is when something has broken and this module will never be able to work again

NUbook update at NUbots/NUbook#277.


Tom0Brien (approved):
This makes a lot of sense.

PR #1042, opened 2023-03-13
Add merge queue github actions

ysims:
I don't know if this is right. Pls work.


TrentHouliston (commented):
Needs to have this on the other workflows that run on pull_request too

TrentHouliston (approved):
Does it work?

PR #1043, opened 2023-03-14
Refactor the Director Look planner into a planner and strategy

ysims:
The planner includes only the LookAround (search).
The strategy includes the LookAtBall and LookAtGoals

This is to match better with the rest of the system, where WalkToBall is a strategy and TurnOnSpot is a planner.

The strategies only fixate on the ball/goals if they exist.

Future strategies will involve a FindBall which uses the LookAround and TurnOnSpot to find the ball.
A WalkToBall will only walk to the ball if the ball is found.
Then in the purpose level, if the WalkToBall and LookAtBall cannot run, the lower priority FindBall will take over and find the ball.

### DIFF
--- a/module/planning/PlanLook/README.md
+++ b/module/planning/PlanLook/README.md
@@ -2,23 +2,14 @@
 
 ## Description
 
-Plans how to look for features.
-
-LookAtBall looks for the ball and fixates on it if it has been seen.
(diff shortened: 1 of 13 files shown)


CMurtagh-LGTM (approved):
LGTM

Issue #10, opened 2015-06-25
Walkengine needs to filter incoming walk command somehow


Jade-Fountain:
The walk path planner sends varying walk speed commands to the walkengine which may be causing a lack of smoothness in the leg motions

josiahw:
The walk skill should be in charge of this. If it's done in the walk engine, we'll end up with the overshoot problems we had in the old walk. You need the component that smooths the walk commands to have some notion of the future plans.

ysims:
Implemented in #1061 in Director

PR #1045, opened 2023-03-14
Initialise Nusense HardwareIO

jpptm:
Things the PR does go here

- Adds necessary HardwareIO code for NUsense, our very own subcontroller for the robots

### Pre-PR Checklist:

Have you

### Things left to do

- [x] Test functionalities as we go so we don't go insane debugging later on
- [x] Replace c code with c++

### Required functionalities
- [x] Serialising protobuf to nbs msg and vice versa
- [x] Spam msgs 90 Hz
- [x] Stream reactor


module/platform/NUsense/HardwareIO/data/config/nugus1/HardwareIO.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+bus_reset_wait_time_us: 0
+byte_wait: 1200
+packet_wait: 100000
+
+servos:
+  - # [0]  R_SHOULDER_PITCH
+    direction: -1
+    offset: 0.0
+    simulated: false
```
TrentHouliston, line 14: This should all be done on the NUsense I think

module/platform/NUsense/HardwareIO/data/config/nugus1/HardwareIO.yaml
```diff
@@ -0,0 +1,92 @@
+battery:
+  charged_voltage: 16.8
+  nominal_voltage: 14.8
+  flat_voltage: 14.0
+
+bus_reset_wait_time_us: 0
+byte_wait: 1200
+packet_wait: 100000
```
TrentHouliston, line 8: These don't make sense for this

module/platform/NUsense/HardwareIO/data/config/nugus1/HardwareIO.yaml
```diff
@@ -0,0 +1,92 @@
+battery:
+  charged_voltage: 16.8
+  nominal_voltage: 14.8
+  flat_voltage: 14.0
```
TrentHouliston, line 4: NUsense itself will handle all this

module/platform/NUsense/HardwareIO/src/NUsense/SIProcessor.cpp
```diff
@@ -0,0 +1,59 @@
...
+        packet_data.push_back(0xE2);
+        packet_data.push_back(0x98);
+        packet_data.push_back(0xA2);
+
+        // Serialise protobuf message to string then add everything to our vector
+        std::string protobuf_bytes = msg.SerializeToString();
+
+        for (const auto& byte : protobuf_bytes) {
+            packet_data.push_back(static_cast<uint8_t>(byte));
+        }
```
TrentHouliston, line 22: Use `vector.insert` function instead of doing a loop

module/platform/NUsense/HardwareIO/src/NUsense/SIProcessor.cpp
```diff
@@ -0,0 +1,59 @@
...
+        // We insert everything to a string stream
+        std::vector<uint8_t> packets;
+
+        // Add header
+        packet_data.push_back(0xE2);
+        packet_data.push_back(0x98);
+        packet_data.push_back(0xA2);
+
+        // Serialise protobuf message to string then add everything to our vector
+        std::string protobuf_bytes = msg.SerializeToString();
```
TrentHouliston, line 18: This is going to be a Neutron not a protobuf, so you need to use the NUClear serialisation method

shared/message/platform/RawSensorsV2.proto
```diff
@@ -0,0 +1,194 @@
...
+ message ButtonLeftDown {}
+
+ message ButtonLeftUp {}
+
+ message ButtonMiddleDown {}
+
+ message ButtonMiddleUp {}
+
+ // Reset sensors and servo targets in Webots
+ message ResetWebotsServos {}
```
TrentHouliston, line 194: I feel like these belong elsewhere

module/platform/NUsense/HardwareIO/src/NUsense/SIProcessor.cpp
jpptm: Does the nuclear one agree with the protobuf one? I just wanna make sure as nanopb agrees with vanilla protobuf only

shared/message/platform/RawSensorsV2.proto
jpptm: They definitely do 👌 Thank you!

jpptm:
TODO open /dev/ttyACM0 or USB0 (config val) using the NUClear uart utility for it and send dummy messages for testing

docker/Dockerfile
```diff
@@ -358,8 +358,10 @@ RUN chmod 440 /etc/sudoers.d/user
 # Create the user, and setup sudo so no password is required
 ARG user_uid=1000
 ARG user_gid=$user_uid
-RUN groupadd --gid $user_gid nubots \
-    && useradd --uid $user_uid --gid $user_gid -m nubots
+ARG dialout_gid=20
+RUN groupadd --gid $dialout_gid dialout \
+    && groupadd --gid $user_gid nubots \
+    && useradd --uid $user_uid --gid nubots -G dialout -m nubots
```
Bidski, line 364:
```suggestion
ARG dialout_gid=20

# useradd -G is not sufficient to add the nubots user to the dialout 
# group as docker manages to lose supplementary groups when changing 
# to the nubots user.
# Also need to add "--group-add dialout" to the docker run command in run.py
RUN groupadd --gid $dialout_gid dialout \
    && groupadd --gid $user_gid nubots \
    && useradd --uid $user_uid --gid nubots -G dialout -m nubots
```

tools/utility/dockerise/run.py
```diff
@@ -144,6 +144,8 @@ def _run(**kwargs):
             "--env",
             f"EDITOR={os.environ.get('EDITOR', 'nano')}",
             "--privileged",
+            "--group-add",
+            "dialout",
```
Bidski, line 148:
```suggestion
            # Also need to create the group in the Dockerfile
            "--group-add",
            "dialout",
```

module/platform/NUsense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -0,0 +1,55 @@
...
+        });
+
+        on<Startup>().then("NUSense HardwareIO Startup", [this] {
+            // Make sure the port is closed before we open it
+            if (nusense.connected()){
+                nusense.close();
+            }
+
+            // If /dev/ttyACM0 is acting up, follow the link below
+            // https://stackoverflow.com/questions/40951728/avrdude-ser-open-cant-open-device-dev-ttyacm0-device-or-resource-busy
```
Bidski, line 27:
```suggestion
```

module/platform/NUsense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -0,0 +1,55 @@
...
+            nusense.open(cfg.nusense.port, cfg.nusense.baud);
```
Bidski, line 28: I would issue caution when setting the baud rate, I don't think the NUSense is using a baud rate and I am not sure what would happen if you try to set one?

Bidski (changes requested):

module/platform/NUsense/HardwareIO/src/HardwareIO.cpp
jpptm: I'm not sure whether we should remove it. I put the link there because it was a fix for nix based machines not seeing the port when we connect the USB to it

tools/utility/dockerise/run.py
TrentHouliston: Also, you can't do this, it makes any computer (i.e. mine) no longer able to build the code at all.

TrentHouliston (changes requested):

tools/utility/dockerise/run.py
jpptm: @TrentHouliston  May I ask why? @Claegtun has been testing and building with this branch and it worked.

TrentHouliston: I'm gonna guess he doesn't use a Mac

jpptm: He uses pop afaik, Im not really sure what I should do instead of this so it builds with anything. Is there a dialout equivalent for mac?

Claegtun: > He uses pop afaik, Im not really sure what I should do instead of this so it builds with anything. Is there a dialout equivalent for mac?

I actually use Mint.

> @TrentHouliston May I ask why? @Claegtun has been testing and building in this branch with these lines and it worked.

Also, I don't think that I have been building from this branch as is. I had to merge with main to get it to build properly. I am not sure if that is to what @TrentHouliston is referring.

Bidski: How does this break it on Mac? Is it just that dialout is the wrong group?

TrentHouliston: Dialout group doesn't exist so it just fails

Bidski: According to this you shouldn't need it on the host?
https://github.com/docker/cli/issues/4714#issuecomment-1881571231

Bidski: Have you rebuilt the image?

Bidski: Although there looks like there is a change to platform.py that is missing (but this missing should probably work in your favour)

module/platform/NUsense/HardwareIO/src/HardwareIO.cpp
JohanneMontano: NUSense just apparently ignores the baud rate set by this.

module/platform/NUsense/HardwareIO/src/util/packet_handler.cpp
Claegtun, line 1: Change the file's name to Pascal case, i.e. PacketHandler.cpp.

module/platform/NUsense/HardwareIO/src/util/packet_handler.hpp
Claegtun, line 1: Change the file's name to Pascal case, i.e. PacketHandler.hpp.

Claegtun (commented):
Should remove all the debug-logs and clean this up so that it is ready to be merged or to merge with Dex's branch. 

Also, the PacketHandler code is very rough and ready given that it is code copied from NUSense. It will have to change quite a bit before this will get merged with main, e.g. the `memcpy`'s were a C-like compromise because of ST's function being written in C. But given that this is entirely C++, it should be `std::copy`. 

Also, I am not sure whether a ring-buffer is needed. We used the ring-buffer in the NUSense since the callback needed to be short, and the decoding was done elsewhere. However, given that both the receiving and the decoding are done in the same scope, i.e. `on<IO>`, it seems very redundant. So, we may just have one function that takes `nusense_usb_bytes`.

However, for the sake of getting this merged with Dex's branch and then working rough and ready only within this branch, I may allow the ring-buffer, the `memcpy`'s, etc. since it is working, and I want us to push forward.

So, watch this space before merging with main.

JohanneMontano:
NOTE: This must go in main together with pr #10 in NUcontroller, otherwise it wont work

JohanneMontano:
People should now be able to  `git pull` safely

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -0,0 +1,222 @@
...
+                    log<NUClear::DEBUG>(fmt::format("    pvelo: {}", val.present_velocity));
+                    log<NUClear::DEBUG>(fmt::format("     ppos: {}", val.present_position));
+                    log<NUClear::DEBUG>(fmt::format("     gpwm: {}", val.goal_pwm));
+                    log<NUClear::DEBUG>(fmt::format("    gcurr: {}", val.goal_current));
+                    log<NUClear::DEBUG>(fmt::format("    gvelo: {}", val.goal_velocity));
+                    log<NUClear::DEBUG>(fmt::format("     gpos: {}", val.goal_position));
+                    log<NUClear::DEBUG>(fmt::format("  voltage: {}", val.voltage));
+                    log<NUClear::DEBUG>(fmt::format("     temp: {}", val.temperature));
+                }
+            }
```
ysims, line 186: theres no need to wrap this in an if

shared/utility/reactor/StreamReactor.hpp
```diff
@@ -427,9 +436,15 @@ namespace utility::reactor {
          * passes it to the parser. Once the parser finds a packet it emits that packet.
          */
         void process(const int& fd) {
+            std::stringstream debug_string;
```
ysims, line 439:
```suggestion
            std::stringstream debug_string{};
```

module/platform/NUSense/HardwareIO/src/NUSenseParser.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <vector>
+
+namespace module::platform::NUSense {
+
+    struct NUSenseFrame {
+        std::array<uint8_t, 3> header;  //< NBS Header, should be 0xE2, 0x98, 0xA2
+        uint32_t size;                  //< Size of the payload after this field
+        uint64_t timestamp;             //< Timestamp of the message
+        uint64_t hash;                  //< Hash of the message type
+        std::vector<uint8_t> payload;   //< The payload of the message
```
ysims, line 16: initialise?

shared/utility/strutil/strutil.hpp
```diff
@@ -1,125 +1,280 @@
...
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
- * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
- * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
- * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
- * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
- * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
- * SOFTWARE.
- */
-
```
ysims, line 27: add licenses

ysims (commented):
Check that you have the latest HardwareIOs from OpenCR for the robots. There's been some changes recently.

shared/utility/reactor/StreamReactor.hpp
```diff
@@ -443,6 +458,9 @@ namespace utility::reactor {
                     log<NUClear::WARN>(ex.what());
                 }
             }
+            if (log_level <= NUClear::TRACE) {
+                log<NUClear::TRACE>("Read:", debug_string.str());
+            }
```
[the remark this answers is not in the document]
TrentHouliston: This one is here on purpose, should probably have a comment.
Since this will be dumping so much data to log you really don't want to be doing all the work involved in making a log line for every single read/write you get.

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
TrentHouliston: As a counter point to my other comment with the trace line, this one doesn't need the if statement

module/platform/NUSense/HardwareIO/src/HardwareIO.hpp
```diff
@@ -59,7 +59,10 @@ namespace module::platform::NUSense {
         NUgus nugus{};
 
     private:
-        /// @brief Send message to StreamReactor so it can write the data to NUSense using uart
+        /// @brief Send a TransmitData message containing an nbs packet to StreamReactor so it can write the data to
+        /// NUSense.
+        /// @tparam packet A const reference to a protobuf message of type T.
+        ///                This gets serialised and turned to a vector of bytes before sending to NUSense.
```
ysims, line 65: tparam is the template parameter (`@tparam T`), different to normal param (`@param packet`)

module/platform/NUSense/HardwareIO/src/HardwareIO.hpp
```diff
@@ -0,0 +1,104 @@
...
+        NUgus nugus{};
+
+    private:
+        /// @brief Send a TransmitData message containing an nbs packet to StreamReactor so it can write the data to
+        /// NUSense.
+        /// @tparam packet A const reference to a protobuf message of type T.
+        ///                This gets serialised and turned to a vector of bytes before sending to NUSense.
+        template <typename T>
+        void send_packet(const T& packet) {
+            // Serialize the packet
```
ysims, line 68:
```suggestion
            // Serialise the packet
```

module/platform/NUSense/HardwareIO/src/NUSenseParser.hpp
```diff
@@ -0,0 +1,40 @@
...
+    };
+
+    struct NUSenseParser {
+        /**
+         * Parses bytes from NUSense one byte at a time and returns a NUSenseFrame when a complete frame is read
+         *
+         * @param byte The next byte to parse
+         *
+         * @return A NUSenseFrame if a complete frame has been read, otherwise nullptr
+         */
```
ysims, line 26:
```suggestion
         /// @brief Parses bytes from NUSense one byte at a time and returns a NUSenseFrame when a complete frame is read
         /// @param byte The next byte to parse
         /// @return A NUSenseFrame if a complete frame has been read, otherwise nullptr
```

module/platform/NUSense/HardwareIO/src/HardwareIO.hpp
JohanneMontano: aha gotcha thank you thats makes more sense

shared/utility/strutil/strutil.hpp
```diff
@@ -29,97 +29,307 @@
...
+     * @brief Checks if a string ends with a specific substring.
+     *
+     * This function compares the given string `str` with the specified `ending` substring
+     * and determines whether `str` ends with `ending`. The comparison is case-sensitive.
+     *
+     * @param str The string to check for the ending substring.
+     * @param ending The substring to search for at the end of the string.
+     * @return `true` if `str` ends with `ending`, `false` otherwise.
+     */
     // http://stackoverflow.com/a/874160/1387006
```
ysims, line 57:
```suggestion
    /**
     * @brief Checks if a string ends with a specific substring.
     *
     * This function compares the given string `str` with the specified `ending` substring
     * and determines whether `str` ends with `ending`. The comparison is case-sensitive.
     * http://stackoverflow.com/a/874160/1387006
     *
     * @param str The string to check for the ending substring.
     * @param ending The substring to search for at the end of the string.
     * @return `true` if `str` ends with `ending`, `false` otherwise.
     */
```

shared/utility/strutil/strutil.hpp
```diff
@@ -29,97 +29,307 @@
...
+     * @brief The overloaded version of ends_with which takes vectors of strings instead of single strings.
+     *
+     * This function compares the given string `str` with the specified `ending` substring in the vector endings
+     * and determines whether `str` ends with `ending`. The comparison is case-sensitive.
+     *
+     * @param str The string to check for the ending substring.
+     * @param ending The substring to search for at the end of the string.
+     * @return `true` if `str` ends with `ending`, `false` otherwise.
+     */
+    // http://stackoverflow.com/a/874160/1387006
```
ysims, line 73: this should also be in the main function description

ysims (approved):

module/input/FakeCamera/src/FakeCamera.cpp
```diff
@@ -82,7 +82,7 @@ namespace module::input {
 
                     // Traverse specified directory and locate all image/lens pairs
                     for (const auto& p : std::filesystem::recursive_directory_iterator(config.image_folder)) {
-                        if (utility::strutil::endsWith(p.path(), std::vector<std::string>{".jpg", ".jpeg"})) {
+                        if (utility::strutil::ends_with(p.path(), std::vector<std::string>{".jpg", ".jpeg"})) {
```
Bidski, line 85: Seems unrelated to this PR

module/platform/NUSense/HardwareIO/data/config/nugus1/HardwareIO.yaml
Bidski, line 1: Remove things from this file that are identical to the parent file, same with the other robot-specific configs

module/platform/NUSense/HardwareIO/src/NUSenseParser.cpp
```diff
@@ -0,0 +1,64 @@
+#include "NUSenseParser.hpp"
+
+namespace module::platform::NUSense {
+
+    uint32_t read_le_32(const uint8_t* ptr) {
```
Bidski, line 5:
```suggestion
    /**
     * @brief .....
     */
    uint32_t read_le_32(const uint8_t* ptr) {
```

module/platform/NUSense/HardwareIO/src/NUSenseParser.cpp
```diff
@@ -0,0 +1,64 @@
...
+        return (uint32_t(ptr[0]) << 0) | (uint32_t(ptr[1]) << 8) | (uint32_t(ptr[2]) << 16) | (uint32_t(ptr[3]) << 24);
+    }
+
+    uint64_t read_le_64(const uint8_t* ptr) {
```
Bidski, line 9:
```suggestion
    /**
     * @brief .....
     */
    uint64_t read_le_64(const uint8_t* ptr) {
```

module/platform/NUSense/HardwareIO/src/NUSenseParser.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include <array>
+#include <cstdint>
+#include <memory>
+#include <vector>
+
+namespace module::platform::NUSense {
+
+    struct NUSenseFrame {
+        std::array<uint8_t, 3> header{};  //< NBS Header, should be 0xE2, 0x98, 0xA2
```
Bidski, line 12: Same for the other fields
```suggestion
        /// NBS Header, should be 0xE2, 0x98, 0xA2
        std::array<uint8_t, 3> header{};
```

module/platform/NUSense/HardwareIO/src/NUgus.cpp
```diff
@@ -0,0 +1,33 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+#include "NUgus.hpp"
+
+namespace module::platform::NUSense {
+
+    NUgus::NUgus() {}
```
Bidski, line 31: Not a lot of point in having this here? Not much point in having a default constructor at all if it is just going to be empty

module/tools/FirmwareInstaller/src/FirmwareInstaller.cpp
```diff
@@ -56,8 +56,8 @@ namespace module::tools {
 
             for (const auto& f : config["firmwares"].config) {
                 std::pair<std::string, std::string> name;
-                name.first  = utility::strutil::toUpper(f["device"].as<std::string>());
-                name.second = utility::strutil::toUpper(f["battery"].as<std::string>());
+                name.first  = utility::strutil::to_upper(f["device"].as<std::string>());
+                name.second = utility::strutil::to_pper(f["battery"].as<std::string>());
```
Bidski, line 60: Seems unrelated to this PR?

Bidski (changes requested):

module/platform/NUSense/HardwareIO/data/config/nugus1/HardwareIO.yaml
jpptm: I think @ysims can explain this better than me but I think we put each robot specific config in hwio because as far as I know, we do alias matching stuff with the robot configs now so that we don't have to replace config values every time a NUC is taken out or replaced

Bidski: The default config should have all values set to sane defaults. The robot-specific configs should only set the values that differ from the default config.

module/platform/NUSense/HardwareIO/src/NUSenseParser.hpp
TrentHouliston: ///< is a totally viable alternate doxygen syntax when you want to be more concise!

module/platform/NUSense/HardwareIO/data/config/nugus1/HardwareIO.yaml
TrentHouliston: I would also say name this file NUSense.yaml as it's going to make your life easier when trying to build for multiple robots

module/tools/FirmwareInstaller/src/FirmwareInstaller.cpp
TrentHouliston: Also that just seems like a bug

jpptm: Cant believe this one got past the build checks lol, should get fixed by tomorrow hopefully

module/platform/NUSense/HardwareIO/src/NUSenseParser.hpp
Bidski: Consistency trumps "valid alternative style"

module/tools/FirmwareInstaller/src/FirmwareInstaller.cpp
ysims: This is because this module isn't in any roles, so it's not being built.

module/platform/NUSense/HardwareIO/data/config/nugus1/HardwareIO.yaml
ysims: What @bidski is saying has nothing to do with the robot name config type. 
Configs are merged from default to more specific, so if a field is always the same for all robots, ie everything that isn't under `servos` here, can be put in default and not included here.

I'm not sure if renaming to NUSense.yaml will do much for building to robots. But it would mean we wouldn't need to restrict to only one HardwareIO module being built and can have roles for each explicitly for CI.  The other thing is that these and the OpenCR HardwareIOs will need to be the same, and updating the same thing in two places is annoying.

module/platform/NUSense/HardwareIO/src/NUSenseParser.hpp
ysims: I kinda like it for short things

module/input/FakeCamera/src/FakeCamera.cpp
JohanneMontano: StreamReactor uses `dedent` which relies on other functions in strutil like `join` and the like. I will try to be consistent but I think I will have to definitely add stuff into strutil and make it a part of this pr otherwise it wont compile

module/platform/NUSense/HardwareIO/data/config/nugus1/HardwareIO.yaml
JohanneMontano: My bad I just didnt understand it at all, not enough coffee

module/platform/NUSense/HardwareIO/src/NUgus.cpp
JohanneMontano: Removed.

module/platform/NUSense/HardwareIO/src/NUSenseParser.hpp
JohanneMontano: I chose consistency

module/tools/FirmwareInstaller/src/FirmwareInstaller.cpp
JohanneMontano: Reverted.

module/platform/NUSense/HardwareIO/data/config/nugus1/HardwareIO.yaml
JohanneMontano: Addressed.

module/platform/NUSense/HardwareIO/src/NUSenseParser.cpp
```diff
@@ -1,11 +1,18 @@
 #include "NUSenseParser.hpp"
 
 namespace module::platform::NUSense {
-
+    /// @brief  Read a buffer of bytes and convert it to an unsigned 32 bit integer in little endian
+    ///         HwIO uses this to read the size field of the nbs packet sent by NUSense
+    /// @param  ptr Pointer to a buffer of bytes
+    /// @return The unsigned 32 bit number decoded from the buffer of bytes
```
Bidski, line 7:
```suggestion
    /**
     * @brief  Read a buffer of bytes and convert it to an unsigned 32 bit integer in little endian
     *         HwIO uses this to read the size field of the nbs packet sent by NUSense
     * @param  ptr Pointer to a buffer of bytes
     * @return The unsigned 32 bit number decoded from the buffer of bytes
     */
```

shared/utility/strutil/strutil.hpp
Bidski, line 1: Why are these changes here?

Bidski (approved):

shared/utility/strutil/strutil.hpp
JohanneMontano: Its because StreamReactor uses dedent which was not in this file in main, but dedent relied on other functions such as join so I had to add those in as well

ysims (approved):

PR #1046, opened 2023-03-15
Fix the role search in the Webots build command for the virtual competitions

ysims:
The script accepts a role name as command line input. It searches all the role files and checks if the requested role exists.

The old role search found the full path of the roles. This means the script fails to find the role in the available list. If you input the full path, then the part where the script turns the role on and builds it fails because it expects the role path and name relative to the role folder. 

This PR fixes it so the role search stores the role names relative to the folder and without the `.role` suffix. 

Also changed the image tag from robocup2021 to robocup.


JesseWilliamson (approved):
LGTM

PR #1047, opened 2023-03-16
Tune the ready state configuration values for webots

ysims:
Hasn't been updated since making a faster walk. Made it walk faster (otherwise it walks on the spot) and reduced the time to walk in ready so it stops before the center.

High priority for the virtual season.


LachlanCourt (approved):
LGTM

PR #1049, opened 2023-03-16
Add a webots global configuration file for the virtual season team ID

ysims:
We updated the team ID to 1 for last RoboCup, but there should still be a webots-specific configuration file with the team ID for the virtual games.


LachlanCourt (approved):
LGTM

PR #1052, opened 2023-03-20
Add a director-style ready module

ysims:
Part of the old system is walking to ready, so this module handles that walk to ready Task. In the ready state, the robot needs to walk for a configured time and then stop.


Tom0Brien (approved):
Lgtm

PR #1053, opened 2023-03-21
Fixes a director bug where a task does not get removed when its sibling takes over

ysims:
Added a test - fails on main - and added a fix to get working. Makes a DirectorTask which lives in Director and contains the BehaviourTask info as well as Director algorithm specific functionality.

### DIFF
--- a/module/extension/Director/src/Director.cpp
+++ b/module/extension/Director/src/Director.cpp
@@ -269,7 +269,7 @@ namespace module::extension {
             // Root level task, make the pack immediately and send it off to be executed as a root task
             if (!providers.contains(task->requester_id)) {
                 // Get the root provider from the task root type
-                auto root_provider = get_root_provider(task->root_type);
+                auto root_provider = get_root_provider(t.root_type);
 
                 // Modify the task we received to look like it came from this provider
                 task->requester_id = root_provider->id;
--- a/module/extension/Director/src/Director.cpp
+++ b/module/extension/Director/src/Director.cpp
@@ -23,6 +23,9 @@
 
 namespace module::extension {
 
+    using component::DirectorTask;
+    using component::Provider;
+    using component::ProviderGroup;
     using ::extension::Configuration;
     using ::extension::behaviour::RunInfo;
     using ::extension::behaviour::commands::BehaviourTask;
@@ -31,8 +34,6 @@ namespace module::extension {
     using ::extension::behaviour::commands::ProviderDone;
     using ::extension::behaviour::commands::ProvideReaction;
     using ::extension::behaviour::commands::WhenExpression;
-    using provider::Provider;
-    using provider::ProviderGroup;
     using Unbind = NUClear::dsl::operation::Unbind<ProvideReaction>;
 
 
@@ -118,7 +119,7 @@ namespace module::extension {
         }
     }
 
-    std::shared_ptr<provider::Provider> Director::get_root_provider(const std::type_index& root_type) {
+    std::shared_ptr<component::Provider> Director::get_root_provider(const std::type_index& root_type) {
         // Create a root provider for this task if one doesn't already exist and use it
         if (!groups.contains(root_type)) {
             groups.emplace(root_type, ProviderGroup(root_type));
@@ -160,7 +161,7 @@ namespace module::extension {
             }
 
             // The when condition object
-            auto w  = std::make_shared<provider::Provider::WhenCondition>(when.type, when.validator, current);
+            auto w  = std::make_shared<component::Provider::WhenCondition>(when.type, when.validator, current);
             auto id = when.reaction->id;
 
             // Add a reaction that will listen for changes to this state and notify the director
@@ -262,8 +263,8 @@ namespace module::extension {
         // the Provider has finished executing
         on<Trigger<BehaviourTask>>().then("Director Task", [this](const BehaviourTask& t) {
             std::lock_guard<std::recursive_mutex> lock(director_mutex);
-            // Make our own mutable copy of the task
-            auto task = std::make_shared<BehaviourTask>(t);
+            // Make our own mutable director task from the behaviour task
+            auto task = std::make_shared<DirectorTask>(t, false);
 
             // Root level task, make the pack immediately and send it off to be executed as a root task
             if (!providers.contains(task->requester_id)) {
--- a/module/extension/Director/src/Director.cpp
+++ b/module/extension/Director/src/Director.cpp
@@ -264,7 +264,7 @@ namespace module::extension {
         on<Trigger<BehaviourTask>>().then("Director Task", [this](const BehaviourTask& t) {
             std::lock_guard<std::recursive_mutex> lock(director_mutex);
             // Make our own mutable director task from the behaviour task
-            auto task = std::make_shared<DirectorTask>(t, false);
+            auto task = std::make_shared<DirectorTask>(t);
 
             // Root level task, make the pack immediately and send it off to be executed as a root task
             if (!providers.contains(task->requester_id)) {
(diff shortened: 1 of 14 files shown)


module/extension/Director/src/Director.cpp
```diff
@@ -262,8 +263,8 @@ namespace module::extension {
         // the Provider has finished executing
         on<Trigger<BehaviourTask>>().then("Director Task", [this](const BehaviourTask& t) {
             std::lock_guard<std::recursive_mutex> lock(director_mutex);
-            // Make our own mutable copy of the task
-            auto task = std::make_shared<BehaviourTask>(t);
+            // Make our own mutable director task from the behaviour task
+            auto task = std::make_shared<DirectorTask>(t, false);
```
TrentHouliston, line 267: I would just make false be a default argument on the DirectorTask constructor rather than passing it in

module/extension/Director/src/component/DirectorTask.hpp
```diff
@@ -0,0 +1,109 @@
...
+         * @param requester_id_         the id of the reaction that emitted this task
+         * @param requester_task_id_    the task_id of the reaction task that emitted this
+         * @param data_                 the task data to be sent to the provider
+         * @param name_                 a string name for this task for use in debugging
+         * @param priority_             the priority that this task is to run with
+         * @param optional_             whether this task is optional or not
+         * @param dying_                whether this task is currently about to be removed
+         */
+        DirectorTask(std::type_index type_,
+                     std::type_index root_type_,
```
TrentHouliston, line 70: I think that at the point where you make this DirectorTask you no longer need the root type...

module/extension/Director/src/component/DirectorTask.hpp
```diff
@@ -0,0 +1,109 @@
...
+            , name(name_)
+            , priority(priority_)
+            , optional(optional_)
+            , dying(dying_) {}
+
+        /// The Provider type this task is for
+        std::type_index type;
+        /// A secondary provider type to use if this is a root task
+        std::type_index root_type;
+        /// The reaction id of the requester (could be the id of a Provider)
```
TrentHouliston, line 92: Will be a provider (or a root provider) at this point

module/extension/Director/src/Director.cpp
```diff
@@ -23,6 +23,9 @@
 
 namespace module::extension {
 
+    using component::DirectorTask;
```
TrentHouliston, line 26: I think rather than component, director might make more sense (then you can also just name it `director::Task`)

module/extension/Director/src/component/DirectorTask.hpp
```diff
@@ -0,0 +1,109 @@
...
+     */
+    struct DirectorTask {
+
+        /**
+         * Construct a new Task object for use in the director algorithm using a given BehaviourTask
+         *
+         * @param task                  the BehaviourTask to use as a base
+         * @param optional_             whether this task is optional or not
+         */
+        DirectorTask(::extension::behaviour::commands::BehaviourTask task, bool dying_) : dying(dying_) {
```
TrentHouliston, line 45:
```suggestion
        DirectorTask(const ::extension::behaviour::commands::BehaviourTask& task) : dying(false) {
```

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -146,6 +147,38 @@ namespace module::extension {
...
+                                       "has active task",
+                                       NUClear::util::demangle(g.active_task->type.name()));
+                }
+                else {
+                    log<NUClear::WARN>("-", NUClear::util::demangle(g.type.name()), "has no active task");
+                }
+            }
+        }
+
+
```
TrentHouliston, line 181: You left some log lines in
```suggestion
```

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -240,19 +273,56 @@ namespace module::extension {
...
         group.watch_handles.clear();
 
+        // This new task pack we are trying to run might be different from the task pack previously ran.
+        // There might be some tasks that weren't emitted in the new task pack.
+        // These tasks need to be removed from the places they were running.
+        // This is at the end, so that if one of our new tasks uses something the old tasks did then we won't run other
+        // tasks with lower priority for a single cycle until this one runs
+        for (const auto& t : group.subtasks) {
+            if (normal_striker_log) {
+                log<NUClear::WARN>("\t-", NUClear::util::demangle(t->type.name()), "was running, let's check");
```
TrentHouliston, line 283: More log lines through here

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -240,19 +273,56 @@ namespace module::extension {
...
         auto run_level      = run_tasks(group, TaskList(tasks.begin(), first_optional), RunLevel::OK);
 
         // If we are not running normally now and were previously we might still have active tasks that need removing
         if (run_level != RunLevel::OK) {
             for (auto& t : group.subtasks) {
+                if (normal_striker_log) {
+                    log<NUClear::WARN>("\t-",
+                                       NUClear::util::demangle(t->type.name()),
+                                       ": something is not okay and the task is being removed");
+                }
```
TrentHouliston, line 316:
```suggestion
```

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -240,19 +273,56 @@ namespace module::extension {
...
                 remove_task(t);
             }
         }
 
         // Run each of the optional tasks but only to the level the main task ran or pushed
         if (run_level <= RunLevel::PUSH) {
+            if (normal_striker_log) {
+                log<NUClear::WARN>("PUSHING");
+            }
```
TrentHouliston, line 325:
```suggestion
```

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -268,18 +338,19 @@ namespace module::extension {
             }
         }
 
-        // This new task pack we are trying to run might be different from the task pack previously ran.
-        // There might be some tasks that weren't emitted in the new task pack.
-        // These tasks need to be removed from the places they were running.
         for (const auto& t : group.subtasks) {
```
TrentHouliston, line 341: Probably should put a comment here as to why it's being removed after (to stop the single frame run)

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -268,18 +338,19 @@ namespace module::extension {
...
             }
         }
 
+        // print tasks
+        if (normal_striker_log) {
+            log<NUClear::WARN>("Tasks:");
+            for (const auto& t : tasks) {
+                log<NUClear::WARN>("\t-", NUClear::util::demangle(t->type.name()));
+            }
+        }
```
TrentHouliston, line 353:
```suggestion
```

module/extension/Director/tests/change_subtask.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+#include "Director.hpp"
+#include "TestBase.hpp"
+#include "util/diff_string.hpp"
+
+// Anonymous namespace to avoid name collisions
+namespace {
+
+    struct ParentTask {
+        ParentTask(bool subtask) : subtask_a(subtask) {}
```
TrentHouliston, line 31: Does this even compile?
Converting a bool to an int, and the variable name is wrong

TrentHouliston (changes requested):

module/extension/Director/tests/change_subtask.cpp
ysims: Teaches me to rush to get a PR up before my laptop dies while I'm half asleep lol

module/extension/Director/src/Director.cpp
ysims: there's already a director folder with implementation for some director functions
it seems not right to put the task inside that folder and add a director namespace when everything else isn't

TrentHouliston: True

PR #1056, opened 2023-04-07
Adds a Director style striker module

ysims:
Creates a Director Striker purpose module that walks to the ball and kicks it. Does pretty much what the old soccer strategy does. 

There'll be another layer on top of this that decides what soccer position the robot does and also does some generic things like fall management.


module/purpose/Striker/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Add this module to play as a striker in soccer!
+
+## Consumes
+
+- `message::input::GameState` to get information about the state of the game
+- `message::input::GameState::Data::Phase` to get specific information about the current game phase (initial, ready, set, playing, etc).
+- `message::strategy::Striker` a Task requesting to play as a Striker
+- `message::strategy::NormalStriker` a Test requesting to play as a Striker in the normal game state
+- `message::strategy::PenaltyShootoutStriker` a Test requesting to play as a Striker in the penalty shootout game state
```
Crafty15, line 23: Task?

module/purpose/Striker/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Emits
+
+- `message::strategy::NormalStriker` a Test requesting to play as a Striker in the normal game state
```
Crafty15, line 27: Task?

module/purpose/Striker/README.md
```diff
@@ -0,0 +1,36 @@
...
+- `message::strategy::PenaltyShootoutStriker` a Test requesting to play as a Striker in the penalty shootout game state
```
Crafty15, line 28: Task?

module/purpose/Striker/src/Striker.cpp
```diff
@@ -0,0 +1,94 @@
...
+        // Default for INITIAL, READY, SET, FINISHED, TIMEOUT
+        on<Provide<PenaltyShootoutStriker>>().then([this] { emit<Task>(std::make_unique<StandStill>()); });
+    }
+
+    void Striker::play() {
+        // Walk to the ball and kick!
+        emit<Task>(std::make_unique<FindBall>(), 1);    // if the look/walk to ball tasks are not running, find the ball
+        emit<Task>(std::make_unique<LookAtBall>(), 2);  // try to track the ball
+        emit<Task>(std::make_unique<WalkToBall>(), 3);  // try to walk to the ball
+        emit<Task>(std::make_unique<KickTo>(Eigen::Vector3f::Zero()), 4);  // kick the ball if possible
```
Crafty15, line 91: What's this syntax with the 1,2,3,4? A priority thing?

shared/message/strategy/Ready.proto
```diff
@@ -20,10 +20,5 @@ syntax = "proto3";
 
 package message.strategy;
 
-import "Vector.proto";
-
 /// Ready state strategy for walking onto the field
-message Ready {
-    /// The target position to walk to relative to the field
-    vec3 rRFf = 1;
```
Crafty15, line 28: Could we use a target position now that we have localisation?

Crafty15 (dismissed):

module/purpose/Striker/README.md
ysims:
```suggestion
- `message::strategy::PenaltyShootoutStriker` a Task requesting to play as a Striker in the penalty shootout game state
```

module/purpose/Striker/README.md
ysims:
```suggestion
- `message::strategy::NormalStriker` a Task requesting to play as a Striker in the normal game state
```

module/purpose/Striker/README.md
ysims:
```suggestion
- `message::strategy::PenaltyShootoutStriker` a Task requesting to play as a Striker in the penalty shootout game state
```

shared/message/strategy/Ready.proto
ysims: yes we can but not doing it in this PR, can come in a future one

module/purpose/Striker/src/Striker.cpp
ysims: yes priority

shared/message/strategy/Ready.proto
ysims: I didn't want to have things lying around that weren't being used, like this. It might not be the best way to organise localisation based ready. We probably want a different message and module, something like `WalkToFieldPosition`. And if we keep the old basic ready, we can easily switch it in if localisation stops working.

module/purpose/Striker/src/Striker.cpp
ysims: I have added in a comment above these since priority is only really used in a few places higher up so ppl may not be super familiar.

PiiDish (approved):
LGTM

PR #1057, opened 2023-04-10
Update the getup module to use the RunState Uses information

ysims:
Getup was made before the Uses RunState information was added, so it had a boolean to track if the getup was running. This updates it to use the Uses information instead. 
It also wasn't really working when I was testing some other stuff in the conversion from the old system and this works :shrug:


PiiDish (approved):
LGTM

PR #1058, opened 2023-04-12
Adds a Director-style soccer purpose module

ysims:
This starts the graph for the soccer scenario (ie robocup). 

It always runs the fall recovery in the background at a higher priority to the other children. 
It assumes the robot starts in a standing state and emits a stability standing message (the fall modules will update this message if it is on the ground to start with). This is needed since some modules rely on this emit to trigger.

The module checks the config and runs the appropriate purpose task (Striker etc) based on that. This FindPurpose could be changed to something more complex in the future. 

It handles penalisation by making the robot stop and stand still if it is penalised, and then rerunning FindPurpose when it is unpenalised.  

Also handles button press to force play.


PiiDish (dismissed):
Looks OK to me???

PR #1059, opened 2023-04-12
Add Director-style Dive and DiveToBall modules for Goalie behaviour

ysims:
Two modules created for goalie behaviour:

- A Dive skill that Dives either left or right when requested and emits Done when completed
- A DiveToBall strategy that requests the Dive when the ball is close enough in the x-axis 

Tested and seems to work okay but would be better with ball velocity information. We probably only want the goalie to dive when the ball is moving, then kick it back to the field when it's still.


module/strategy/DiveToBall/src/DiveToBall.hpp
PiiDish, line 1: This threshold that the robot dives when the ball is zero meters away. Wouldn't be something like 0.2meters?

module/strategy/DiveToBall/data/config/DiveToBall.yaml
```diff
@@ -0,0 +1,8 @@
+# Controls the minimum log level that NUClear log will display
+log_level: INFO
+
+# Distance to begin the dive to stop the ball (in meters)
+diving_distance_threshold: 1.0
+
+# Time before the ball measurement is too old
+ball_timeout: 0.5
```
ysims, line 8:
```suggestion
```

module/strategy/DiveToBall/README.md
```diff
@@ -0,0 +1,22 @@
+# DiveToBall
+
+## Description
+
+Dives when the ball is close enough on the x-axis in the direction of the ball. Intended for use by a goalie to stop the ball.
```
ysims, line 5:
```suggestion
Dives when the ball is close enough on the x-axis, diving in the direction of the ball. Intended for use by a goalie to stop the ball.
```

PiiDish (approved):
Looks good

PR #1060, opened 2023-04-12
Quickens the look search fixation time from 2 seconds to half a second

ysims:
The look search was a bit slow. This quickens it up so it's more likely to see the ball near its feet when it's lost it.


PiiDish (approved):
That's fine

PR #1061, opened 2023-04-12
Add director ramp up speed 

ysims:
Adds 

- Director ramping (slowing near ball, ramping up speed far away from ball)
- Comment out stand still in penalisation bc it causes a director bug (commented because we need to put it back in for the real robot)
- Adds a better right kick for webots

### DIFF
--- a/module/purpose/Soccer/src/Soccer.cpp
+++ b/module/purpose/Soccer/src/Soccer.cpp
@@ -68,14 +68,14 @@ namespace module::purpose {
             if (self_penalisation.context == GameEvents::Context::SELF) {
                 emit(std::make_unique<ResetWebotsServos>());
                 emit<Task>(std::unique_ptr<FindPurpose>(nullptr));
-                emit<Task>(std::make_unique<StandStill>());
+                // emit<Task>(std::make_unique<StandStill>());
             }
         });
 
         on<Trigger<Unpenalisation>>().then([this](const Unpenalisation& self_unpenalisation) {
             // If the robot is unpenalised, stop standing still and find its purpose
             if (self_unpenalisation.context == GameEvents::Context::SELF) {
-                emit<Task>(std::unique_ptr<StandStill>(nullptr));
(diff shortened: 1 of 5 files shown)


module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -87,7 +87,7 @@ namespace module::purpose {
                 log<NUClear::INFO>("Force playing started.");
                 cfg.force_playing = true;
                 emit<Task>(std::make_unique<FindPurpose>());
-                emit<Task>(std::unique_ptr<StandStill>(nullptr));
+                // emit<Task>(std::unique_ptr<StandStill>(nullptr));
```
PiiDish, line 90: Commented out?? Future development?

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -68,14 +68,14 @@ namespace module::purpose {
...
-                emit<Task>(std::make_unique<StandStill>());
+                // emit<Task>(std::make_unique<StandStill>());
             }
         });
 
         on<Trigger<Unpenalisation>>().then([this](const Unpenalisation& self_unpenalisation) {
             // If the robot is unpenalised, stop standing still and find its purpose
             if (self_unpenalisation.context == GameEvents::Context::SELF) {
-                emit<Task>(std::unique_ptr<StandStill>(nullptr));
+                // emit<Task>(std::unique_ptr<StandStill>(nullptr));
```
PiiDish, line 78: Commented out?? Is this for future development?

PiiDish (changes requested):

module/purpose/Soccer/src/Soccer.cpp
ysims: see second dot point in the main PR comment

module/purpose/Soccer/src/Soccer.cpp
ysims: see second dot point in the main PR comment

PR #1063, opened 2023-04-15
Fix for Dockerfile smell DL3003

grosa1:
Hi!
The Dockerfile placed at "docker/Dockerfile" contains the best practice violation [DL3003](https://github.com/hadolint/hadolint/wiki/DL3003) detected by the [hadolint](https://github.com/hadolint/hadolint) tool.

The smell DL3003 occurs if the pattern `RUN cd ...` is used to change directories instead of the dedicated WORKDIR instruction.
This pull request proposes a fix for that smell generated by my fixing tool. The patch was manually verified before opening the pull request. To fix this smell, specifically, the pattern `RUN cd ...` is replaced with the equivalent WORKDIR instruction.

This change is only aimed at fixing that specific smell. If the fix is not valid or useful, please briefly indicate the reason and suggestions for possible improvements.

Thanks in advance

### Pre-PR Checklist:

Have you

### Things left to do

None

### Labels
- `-Dependencies`
- `L-Docker`

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -53,7 +53,8 @@ COPY etc/ld.so.conf.d/usrlocal.conf /etc/ld.so.conf.d/usrlocal.conf
 RUN ldconfig
 
 # Make a symlink from /usr/local/lib to /usr/local/lib64 so library install location is irrelevant
-RUN cd /usr/local && ln -sf lib lib64
+WORKDIR /usr/local 
+RUN  ln -sf lib lib64
 
 # Generate toolchain files for the generic platform
 COPY usr/local/toolchain/generate_toolchains.py /usr/local/generate_toolchains.py


ysims:
Fix is not valid.

TrentHouliston:
Hi @grosa1 

Just to add some more since it seems you're trying to make a linting tool for docker which I think is super cool.
In my opinion, the rule you're trying to enforce here is more of an antipattern to be avoided than something that should be done.
By adding a `WORKDIR` line before this command, you have not only edited the command but also changed the working directory for every subsequent command.

In this case, we want to change the current directory for a single command (the symlink command) since the working directory changes the result.
We do not want every subsequent command to be running in this new directory as that would break things.
That leaves two options, putting a `WORKDIR` before and after this command
This option means we need to know what the WORKDIR was before our first command and if it were to change (potentially higher in the file) we now need to modify both that line AND this line.
Or do an inline cd like we have done which only impacts this one line and not the remainder of the file.

Secondly, there is a limit to the number of layers you can do in docker, and by adding a workdir command each time you have wasted a layer (or two to change back) simply to use a feature of docker rather than an inline cd.

In my opinion, this rule is like saying, Hey I noticed that every for loop seems to use an `i` variable, lets's just join them all together and have a single `i` variable for all loops.

PR #1070, opened 2023-05-01
UKF Ball Filter

Tom0Brien:
This PR replaces the exponential filter within BallFilter with an Unscented Kalman Filter (UKF). 

In theory this should perform much better, and it also gives us an estimate of the ball velocity if we wish to use that in the future.


module/localisation/BallFilter/README.md
```diff
@@ -2,12 +2,12 @@
 
 ## Description
 
-This module takes in a list of vision balls, uses the ball measurement closest to our current estimate and applies a
-rudimentary exponential filter to the measurement.
+This module takes in a list of vision balls, uses the ball measurement closest to our current estimate and applies an
+Unscented Kalman Filter to estimate the balls position and velocity in world space.
```
ysims, line 6: if we have a ball velocity can we include that in the message.

module/localisation/BallFilter/README.md
```diff
@@ -2,12 +2,12 @@
...
 
 ## Usage
 
-Include this module to allow the robot to filter vision ball position measurements.
+Include this module to allow the robot to estimate the balls position and velocity.
```
ysims, line 10: can't comment on it but also consumes FieldDescription

ysims (commented):
can you test this on the real robot and make sure we can still track the ball

ysims (commented):
can you test this on the real robot and make sure we can still track the ball

Tom0Brien:
> can you test this on the real robot and make sure we can still track the ball

works 👍

module/localisation/BallFilter/src/BallFilter.cpp
```diff
@@ -35,53 +37,107 @@ namespace module::localisation {
...
-                // Loop through all our ball measurements and find the closest measurement to our current estimate
-                for (const auto& ball : balls.balls) {
-                    Eigen::Vector3f temp_rBCc =
-                        reciprocalSphericalToCartesian(ball.measurements[0].srBCc.cast<float>());
-                    if ((temp_rBCc - filtered_rBCc).norm() < lowest_distance) {
-                        lowest_distance = (temp_rBCc - filtered_rBCc).norm();
-                        rBCc            = temp_rBCc;
+        on<Trigger<VisionBalls>, With<Sensors>, With<FieldDescription>>().then(
+            [this](const VisionBalls& balls, const Sensors& sensors, const FieldDescription& fd) {
+                if (balls.balls.size() > 0) {
```
Bidski, line 77:
```suggestion
                if (!balls.balls.empty()) {
```

module/localisation/BallFilter/src/BallFilter.cpp
```diff
@@ -35,53 +37,107 @@ namespace module::localisation {
...
+                    Eigen::Isometry3f Hcw = Eigen::Isometry3f(balls.Hcw.cast<float>());
+                    Eigen::Isometry3f Hwc = Hcw.inverse();
+                    auto state            = BallModel<float>::StateVec(ukf.get_state());
+
+                    // Get the first ball measurement and treat it as the closest measurement to our current estimate
+                    Eigen::Vector3f rBWw =
+                        Hwc * reciprocalSphericalToCartesian(balls.balls[0].measurements[0].srBCc.cast<float>());
+                    float lowest_distance = (rBWw.head<2>() - state.rBWw).norm();
```
Bidski, line 85: Use `squaredNorm()` instead (here, as well as below). It will save you doing square root operations and will have no impact on the logic

module/localisation/BallFilter/src/BallFilter.cpp
```diff
@@ -35,53 +37,107 @@ namespace module::localisation {
...
+
+                    // Generate and emit message
+                    auto ball = std::make_unique<FilteredBall>();
+                    Eigen::Isometry3f Htw(sensors.Htw.cast<float>());
+                    Eigen::Isometry3f Hrw(sensors.Hrw.cast<float>());
+                    Eigen::Isometry3f Htc     = Eigen::Isometry3f(sensors.Htw.cast<float>()) * Hwc;
+                    ball->rBWw                = rBWw;
+                    ball->rBTt                = Htw * rBWw;
+                    ball->rBRr                = Hrw * rBWw;
+                    ball->rBCt                = Htc.rotation() * (Hcw * rBWw);
```
Bidski, line 124: Seems like overkill to compute all of these variations on `rBWw` when all of the transformations just come directly from `Sensors`. It would be better (and ultimately more accurate) yo just compute and store `rBWw`, whoever is triggering on your emitted message can then `With<Sensors>` if they need one of the other variations.

Bidski (changes requested):

module/localisation/BallFilter/src/BallFilter.cpp
Tom0Brien: While I very much agree with this, making this change is going to require a considerable amount of changes so I would prefer to leave updating this to a later PR. Also highlights the need to focus on getting the Director transition completed so we can phase out a lot of duplicate code and clean everything up.

module/purpose/Soccer/data/config/webots/Soccer.yaml
```diff
@@ -0,0 +1,9 @@
+# Controls the minimum log level that NUClear log will display
+log_level: INFO
+
+# For the robot playing without listening to GameController
+force_playing: false
+
+# The position of the robot
+# GOALIE, STRIKER, DEFENDER
+position: STRIKER
```
Bidski, line 9: Is this relevant to this PR?

module/localisation/BallFilter/src/BallFilter.cpp
Bidski: Just remember that these computed positions can (and probably will) be out of date when you use them, so there will likely be more error in them than there actually is.

Bidski (changes requested):

module/purpose/Soccer/data/config/webots/Soccer.yaml
Tom0Brien: Not entirely, but it got removed when it shouldn't have in another PR prior and is needed to test the filter in webots

Bidski (approved):

PR #1073, opened 2023-05-17
Perform updates to many NUsight systems to synchronise feature sets

TrentHouliston:
This PR updates many components of NUsight.

Some of the highlights are
- Scrub bar: you can open NBS files and scrub through them
- Networking improvement: You can do RPC calls in NUsight so you can run functions on the robots and get things back from NUbots
- Zoom pan and focus in the camera views
- Other things @JosephusPaye could tell you about


Bidski (changes requested):
Not enough lines changed. Please increase line count and resubmit

Tom0Brien:
Networking seems to be working much better and there doesn't seem to be any of the old issues where I would have to refresh page etc. Scrubber looks awesome! Will be very useful. However, when I tried to use it, it didn't seem to work, not sure if an issue with NUsight or my NBS file though. 

The ability to send commands back to the robot opens up a lot of possibility for extra functionality.

JosephusPaye:
@Tom0Brien Was there an error message when you tried using the scrubber and it didn't work?

Currently the scrubber needs a corresponding `.nbs.idx` file for the `.nbs` file you're trying to use, otherwise it should show an error about the index file not being found. You can generate an index for a nbs file that doesn't have one using `./b nbs stats`.

Tom0Brien:
> @Tom0Brien Was there an error message when you tried using the scrubber and it didn't work?
> 
> Currently the scrubber needs a corresponding `.nbs.idx` file for the `.nbs` file you're trying to use, otherwise it should show an error about the index file not being found. You can generate an index for a nbs file that doesn't have one using `./b nbs stats`.

It ended up being an issue related to the nbs file being generated incorrectly. (time related)

PR #1076, opened 2023-05-20
NUsight localisation: Add world to field transformation

BrendanAnnable:
Currently the localisation tab in NUsight doesn't show localisation. This PR fixes that!

It applies `message.localisation.Field.Hfw` to the robot's position.

Confirmed as matching the overview tab, which I'm assuming to be correct.

Attached is an nbs file recorded from running webots: [nbs_example.zip](https://github.com/NUbots/NUbots/files/11521928/nbs_example.zip)


ysims (approved):
lgtm

PR #1077, opened 2023-05-21
Fix rotation in localisation view

Tom0Brien:
This PR applies simple fix to rotation in NUsight localisation view.

image

### DIFF
--- a/nusight2/src/client/components/localisation/robot_model.ts
+++ b/nusight2/src/client/components/localisation/robot_model.ts
@@ -167,7 +167,7 @@ export class LocalisationRobotModel {
     return this.position.rTFf;
   }
 
-  /* Field to torso rotation in field space. */
+  /* Rotation matrix from torso space to field space. */
   @computed get Rft(): Quaternion {
     return this.position.Rft;
   }
--- a/nusight2/src/client/components/localisation/robot_model.ts
+++ b/nusight2/src/client/components/localisation/robot_model.ts
@@ -168,18 +168,17 @@ export class LocalisationRobotModel {
   }
 
   /* Field to torso rotation in field space. */
-  @computed get Rtf(): Quaternion {
-    return this.position.Rtf;
+  @computed get Rft(): Quaternion {
+    return this.position.Rft;
   }
 
   @computed
   private get position() {
     const Hwf = this.Hfw.toThree().invert();
     const Htf = this.Htw.toThree().multiply(Hwf);
-    const { rotation: Rtf } = decompose(Htf);
     const Hft = Htf.invert();
-    const { translation: rTFf } = decompose(Hft);
-    return { Htf, rTFf, Rtf };
+    const { translation: rTFf , rotation: Rft } = decompose(Hft);
+    return { Htf, rTFf, Rft };
   }
 }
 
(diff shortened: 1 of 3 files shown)


ysims (dismissed):

nusight2/src/client/components/localisation/robot_model.ts
```diff
@@ -168,18 +168,17 @@ export class LocalisationRobotModel {
   }
 
   /* Field to torso rotation in field space. */
```
BrendanAnnable, line 170: Can we update this comment as well?

nusight2/src/client/components/localisation/robot_model.ts
```diff
@@ -168,18 +168,17 @@ export class LocalisationRobotModel {
...
 
   @computed
   private get position() {
     const Hwf = this.Hfw.toThree().invert();
     const Htf = this.Htw.toThree().multiply(Hwf);
-    const { rotation: Rtf } = decompose(Htf);
     const Hft = Htf.invert();
-    const { translation: rTFf } = decompose(Hft);
-    return { Htf, rTFf, Rtf };
+    const { translation: rTFf , rotation: Rft } = decompose(Hft);
```
BrendanAnnable, line 180:
```suggestion
    const { translation: rTFf, rotation: Rft } = decompose(Hft);
```

nusight2/src/client/components/localisation/robot_model.ts
BrendanAnnable, line 170: Which I guess would be:

```suggestion
  /* Torso to field rotation in torso space. */
```

Been a while since I've dealt with these... but is it typical for translation and rotation to be in separate spaces? 🤨

Tom0Brien: They are in the same space. 

image

Tom0Brien: So comment is correct as is

BrendanAnnable: Screen Shot 2023-05-21 at 12 35 11 pm

Since its named `Rft`, doesn't that mean we're going from torso to field?

TrentHouliston: The new code `Rft` is absolutely converting vectors in torso space to field space
so it is "Torso to field"

TrentHouliston, line 170: Describing it as "in a space" is kinda misleading anyway

```suggestion
  /* Torso to field rotation. */
```

TrentHouliston (dismissed):

nusight2/src/client/components/localisation/robot_model.ts
Tom0Brien: I guess it depends on what context when considering the statement X to Y. Imo, since we are using these functions to position and orientation the robot in field space, it makes sense to leave the comment as is, and not interpret the comment as the transform but as the orientation information it provides.

TrentHouliston: But the information it provides is the transform 🤨

Bidski: `Rft` describes a rotation from torso space to field space. Describing it as anything else is just confusing.

TrentHouliston: If you did want to flip the language you could say that Rft describes the torso basis in field space

Tom0Brien: I agree but the very reason this PR exists and some others in the past is because the language "X to Y rotation" is very easily misinterpreted.

Tom0Brien: I will update comment

BrendanAnnable:
Alternative approach #1082 1082

PR #1079, opened 2023-05-21
Add OpenCR and DYNAMIXEL Protocol 2.0 compatibility 

dex-k:
This PR is a relatively minor patch which will:
* Rewrite HardwareIO to implement:
  * DYNAMIXEL Protocol 2.0 communication (including support for new X-Series servos)
  * Support for the OpenCR subcontroller
* Adjust the RawSensors message to reflect DYNAMIXEL Protocol 2.0 changes
* Make minor changes to the existing CM740 HardwareIO and HardwareSimulator modules to ensure forward compatibility with the updated RawSensors message
* Add a CMake build flag to allow for subcontroller selection (**note:** the default value is `OpenCR`)
* Edit existing roles to use the new CMake subcontroller variable
* Add a stub option for OpenCR in FirmwareInstaller
* Fix `shared/utility/io/uart` read and write template functions (to be used with DYNAMIXEL Protocol 2.0 command structs)

### Pre-PR Checklist:

Have you


ysims:
Run `./b format` pass the formatting check

roles/test/visualmesh.role
```diff
@@ -14,8 +14,7 @@ nuclear_role(
   support::configuration::SoccerConfig
   # Sensors
   input::Camera
-  platform::cm740::HardwareIO
-  # platform::cm740::HardwareSimulator
+  platform::<subcontroller>::HardwareIO
```
Bidski, line 17: Yea, this seems like a horrible way to do this

CMakeLists.txt
```diff
@@ -103,6 +103,14 @@ elseif(CMAKE_CXX_COMPILER_ID MATCHES Clang)
...
 endif()
 
+# Allow selection of subcontroller for conditional build
+set(SUBCONTROLLER
+    "OpenCR"
+    CACHE STRING "Select what subcontroller to build for"
+)
+# Set the possible values of subcontroller for cmake-gui
+set_property(CACHE SUBCONTROLLER PROPERTY STRINGS "OpenCR" "cm740" "NUSense")
+
```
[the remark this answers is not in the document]
ysims: I'd much rather change one cmake variable than go through a heap of roles and probably miss some if I'm doing it on the fly. As for keeping roles for both, they would be completely duplicated apart from the subcontroller and that just makes it less likely that they will all be kept up to date. I imagine this would be more likely to be used at robocup, if we need to go back to the CM740 for whatever reason, and that's already a high pressure environment where things can easily go wrong. I really don't see any downside of having a quick flag to change if we swap subcontrollers.

module/platform/OpenCR/HardwareIO/src/HardwareIO.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+            // Service the watchdog because we recieved a valid packet
+            emit<Scope::WATCHDOG>(ServiceWatchdog<HardwareIO>());
+
+            // Pop the front of the packet queue
+            auto& info = packet_queue[packet.id].front();
+            packet_queue[packet.id].erase(packet_queue[packet.id].begin());
+
+            // Check for packet errors
+            /// @todo Do we want to handle packets differently if they have errors?
```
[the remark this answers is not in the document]
ysims: There is an issue open for OpenCR. Can still get rid of the todo comments in favour of the issue but I wouldn't say we should just do them, because this PR is already huge and I want to get this into main so Dex and I can work on issues concurrently in branches.

CMakeLists.txt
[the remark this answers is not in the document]
TrentHouliston: I agree with @Bidski that this is not the right way to do this, it's gonna cause chaos.

I also do see your point that changing the subcontroller shouldn't result in duplicated roles all over the place.
My suggestion for this is take advantage of the fact that .role files are still cmake files and declare your hardware io etc like this.

```cmake
nuclear_role(
    ...
    platform::${NUBOTS_PLATFORM)::HardwareIO
)
```

Then have the platform-specific modules for each in that directory and you can select which module you use by changing that variable (set it up earlier in the same way we set up the choices for the build mode)

module/platform/OpenCR/HardwareIO/src/hardwareio/send_request.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "HardwareIO.hpp"
+
+namespace module::platform::OpenCR {
+
+    using message::platform::RawSensors;
+
+    void HardwareIO::send_servo_request() {
+
+        // Write out servo data
+        // SYNC_WRITE (write the same memory addresses on all devices)
```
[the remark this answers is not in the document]
dex-k: Nah I was working on only adding relevant devices to the Sync Write packet to make the packets shorter, which I think may be a better solution (although, comes with it's own problems for implementation).

Using BulkWrite may make instruction packets even longer, as the write addresses would need to be specified for every servo (rather than once at the start). Additionally, we wouldn't be able to construct packets easily using the `DynamixelServoWriteData` structs, as write data would be different, which would be a pain to work with.

As it stands, we don't have reason to be incredibly concerned about efficiency (before RoboCup anyway). The loop ran at >100Hz when I last tested it.

roles/test/visualmesh.role
TrentHouliston: These are cmake files, just use cmake variables like `${NUBOTS_SUBCONTROLLER}`

module/platform/OpenCR/HardwareIO/src/HardwareIO.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+            opencr.open(config["opencr"]["device"], config["opencr"]["baud"]);
+            byte_wait   = config["opencr"]["byte_wait"];
+            packet_wait = config["opencr"]["packet_wait"];
+
+            // Initialise packet_queue map
+            // OpenCR
+            packet_queue[uint8_t(NUgus::ID::OPENCR)] = std::vector<PacketTypes>();
+            // Servos - NOT zero indexed. First servo is ID 1.
+            // Required because response packets have Dynamixel ID
```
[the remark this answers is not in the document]
dex-k: They aren't zero indexed because the packet queue tracks packets based on the dynamixel ID of the device it's communicating with i.e. packet_queue[1] is the queue for ID 1 etc.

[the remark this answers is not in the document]
ysims: I've updated the comment to hopefully make it make more sense

CMakeLists.txt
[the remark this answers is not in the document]
ysims: That's cleaner, I didn't think about the fact that roles are just cmake files.
Have updated it.

Bidski (changes requested):

module/platform/OpenCR/HardwareIO/src/hardwareio/handle_response.cpp
```diff
@@ -0,0 +1,182 @@
...
+        static uint16_t packet_length = 0;
+
+        // Read from the OpenCR device
+        std::array<uint8_t, 128> buf;
+        uint8_t num_bytes;
+        num_bytes = opencr.read(buf.data(), 128);
+
+        // Emits a completed StatusReturn message
+        auto emit_msg = [&]() -> void {
+            auto msg = std::make_unique<StatusReturn>();
```
Bidski, line 25: For reference, if you are only emitting this locally (its never going to go to nusight, probably never going to leave this module) you can just emit the `dynamixel::v2::StatusReturn` struct directly, you don't need to create the `StatusReturn` protobuf message.

Bidski (changes requested):

module/platform/OpenCR/HardwareIO/src/hardwareio/handle_response.cpp
ysims: I just had a look at it - it looks like there's some processing happening on the StatusReturn message, including another field. I could refactor it a bit to do that processing later but it might be better to add it to the issue and get this merged.

Bidski:
Why are you using c-style arrays now?

ysims:
> Why are you using c-style arrays now?

Couldn't pack std::array, was getting warnings on it not being POD.

module/platform/OpenCR/HardwareIO/src/hardwareio/send_request.cpp
```diff
@@ -19,8 +19,8 @@ namespace module::platform::OpenCR {
 
         if (servos_dirty) {
             // Write data is split into two components
-            std::array<dynamixel::v2::SyncWriteData<DynamixelServoWriteDataPart1>, 20> data1;
-            std::array<dynamixel::v2::SyncWriteData<DynamixelServoWriteDataPart2>, 20> data2;
+            dynamixel::v2::SyncWriteData<DynamixelServoWriteDataPart1> data1[20];
+            dynamixel::v2::SyncWriteData<DynamixelServoWriteDataPart2> data2[20];
```
Bidski, line 23: You should be able to just do this
```suggestion
            dynamixel::v2::SyncWriteCommand<DynamixelServoWriteDataPart1, 20> command1;
            dynamixel::v2::SyncWriteCommand<DynamixelServoWriteDataPart1, 20> command2;
```
Then `command1.data[i]` is the `i`th `SyncWriteData` so you can just write the servo data directly into the `SyncWriteCommand` struct. If this doesn't work then we should look at fixing this since copying that data should be unnecessary.

Bidski (dismissed):

module/platform/OpenCR/HardwareIO/src/hardwareio/send_request.cpp
ysims: The fields in all the messages are const - is this important for messages for any particular reason or is it fine to not have it const to be able to do this?

ysims: The array is pass by reference so there shouldn't be any copying right?

Bidski: It is passed by `const&`, so it will be copied. All the fields are `const` to prevent the message fields from being accidentally corrupted, but I think it should be ok for `data` to be non-const

ysims: I changed it but now it doesn't work :shrug:

ysims: I've added this problem to the issue. Unfortunately I won't be able to work on this again until I get back in two weeks, but we need to get it merged asap since it is critical to the robots working.

Tom0Brien (approved):
Need this merged, we can fix issues later

PR #1082, opened 2023-05-22
NUsight Localisation: Fix rotation by simplification

BrendanAnnable:
An alternative to #1077 which fixes the problem by simplification.

We instead use the matrix directly to the three.js model, reducing this risk of anything being wrong.

Deletes a bunch of code and simplifies what is left.

### DIFF
--- a/nusight2/src/client/components/localisation/controller.ts
+++ b/nusight2/src/client/components/localisation/controller.ts
@@ -217,9 +217,9 @@ export class LocalisationController {
 
     // This camera position hack will not work with orientation/head movement.
     // TODO (Annable): Sync camera position/rotation properly using kinematic chain.
-    model.camera.position = new Vector3(target.rTFf.x, target.rTFf.y, target.rTFf.z + 0.15);
-    const Rtf = new THREE.Quaternion(target.Rtf.x, target.Rtf.y, target.Rtf.z, target.Rtf.w);
-    model.camera.yaw = new THREE.Euler().setFromQuaternion(Rtf).z;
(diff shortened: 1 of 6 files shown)


Tom0Brien (approved):
lgtm

nusight2/src/client/components/localisation/controller.ts
```diff
@@ -217,9 +217,9 @@ export class LocalisationController {
 
     // This camera position hack will not work with orientation/head movement.
     // TODO (Annable): Sync camera position/rotation properly using kinematic chain.
-    model.camera.position = new Vector3(target.rTFf.x, target.rTFf.y, target.rTFf.z + 0.15);
-    const Rtf = new THREE.Quaternion(target.Rtf.x, target.Rtf.y, target.Rtf.z, target.Rtf.w);
-    model.camera.yaw = new THREE.Euler().setFromQuaternion(Rtf).z;
+    const { translation, rotation } = target.Hft.decompose();
```
TrentHouliston, line 220: I would try to name these variables their proper labelled name as soon as possible to help confusion later.
so these would be `rTFf` and `Rft`, I suppose that might mean that the model should be storing those names too.

Not sure if that should be for another PR or this one.

BrendanAnnable: This code needs to go, does anyone actually use first person mode? It's basically just torso-view since it doesn't actually follow the head.

Tom0Brien: I think the idea is cool, doesn't hurt to keep

PR #1083, opened 2023-05-22
NUsight: Add matrix methods

BrendanAnnable:
This PR adds multiply and invert methods to `Matrix4` and `Matrix3`.

No need to complicate consumers when we can hide the THREE conversions internally.

If performance was ever an issue we could also easily inline the math operations.

### DIFF
--- a/nusight2/src/client/components/localisation/robot_model.ts
+++ b/nusight2/src/client/components/localisation/robot_model.ts
@@ -164,21 +164,7 @@ export class LocalisationRobotModel {
 
   /** Torso to field transformation */
   @computed
-  get Hft() {
-    const Hwf = this.Hfw.toThree().invert();
-    const Htf = this.Htw.toThree().multiply(Hwf);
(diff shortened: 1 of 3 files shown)


nusight2/src/client/components/localisation/robot_model.ts
```diff
@@ -164,21 +164,7 @@ export class LocalisationRobotModel {
...
-  get Hft() {
-    const Hwf = this.Hfw.toThree().invert();
-    const Htf = this.Htw.toThree().multiply(Hwf);
-    return Matrix4.fromThree(Htf.invert());
+  get Hft(): Matrix4 {
+    return this.Hfw.multiply(this.Htw.invert());
   }
 }
-
-function decompose(m: THREE.Matrix4): { translation: Vector3; rotation: Quaternion; scale: Vector3 } {
```
BrendanAnnable, line 174: Seems I forgot to delete this in the last PR.

TrentHouliston (approved):

PR #1088, opened 2023-05-27
Add behaviour to align to the goal before kicking

ysims:
Adds behaviour where the robot will align itself with the ball to the goal before it kicks.

- Adds the AlignBallToGoal module, which will rotate the robot so that it is facing the middle of the goals. The rotation direction is based on whichever is quicker.
- Adds the KickToGoal module, which just finds where the goal is relative to the robot and emits a KickTo task with that direction.
- Updates the PlanKick module to enforce alignment before kicking
- Updates the Striker module to emit AlignBallToGoal and KickToGoal Tasks, and removes KickTo

### DIFF
--- a/module/purpose/Striker/data/config/Striker.yaml
+++ b/module/purpose/Striker/data/config/Striker.yaml
@@ -2,4 +2,4 @@
 log_level: INFO
 
 # Position on field to walk to (x,y,theta)
-ready_position: [1.2, 0, -3.14]
+ready_position: [4, 0, -3.14]
--- a/module/purpose/Striker/data/config/Striker.yaml
+++ b/module/purpose/Striker/data/config/Striker.yaml
@@ -1,5 +1,5 @@
 # Controls the minimum log level that NUClear log will display
-log_level: DEBUG
+log_level: INFO
 
 # Position on field to walk to (x,y,theta)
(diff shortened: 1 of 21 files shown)


module/purpose/Striker/data/config/Striker.yaml
```diff
@@ -2,4 +2,4 @@
 log_level: INFO
 
 # Position on field to walk to (x,y,theta)
-ready_position: [4, 0, -3.14]
+ready_position: [1.2, 0, -3.14]
```
TrentHouliston, line 5: Honestly I might consider putting a further comment as to what this means for future people changing this, is -3.14 towards the enemy goal? towards ours?
What is the coordinate system?
Either reference another coordinate system (e.g. make this variable rPFf or something) or some other way describe it.

Also once again if this is an expression you can just put `pi` rather than `3.14`

TrentHouliston (commented):
A few things I noticed

module/purpose/Striker/data/config/Striker.yaml
ysims: this isn't even relevant for this PR :weary:

Tom0Brien (approved):
We do need to improve localisation behaviour when looking at ground with no field lines visible, but will look at in another PR.

PR #1089, opened 2023-05-29
Localisation tweaks

Tom0Brien:
This PR does the following:

- Adds `cfg.min_observations` which specifies a minimum number of observations needed to run a measurement update
- Adds uncertainty property to field message which we can use for behaviour. For example we could add behavior such as:
    -  Only begin walking around if uncertainty is low
    -  If uncertainty value becomes to large, stop walking and look around for field lines
    - If uncertainty is too large for X amount of time, reset the filter
- Updates penalising particles to use penalising factor
- Renames module from RobotLocalisation to FieldLocalisation
- Moves projecting field line points onto field to the FieldLineDetector module


ysims (dismissed):

ysims (dismissed):

module/localisation/FieldLocalisation/src/FieldLocalisation.cpp
```diff
@@ -196,9 +178,6 @@ namespace module::localisation {
             int centre_circle_r = (fd.dimensions.center_circle_diameter / 2) / cfg.grid_size;
             fieldline_map.add_circle(centre_circle_x0, centre_circle_y0, centre_circle_r, line_width);
 
-            // Fill the surrounding cells close to the field lines with decreasing occupancy values
-            // fieldline_map.fill_surrounding_cells(0.25 / cfg.grid_size);
```
ysims, line 200: commented out?

module/localisation/FieldLocalisation/src/FieldLocalisation.cpp
```diff
@@ -221,6 +200,32 @@ namespace module::localisation {
...
 
+            // Left side penalty mark
+            cfg.initial_state.emplace_back((fd.dimensions.field_length / 2.0),
+                                           (fd.dimensions.field_width / 2.0),
+                                           -M_PI_2);
+
+            // Right side penalty mark
+            cfg.initial_state.emplace_back((fd.dimensions.field_length / 2.0),
+                                           (-fd.dimensions.field_width / 2.0),
+                                           M_PI_2);
```
ysims, line 211: if we're specifying that positive half is our starting position, we should document this somewhere. in the readme, nubook, or a comment here

ysims (dismissed):

ysims (approved):

PR #1091, opened 2023-06-01
Fix systemconfigurator for nuc12

JesseWilliamson:
This PR replaces hardcoded values for network interface names with user-configurable values and sets the device names used by the new nuc12s as the default.


Bidski (changes requested):
I don't understand the point of this PR. The `SystemConfiguration` module already handles varying configs on different platforms and this PR only seems to further complicate an already complicated system in order to achieve the same results that the original system already achieved.

JesseWilliamson:
> I don't understand the point of this PR. The `SystemConfiguration` module already handles varying configs on different platforms and this PR only seems to further complicate an already complicated system in order to achieve the same results that the original system already achieved.

Does it? [Currently](https://github.com/NUbots/NUbots/tree/main/module/tools/SystemConfiguration/data/system), systemconfigurator has different config files for different hostnames, but as far as I can tell there isn't a way to handle different network interface names.

I don't know if there's a simpler way to handle this.

Bidski:
> Does it? [Currently](https://github.com/NUbots/NUbots/tree/main/module/tools/SystemConfiguration/data/system), systemconfigurator has different config files for different hostnames, but as far as I can tell there isn't a way to handle different network interface names.
> 
> I don't know if there's a simpler way to handle this.

You just rename `nugus1/etc/systemd/network/40-wlp58s0-robocup.network` to match the interface name you want (this part is optional), and you edit the `Name=<interface>` section in that file to match the name of the wireless interface you need to target. Then whenever `system_configuration` is run on `nugus1` its network configuration will be updated. 

The hostname will match the particular nuc you are running on and that particular nuc will have a specific network interface name

JesseWilliamson:
> > Does it? [Currently](https://github.com/NUbots/NUbots/tree/main/module/tools/SystemConfiguration/data/system), systemconfigurator has different config files for different hostnames, but as far as I can tell there isn't a way to handle different network interface names.
> > I don't know if there's a simpler way to handle this.
> 
> You just rename `nugus1/etc/systemd/network/40-wlp58s0-robocup.network` to match the interface name you want (this part is optional), and you edit the `Name=<interface>` section in that file to match the name of the wireless interface you need to target. Then whenever `system_configuration` is run on `nugus1` its network configuration will be updated.
> 
> The hostname will match the particular nuc you are running on and that particular nuc will have a specific network interface name

I guess this works. It still seems strange to me that having the user manually edit these files would be the intended way to use systemconfigurator. Maybe still easier to maintain than what I've done here :shrug:

Bidski:
> I guess this works. It still seems strange to me that having the user manually edit these files would be the intended way to use systemconfigurator. Maybe still easier to maintain than what I've done here shrug

That was literally the way `SystemConfiguration` was intended to be used. Edit the configuration files to match the expected system configuration of each robot, run `system_configuration` on the robot and all of the appropriate files are copied into their appropriate locations. Allowing you to edit the configuration files as needed allows you to create entirely custom configurations for each robot (if you so desired, or if it was required -- perhaps NUC12 requires an extra package to be installed to handle newer hardware that the older NUCs dont have for instance)

The [configuration system](https://nubook.nubots.net/system/foundations/config-script#the-configuration-system) is also designed to allow you to have per-robot configuration so you can even have a different settings applied on different robots. The combination of this system with the already existing functionality in `SystemConfiguration` achieves everything that you are trying to achieve in this PR. The wheel has already been invented, you just got to let it roll.

TrentHouliston:
that said, it might be a good idea to change all the ethernet files from 10-<devicename>.network to just `10-ethernet.network` and `10-wifi.network`.

The name of the files themselves doesn't matter, just the text in the match section in the file.

PR #1092, opened 2023-06-01
Remove locally built zlib

JesseWilliamson:
This pr should remove the locally built zlib, relying instead on the zlib version provided by the arch package. The purpose of this is to get around a recent problem with zlib not being accessible in the initramfs.

I still don't understand the root cause of this problem, and we didn't have this problem on the nuc7's, so it is probably worth fixing this properly at some point.

However, in order to have things working for robocup I think this is probably the best way forward.

This builds and seems to fix the problem, but I don't know if there's more I need to do, or if this is the wrong way to do it.


docker/Dockerfile
```diff
@@ -308,7 +304,8 @@ RUN install-package \
     vim \
     nano \
     clinfo \
-    doxygen
+    doxygen \
+    zlib
```
TrentHouliston, line 308: If you're installing zlib with arch instead of building it (which for this library is probably fine) you should make sure it's installed at the start of this file (before all the things that need it) rather than at the end.

That said I think it's included by default so this probably does nothing anway

TrentHouliston (changes requested):
Not ideal but I don't think not having it here makes a huge difference.

Move the install to the start though to ensure it's there before the things that need it

Bidski (approved):
LGTM

TrentHouliston (approved):

PR #1094, opened 2023-06-05
Update scripttuner for director

Tom0Brien:
This PR updates scripttuner to use director. 

Tested and working in both webots and real robot.


ysims (approved):
Yeah probably fine. It would be better to use the servo providers than servo targets but there wouldn't really be any clashes in servo motions anyway since there is just one specific use case for this module. And the code would probably not look very nice.

ysims (approved):

PR #1097, opened 2023-06-05
Remove unused and duplicate code

Tom0Brien:
This PR removes all modules replaced by Director equivalents. Additionally, I have removed a number of other modules and utilities which I believe are no longer used.

TODO:
- [ ] Ensure Director on NUC12 has no computational issues


TrentHouliston:
For the sanity of git history, I'd do separate PRs for the removal of the non-director system and other unused code.

Also, this PR is clearly trying to sneak in other code! Like in `Ball.proto`

I'd keep odometry localisation, it's useful for testing behaviour etc when localisation is not working properly (since it just sets field to world and emits it)

ysims:
> I'd keep odometry localisation, it's useful for testing behaviour etc when localisation is not working properly (since it just sets field to world and emits it)
We test behaviour with our simulator, Webots :stuck_out_tongue: 
Tbh no one on the current team have ever used odometry localisation.

TrentHouliston:
Fair enough, delete it then. My point about removing the non director and then other unused stuff still stands though!

I'd also change the names of the roles from `director<thing>.role` to just `<thing>.role` again too since the director part is now redundant

ysims:
Agree on seperating those aspects. Defs should keep PRs minimal. It also makes it easier to get reviews. And the director stuff won't be removed until after robocup, whereas other things can be cleaned up.

PR #1090, opened 2023-05-30
Update `./b` tools and NUClear roles

TrentHouliston:
Another minor patch that updates lots of the ./b tools and NUClear roles features

- Adds a platform to the docker image so poor m1/m2 mac people can build it
- Upgrades catch so that empty tests can be removed
- Auto installs dependencies when using ./b
- ./b can now have tools as folders with `__init__.py` rather than just files
- oneof support is fixed if you ever need to use it
- adds neutron support for isometry so you don't need to cast them in/out when emitting
- adds ccache for faster rebuilding of the code
- fixes some issues with ./b build not setting toolchains properly
- updates formatting tools to do eslint too (and allow for multiple formatters per file)
- updates nbs encoder/decoder tools in python
- changes docker build to cache properly for real this time rather than pulling
- puts ./b yarn into the docker container for easier nusight development
- other minor fixes to lots of things

PR #934, opened 2022-08-20
Update the Docker Image and fix the issues that resulted from it

TrentHouliston:
This PR is a metabranch to gather all the PRs (#909 #926) related to updating the docker image so they can be reviewed individually before merging to main

PR #1101, opened 2023-06-06
Update protobuf to fix repeated message building

TrentHouliston:
Unfortunately one of the updates in #1090 relied on changes that are still stuck in #934 specifically the protocol buffer updates that include this patch https://github.com/protocolbuffers/protobuf/pull/10414

Without these ninja gets super confused and keeps rebuilding the protocol buffers over and over again.

This PR extracts just that change to fix the issue.


Bidski (approved):
LGTM

PR #1104, opened 2023-06-08
Add binary folder specific config

ysims:
Currently webots config is broken on main - the hostname used to be 'webots' if the binary has the word 'webots' in it, but changes to the tools set up and run command have eliminated this functionality. Instead the config would have been defaulting to the docker hostname config or the default config. 

Rather than continuing to do this in the build system, I've move it to the configuration extension.

Before, config was looked at and overwritten in the order of: default, platform (eg nugus), hostname (eg nugus1), binary name (eg robocup). 
Now it is: default, platform (eg nugus), binary folder (eg webots), hostname (eg nugus1), binary name (eg robocup). 

I originally had the binary folder override hostname to overwride any docker config, but we should remove any docker config left and have it this way so that when we have eg `webots1` hostnames they are not overriden by webots binary folder config.

I've tested in webots and all looks fine. Might be good to test on real robot but I did take that into account and should be fine. With a robot specific hostname such as webots4, the relevant config kicks in (ie goalie config, I changed the ready positions of the striker and goalie).

Also updated from camelCase to snake_case.

NUbook needs updating so I'll make an issue to do that later.

edit: I'm tired and the docker thing might not be an issue bc it's platform not robot and we still override platform. pls sanity check..
edit2: Ah bother the script system also needs updating accordingly. I won't be back doing things until maybe tomorrow night my time, so feel free to continue this without me. altho it defs got up as-is :shrug:


Tom0Brien:
Another solution could be we just rename all the `webots` configs to docker and move scripts from `webots` to `docker` folder.

ysims:
That will cause problems during proper games since we change the hostname in that scenario to webots1, webots2, etc (altho this will need checking, it might not work in the new system?).
I've updated the script system, but it doesn't really change much because all our webots scripts are duplicated in the docker script folder anyway. It will make things work though in games. Have tested by removing the docker script folder and it does use the webots one instead. I've left the docker folder in because without it I'd assume fake keyboardwalk would break for the recruitment task.

ysims:
There is a good chance that the docker folder and webots folders are out of sync though.

ysims:
:no_good:

PR #1108, opened 2023-06-12
Add a test for multithreaded Director

TrentHouliston:
Adds a unit test to the director that tests it under multithreaded workloads.

Runs the test repeatedly many times to hopefully reveal any multithreading bugs.

Includes an upgrade to NUClear and use of Sync to ensure that tasks are executed correctly in order


Bidski (dismissed):
LGTM

PR #1110, opened 2023-06-13
Make docker hostname a little neater

Bidski:
An attempt to make setting the docker hostname to `webots` a little neater. Still not a great solution

### DIFF
--- a/tools/utility/dockerise/run.py
+++ b/tools/utility/dockerise/run.py
@@ -120,6 +120,7 @@ def run(func, image, hostname="docker", ports=[], docker_context=None):
         # the hostname should be docker or webots
         # Binaries containing 'webots' (ie in the webots folder) should be given the hostname 'webots'
         # to ensure the config files are chosen correctly
+        hostname = "docker"
         if kwargs["command"] == "run":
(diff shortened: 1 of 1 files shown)


Bidski:
> This breaks when using `./b install` or `./b configure` due to `UnboundLocalError: local variable 'hostname' referenced before assignment`.

Boo! ok, gimme a sec

tools/utility/dockerise/run.py
Tom0Brien, line 1: Should we remove hostname from `def run(func, image, hostname="docker", ports=[], docker_context=None):` since its always overwritten with `hostname = "docker"` right?

Tom0Brien: Not a big deal though

Tom0Brien (approved):
lgtm

tools/utility/dockerise/run.py
Bidski: No, we need to integrate it in properly since this could be set from the `@run_on_docker` decoration if I am not mistaken, I need to think on this a little more

Tom0Brien: Looks good, I think formatting will complain about `docker_hostname="webots"` again though

PR #1071, opened 2023-05-09
Williamson/nuc12 support (toolchain and install script)

JesseWilliamson:
Allows changing the names of network interfaces in arch-chroot install script, adds new generate toolchain file for nuc12, and adds nuc12 platform to nubots github workflow.

I can build with the new platform fine locally.
Systemconfigurator still needs changes.

I just want to merge this PR before finishing / testing SystemConfigurator because I want to be able to pull the docker image rather than building again on my laptop.

PR #1112, opened 2023-06-15
Nuc12 network defaults

JesseWilliamson:
This pr replaces network device defaults with values for the NUC12s, replaces network interface-specific file names with generic wifi / ethernet file names and fixes my incorrect link file path from #1071 . This is a better / simpler approach than #1091


module/tools/SystemConfiguration/data/config/SystemConfiguration.yaml
```diff
@@ -35,7 +35,7 @@ systemd:
   - systemd-resolved.service
   - systemd-timesyncd.service
   - sshd.service
-  - wpa_supplicant@wlp58s0.service
+  - wpa_supplicant@wlp0s20f3.service
```
Tom0Brien, line 38: Do we know why this is `wlp0s20f3` and not  `wlan0`? I have no understanding on any of these interface names, just curious as we changed from `WIFI_INTERFACE:-"wlp58s0"` to `WIFI_INTERFACE=${WIFI_INTERFACE:-"wlan0"}`

JesseWilliamson: The next line in `arch-chroot_install.sh` runs `udevadm test-builtin net_id` to find the [predictable network name](https://systemd.io/PREDICTABLE_INTERFACE_NAMES/) udev will give the wlan0 device. For the nuc12s, this is `wlp0s20f3`. After running the scripts and rebooting, the `99-default.link` file should tell udev to use the predictable network names (now the standard) so the device which was `wlan0` in the live install environment is now `wlp0s20f3`.

JesseWilliamson: Since you can give interface names as environment variables to  `arch-chroot_install.sh` now, using whatever the names are in the installation environment should be fine. `udevadm` will find the right name.

Tom0Brien (dismissed):

JesseWilliamson:
Whoops, sorry, missed the wpa_supplicant file. Otherwise tested and working now.

ysims (approved):

PR #1114, opened 2023-06-17
Fix stack overflow in Director where StandStill runs task pack in infinite loop with its dying child

ysims:
We've had a problem in Director where in webots, if keyboardwalk is started with the robot fallen over there is a stack overflow.
This same problem appears on the real robot in both robocup and keyboardwalk pretty much all of the time.

The problem is to do with dying tasks. When the dying tasks are removed, the groups are reevaluated. The group with the dying task still has it as a subtask. This causes an infinite loop of trying to run the task pack with the dying task, only to do nothing except delete the dying task, running the reevaluate group again... running the task pack with a dying task again...

The fix I found was to make sure that our group's subtasks are updated, then we remove the dying tasks.
With this, all existing tests pass, webots keyboardwalk on ground works, and the real robot works.

@TrentHouliston feel free to figure out a test for this :stuck_out_tongue:

### DIFF
--- a/module/extension/Director/src/director/run_task_pack.cpp
+++ b/module/extension/Director/src/director/run_task_pack.cpp
@@ -282,17 +282,20 @@ namespace module::extension {
             }
         }
 
+
+        // Make a copy of group.subtasks so we can remove tasks from it with updated subtasks
+        auto old_subtasks = group.subtasks;
+        // Update the group's subtasks to the new subtasks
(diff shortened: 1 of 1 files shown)


Crafty15 (approved):
LGTM

TrentHouliston:
I'll try come up with a test but I don't see any other issues with this change

PR #1115, opened 2023-06-17
Add react three fiber

BrendanAnnable:
This PR adds react three fiber as a nusight dependency, and demonstrates its utility in a storybook.

It takes advantage of React's JSX, providing us a declarative language to describe threejs scenes.

I'm pretty convinced its superior to our current implementation is nearly every way, but the real proof will come when we rewrite the various components that use threejs (e.g. vision and localisation). I'm pretty confident it will simplify a lot, and that we should be able to maintain similar performance. See #938 for a previous attempt at porting vision.

Electing to keep this PR as minimal as possible for now.

### DIFF
--- a/nusight2/NUsight2/package.json
+++ b/nusight2/NUsight2/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "NUsight2",
+  "version": "1.0.0",
+  "dependencies": {}
+}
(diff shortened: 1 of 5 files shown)


TrentHouliston (approved):

JosephusPaye:
Exciting!

PR #1118, opened 2023-06-17
Fix protobuf on windows

BrendanAnnable:
pbjs was unhappy about windows directory separators.


TrentHouliston (approved):
The real solution would be to remove windows :P

PR #1119, opened 2023-06-17
Ball localisation

Tom0Brien:
This PR does the following:
- Removes the particle filter ball localisation module
- Renames `BallFilter` to `BallLocalisation`
- Renames message `FilteredBall` to `Ball`
- Removes the `Sensors` based calculations from `BallLocalisation` module and implements them within the module they are used in with the latest `Sensors` message

Waiting on https://github.com/NUbots/NUbots/pull/1121

PR #1121, opened 2023-06-17
Remove subsumption behaviour modules

Tom0Brien:
This PR removes the subsumption based behaviour modules. 

Currently, it is very annoying to have to fix issues which arise in these modules when the code will soon be removed. For example, PR https://github.com/NUbots/NUbots/pull/1119 would require a considerable amount of changes within these modules to compile and imo is a waste of time to get working atm. If we need to use subsumption at robocup we can easily revert this PR and fix any issues.


ysims (approved):
I've made a branch on the docker changes so if we need to use subsumption it will have the opencl fix.

PR #1132, opened 2023-06-18
Odometry ground truth method

Tom0Brien:
This PR adds a ground truth method to `SensorFilter` which allows the use of webots ground truth information for odometry which is useful for testing purposes. 

Note, this will only work when using our webots controller.

Something I'm unsure on is it seems the recent timestamp changes https://github.com/NUbots/NUbots/pull/1116/files#diff-fc0b2b2d580a7b5d606168708dc45d1b335285a24d5c80f1f4ef7846672cc8b4 required the change in `Webots.cpp`. Maybe we should still be using `sensor_measurements.time` for the comparison, I need to look into it more.

### DIFF
--- a/module/input/SensorFilter/data/config/webots/SensorFilter.yaml
+++ b/module/input/SensorFilter/data/config/webots/SensorFilter.yaml
@@ -77,5 +77,5 @@ mahony:
 # Tuning parameters to scale walk command to match actual achieved velocity for deckreckoning x, y, theta
 deadreckoning_scale: [1, 1, 0.6]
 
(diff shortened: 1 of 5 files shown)


ysims (commented):
Update the comment in the default config too

Tom0Brien:
> Update the comment in the default config too

The method should never be used in the default config because it will only work in webots so thought it was best not to include it

ysims:
I guess. It would be cool to get mocap ground truth.

ysims (approved):

PR #1136, opened 2023-06-19
Update kinematics to use URDF model

Tom0Brien:
The following PR adds external library [tinyrobotics](https://github.com/Tom0Brien/tinyrobotics).

This library is used for parsing our [urdf](https://github.com/NUbots/NUgus) file which is then used for:

- Improving IK accuracy using a combination of a numerical optimisation based method and the previous analytical solution. The optimisation method is warm started with the analytical solution which ensures convergence and ensures that knees are bent correctly.
- Forward kinematics and centre of mass calculations within `SensorFilter`
- Camera kinematics `Hpc`

This PR also acts as the first step in unifying our robot model description to allow our onshape model to provide us up to date kinematics and inertial properties of our hardware platform. My end goal is roughly illustrated by diagram below:

image

Future improvements/features:

- Use tinyrobotics IK method with constrained optimization to allow addition of constraints such as CoM position/deviation, joint limits etc.
- Use of tinyrobotics dynamics algorithms
- Move nugus.urdf to it's own repo/folder with `onshape-to-urdf` config


module/actuation/Kinematics/src/Kinematics.cpp
```diff
@@ -26,15 +27,37 @@ namespace module::actuation {
...
+
+            // Show the model if debug is enabled
+            if (log_level <= NUClear::DEBUG) {
+                nugus_model_left.show_details();
+            }
+
+            // tinyrobotics IK options
+            options.tolerance      = config["ik_tolerance"].as<double>();
+            options.max_iterations = config["ik_max_iterations"].as<int>();
+            options.method         = tinyrobotics::InverseKinematicsMethod::LEVENBERG_MARQUARDT;
```
ysims, line 55: this seems like something we could put in config

module/actuation/Kinematics/src/Kinematics.cpp
```diff
@@ -26,15 +27,37 @@ namespace module::actuation {
...
+
+            // Parse NUgus URDF file description
+            cfg.urdf_path     = config["urdf_path"].as<std::string>();
+            nugus_model_left  = tinyrobotics::import_urdf<double, n_joints>(cfg.urdf_path);
+            nugus_model_right = tinyrobotics::import_urdf<double, n_joints>(cfg.urdf_path);
+
+            // Show the model if debug is enabled
+            if (log_level <= NUClear::DEBUG) {
+                nugus_model_left.show_details();
+            }
```
ysims, line 50: why only the left

module/input/Camera/data/config/nugus1/Cameras/Left.yaml
```diff
@@ -24,7 +24,7 @@ lens:
   fov: 180 * pi / 180
   # The homogenous transform from the rigid platform this camera is attached to to its optical location
   Hpc:
-    - [0.9997403694613228, -0.016551656784825577, -0.01566002262642871, 0.08893612063588988]
-    - [0.01728764296000275, 0.9986932814319833, 0.04809227613146373, 0.03529909622045881]
-    - [0.014843552535558163, -0.04835051478781678, 0.9987201292902445, 0.0713674563254241]
+    - [0.0, 0.0, 1.0, 0.08893612063588988]
```
ysims, line 27: revert

module/input/Camera/data/config/nugus1/Cameras/Right.yaml
```diff
@@ -24,7 +24,7 @@ lens:
   fov: 180 * pi / 180
   # The homogenous transform from the rigid platform this camera is attached to to its optical location
   Hpc:
-    - [0.9997403694613228, -0.016551656784825577, -0.01566002262642871, 0.08893612063588988]
```
ysims, line 27: revert

module/input/Camera/data/config/nugus2/Cameras/Left.yaml
```diff
@@ -24,7 +24,7 @@ lens:
   fov: 180 * pi / 180
   # The homogenous transform from the rigid platform this camera is attached to to its optical location
   Hpc:
-    - [0.9997403694613228, -0.016551656784825577, -0.01566002262642871, 0.08893612063588988]
-    - [0.01728764296000275, 0.9986932814319833, 0.04809227613146373, 0.03529909622045881]
-    - [0.014843552535558163, -0.04835051478781678, 0.9987201292902445, 0.0713674563254241]
+    - [0.0, 0.0, 1.0, 0.08893612063588988]
```
ysims, line 27: revert

module/input/Camera/data/config/nugus3/Cameras/Left.yaml
```diff
@@ -24,7 +24,7 @@ lens:
   fov: 180 * pi / 180
   # The homogenous transform from the rigid platform this camera is attached to to its optical location
   Hpc:
-    - [0.9997403694613228, -0.016551656784825577, -0.01566002262642871, 0.08893612063588988]
-    - [0.01728764296000275, 0.9986932814319833, 0.04809227613146373, -0.03529909622045881]
-    - [0.014843552535558163, -0.04835051478781678, 0.9987201292902445, 0.0713674563254241]
+    - [0.0, 0.0, 1.0, 0.08893612063588988]
```
ysims, line 27: revert

module/input/Camera/data/config/nugus4/Cameras/Left.yaml
```diff
@@ -24,7 +24,7 @@ lens:
   fov: 180 * pi / 180
   # The homogenous transform from the rigid platform this camera is attached to to its optical location
   Hpc:
-    - [0.9997403694613228, -0.016551656784825577, -0.01566002262642871, 0.08893612063588988]
-    - [0.01728764296000275, 0.9986932814319833, 0.04809227613146373, 0.03529909622045881]
```
ysims, line 28: revert

module/input/Camera/data/config/nugus4/Cameras/Right.yaml
```diff
@@ -24,7 +24,7 @@ lens:
   fov: 180 * pi / 180
   # The homogenous transform from the rigid platform this camera is attached to to its optical location
   Hpc:
-    - [0.9997403694613228, -0.016551656784825577, -0.01566002262642871, 0.08893612063588988]
-    - [0.01728764296000275, 0.9986932814319833, 0.04809227613146373, -0.03529909622045881]
-    - [0.014843552535558163, -0.04835051478781678, 0.9987201292902445, 0.0713674563254241]
+    - [0.0, 0.0, 1.0, 0.08893612063588988]
+    - [1.0, 0.0, 0.0, -0.03529909622045881]
```
ysims, line 28: revert

module/input/SensorFilter/src/SensorFilter.hpp
```diff
@@ -42,12 +43,13 @@ using extension::Configuration;
 
 namespace module::input {
 
-    using message::actuation::KinematicsModel;
     using message::behaviour::state::Stability;
     using message::behaviour::state::WalkState;
     using message::input::Sensors;
     using message::platform::RawSensors;
 
+    using namespace tinyrobotics;
```
ysims, line 51: hmmm do we want to do this

module/platform/Webots/data/config/WebotsCameras/left_camera.yaml
```diff
@@ -23,9 +23,9 @@ lens:
   fov: 2.0944
   # The homogenous transform from the rigid platform this camera is attached to to its optical location
   Hpc:
-    - [1.0, 0.0, 0.0, 0.0889361]
-    - [0.0, 1.0, 0.0, 0.0352991]
-    - [0.0, 0.0, 1.0, 0.0713675]
+    - [0.0, 0.0, 1.0, 0.0889361]
+    - [1.0, 0.0, 0.0, 0.0352991]
+    - [0.0, 1.0, 0.0, 0.0713675]
```
ysims, line 28: wait I just realised why you changed all the cameras. can't we just rotate in code somewhere so we don't do this? Our calibration tool/s won't work anymore if we have it like this. and if we can't calibrate the cameras (extrinsics) then localisation will be off.

shared/utility/input/ServoID.hpp
```diff
@@ -47,7 +47,9 @@ namespace utility::input {
             L_ANKLE_ROLL     = 17,
             HEAD_YAW         = 18,
             HEAD_PITCH       = 19,
-            NUMBER_OF_SERVOS = 20
+            L_FOOT_BASE           = 20,
+            R_FOOT_BASE           = 21,
+            NUMBER_OF_SERVOS = 22
```
ysims, line 52: but we only have 20 servos!

module/actuation/Kinematics/src/Kinematics.cpp
Tom0Brien: They are the same model, but you can't perform IK for two limbs at same time using the same model instance because it'll have issues I have defined seperate to avoid the need for mutex type stuff

module/input/Camera/data/config/nugus1/Cameras/Right.yaml
Tom0Brien: This will be needed as pitch frame is orientated differently now. I haven't tested on the real robot yet though but was rotation needed in webots. Ideally, we should add a frame in the onshape/urdf model of where the cameras are so we can obtain this matrix dynamically.

module/input/SensorFilter/src/SensorFilter.hpp
Tom0Brien: Hmm, probably best to not tbh. Will update.

module/platform/Webots/data/config/WebotsCameras/left_camera.yaml
Tom0Brien: How is `Hpc` calculated?

shared/utility/input/ServoID.hpp
Tom0Brien: Hmm, we could add something like LinkID if we want to keep ServoID's to only include the actuatable servo joints. With the URDF we can add fixed joints/frames to our model which is useful for specifying additional frames of interest such as the base of the foot, hands, camera locations etc.

module/platform/Webots/data/config/WebotsCameras/left_camera.yaml
ysims: You'd have to check the code for the calibration tool.

Tom0Brien: I think it is given quantity to calibration so we either calibrate all the cameras again with an updated Hpc, or we apply this rotation offset to the old Hpc.

module/actuation/Kinematics/src/Kinematics.cpp
Tom0Brien: Yeah we could.

ysims:
where's this at? we can manually do the extrinsics now, but we'll need it documented how the matrix works/what the axes are

module/input/SensorFilter/data/config/SensorFilter.yaml
```diff
@@ -84,3 +84,5 @@ initial_rpy: [0, 0.3839, 0]
 
 # Specify which filtering method to use, either UKF, KF or MAHONY
 filtering_method: MAHONY
+
+urdf_path: "/home/nubots/nugus.urdf"
```
ysims, line 88: I don't think this will work if SensorFilter is in the role but not Kinematics
For now maybe just make sure all the roles with SensorFilter have Kinematics, but I can put in a PR later to do something similar to the scripts where the urdf is always copied over.

shared/message/input/Sensors.proto
```diff
@@ -131,13 +131,13 @@ message Sensors {
     float battery = 9;
 
     /// Cartesian position of the CoM {M} of the robot from the torso {T} in the torso space {t}
-    vec4 rMTt = 10;
+    vec3 rMTt = 10;
 
     /// Inertia tensor measured in particle space and relative to the CoM of the particle
     mat3 inertia_tensor = 11;
 
     /// A set of transforms; one for each servo
```
ysims, line 139: add something to this comment so it's not confusing for ppl - ie specify what the extra 4 are for (it's the cameras and feet right?)

ysims (approved):
LGTM

ysims (dismissed):

module/input/SensorFilter/data/config/webots/SensorFilter.yaml
```diff
@@ -80,4 +80,4 @@ initial_rpy: [0, 0.3839, 0]
 # Specify which filtering method to use, either UKF, KF, MAHONY or GROUND_TRUTH
 filtering_method: MAHONY
 
-urdf_path: "/home/nubots/NUbots/module/actuation/Kinematics/data/nugus.urdf"
+urdf_path: "/home/nubots/NUbots/shared/utility/platform/models/nugus.urdf"
```
TrentHouliston, line 83: All these absolute paths are very suspect

ysims: it's the path in docker

ysims (approved):

Issue #1078, opened 2023-05-21
OpenCR HardwareIO


ysims:
- [ ] Get initial HardwareIO set up
- [ ] Set the starting state by reading the servos
- [ ] Error handling
- [ ] Check the velocity
- [ ] Update gains conversions based on todo in code
- [ ] Set more servo things on startup

ysims:
Old, and superceded by NUSense development.

PR #1138, opened 2023-06-19
Implement proper dynamixel v2 error handling in HardwareIO/RawSensors

dex-k:
Implements checkbox 3 of issue #1078, adding proper error handling for protocol v2 errors.

This maintains compatibility with v1, although it's done in a bit of a yucky way (redundant fields in the RawSensors message). This is only needed as long as we want v1 compatibility which hopefully isn't very long.

Note: protocol v2 introduces two different error types, the standard packet errors (bad instruction, bad checksum etc.), as well as hardware errors (overheating, overloaded etc.). RawSensors will report hardware errors individually for each servo, but packet errors are all combined (with logical or) into the one subcontroller/platform error, as this most closely matches the existing behaviour so it was much easier to implement.


shared/message/platform/RawSensors.proto
```diff
@@ -32,21 +32,44 @@ message RawSensors {
     /// Set of error flags/statuses.
     /// bitmask values.
 
-    /// @note these are kind of out of date or are triggered elsewhere in Dynamixel v2
-    ///       ANGLE_LIMIT, CORRUPT_DATA and TIMEOUT won't get triggered
-    ///       RANGE, CHECKSUM and INSTRUCTION are processed per status-return packet so
-    ///       will never make it to a RawSensors message.
+    /// @note RawSensors::Error is kept for protocol v1 compatibility only.
+    /// Values are prefixed with an underscore to resolve naming conflicts.
```
ysims, line 36: I think leading underscores are reserved or something? What is it conflicting with?

dex-k: oh whoops, i changed the name later so the conflict wasn't an issue anymore, forgot to change the underscores back

dex-k: the RawSensors::Error enum has names shared with RawSensors::PacketError and RawSensors::HardwareError, Can change the underscore to something else

dex-k: leading underscore has been replaced by a trailing underscore to avoid using reserved identifiers

ysims, line 36:
```suggestion
    /// Values are suffixed with an underscore to resolve naming conflicts.
```

ysims (approved):

PR #1144, opened 2023-06-20
Add new kick module and motion tuning

Tom0Brien:
### SplineKick

The PR adds a module similar to KickScript but it doesn't use a joint keyframe style script. A quintic spline based trajectory is generated between user specified waypoints in Cartesian space relative to the support foot for both the swing foot and torso. 

### Walk Engine
The PR also updates the WalkEngine:
- Makes use of the `Waypoints` struct added in `Trajectory.hpp` for generating walking trajectories. 
- Add config param `torso_final_position_ratio` for tuning the torso placement at end of step.


module/skill/SplineKick/src/SplineKick.cpp
```diff
@@ -0,0 +1,145 @@
...
+    using message::actuation::LeftArm;
+    using message::actuation::LeftLegIK;
+    using message::actuation::RightArm;
+    using message::actuation::RightLegIK;
+    using message::actuation::ServoCommand;
+    using message::actuation::ServoState;
+    using message::skill::ControlLeftFoot;
+    using message::skill::ControlRightFoot;
+    using message::skill::Kick;
+    using message::skill::KickFinished;
```
ysims, line 33: is this used somehow

module/skill/SplineKick/data/config/SplineKick.yaml
```diff
@@ -0,0 +1,38 @@
+# Controls the minimum log level that NUClear log will display
```
ysims, line 1: is this tuned for webots? do you wanna put it in a webots folder and just duplicate it for the default config so when we tune it for the real robot we don't lose the webots tune. If you remember some of the values we did the other day could add that in here. Afaik we made the time longer and made the torso move across less.

module/skill/Walk/data/config/Walk.yaml
```diff
@@ -2,25 +2,31 @@ log_level: INFO
 
 walk:
   # Time for one complete step (in seconds).
-  period: 0.35
+  period: 0.375
```
ysims, line 5: tuning should go in another PR, but if you want to get it through in this at least update the PR title

ysims (approved):

PR #1147, opened 2023-06-22
Logging configs

Tom0Brien:
This PR does the following:

- Updates `NetworkForwarder.yaml` config to send all messages for debugging by default. Most cases if NetworkForwarder is in the role, we want to be debugging data in nusight.

- Adds all the messages to `DataLogging.yaml` to make it easier to enable/disable messages in DataLogging. 

- Disables the button spoof in HardwareIO


module/support/logging/DataLogging/data/config/DataLogging.yaml
```diff
@@ -6,7 +6,18 @@ output:
 
 # You can put any messages that is emitted in the system here and it will be logged
 # The below ones are just the most common and here as an example
+# You can put any messages that is emitted in the system here and it will be logged
 messages:
-  message.output.CompressedImage: false
+  message.support.nusight.Overview: false
```
ysims, line 11: is there any difference between this and the webots config? if not then delete the webots one

module/platform/OpenCR/HardwareIO/src/HardwareIO.cpp
```diff
@@ -13,7 +13,7 @@
 #include "utility/math/comparison.hpp"
 #include "utility/support/yaml_expression.hpp"
 
-#define DEBUG_ENABLE_BUTTON_SPOOF 1
+#define DEBUG_ENABLE_BUTTON_SPOOF 0
```
ysims, line 16: I feel like this is out of scope for this PR and also we should just get rid of the macro entirely

module/network/NetworkForwarder/data/config/NetworkForwarder.yaml
```diff
@@ -21,13 +22,13 @@ targets:
...
-    message.vision.GreenHorizon: false
-    message.vision.Lines: false
-    message.vision.Obstacles: false
-    message.vision.VisualMesh: false
+    message.vision.Balls: true
+    message.vision.Goals: true
+    message.vision.GreenHorizon: true
+    message.vision.FieldLines: true
+    message.vision.Obstacles: true
+    message.vision.VisualMesh: true
```
ysims, line 30: I'm not sure if I like this, I feel like it could go wrong easily. Like saturating the network. Does this work, does message.vision.Obstacles exist?

ysims (approved):

PR #1148, opened 2023-06-22
Path planning tweaks

Tom0Brien:
This PR does the following:

- Fixes an issue with path pivot not inverting y translational velocity 
- Renames path planning `speed` to `velocity`/`velocity_magnitude` to keep consistent with terminology used elsewhere
- Adds check in path planner to reset the walk to minimum velocity if there is no walk task yet
- Various tuning of configs for webots behaviour


module/planning/PlanWalkPath/src/PlanWalkPath.cpp
```diff
@@ -74,15 +76,20 @@ namespace module::planning {
...
-            emit<Task>(std::make_unique<Walk>(
-                Eigen::Vector3d(cfg.pivot_ball_speed_x, cfg.pivot_ball_speed_y, sign * cfg.pivot_ball_speed)));
+            emit<Task>(std::make_unique<Walk>(Eigen::Vector3d(cfg.pivot_ball_velocity_x,
+                                                              sign * cfg.pivot_ball_velocity_y,
+                                                              sign * cfg.pivot_ball_velocity)));
         });
+
+        // Reset the velocity to minimum velocity after a kick
+        on<Trigger<KickFinished>>().then(
+            [this](const KickFinished&) { velocity_magnitude = cfg.min_translational_velocity_magnitude; });
```
ysims, line 93: we shouldn't have to do anything like this with Director, all modules should be self contained. Just check if there's no walk task, and if so then reset the velocity.

module/strategy/WalkToFieldPosition/src/WalkToFieldPosition.cpp
```diff
@@ -42,20 +43,29 @@ namespace module::strategy {
...
-                const Eigen::Vector3d rPRr(Hrf * rPFf);
+                const Eigen::Vector3d rPRr(Hfr.inverse() * rPFf);
 
                 // Compute the current position error and heading error in field {f} space
                 const double position_error = (Hfr.translation().head(2) - rPFf.head(2)).norm();
                 const Eigen::Vector3d uXRf  = Hfr.linear().col(0).head(2);
                 const double heading_error  = std::acos(std::max(-1.0, std::min(1.0, uXRf.dot(uHFf.head(2)))));
-                log<NUClear::DEBUG>("Position error: ", position_error, " Heading error: ", heading_error);
+
+                // If we have stopped but our position error is too high, then we need to resume walking to position
```
ysims, line 53: this comment seems to be the opposite

roles/webots/robocup.role
```diff
@@ -9,6 +9,8 @@ nuclear_role(
   support::configuration::GlobalConfig
   # Network
   network::PlotJuggler
+  network::NUClearNet
+  network::NetworkForwarder
```
ysims, line 13: do we want these in here

Tom0Brien: Probably best for testing behaviour, we can remove later

module/strategy/WalkToFieldPosition/src/WalkToFieldPosition.cpp
Tom0Brien: true

ysims (approved):

PR #1149, opened 2023-06-23
Fix rendered visual mesh aspect ratio in NUsight

JosephusPaye:
Currently when the GL view's aspect ratio is wider than the image (usually the case when zoomed in), the mesh is rendered past the edges of the image. This PR fixes that by correcting the aspect ratio of the rendered mesh to match the image.

This is what it looked like before: 

image

And here's after:

image


BrendanAnnable (approved):
Nice!

PR #1151, opened 2023-06-24
Separate model and packet watchdogs to prevent dropped packets stopping the robot's motions

ysims:
This PR splits the watchdog into two so if we get dropped servo packets we recover quickly.

At the beginning, we get the OpenCR model. A model watchdog watches for this and times out after 300ms. This watchdog is disabled once the model is received.
When the model watchdog is disabled and the model is received, we enable the packet watchdog. This times out after 20ms of not getting a packet.

I've tested this under normal circumstances and it looks good. I've tested it under packet dropping circumstances and it also looks good, minimal impact on the walk. I did notice a slight impact on the walk but I'd say that is expected when all packets are dropping constantly.

There is an issue where the packet watchdog isn't disabling quickly enough at the beginning before the watchdog triggers (twice). This may just be NUClear being slow or maybe it's the NUClear bug/feature with out of order packets. We can check when the bug fix goes in. I have a work around that works fine and shouldn't cause any issues.

### DIFF
--- a/module/platform/OpenCR/HardwareIO/src/HardwareIO.cpp
+++ b/module/platform/OpenCR/HardwareIO/src/HardwareIO.cpp
@@ -57,9 +57,56 @@ namespace module::platform::OpenCR {
                 nugus.servo_direction[i]  = config["servos"][i]["direction"].as<Expression>();
                 servo_states[i].simulated = config["servos"][i]["simulated"].as<bool>();
             }
+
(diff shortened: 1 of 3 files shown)


TrentHouliston:
If you're not ever re-enabling the watchdog you should unbind it instead of disabling it

Tom0Brien (approved):

PR #1152, opened 2023-06-24
Make the WalkState and Stability messages optional in SensorFilter

ysims:
SensorFilter couldn't run without these messages before, making it hard to run smaller roles testing singular things eg the test/visualmesh role. This PR makes those optional in SensorFilter so it doesn't need them to run. The messages are only used to utilise the walk for positional odometry so if they don't exist it will skip that step. This works similar to how it skips it when the robot isn't walk/stable.

### DIFF
--- a/module/input/SensorFilter/src/SensorFilter.cpp
+++ b/module/input/SensorFilter/src/SensorFilter.cpp
@@ -153,6 +153,19 @@ namespace module::input {
                 .disable();
     }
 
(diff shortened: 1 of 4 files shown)


Tom0Brien (approved):

TrentHouliston (commented):
These feel like a whole new module to me, some sort of walk odometry that would make it's own emit and send it that Sensors would potentially listen on.

PR #1153, opened 2023-06-25
Receive is spelt EI not IE

ysims:
It was annoying me.


TrentHouliston (approved):
I before E except after C. 
Although that rule is disproven by science!

PR #1157, opened 2023-06-26
Fix images being out of sync with sensors

Tom0Brien:
This PR fixes an issue with the latching method in `Camera.cpp` having incorrect logic for associating a sensors message to an image.


TrentHouliston (approved):
</3

PR #1159, opened 2023-06-26
Add a unit test for Director self intersection and a fix

TrentHouliston:
Adds a unit test for Director self-intersection and provides a fix.

When a single Provider intersected with itself before with optional tasks, all of them would run even though they shouldn't.

E.g. if two optional tasks A and B were both emitted that needed C they would both run even though they intersected with each other.

This test was pulled out from #1108 as the way that test was written relies on this behaviour being correct

### DIFF
--- a/module/extension/Director/tests/self_intersection.cpp
+++ b/module/extension/Director/tests/self_intersection.cpp
@@ -1,0 +1,141 @@
+/*
+ * This file is part of the NUbots Codebase.
+ *
+ * The NUbots Codebase is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * The NUbots Codebase is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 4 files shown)


module/extension/Director/tests/self_intersection.cpp
```diff
@@ -0,0 +1,141 @@
...
+            /**************
+             * TEST STEPS *
+             **************/
+            on<Trigger<Step<1>>, Priority::LOW>().then([this] {
+                events.push_back("emitting main task");
+                emit<Task>(std::make_unique<MainTask>());
+            });
+            on<Trigger<Step<2>>, Priority::LOW>().then([this] {
+                events.push_back("removing main task");
+                emit<Task>(std::unique_ptr<MainTask>());
```
ysims, line 88:
```suggestion
                emit<Task>(std::unique_ptr<MainTask>(nullptr));
```

module/extension/Director/tests/self_intersection.cpp
```diff
@@ -0,0 +1,141 @@
...
+            });
+            on<Trigger<Step<3>>, Priority::LOW>().then([this] {
+                events.push_back("emitting main task");
+                emit<Task>(std::make_unique<MainTask>());
+            });
+            on<Trigger<Step<4>>, Priority::LOW>().then([this] {
+                events.push_back("removing main task");
+                emit<Task>(std::unique_ptr<MainTask>());
```
ysims, line 96:
```suggestion
                emit<Task>(std::unique_ptr<MainTask>(nullptr));
```

module/extension/Director/tests/self_intersection.cpp
TrentHouliston: These do the same thing but I'll change it if you think it makes it clearer

ysims (approved):

PR #1163, opened 2023-06-27
Set servo max voltage to 16 volts on startup

ysims:
The servos may not have their max voltage set to 16 volts yet. If they are less, we will likely get errors on battery. This PR writes to the servos on startup to set the max voltage to 16 volts.

This has been tested and it eliminates the errors we are seeing on the robots with batteries.

### DIFF
--- a/module/platform/OpenCR/HardwareIO/src/hardwareio/startup.cpp
+++ b/module/platform/OpenCR/HardwareIO/src/hardwareio/startup.cpp
@@ -45,6 +45,27 @@ namespace module::platform::OpenCR {
         opencr.write(
             dynamixel::v2::SyncWriteCommand<uint8_t, 20>(uint16_t(DynamixelServo::Address::RETURN_DELAY_TIME), data));
 
+        // Set the maximum voltage for the servos to 16 volts, so that they will not error on battery
+        // For 16 volts, we need to write a value of 160 to the max voltage limit registers
+        uint8_t max_voltage_limit_l = 160 & 0xFF;
+        uint8_t max_voltage_limit_h = (160 >> 8) & 0xFF;
+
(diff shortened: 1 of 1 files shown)


tayloryoung6396:
This should probably be 16.8v, as a fully charged battery could be 16.8v (ignoring the likely voltage droop that will occur)

ysims:
> This should probably be 16.8v, as a fully charged battery could be 16.8v (ignoring the likely voltage droop that will occur)

Dynamixel maximum seems to be 16 volts on the specs

tayloryoung6396:
The max that they can "handle" isn't the value you're giving them though. You're giving them 16.8 at startup. Unless the max voltage field only accepts a maximum of 16 then this won't help

ysims:
> The max that they can "handle" isn't the value you're giving them though. You're giving them 16.8 at startup. Unless the max voltage field only accepts a maximum of 16 then this won't help

I can test in dynamixel wizard to see if it will go above, but the docs 
https://emanual.robotis.com/docs/en/dxl/mx/mx-106/#max-voltage-limit
say the value can only be in the range of 5v to 16v

ysims:
> > The max that they can "handle" isn't the value you're giving them though. You're giving them 16.8 at startup. Unless the max voltage field only accepts a maximum of 16 then this won't help
> 
> I can test in dynamixel wizard to see if it will go above, but the docs https://emanual.robotis.com/docs/en/dxl/mx/mx-106/#max-voltage-limit say the value can only be in the range of 5v to 16v

It is not possible to increase it more than 160 in dynamixel wizard

Tom0Brien (approved):

ysims:
Pretty sure this isn't relevant anymore after some fixes recently to account for the voltage issues.

PR #1167, opened 2023-06-28
Add behaviour tester purpose

Tom0Brien:
This PR adds a module which allows you to isolate running different combinations of strategies and purposes with user specified priority. 

This will be very useful for tuning and testing behavior components individually and saves us having to create individual roles for each behavior component.


ysims (approved):

module/purpose/Tester/README.md
```diff
@@ -0,0 +1,16 @@
+# Tester
+
```
ysims, line 2: maybe fill this in

ysims (approved):

PR #1169, opened 2023-06-28
Walk bug fixes and tuning

Tom0Brien:
This PR does the following:

- Fixes a bug with the `WalkGenerator` introduced when switching to the use of `Waypoint`. The z translational velocity of the swing foot and torso middle waypoint  was being set to `velocity_target.z()` when it is actually the angular yaw velocity.
- Simplifies generating trajectories slightly by merging three functions into a single one as they are only used once
- Slightly tunes billie walk params by adjusting pitch
- Tunes walk path planner (for walking to ball)

### DIFF
--- a/roles/test/behaviour.role
+++ b/roles/test/behaviour.role
@@ -1,0 +1,53 @@
+nuclear_role(
+  # FileWatcher, ConsoleLogHandler and Signal Catcher Must Go First and without it many roles do not run
+  extension::FileWatcher
+  support::SignalCatcher
+  support::logging::ConsoleLogHandler
+  # Config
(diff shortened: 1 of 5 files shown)


roles/test/behaviour.role
```diff
@@ -0,0 +1,53 @@
+nuclear_role(
```
ysims, line 1: how come this is in this PR?

ysims (approved):

roles/test/behaviour.role
Tom0Brien: was needed to test path planning and wasn't to added when Tester module was created

PR #1173, opened 2023-06-29
Align robot to ball

Tom0Brien:
This PR adds a strategy to align to the ball.

This is used to ensure the robot is facing the ball before walking towards it as path planning is not so great on the real robot.

Misc:

- Tune path planner and walk for nugus4
- Adds ability to toggle y velocity on/off in the WalkTo path planner


module/strategy/AlignRobotToBall/src/AlignRobotToBall.cpp
```diff
@@ -0,0 +1,64 @@
...
+                // If the angle to the ball is greater than maximum threshold, rotate to face the ball
+                if (rotate) {
+                    // Check if we need to rotate clockwise or anti-clockwise
+                    if (rBRr.y() < 0) {
+                        emit<Task>(std::make_unique<TurnOnSpot>(true));
+                    }
+                    else {
+                        emit<Task>(std::make_unique<TurnOnSpot>(false));
+                    }
+                }
```
ysims, line 60:
```suggestion
                if (rotate && turn.run_state == GroupInfo::RunState::NO_TASK) {
                    // Check if we need to rotate clockwise or anti-clockwise
                    if (rBRr.y() < 0) {
                        emit<Task>(std::make_unique<TurnOnSpot>(true));
                    }
                    else {
                        emit<Task>(std::make_unique<TurnOnSpot>(false));
                    }
                }
                else if (rotate) { // already rotating and still rotating
                    emit<Task>(std::make_unique<Idle>());
                }
```

module/strategy/AlignRobotToBall/src/AlignRobotToBall.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+        on<Configuration>("AlignRobotToBall.yaml").then([this](const Configuration& config) {
+            // Use configuration here from file AlignRobotToBall.yaml
+            this->log_level                    = config["log_level"].as<NUClear::LogLevel>();
+            cfg.start_rotating_angle_threshold = config["start_rotating_angle_threshold"].as<double>();
+            cfg.stop_rotating_angle_threshold  = config["stop_rotating_angle_threshold"].as<double>();
+        });
+
+        on<Provide<AlignRobotToBallTask>, With<Ball>, With<Sensors>, Every<30, Per<std::chrono::seconds>>>().then(
+            [this](const Ball& ball, const Sensors& sensors) {
```
ysims, line 32:
```suggestion
        on<Provide<AlignRobotToBallTask>,  Uses<TurnOnSpot>, With<Ball>, With<Sensors>, Every<30, Per<std::chrono::seconds>>>().then(
            [this](const Uses<TurnOnSpot>& turn, const Ball& ball, const Sensors& sensors) {
```

module/strategy/AlignRobotToBall/src/AlignRobotToBall.cpp
Tom0Brien: This breaks my hack of resetting walk velocity magnitude to minimum in the regular path planner using the turn on spot.

PR #1102, opened 2023-06-07
Fix SSH key issue

Tom0Brien:
This PR fixes issue with SSH key from https://github.com/NUbots/NUbots/pull/1090

PR #1174, opened 2023-06-29
Fix Arch install scripts

JesseWilliamson:
This pr fixes a few small problems in the arch install scripts. Our nuc12s are working fine, but this will make things easier when nucs need to be flashed in future.

Changes:
1. Add a pacman.conf file to ignore keychain when installing packages (we do this in docker since #896 so I don't think there's a good reason not to do it here)
2. Remove request for missing file `docker/home/nubots/.ssh/id_rsa.pub` (removed in #1102 )
3. Add user libraries to `/etc/ld.so.conf` to allow systemconfiguration to run
4. Add a blank zsh config file to prevent the prompt when restarting after post-install script


module/tools/SystemConfiguration/data/system/default/etc/pacman.conf
```diff
@@ -0,0 +1,19 @@
+[options]
+HoldPkg      = pacman glibc
+Architecture = auto
+ParallelDownloads = 5
+CheckSpace
+Color
+ILoveCandy
```
CMurtagh-LGTM, line 7: :candy:

doc/ArchInstall/arch-chroot_install.sh
```diff
@@ -162,22 +162,16 @@ systemctl enable wpa_supplicant@${WIFI_INTERFACE}
...
-############
-# SSH KEYS #
-############
-
-# Create user ssh directory
-mkdir -p ${HOME}/.ssh
-
-# Register docker client as an authorized user
-wget https://raw.githubusercontent.com/NUbots/NUbots/main/docker/home/nubots/.ssh/id_rsa.pub \
-    -O ${HOME}/.ssh/authorized_keys
```
CMurtagh-LGTM, line 177: I thought we needed this, how do we ssh into the robot without this?

TrentHouliston: Using the password I assume.
Also the new docker image just forwards the ssh credentials from the host

doc/ArchInstall/arch_install.sh
```diff
@@ -29,8 +29,30 @@ mount ${ROOT} /mnt
...
+
+[core]
+Include = /etc/pacman.d/mirrorlist
+
+[extra]
+Include = /etc/pacman.d/mirrorlist
+
+[community]
+Include = /etc/pacman.d/mirrorlist
+EOF
```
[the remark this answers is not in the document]
JesseWilliamson: Why do you want to remove this but not the copy in systemconfigurator? It's not just for pacstrap, it's copied over for the new root location two lines later. I don't understand.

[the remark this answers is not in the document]
TrentHouliston: In the past we've had issues where you need to install using the mirrorlist from the dockerfile too... so you might need to put that in so pacstrap can use it.

As for this file, systemconfiguration will overwrite it anyway later

doc/ArchInstall/arch-chroot_install.sh
```diff
@@ -162,22 +162,16 @@ systemctl enable wpa_supplicant@${WIFI_INTERFACE}
 # ZSH SHELL #
 #############
 
-# Change user shell to zsh
+# Change user shell to zsh and create a blank config file to prevent user prompt
 chsh -s /usr/bin/zsh ${USER}
+touch ${HOME}/.zshrc
```
Bidski, line 167: This isn't needed because `SystemConfiguration` will setup this file correctly. So you should only ever see this prompt once. `SystemConfiguration` also sets the user shell to `zsh`, so you can probably get rid of all of this and just let `SystemConfiguration` handle it all.

doc/ArchInstall/arch_install.sh
```diff
@@ -29,8 +29,9 @@ mount ${ROOT} /mnt
 mkdir -p /mnt/boot/efi
 mount ${BOOT} /mnt/boot/efi
 
-# Bootstrap Pacman
+# Bootstrap Pacman, and trust everything to prevent pgp key errors
 pacstrap /mnt base linux linux-firmware
+cp /etc/pacman.conf /mnt/etc/pacman.conf
```
Bidski, line 34: You can do the `pacman.conf` with `SystemConfiguration` too. Copying `pacman.conf` is just creating another copy of the file that you need to rememebr to maintain.

Bidski (changes requested):

doc/ArchInstall/arch-chroot_install.sh
TrentHouliston: Yep, you could just leave the install as bash and not install zsh at all during setup. System configuration will install zsh for you and setup the zshrc stuff

doc/ArchInstall/arch_install.sh
TrentHouliston: One thing you do need to do is make sure that when you're installing your mirror list for pacstrap is the same as the one on docker, otherwise you need to downgrade later

JesseWilliamson: That makes sense. But won't using that old mirrorlist likely give us expired pgp key errors now that we're not using the [pacman.conf ](https://github.com/NUbots/NUbots/pull/1174#discussion_r1295756966) file until systemconfigurator copies it over?

doc/ArchInstall/arch-chroot_install.sh
JesseWilliamson: I think the best thing to do is just use bash until systemconfigurator. This avoids the prompt, and systemconfigurator can set it up properly when it's actually going to be useful.

JesseWilliamson:
NUbots/NUbook#281, I think mirror list date is relevant here too

doc/ArchInstall/arch_install.sh
TrentHouliston: It will, you'll need to make sure at least that setting is set in the conf file

Bidski: You can always make your install script download [pacman.conf](https://raw.githubusercontent.com/NUbots/NUbots/main/docker/etc/pacman.conf) from the repository. 

You could also make a similar `mirrorlist` file for docker (and update [Dockerfile](https://github.com/NUbots/NUbots/blob/main/docker/Dockerfile#L20) to use it). This will make it easier for you to keep to the docker image and the install scripts in sync

doc/ArchInstall/arch_install.sh
```diff
@@ -29,9 +29,11 @@ mount ${ROOT} /mnt
 mkdir -p /mnt/boot/efi
 mount ${BOOT} /mnt/boot/efi
 
-# Bootstrap Pacman, and trust everything to prevent pgp key errors
+# Bootstrap Pacman
 pacstrap /mnt base linux linux-firmware
-cp /etc/pacman.conf /mnt/etc/pacman.conf
+# Set mirrorlist to match Dockerfile
+# ↓↓↓ UPDATE MIRRORS HERE ↓↓↓
+echo -e "Server=https://archive.archlinux.org/repos/2022/07/10/\$repo/os/\$arch" | tee /etc/pacman.d/mirrorlist
```
Bidski, line 36: I think this would be better if you made a file at `docker/etc/pacman.d/mirrorlist` that had the correct contents and then you can `wget` it when installing here and you can update the Dockerfile to copy that file into the docker image. That way there is only 1 location where the mirrorlist is set and you can't forget to update it somewhere

Bidski (changes requested):

doc/ArchInstall/arch_install.sh
JesseWilliamson: Thanks, that sounds like a good way to do it. I've also added updating that file as an option to systemconfigurator. I'm not sure if it's better to do it that way or just copy the file, but I do think it should be in systemconfigurator somewhere.

module/tools/SystemConfiguration/src/SystemConfiguration.cpp
```diff
@@ -181,6 +181,16 @@ namespace module::tools {
...
+            /*************
+             * MIRRORLIST *
+             **************/
+            if (config["update_mirrorlist"].as<bool>()) {
+                log<NUClear::INFO>("Updating mirror list to match dockerfile");
+                std::system(
+                    "wget https://raw.githubusercontent.com/NUbots/NUbots/main/docker/etc/pacman.d/mirrorlist \
+                 -O /etc/pacman.d/mirrorlist");
+            }
+
```
Bidski, line 193: This is unnecessary, [this file](https://github.com/NUbots/NUbots/blob/69e6f4f12fc5c2468108eeb0ad726f08dca48791/module/tools/SystemConfiguration/CMakeLists.txt) already causes the docker images mirror list to be copied into `SystemConfiguration`s data files, which will automatically result in the mirror list being updated whenever `SystemConfiguration` is run. 

You need to do the extra work in doc/ArchInstall/arch-chroot_install.sh to ensure that `SystemConfiguration` can run in the first place (if you don't sometimes you can't even run `pacman` to downgrade packages to the right versions because there are too many differences in system libraries and package dependencies)

Bidski (changes requested):

module/tools/SystemConfiguration/data/system/default/etc/pacman.conf
```diff
@@ -0,0 +1,19 @@
+[options]
```
Bidski, line 1: Wait, this file shouldn't be here. In `module/tools/SystemConfiguration/CMakeLists.txt` you should have 
```cmake
nuclear_module(DATA_FILES "/etc/pacman.d/mirrorlist:system/default/etc/pacman.d" "/etc/pacman.conf:system/default/etc")
```

This will copy the `pacman.conf` from inside the docker image into teh data files for `SystemConfiguration`. This will then install result in the docker image's `pacman.conf` being installed on to the device

Bidski (changes requested):

module/tools/SystemConfiguration/data/system/default/etc/pacman.conf
JesseWilliamson: Alright, thanks. That makes sense :+1:

ysims (approved):

Bidski (approved):

PR #1176, opened 2023-07-04
Update field description

Tom0Brien:
Update field description to use full sized field by default.


ysims:
Changed the base to robocup2023, we want it in the robocup branch. this can be merged in

PR #1178, opened 2023-07-04
Update camera params for the robocup field

ysims:
looks better and the ball detector stuff makes it find the ball more


ysims:
I guess this is obsolete

PR #1180, opened 2023-07-04
Add buzzer functionality for the OpenCR board

jpptm:
This pull request aims to add a functionality for the OpenCR's buzzer. It cycles through all the servo motors and if it finds one servo that has a temperature that is above the configurable value, it writes the frequency value to the board and plays the sound of the buzzer.

Currently, a frequency of 0 is written to the board's buzzer but the functionality for the firmware  is already implemented in the firmware


module/platform/OpenCR/HardwareIO/src/hardwareio/send_request.cpp
```diff
@@ -88,7 +88,18 @@ namespace module::platform::OpenCR {
...
+            // Run through all of the servo states and if even one motor is above the temeperature
+            // threshold, fill the buzzer field
+            data.buzzer = std::any_of(
+                servo_states.begin(),
+                servo_states.cend(),
+                [](const ServoState& servo) -> bool { return servo.temperature > 70.0; }
+            ) ? 880u : 0u;
+
+            // This line serves as a naive test
+            data.buzzer = 880u;
```
jpptm, line 102: Will remove this later once properly tested, will also make several values configurable and put them to their corresponding yaml files.

Tom0Brien (approved):

PR #1186, opened 2023-07-05
Ball Detector Tuning Robocup2023

BaileyJoe:
Tuning for ball detector for Robocup2023.
Slightly reduces confidence threshold and increases distance disagreement substantially.


ysims:
LGTM

PR #1202, opened 2023-07-06
Add the ability to reference cameras by bus/device rather than serial

### LINKED ISSUE #1200: Camera module should base connections on the port number rather than serial number
Swapping camera parts at RoboCup is difficult as when we do we need to change the serial numbers in the configuration.
Instead, we should scan which cameras are connected and find some way to assign them left/right based on which USB port they are attached to.

TrentHouliston:
Currently, the only way to reference cameras in the Camera module is to provide their serial number.

This change allows them to be defined by USB bus and device numbers.
This way you can swap cameras around between robots and so long as they are plugged into the same ports they will work without updating serial numbers.

Fixes #1200

### DIFF
--- a/module/input/Camera/src/Camera.cpp
+++ b/module/input/Camera/src/Camera.cpp
@@ -6,6 +6,8 @@ extern "C" {
 
 #include <cmath>
 #include <fmt/format.h>
+#include <libusb-1.0/libusb.h>
+#include <memory>
 
 #include "aravis_wrap.hpp"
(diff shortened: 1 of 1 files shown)


tayloryoung6396:
Would require actually having the cameras plugged into the left and right port 😛 Would be much more convenient if you could look for the nose or do some calibration at startup to determine left and rightness based on an image.

The camera cables are nontrivial to know leftness and rightness just by looking at them

PR #1205, opened 2023-07-06
Add localisation reset functionality

jpptm:
This pull request aims to add a localisation reset feature. The user has to press the left (black) button on the back panel of the robot. A buzzer will ring for a configurable amount of time in milliseconds to let the user know that the localisation process has been successfully reset.


ysims (commented):
This should probably be decoupled so that when it's pressed, buzz and emit a button down message. Then that button down message triggers in the localisation module and runs the reset.

jpptm:
PR will be added to main with the right files instead

ysims:
Don't forget to delete this branch if not needed anymore

PR #1208, opened 2023-07-07
Correct the implementation of the buzzer feature

jpptm:
Generalise the access to the buzzer message by adding a frequency field so other modules can use this message as required.

### DIFF
--- a/module/platform/OpenCR/HardwareIO/src/hardwareio/process_data.cpp
+++ b/module/platform/OpenCR/HardwareIO/src/hardwareio/process_data.cpp
@@ -133,9 +133,7 @@ namespace module::platform::OpenCR {
 
         for (const auto& servo : servo_states) {
             if (servo.temperature > cfg.max_tol_temp) {
-                auto buzzer_msg = std::make_unique<Buzzer>();
-                buzzer_msg -> buzzer_frequency = cfg.buzzer_freq;
-                emit(buzzer_msg);
+                emit(std::make_unique<Buzzer>(cfg.buzzer_freq));
                 break;
             }
         }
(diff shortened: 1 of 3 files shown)


module/platform/OpenCR/HardwareIO/src/hardwareio/process_data.cpp
```diff
@@ -133,7 +133,9 @@ namespace module::platform::OpenCR {
 
         for (const auto& servo : servo_states) {
             if (servo.temperature > cfg.max_tol_temp) {
-                emit(std::make_unique<Buzzer>());
+                auto buzzer_msg = std::make_unique<Buzzer>();
+                buzzer_msg -> buzzer_frequency = cfg.buzzer_freq;
+                emit(buzzer_msg);
```
ysims, line 138:
```suggestion
                emit(std::make_unique<Buzzer>(cfg.buzzer_freq));
```

shared/message/output/Buzzer.proto
```diff
@@ -2,4 +2,6 @@ syntax = "proto3";
 
 package message.output;
 
-message Buzzer {}
+message Buzzer {
+    int32 buzzer_frequency = 1;
+}
```
ysims, line 7: Put a comment on the buzzer message. It isn't clear what is buzzing to someone who isn't that familiar with this kind of stuff.

jpptm: Fixed

ysims:
This PR is to merge into robocup2023. It doesn't make sense to add to this anymore, you may as well make a PR to main with the buzzer functionality.

jpptm:
@ysims just change the merge to main?

jpptm:
To merge properly tested code into main instead

ysims:
Branch should be deleted if not needed anymore

PR #1209, opened 2023-07-09
Robocup parkour scripts

Crafty15:
Was made under duress. Mostly works


ysims:
Fix up the files changed

ysims (commented):
Should fix this up so we can merge it. There are a lot of extra files changed in this. Is the script also the mirrored script?

PR #1213, opened 2023-07-10
Correct the visual mesh triangle index to correct for the new neighbour order

TrentHouliston:
When the visual mesh was upgraded to a new version a while ago the order of neighbours changed.
Previously the order was

```
  1  2
   \/
3 ---- 4
   /\
  5  6
```

The new order of indexes is like this
```
  2  3
   \/
1 ---- 4
   /\
  6  5
```

This PR fixes the triangle indexes so it correctly shows the visual mesh as a coloured mesh again

Before
image
After
image


JosephusPaye:
Looks good, but the build seems to be failing.

TrentHouliston:
Lighter blue
image
image

BrendanAnnable (approved):

PR #1216, opened 2023-08-02
Add robot mesh detections to NUsight visual mesh display

ysims:
The NUsight mesh colours are hardcoded with environment in the fourth position. Now that the network has an extra class, robots are the fourth index and environment is the fifth. This is consistent across the networks we have currently. 
NUsight has been displaying robots as black and ignoring environment (resulting in also black). 
This PR changes this so that robots appear in a different colour to the environment. 
Robots set to magenta.

image
image

### DIFF
--- a/nusight2/src/client/components/vision/vision_camera/shaders/visual_mesh.frag
+++ b/nusight2/src/client/components/vision/vision_camera/shaders/visual_mesh.frag
@@ -12,12 +12,12 @@ varying float vEnvironment;
 
 void main() {
 
-    vec3 colour = vBall * vec3(1.0, 0.0, 0.0)           // Red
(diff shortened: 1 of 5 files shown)


BrendanAnnable (commented):
Doesn't this invalidate old nbs recordings? Is there a reason robot wasn't added as the 5th/last class?

ysims:
> Doesn't this invalidate old nbs recordings? Is there a reason robot wasn't added as the 5th/last class?

That's the way the mesh was trained. It makes sense to me to have background as the last one. This has been like this for a long time. NUsight has been technically broken for that long. This is also why I want positions of classes to be dynamic based on the proto, not hardcoded. Any nbs file that is that old will have more problems than that.

BrendanAnnable (approved):

PR #1217, opened 2023-08-09
Delete directorkeyboardwalk.role

Tom0Brien:
This PR deletes unneeded (keyboardwalk is the same now) directorkeyboardwalk role.


ysims (approved):
ah missed this one in the purge

PR #1218, opened 2023-08-09
Change webots image format from BGR3 to RGBA

Tom0Brien:
This PR changes webots image format from BGR3 to RGBA.

Requires https://github.com/NUbots/NUWebots/pull/103

### DIFF
--- a/module/platform/Webots/src/Webots.cpp
+++ b/module/platform/Webots/src/Webots.cpp
@@ -801,7 +801,7 @@ namespace module::platform {
             image->name           = camera.name;
             image->dimensions.x() = camera.width;
             image->dimensions.y() = camera.height;
-            image->format         = fourcc("BGR3");  // Change to "JPEG" when webots compression is implemented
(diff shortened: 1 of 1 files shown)


ysims (commented):
The extract images tool would also need to be updated iirc.
This looks like it would break vision when using the official robocup world?

Tom0Brien:
> The extract images tool would also need to be updated iirc. This looks like it would break vision when using the official robocup world?

Looks like extract image tool saves images as JPEG so should be fine. 

Vision shouldn't break, let me confirm.

Tom0Brien:
> > The extract images tool would also need to be updated iirc. This looks like it would break vision when using the official robocup world?
> 
> Looks like extract image tool saves images as JPEG so should be fine.
> 
> Vision shouldn't break, let me confirm.

I see what you mean now with the `player` controller in `hlvs_webots`. We could make a fork and fix it or do the format conversion on the NUbots side only. Let's put this PR on hold for now until we can get `opencl` working in docker because its not important unless it helps with that

ysims:
Can the branch be deleted too

PR #1222, opened 2023-08-12
NUsight: Add R3F polylines

BrendanAnnable:
Renames the vision objects "straight lines" renderer to polylines and refactors it to use R3F components.

Removed the ray version, as we can easily map them on the CPU via `project` if we need.

Screenshot 2023-08-12 at 9 07 00 pm

After I port vision (likely the next PR), there will be an image view that matches the highlights:

Screenshot 2023-08-14 at 3 31 30 pm


nusight2/src/client/components/camera/objects/polyline/polyline.tsx
TrentHouliston, line 1: I think it makes more sense to have a bunch of `<PolyLine>` elements with one for each series of line segments rather than a giant `<PolyLines>` that has all of them!

nusight2/src/client/components/three/three_fiber.tsx
```diff
@@ -44,3 +43,33 @@ export const PerspectiveCamera = ({
...
+  props: JSX.IntrinsicElements["rawShaderMaterial"] & {
+    uniforms: { [k: string]: any };
+  },
+) => {
+  const uniforms = React.useMemo(() => props.uniforms, []);
+  for (const key in props.uniforms) {
+    uniforms[key].value = props.uniforms[key].value;
+  }
+  return <rawShaderMaterial {...props} uniforms={uniforms} />;
+};
```
TrentHouliston, line 75: Does this reuse the same shader material for the same uniforms?
Shouldn't you keep using a clone of the same shader and update the uniforms?
Or does three fibre handle not recompiling the shader each time.

BrendanAnnable: tl;dr Yes it reuses the compiled shaders. No need to clone.

R3F does nothing special, all it does is match the lifetime of rendered objects to the lifetime of the component.

If we ignore uniforms, this component is essentially equivalent to:

```ts
export const RawShaderMaterial = (props) => {
  const shader = React.useMemo(() => new THREE.RawShaderMaterial(), []);
  shader.vertexShader = props.vertexShader;
  shader.fragmentShader = props.fragmentShader;
  return <primitive object={shader}/>;
}
```

i.e. all it does it cache the instance.

So your real question is, does three.js recompile the shaders every time we render?

And the answer is no, it has an internal [WebGLShaderCache](https://github.com/mrdoob/three.js/blob/master/src/renderers/webgl/WebGLShaderCache.js), which thanks to string interning, will always hit, given the same string:

https://github.com/mrdoob/three.js/blob/c02fb47ade2a3666b9b8845cb159dcbe3d7bc932/src/renderers/webgl/WebGLShaderCache.js#L96

However three.js does not like you replacing `shader.uniforms` with a new object after first-render, which is why I cache the uniforms object itself and instead mutate individual values. I'll see if I can write a test for it.

TrentHouliston (approved):

PR #1225, opened 2023-08-17
Improve RawSensors error handling [HW-16]

dex-k:
Change RawSensors error handling to, broadly, stop odometry from exploding when we get any error. Now:
- Odometry will update regardless of *hardware* errors, and will only be conditional on subcontroller packet errors (which are the only ones relevant to the IMU)
- Other sensors message values will be conditional only to the specific errors which might be relevant, not just any error.

It also re-implements a check which disables the servo input voltage error as long as it's below the battery voltage. This was in the code previously but I removed it a while ago because I didn't know what it was for. Turns out it was important.

This also removes all references to Dynamixel protocol v1 errors which I'd left in pre-robocup for just-in-case compatibility, but are no longer needed. Hopefully there isn't anything still hanging around, but there may be. It compiles at least.

### Reviewers, Note

~~I'm a bit uncertain about some of the error handling for overload, input voltage and overheating errors. I've kept the same logic as before for now, assuming that what was previously there was there because it works/was written by people much smarter than me. See lines 73-87 in RawSensors.cpp for the comments there (line numbers might change if I change stuff)~~

### Things left to do

- [x] Test it on a real robot (right now it just compiles)

### DIFF
--- a/module/input/SensorFilter/src/sensorfilter/RawSensors.cpp
+++ b/module/input/SensorFilter/src/sensorfilter/RawSensors.cpp
@@ -35,8 +35,13 @@ namespace module::input {
     void SensorFilter::update_raw_sensors(std::unique_ptr<Sensors>& sensors,
                                           const std::shared_ptr<const Sensors>& previous_sensors,
                                           const RawSensors& raw_sensors) {
+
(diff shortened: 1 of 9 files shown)


ysims (approved):
lgtm

PR #1230, opened 2023-09-04
Tune webots scripts for new kinematics

ysims:
With the new up-to-date model in Webots, our scripts needed tuning. 
Getups were not working, minor fixes to the knee angles fixed that.
Stand has been set to the same as the walk at idle, and the final frame of the getups matches this Stand.


Tom0Brien (approved):

Tom0Brien:
Just needs formatting

Tom0Brien (approved):

PR #1242, opened 2023-09-24
Make the dockerise pass the broadcast address through to the container

TrentHouliston:
Currently when you run ./b yarn and friends on windows you need to work out what address to use so that the containers can find eachother.
This pr uses docker inspect to see what network you're running on (if it's not the hosts stack) and calcualtes the broadcast address.
This is then put in an environment variable where future tools can read it and use it.

### Pre-PR Checklist:

Have you

### Reviewers, Note

This doesn't use the environment variable for anything yet, that will need to be added to nusight and the NUClearNet module.
However in order to test that it's working you can run `./b shell` on windows or mac or `./b shell --network bridge` on linux.

Then inside the shell running `echo ${BROADCAST_ADDRESS}` should print out the broadcast address of the internal docker network.

### Things left to do

- [ ] Use this environment variable for something
- [ ] Maybe work out the broadcast address for the host networking mode too

### DIFF
--- a/tools/utility/dockerise/run.py
+++ b/tools/utility/dockerise/run.py
@@ -1,8 +1,11 @@
 #!/usr/bin/env python3
 
+import json
 import os
 import re
 import shutil
+import socket
+import struct
 import subprocess
 import sys
 from subprocess import DEVNULL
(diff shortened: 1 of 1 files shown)


TrentHouliston:
My suggestion for how to use this information is the following.

Change the default config in NUClearNet.yaml to have an empty string for IP (or a missing field)

Then do the following to determine the NUClearNet IP

1. See if there is a config value and if so use it
1. If not see if there is an environment variable and if so use it
1. If not try to work out what the broadcast IP of the default interface is
1. If not fall back to some default IP (or crash)

Then do the same type of logic for NUsight

1. See if an argument is provided by --network
1. See if an environment variable exists
1. Try to get the default broadcast address
1. Fallback to some default IP (or crash)

PR #1243, opened 2023-09-25
Improve odometry with anchor point method

Tom0Brien:
The current method for estimating the x-y translation and yaw of odometry involves integrating the walk command. This PR replaces this with an "anchor point" method. We select an anchor point A which is a left/right foot base frame and assume that this point is in contact, fixed to the ground in world space, giving us Hwa. We then simply compute the pose of the torso from this frame (Hat) using kinematics. This allows us to compute Hwt = Hwa * Hat. Finally, we replace the roll and pitch components of the rotation using the Mahony filter estimation. After a step has been taken the anchor point is switched to the next planted foot, illustrated below.

image

A nice blog post can be found here discussing the idea: https://scaron.info/robotics/floating-base-estimation.html
This method can be improved with a more advanced filtering technique, however, this simple idea provides quite good results.

### DIFF
--- a/module/input/SensorFilter/data/config/SensorFilter.yaml
+++ b/module/input/SensorFilter/data/config/SensorFilter.yaml
@@ -77,6 +77,7 @@ mahony:
 
 # Tuning parameters to scale walk command to match actual achieved velocity for deckreckoning x, y, theta
 deadreckoning_scale: [1, 1, 0.6]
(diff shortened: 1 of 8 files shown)


ysims:
Can you add this info and that link to NUbook on the odometry page

ysims (approved):

PR #1244, opened 2023-09-25
Update initial x position in ball localisation

Tom0Brien:
Currently the ball is initialized 0 meters away from the robot which sometimes causes the kick to be executed when the ball is first seen. This PR updates the initial x position of the ball to 5m.


TrentHouliston (approved):
Robot turns on and tries to get up walk and kick all at the same time!

PR #1245, opened 2023-09-25
Add visualisation of the ball in NUsight localisation view

Tom0Brien:
This PR adds ball visualisation to the localisation view. This PR also updates the field space transform of both field line points and ball to be computed property.

image

### DIFF
--- a/nusight2/src/client/components/localisation/network.ts
+++ b/nusight2/src/client/components/localisation/network.ts
@@ -39,13 +39,13 @@ export class LocalisationNetwork {
   @action.bound
   private onFieldLines(robotModel: RobotModel, fieldLines: message.vision.FieldLines) {
     const robot = LocalisationRobotModel.of(robotModel);
-    robot.fieldLinesDots.rPWw = fieldLines.rPWw.map((rPWw) => Vector3.from(rPWw).applyMatrix4(robot.Hfw));
+    robot.fieldLinesDots.rPFf = fieldLines.rPWw.map((rPWw) => Vector3.from(rPWw).applyMatrix4(robot.Hfw));
   }
 
   @action.bound
   private onBall(robotModel: RobotModel, ball: message.localisation.Ball) {
     const robot = LocalisationRobotModel.of(robotModel);
-    robot.ball = Vector3.from(ball.rBWw).applyMatrix4(robot.Hfw);
+    robot.ball.rBFf = Vector3.from(ball.rBWw).applyMatrix4(robot.Hfw);
   }
 
   @action
--- a/nusight2/src/client/components/localisation/network.ts
+++ b/nusight2/src/client/components/localisation/network.ts
@@ -45,7 +45,7 @@ export class LocalisationNetwork {
   @action.bound
   private onBall(robotModel: RobotModel, ball: message.localisation.Ball) {
     const robot = LocalisationRobotModel.of(robotModel);
-    robot.ball.rBWw = Vector3.from(ball.rBWw);
+    robot.ball = { rBWw: Vector3.from(ball.rBWw) };
   }
 
   @action
--- a/nusight2/src/client/components/localisation/network.ts
+++ b/nusight2/src/client/components/localisation/network.ts
@@ -39,13 +39,13 @@ export class LocalisationNetwork {
   @action.bound
   private onFieldLines(robotModel: RobotModel, fieldLines: message.vision.FieldLines) {
     const robot = LocalisationRobotModel.of(robotModel);
-    robot.fieldLinesDots.rPFf = fieldLines.rPWw.map((rPWw) => Vector3.from(rPWw).applyMatrix4(robot.Hfw));
+    robot.fieldLinePoints.rPWw = fieldLines.rPWw.map((rPWw) => Vector3.from(rPWw));
   }
 
   @action.bound
   private onBall(robotModel: RobotModel, ball: message.localisation.Ball) {
     const robot = LocalisationRobotModel.of(robotModel);
-    robot.ball.rBFf = Vector3.from(ball.rBWw).applyMatrix4(robot.Hfw);
+    robot.ball.rBWw = Vector3.from(ball.rBWw);
   }
 
   @action
--- a/nusight2/src/client/components/localisation/network.ts
+++ b/nusight2/src/client/components/localisation/network.ts
@@ -18,6 +18,7 @@ export class LocalisationNetwork {
     this.network.on(message.input.Sensors, this.onSensors);
     this.network.on(message.localisation.Field, this.onField);
     this.network.on(message.vision.FieldLines, this.onFieldLines);
+    this.network.on(message.localisation.Ball, this.onBall);
   }
 
   static of(nusightNetwork: NUsightNetwork, model: LocalisationModel): LocalisationNetwork {
@@ -39,6 +40,12 @@ export class LocalisationNetwork {
   private onFieldLines(robotModel: RobotModel, fieldLines: message.vision.FieldLines) {
     const robot = LocalisationRobotModel.of(robotModel);
     robot.fieldLinesDots.rPWw = fieldLines.rPWw.map((rPWw) => Vector3.from(rPWw).applyMatrix4(robot.Hfw));
+  }
+
+  @action.bound
+  private onBall(robotModel: RobotModel, ball: message.localisation.Ball) {
+    const robot = LocalisationRobotModel.of(robotModel);
+    robot.ball = Vector3.from(ball.rBWw).applyMatrix4(robot.Hfw);
   }
 
   @action
--- a/nusight2/src/client/components/localisation/robot_model.ts
+++ b/nusight2/src/client/components/localisation/robot_model.ts
@@ -141,7 +141,7 @@ export class LocalisationRobotModel {
     Rwt: Quaternion;
     motors: ServoMotorSet;
     fieldLinePoints: { rPWw: Vector3[] };
-    ball: { rBWw: Vector3 };
+    ball?: { rBWw: Vector3 };
   }) {
     this.model = model;
     this.name = name;
@@ -163,7 +163,6 @@ export class LocalisationRobotModel {
       Rwt: Quaternion.of(),
       motors: ServoMotorSet.of(),
       fieldLinePoints: { rPWw: [] },
-      ball: { rBWw: Vector3.of() },
     });
   });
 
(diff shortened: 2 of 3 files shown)


nusight2/src/client/components/localisation/robot_model.ts
```diff
@@ -120,6 +120,7 @@ export class LocalisationRobotModel {
   @observable Rwt: Quaternion; // Torso to world rotation.
   @observable motors: ServoMotorSet;
   @observable fieldLinesDots: { rPWw: Vector3[] };
+  @observable ball: Vector3;
```
BrendanAnnable, line 123: Can we match field lines so that there is a proper name

```suggestion
  @observable ball: { rBFf: Vector3 };
```

nusight2/src/client/components/localisation/network.ts
```diff
@@ -41,6 +42,12 @@ export class LocalisationNetwork {
     robot.fieldLinesDots.rPWw = fieldLines.rPWw.map((rPWw) => Vector3.from(rPWw).applyMatrix4(robot.Hfw));
```
BrendanAnnable, line 42:
```suggestion
    robot.fieldLinesDots.rPFf = fieldLines.rPWw.map((rPWw) => Vector3.from(rPWw).applyMatrix4(robot.Hfw));
```

Aside... shouldn't this be named `rPFf`?

nusight2/src/client/components/localisation/view.tsx
```diff
@@ -201,3 +202,19 @@ const FieldLineDots = ({ model }: { model: LocalisationModel }) => (
...
+  <>
+    {model.robots.map(
+      (robot) =>
+        robot.visible && (
+          <object3D key={robot.id}>
+            <mesh position={robot.ball.toArray()}>
+              <sphereBufferGeometry args={[robot.ball.z, 20, 20]} />
+              <meshStandardMaterial color="#1E90FF" />
+            </mesh>
+          </object3D>
```
BrendanAnnable, line 216: No need for an `Object3D` container here since we only have one mesh:

```suggestion
          <mesh position={robot.ball.toArray()} key={robot.id}>
            <sphereBufferGeometry args={[robot.ball.z, 20, 20]} />
            <meshStandardMaterial color="#1E90FF" />
          </mesh>
```

nusight2/src/client/components/localisation/view.tsx
```diff
@@ -201,3 +202,19 @@ const FieldLineDots = ({ model }: { model: LocalisationModel }) => (
...
 );
+
+const Ball = ({ model }: { model: LocalisationModel }) => (
+  <>
+    {model.robots.map(
+      (robot) =>
+        robot.visible && (
+          <object3D key={robot.id}>
+            <mesh position={robot.ball.toArray()}>
+              <sphereBufferGeometry args={[robot.ball.z, 20, 20]} />
```
BrendanAnnable, line 213: It is expensive to have a dynamic geometry definition, it is more efficient to instead scale the mesh dynamically:

```suggestion
            <mesh position={robot.ball.toArray()} scale={[robot.ball.z, robot.ball.z, robot.ball.z]}>
              <sphereBufferGeometry args={[1, 20, 20]} />
```

Although, why exactly are we scaling by `ball.z`?

nusight2/src/client/components/localisation/view.tsx
```diff
@@ -201,3 +202,19 @@ const FieldLineDots = ({ model }: { model: LocalisationModel }) => (
     )}
   </>
 );
+
+const Ball = ({ model }: { model: LocalisationModel }) => (
```
BrendanAnnable, line 206: This component renders multiple balls:

```suggestion
const Balls = ({ model }: { model: LocalisationModel }) => (
```

nusight2/src/client/components/localisation/view.tsx
```diff
@@ -201,3 +202,19 @@ const FieldLineDots = ({ model }: { model: LocalisationModel }) => (
...
+  <>
+    {model.robots.map(
+      (robot) =>
+        robot.visible && (
+          <object3D key={robot.id}>
+            <mesh position={robot.ball.toArray()}>
+              <sphereBufferGeometry args={[robot.ball.z, 20, 20]} />
+              <meshStandardMaterial color="#1E90FF" />
```
BrendanAnnable, line 214: Any reason for this particular colour? Should it be white?

BrendanAnnable (commented):
Nice one! Some quick feedback but its looking good

nusight2/src/client/components/localisation/robot_model.ts
TrentHouliston: The localisation ball is in world space right?

BrendanAnnable: I'm just going off `network.ts` which has `robot.ball = Vector3.from(ball.rBWw).applyMatrix4(robot.Hfw);`

So isn't that `rBFf = Hfw * rBWw`?

TrentHouliston: Good point! This should be applied in real-time not in network!

nusight2/src/client/components/localisation/network.ts
TrentHouliston: The network layer shouldn't be converting to field space, that should happen constantly so that if robot.Hfw updates (localisation updates) then it's recalculated

nusight2/src/client/components/localisation/view.tsx
TrentHouliston: That would make it so the ball is always on the ground and just grows to do so?
I'm not sure that's what was intended but I think that's what it would do

nusight2/src/client/components/localisation/view.tsx
TrentHouliston: I thought the balls were orange

Tom0Brien: Just chose a random colour. White is probably a better option.

nusight2/src/client/components/localisation/network.ts
Tom0Brien: Where abouts would we move the transform to have it updated online with a localisation update?

nusight2/src/client/components/localisation/view.tsx
Tom0Brien: Hmm, white doesn't work well with lighting, orange looks good

nusight2/src/client/components/localisation/network.ts
BrendanAnnable: Correct, it moves into the model as a computed property, similar to how `Hft` is defined in `robot_model.ts`:

```ts
export class LocalisationRobotModel {

  // ...

  /** Ball position in field space */
  @computed
  get rBFf(): Vector3 {
    return this.ball.rBWw.applyMatrix4(robot.Hfw);
  }
}
```

Now you can use `robot.rBFf.toArray()` in the view.

nusight2/src/client/components/localisation/robot_model.ts
```diff
@@ -158,7 +162,8 @@ export class LocalisationRobotModel {
       Hfw: Matrix4.of(),
       Rwt: Quaternion.of(),
       motors: ServoMotorSet.of(),
-      fieldLinesDots: { rPWw: [] },
+      fieldLinePoints: { rPWw: [] },
+      ball: { rBWw: Vector3.of() },
```
BrendanAnnable, line 166: This means all robots will draw a ball at the center of the field, even if it hasn't seen a ball.

Lets make it optional on the model instead. i.e.

```ts
@observable ball?: { rBWw: Vector3 };
```

nusight2/src/client/components/localisation/robot_model.ts
```diff
@@ -175,4 +180,16 @@ export class LocalisationRobotModel {
...
+  @computed
+  get rPFf(): Vector3[] {
+    return this.fieldLinePoints.rPWw.map(rPWw => rPWw.applyMatrix4(this.Hfw));
+  }
+
+  /** Ball position in field space */
+  @computed
+  get rBFf(): Vector3 {
+    return this.ball.rBWw.applyMatrix4(this.Hfw);
+  }
```
BrendanAnnable, line 194: Which means this becomes

```suggestion
  get rBFf(): Vector3 | undefined {
    return this.ball?.rBWw.applyMatrix4(this.Hfw);
  }
```

nusight2/src/client/components/localisation/view.tsx
```diff
@@ -201,3 +202,17 @@ const FieldLineDots = ({ model }: { model: LocalisationModel }) => (
     )}
   </>
 );
+
+const Balls = ({ model }: { model: LocalisationModel }) => (
+  <>
+    {model.robots.map(
+      (robot) =>
+        robot.visible && (
```
BrendanAnnable, line 210: and this becomes

```suggestion
        robot.visible && robot.rBFf && (
```

nusight2/src/client/components/localisation/robot_model.ts
BrendanAnnable, line 166:
```suggestion
```

Tom0Brien: In `network.ts` where we update `onBall` I am getting a warning `robot.ball is possibly undefined`, do I need to do somewhere there as well now that ball is optional?

BrendanAnnable: Update the whole thing:

```ts
robot.ball = { rBWw: Vector3.from(ball.rBWw) };
```

BrendanAnnable (dismissed):
Nice one! LGTM — not sure whats going on with the build though.

Tom0Brien:
> Nice one! LGTM — not sure whats going on with the build though.

Not sure either, @TrentHouliston any ideas?

TrentHouliston:
We've been seeing it a bit, I'm not sure since node-gyp is definitely in that image

JosephusPaye:
For more context:

- It seems like something in the Node 18 image we use in CI has changed such that [yarn can no longer find a globally installed `node-gyp`](https://github.com/NUbots/NUbots/actions/runs/6347202110/job/17241866951?pr=1245#step:6:19) to build native dependencies.
- It [tries to install it on the fly but fails](https://github.com/NUbots/NUbots/actions/runs/6347202110/job/17241866951?pr=1245#step:6:21).

The temporary fix that worked for us is to add a workflow step before "Install dependencies" to install `node-gyp`. Something like:

```yaml
      # Install node-gyp to build native addons
      - name: Install node-gyp
        run: yarn global add node-gyp
```

Or as part of the "Install dependencies" step:

```yaml
      - name: Install Dependencies
        working-directory: nusight2
        run: yarn global add node-gyp && yarn --prefer-offline
```

BrendanAnnable (approved):

PR #1248, opened 2023-09-30
Install node-gyp dependency manually

Tom0Brien:
It seems like something in the Node 18 image we use in CI has changed such that [yarn can no longer find a globally installed node-gyp](https://github.com/NUbots/NUbots/actions/runs/6347202110/job/17241866951?pr=1245#step:6:19) to build native dependencies.

It [tries to install it on the fly but fails](https://github.com/NUbots/NUbots/actions/runs/6347202110/job/17241866951?pr=1245#step:6:21) since we have --prefer-offline (to use the cache) and it doesn't find node-gyp in the cache.

This PR adds the manual installation of `node-gyp` to the "Install dependencies" step to temporarily solve the issue. 

Thanks @JosephusPaye for fix.

### DIFF
--- a/.github/workflows/nusight.yaml
+++ b/.github/workflows/nusight.yaml
@@ -93,7 +93,7 @@ jobs:
       # Check dependencies from cache and install any which weren't cached or were corrupted
       - name: Install Dependencies
         working-directory: nusight2
-        run: yarn --prefer-offline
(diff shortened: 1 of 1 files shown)


TrentHouliston (approved):
I think longer term you should just use the same image that we use for the NUbots

JosephusPaye (approved):

PR #1249, opened 2023-10-04
Clean up analytical IK utility

Tom0Brien:
This PR cleans up the analytical IK utility, adds comments and fixes up conventions. Additionally, it makes use of the numerically stable `angle_between` function added in https://github.com/NUbots/NUbots/pull/1250

### DIFF
--- a/shared/utility/actuation/InverseKinematics.hpp
+++ b/shared/utility/actuation/InverseKinematics.hpp
@@ -97,9 +97,9 @@ namespace utility::actuation::kinematics {
         }
 
         // Knee pitch
-        const Scalar sqr_length    = std::pow(length, 2);
-        const Scalar sqr_upper_leg = std::pow(model.leg.UPPER_LEG_LENGTH, 2);
-        const Scalar sqr_lower_leg = std::pow(model.leg.LOWER_LEG_LENGTH, 2);
+        const Scalar sqr_length    = length * length;
+        const Scalar sqr_upper_leg = model.leg.UPPER_LEG_LENGTH * model.leg.UPPER_LEG_LENGTH;
+        const Scalar sqr_lower_leg = model.leg.LOWER_LEG_LENGTH * model.leg.LOWER_LEG_LENGTH;
         const Scalar cos_knee      = (sqr_upper_leg + sqr_lower_leg - sqr_length)
                                 / (2.0 * model.leg.UPPER_LEG_LENGTH * model.leg.LOWER_LEG_LENGTH);
         const Scalar knee_pitch = M_PI - acos_clamped(cos_knee);
--- a/shared/utility/actuation/InverseKinematics.hpp
+++ b/shared/utility/actuation/InverseKinematics.hpp
@@ -41,6 +41,8 @@ namespace utility::actuation::kinematics {
 
     using utility::input::LimbID;
     using utility::input::ServoID;
+    using utility::math::angle::acos_clamped;
+    using utility::math::angle::angle_between;
 
     /** @brief Calculates the leg joints for a given input ankle position. The robot coordinate system has origin a
      * distance DISTANCE_FROM_BODY_TO_HIP_JOINT above the midpoint of the hips. Robot coordinate system: x is out of the
@@ -56,12 +58,6 @@ namespace utility::actuation::kinematics {
         const message::actuation::KinematicsModel& model,
         const Eigen::Transform<Scalar, 3, Eigen::Isometry>& Htf_,
         const LimbID& limb) {
-        const Scalar LENGTH_BETWEEN_LEGS             = model.leg.LENGTH_BETWEEN_LEGS;
-        const Scalar DISTANCE_FROM_BODY_TO_HIP_JOINT = model.leg.HIP_OFFSET_Z;
-        const Scalar HIP_OFFSET_X                    = model.leg.HIP_OFFSET_X;
-        const Scalar UPPER_LEG_LENGTH                = model.leg.UPPER_LEG_LENGTH;
-        const Scalar LOWER_LEG_LENGTH                = model.leg.LOWER_LEG_LENGTH;
-
         std::vector<std::pair<ServoID, Scalar>> joints;
 
         // Correct for input referencing the bottom of the foot and convention used in IK implementation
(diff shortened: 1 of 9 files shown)


ysims (approved):

shared/utility/actuation/InverseKinematics.hpp
```diff
@@ -42,177 +42,143 @@ namespace utility::actuation::kinematics {
...
-                std::make_pair(ServoID::R_ANKLE_PITCH, (model.leg.LEFT_TO_RIGHT_ANKLE_PITCH) * -anklePitch));
-            positions.push_back(
-                std::make_pair(ServoID::R_ANKLE_ROLL, (model.leg.LEFT_TO_RIGHT_ANKLE_ROLL) * ankleRoll));
+            joints.push_back(std::make_pair(ServoID::R_HIP_YAW, (model.leg.LEFT_TO_RIGHT_HIP_YAW) * hip_yaw));
+            joints.push_back(std::make_pair(ServoID::R_HIP_ROLL, (model.leg.LEFT_TO_RIGHT_HIP_ROLL) * hip_roll));
+            joints.push_back(std::make_pair(ServoID::R_HIP_PITCH, (model.leg.LEFT_TO_RIGHT_HIP_PITCH) * hip_pitch));
+            joints.push_back(std::make_pair(ServoID::R_KNEE, (model.leg.LEFT_TO_RIGHT_KNEE) * knee_pitch));
+            joints.push_back(
+                std::make_pair(ServoID::R_ANKLE_PITCH, (model.leg.LEFT_TO_RIGHT_ANKLE_PITCH) * ankle_pitch));
+            joints.push_back(std::make_pair(ServoID::R_ANKLE_ROLL, (model.leg.LEFT_TO_RIGHT_ANKLE_ROLL) * ankle_roll));
```
TrentHouliston, line 171: The parenthesis they do nothing!

shared/utility/actuation/InverseKinematics.hpp
```diff
@@ -42,177 +42,143 @@ namespace utility::actuation::kinematics {
...
-            positions.push_back(std::make_pair(ServoID::L_HIP_PITCH, -hipPitch));
-            positions.push_back(std::make_pair(ServoID::L_KNEE, M_PI - knee));
-            positions.push_back(std::make_pair(ServoID::L_ANKLE_PITCH, -anklePitch));
-            positions.push_back(std::make_pair(ServoID::L_ANKLE_ROLL, ankleRoll));
+            joints.push_back(std::make_pair(ServoID::L_HIP_YAW, hip_yaw));
+            joints.push_back(std::make_pair(ServoID::L_HIP_ROLL, hip_roll));
+            joints.push_back(std::make_pair(ServoID::L_HIP_PITCH, hip_pitch));
+            joints.push_back(std::make_pair(ServoID::L_KNEE, knee_pitch));
+            joints.push_back(std::make_pair(ServoID::L_ANKLE_PITCH, ankle_pitch));
+            joints.push_back(std::make_pair(ServoID::L_ANKLE_ROLL, ankle_roll));
```
TrentHouliston, line 162:
```suggestion
            joints.emplace_back(ServoID::L_HIP_YAW, hip_yaw);
            joints.emplace_back(ServoID::L_HIP_ROLL, hip_roll);
            joints.emplace_back(ServoID::L_HIP_PITCH, hip_pitch);
            joints.emplace_back(ServoID::L_KNEE, knee_pitch);
            joints.emplace_back(ServoID::L_ANKLE_PITCH, ankle_pitch);
            joints.emplace_back(ServoID::L_ANKLE_ROLL, ankle_roll);
```

shared/utility/actuation/InverseKinematics.hpp
```diff
@@ -42,177 +42,143 @@ namespace utility::actuation::kinematics {
...
+        // Hip pitch
+        const Scalar phi4               = knee_pitch - lower_leg;
+        const Scalar sin_pi_minus_phi_2 = std::sin(M_PI - phi2);
+        const Eigen::Matrix<Scalar, 3, 1> unit_upper_leg =
+            unit_target_leg * (std::sin(phi2 - phi4) / sin_pi_minus_phi_2)
+            + ankle_y * (std::sin(phi4) / sin_pi_minus_phi_2);
+        const bool is_hip_pitch_positive = hip_x.dot(unit_upper_leg.cross(leg_plane_global_z)) >= 0;
+        const Scalar hip_pitch =
+            (is_hip_pitch_positive ? -1 : 1)
+            * std::acos(std::max(std::min(leg_plane_global_z.dot(unit_upper_leg), Scalar(1)), Scalar(-1)));
```
TrentHouliston, line 143: If you're refactoring, instead of using acos of a dot product it's much more numerically stable to use the "angle between" utility function

TrentHouliston: Wait you don't have that function!
You need this! https://github.com/NUbots/NUbots/pull/1250

shared/utility/actuation/InverseKinematics.hpp
```diff
@@ -42,177 +42,143 @@ namespace utility::actuation::kinematics {
...
+        const Scalar sqr_length    = length * length;
+        const Scalar sqr_upper_leg = UPPER_LEG_LENGTH * UPPER_LEG_LENGTH;
+        const Scalar sqr_lower_leg = LOWER_LEG_LENGTH * LOWER_LEG_LENGTH;
+        const Scalar cos_knee =
+            (sqr_upper_leg + sqr_lower_leg - sqr_length) / (2.0 * UPPER_LEG_LENGTH * LOWER_LEG_LENGTH);
+        const Scalar knee_pitch = M_PI - std::acos(std::max(std::min(cos_knee, Scalar(1.0)), Scalar(-1.0)));
+
+        // Ankle pitch
+        const Scalar cos_lower_leg = (sqr_lower_leg + sqr_length - sqr_upper_leg) / (2.0 * LOWER_LEG_LENGTH * length);
+        const Scalar lower_leg     = std::acos(std::max(std::min(cos_lower_leg, Scalar(1.0)), Scalar(-1.0)));
```
TrentHouliston, line 108: If you have to use this (because you can't use the dot product numerically stable version) there is a function acos_clamped in the angle utility that does all this for you

shared/utility/actuation/InverseKinematics.hpp
```diff
@@ -42,177 +42,143 @@ namespace utility::actuation::kinematics {
...
+     * @param Htf_ Desired pose of foot {f} in torso {t} space
+     * @param limb Which leg to solve IK for (left or right)
+     * @return Vector of pairs of servo ID and joint angles
+     */
     template <typename Scalar>
-    [[nodiscard]] std::vector<std::pair<ServoID, Scalar>> calculateLegJoints(
+    [[nodiscard]] std::vector<std::pair<ServoID, Scalar>> calculate_leg_joints(
         const message::actuation::KinematicsModel& model,
-        const Eigen::Transform<Scalar, 3, Eigen::Isometry>& target_,
+        const Eigen::Transform<Scalar, 3, Eigen::Isometry>& Htf_,
```
TrentHouliston, line 57: If you're committed to using doubles everywhere (which is probably a good idea) you could replace all this math with a non template double version. That'd get rid of a lot of the dependent templates and let you use Eigen::Isometry3d and Eigen::Vector3d to make the later code easier to read

shared/utility/actuation/InverseKinematics.hpp
```diff
@@ -42,177 +42,143 @@ namespace utility::actuation::kinematics {
...
-
-        // If leg is above waist then hipX is pointing in the wrong direction in the xy plane
-        Eigen::Matrix<Scalar, 3, 1> hipXProjected = (isAnkleAboveWaist ? -1 : 1) * hipX;
-        hipXProjected.z()                         = 0;
-        hipXProjected.normalize();
-        const bool isHipYawPositive = hipXProjected.dot(Eigen::Matrix<Scalar, 3, 1>::UnitY()) >= 0;
-        hipYaw                      = (isHipYawPositive ? 1 : -1)
-                 * std::acos(std::max(std::min(hipXProjected.dot(Eigen::Matrix<Scalar, 3, 1>::UnitX()), Scalar(1)),
-                                      Scalar(-1)));
+        const Scalar cos_hip_roll = leg_plane_global_z.dot(Eigen::Matrix<Scalar, 3, 1>::UnitZ());
```
TrentHouliston, line 130: Dotting a vector with unit z is exactly the same as just accessing the z component
```suggestion
        const Scalar cos_hip_roll = leg_plane_global_z.z();
```
Although maybe this should be angle between too

shared/utility/actuation/InverseKinematics.hpp
```diff
@@ -42,177 +42,143 @@ namespace utility::actuation::kinematics {
...
+        const Scalar hip_pitch =
+            (is_hip_pitch_positive ? -1 : 1)
+            * std::acos(std::max(std::min(leg_plane_global_z.dot(unit_upper_leg), Scalar(1)), Scalar(-1)));
+
+        // Hip yaw
+        // If leg is above waist then hip_x is pointing in the wrong direction in the xy plane
+        Eigen::Matrix<Scalar, 3, 1> hip_x_projected = (is_ankle_above_waist ? -1 : 1) * hip_x;
+        hip_x_projected.z()                         = 0;
+        hip_x_projected.normalize();
+        const bool is_hip_yaw_positive = hip_x_projected.dot(Eigen::Matrix<Scalar, 3, 1>::UnitY()) >= 0;
```
TrentHouliston, line 150: Dotting with unit Y is just the same as accessing the y component
```suggestion
        const bool is_hip_yaw_positive = hip_x_projected.y() >= 0;
```

shared/utility/actuation/InverseKinematics.hpp
Tom0Brien: I would prefer utilities to be templated for generality. I guess in this case the implementation is fairly restricted to a single use case but I like the convention of utilities always being templated

shared/utility/actuation/InverseKinematics.hpp
```diff
@@ -86,46 +82,45 @@ namespace utility::actuation::kinematics {
...
         // Knee pitch
-        const Scalar sqr_length    = length * length;
-        const Scalar sqr_upper_leg = UPPER_LEG_LENGTH * UPPER_LEG_LENGTH;
-        const Scalar sqr_lower_leg = LOWER_LEG_LENGTH * LOWER_LEG_LENGTH;
-        const Scalar cos_knee =
-            (sqr_upper_leg + sqr_lower_leg - sqr_length) / (2.0 * UPPER_LEG_LENGTH * LOWER_LEG_LENGTH);
-        const Scalar knee_pitch = M_PI - std::acos(std::max(std::min(cos_knee, Scalar(1.0)), Scalar(-1.0)));
+        const Scalar sqr_length    = std::pow(length, 2);
+        const Scalar sqr_upper_leg = std::pow(model.leg.UPPER_LEG_LENGTH, 2);
+        const Scalar sqr_lower_leg = std::pow(model.leg.LOWER_LEG_LENGTH, 2);
```
TrentHouliston, line 102: `std::pow(x, 2)` is less performant than x*x since one uses a power function while the other is simply multiplication. Sometimes it doesn't matter but generally I'd use multiplication to square over powers

ysims (approved):

PR #1252, opened 2023-10-07
Fix the script skipping issue by not emitting Done in the Servo Providers when already Done 

ysims:
Scripts were acting quite strange, such as skipping frames, not going at the right speed.

After having a look, it seems that the Sequence Provider is running the emit Task Group / update Count and emit multiple times for the same group/count. This will restart that script frame each time it does this - if it spams for the whole length of the frame, then the servo won't ever get a chance to move.

It does this because the Group is emitting Done multiple times.

The Group Provider emits Done multiple times because the Servos emit Done multiple times. This is because the Servos run on Sensors (90 times a second) and once the time is over for the servo it keeps emitting Done... 90 times a second. Causing too many triggers of the Group Provider on the Servo's Done.

The solution in this PR is to add information so that the Servo Providers know that they are already in a Done state and don't need to re-emit Done.

I've tested in Webots and it looks good. Real robot looks good too.

### Things left to do

- [x] Test real robot

### DIFF
--- a/module/actuation/Servos/src/Servos.hpp
+++ b/module/actuation/Servos/src/Servos.hpp
@@ -105,7 +105,7 @@ namespace module::actuation {
             // Make an initial count message
             emit<Scope::DIRECT>(std::make_unique<Count<Sequence>>(0));
 
-            on<Provide<Sequence>, Needs<Group>, With<Count<Sequence>>>().then(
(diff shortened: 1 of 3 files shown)


ysims:
Replaced by #1256.

PR #1255, opened 2023-10-10
Add URDF model to nusight

Tom0Brien:
This PR replaces the current model of the NUgus robot in the localisation view with a model generated from our URDF file of the NUgus using the urdf-loader library https://www.npmjs.com/package/urdf-loader.

image

Addionally, this PR consolidates the URDF model into the same folder within `shared` and symlinks it to the `assets` folder within `nusight`. Now only `.urdf` files are copied over to the build folder in `libraries.cmake` as files such as `.stl` are not required for the main code.


module/input/Camera/data/config/nugus4/Cameras/Left.yaml
```diff
@@ -24,7 +24,7 @@ lens:
   fov: 180 * pi / 180
 
 # Camera kinematics
-urdf_path: "/home/nubots/models/nugus.urdf"
+urdf_path: "/home/nubots/models/robot.urdf"
```
TrentHouliston, line 27: Definitely don't want to give absolute paths like this

ysims (approved):

PR #1256, opened 2023-10-10
Stops subtasks spamming Done to their parent (fix issue with scripts not running properly)

ysims:
Previously if a subtask was Done, the parent could keep getting Done emits from the child which runs the parent Provider. We don't want to spam the parent. This was particularly a problem for the Servo Providers and caused issues with Sequences (and hence Scripts).


TrentHouliston (approved):
The fact that you have the ROOT check scares me!
But LGTM

PR #1257, opened 2023-10-19
Add skill to allow robot to speak

Tom0Brien:
This PR adds a skill which allows the robot to speak. The text-to-speech library used is a python based command-line tool https://github.com/MycroftAI/mimic3 which is not ideal as `system` call is needed, the quality of this library is really good though and runs locally.


ysims:
I feel like audio testing shouldn't be in the behaviour testing module. The testing could be simplified with a config value in the Say module containing a test sentence and then ignoring it if it's empty otherwise emit the task with the string.

module/skill/Say/src/Say.cpp
```diff
@@ -0,0 +1,42 @@
...
+        on<Provide<SayTask>>().then([this](const SayTask& say, const RunInfo& info) {
+            // Only say text if it is a new task
+            if (info.run_reason == RunInfo::NEW_TASK) {
+                // Play the requested audio using python command-line tool mimic3 and aplay
+                system(std::string("mimic3 '" + say.text + "' | aplay").c_str());
+            }
+
+            // Nod head to indicate that the robot is "talking"
+            if (say.nod) {
+                emit<Task>(load_script<LimbsSequence>("NodYes.yaml"));
```
ysims, line 37: Does this work?? `LimbsSequence` should only do something with servo commands for the arms and legs. This should be `HeadSequence`.

Tom0Brien:
> I feel like audio testing shouldn't be in the behaviour testing module. The testing could be simplified with a config value in the Say module containing a test sentence and then ignoring it if it's empty otherwise emit the task with the string.

Hmmm, I like the tester module having all the available tasks for testing running combinations of providers simultaneously and having it within the module itself would require a role which is setup to use Say which currently doesn't exist.

ysims:
> > I feel like audio testing shouldn't be in the behaviour testing module. The testing could be simplified with a config value in the Say module containing a test sentence and then ignoring it if it's empty otherwise emit the task with the string.
> 
> Hmmm, I like the tester module having all the available tasks for testing running combinations of providers simultaneously and having it within the module itself would require a role which is setup to use Say which currently doesn't exist.

I don't really see the point though, there's not really a use case where it's necessary and speaking doesn't conflict with other modules. I like the idea of another role because then you don't have to compile so much just to run speech.

module/skill/Say/src/Say.cpp
```diff
@@ -0,0 +1,61 @@
...
+                                         sanitized_text.end());
+                }
+                log<NUClear::DEBUG>("Saying: ", sanitized_text);
+                system(std::string("mimic3 '" + sanitized_text + "' --voice '" + cfg.voice + "' | aplay -D"
+                                   + cfg.device_name)
+                           .c_str());
+
+                // Nod head to indicate that the robot is "talking"
+                if (say.nod) {
+                    emit<Task>(load_script<HeadSequence>("Say.yaml"));
```
ysims, line 55: shouldn't this just be the NodYes script?

module/skill/Say/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Include this module whenever the robot needs to vocalize or acknowledge messages via speech-like actions.
+
+## Consumes
+
+- `message::skill::Say`: A Task requesting to vocalize a text string. Contains the text to be spoken and an optional nod gesture request.
+
+## Emits
+
+- `message::actuation::LimbsSequence`: A Task to instruct the robot's limbs to perform specific sequences, such as a nodding gesture.
```
ysims, line 17:
```suggestion
- `message::actuation::HeadSequence`: A Task to instruct the robot's head to perform specific sequences, such as a nodding gesture.
```

module/skill/Say/src/Say.cpp
Tom0Brien: This new script is essentially NodYes but with multiple nods

ysims (dismissed):

ysims:
You need to merge main and run format again

ysims (approved):

PR #1258, opened 2023-10-19
Add ability to make requests to OpenAI api

Tom0Brien:
This PR adds a module that integrates with OpenAI's api to send text prompts and receive generated responses. Additionally, the responses can be "spoken" using the "Say" skill. I am not sure whether or not it makes sense to integrate this within the Director system or not.

### DIFF
--- a/module/skill/GPT/README.md
+++ b/module/skill/GPT/README.md
@@ -1,0 +1,22 @@
+# GPT
+
+## Description
+
+A module that integrates with OpenAI's GPT-3.5 model to send text prompts and receive generated responses.
+
+## Usage
+
+Include this module to allow your system to interact with OpenAI's GPT for natural language processing tasks.
+
+## Consumes
+
+- `message::skill::GPTRequest`: A request containing text that needs to be processed by GPT.
+
+## Emits
+
+- `message::skill::GPTResponse`: A response from GPT containing the generated text based on the input request.
+
+## Dependencies
+
+- **nlohmann::json**: For JSON data serialization and deserialization.
+- **utility::openai::openai**: A utility to interface with the OpenAI API.
--- a/module/skill/GPT/README.md
+++ b/module/skill/GPT/README.md
@@ -14,7 +14,7 @@ Include this module to allow your system to interact with OpenAI's GPT for natur
 
 ## Emits
 
-- `message::skill::GPTResponse`: A response from GPT containing the generated text based on the input request.
+ `message::skill::Say`: A Task requesting to speak using the output of the GPT model based on the input request.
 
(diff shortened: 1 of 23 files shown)


module/skill/GPT/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Include this module to allow your system to interact with OpenAI's GPT for natural language processing tasks.
+
+## Consumes
+
+- `message::skill::GPTRequest`: A request containing text that needs to be processed by GPT.
+
+## Emits
+
+- `message::skill::GPTResponse`: A response from GPT containing the generated text based on the input request.
```
ysims, line 17:
```suggestion
- `message::skill::Say`: A Task requesting to speak using the output of the GPT model based on the input request.
```

ysims (commented):
Same comment about the tester here from the speech PR
The GPT module can have a test config value to use for simpler testing. 
A role with just Say and GPT wuld be useful. then leave `behaviour.role` for soccer behaviour testing.

module/skill/GPT/data/config/GPT.yaml
```diff
@@ -0,0 +1,5 @@
+# Controls the minimum log level that NUClear log will display
+log_level: INFO
+
+# The OpenAI API costs $$$ to use, so we need to use a key to access it. DO NOT COMMIT THIS
+openai_api_key: ""
```
jpptm, line 5:
```suggestion
# The OpenAI API costs $$$ to use, so we need to use a key to access it. This must be filled before compiling. DO NOT COMMIT THE KEY IN THE REPOSITORY.
openai_api_key: ""
```

module/skill/GPT/src/GPT.cpp
```diff
@@ -0,0 +1,49 @@
...
+        });
+
+        on<Provide<GPTRequest>>().then([this](const GPTRequest& gpt_request, const RunInfo& info) {
+            if (info.run_reason == RunInfo::NEW_TASK) {
+                // Send request to OpenAI API
+                nlohmann::json request = {
+                    {"model", "gpt-3.5-turbo"},
+                    {"messages", nlohmann::json::array({{{"role", "user"}, {"content", gpt_request.text}}})},
+                    {"max_tokens", 100},
+                    {"temperature", 0}};
```
jpptm, line 36: Should these be configurable values?

jpptm (changes requested):
Nitpicking

module/skill/GPT/src/GPT.cpp
Tom0Brien: Yes, good point.

PR #1259, opened 2023-10-19
Add ability to chat with ChatGPT using mic (speech-to-text)

Tom0Brien:
This PR adds a speech-to-text using Openai's whisper model https://openai.com/research/whisper. 

Audio is recorded for a specified duration, which is then transcribed into text using Whisper. This text can be forwarded to ChatGPT for a response. Additionally, ChatGPT's responses can be audibly relayed back to the user through the 'Say' task.

### DIFF
--- a/roles/webots/behaviour.role
+++ b/roles/webots/behaviour.role
@@ -35,6 +35,7 @@ nuclear_role(
   skill::GetUp
   skill::Look
   skill::Dive
+  skill::Say
   # Planners
   planning::PlanWalkPath
   planning::PlanLook
--- a/roles/webots/behaviour.role
+++ b/roles/webots/behaviour.role
@@ -36,6 +36,7 @@ nuclear_role(
   skill::Look
   skill::Dive
   skill::Say
+  skill::GPT
   # Planners
   planning::PlanWalkPath
   planning::PlanLook
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -313,6 +313,8 @@ RUN install-from-source https://github.com/stevengj/nlopt/archive/refs/tags/v2.7
 RUN install-from-source https://github.com/Tom0Brien/tinyrobotics/archive/refs/tags/0.0.1.tar.gz \
     -DBUILD_TESTS=OFF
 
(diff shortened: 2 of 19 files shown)


roles/webots/behaviour.role
```diff
@@ -35,6 +35,8 @@ nuclear_role(
   skill::GetUp
   skill::Look
   skill::Dive
+  skill::Say
+  skill::GPT
```
JohanneMontano, line 39: Why do we want this functionality in webots? Does it provide audio interfaces as well?

JohanneMontano (commented):
1 comment, otherwise it seems that most of the code is copied and pasted somewhere and since it works and is well documented,LGTM

Also, just curious if some files are copied from somewhere?

Tom0Brien:
> 1 comment, otherwise it seems that most of the code is copied and pasted somewhere and since it works and is well documented,LGTM
> 
> Also, just curious if some files are copied from somewhere?

All the functionality to interact with Openai's API is dealt with using implementation provided by https://github.com/olrea/openai-cpp

roles/webots/behaviour.role
Tom0Brien: Audio interface would just be device on your laptop when running in Docker, I was just using behaviour.role as a quick and easy way to test things

jpptm (dismissed):
LGTM

jpptm:
It seems like the improvement can be put in another PR after this PR is merged

roles/test/behaviour.role
```diff
@@ -36,6 +36,9 @@ nuclear_role(
   skill::SplineKick
   skill::GetUp
   skill::Look
+  skill::Dive
```
ysims, line 39: dive?

roles/webots/behaviour.role
```diff
@@ -37,7 +37,9 @@ nuclear_role(
   skill::SplineKick
   skill::GetUp
   skill::Look
+  skill::Dive
```
ysims, line 40: dive?

ysims (approved):

PR #1260, opened 2023-10-23
Add delay and starting side options for field localisation

Tom0Brien:
This PR adds a config value for the following:

- Delaying the start of the filter (useful for letting odometry converge before beginning)
- Selecting a starting side (left, right or either)

### DIFF
--- a/module/localisation/FieldLocalisation/data/config/webots/FieldLocalisation.yaml
+++ b/module/localisation/FieldLocalisation/data/config/webots/FieldLocalisation.yaml
@@ -31,5 +31,5 @@ outside_map_penalty_factor: 0.9
 # Delay before the particle filter starts (allows odometry to settle)
 start_time_delay: 5
 
-# Starting side of the field (LEFT, RIGHT or EITHER)
+# Starting side of the field looking towards own goal from centre of field (LEFT, RIGHT or EITHER)
 starting_side: EITHER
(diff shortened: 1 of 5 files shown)


module/localisation/FieldLocalisation/data/config/webots/FieldLocalisation.yaml
```diff
@@ -27,3 +27,9 @@ min_observations: 100
 
 # Penalty factor for observations being outside map
 outside_map_penalty_factor: 0.9
+
+# Delay before the particle filter starts (allows odometry to settle)
+start_time_delay: 5
+
+# Starting side of the field (LEFT, RIGHT or EITHER)
```
ysims, line 34: Is this left/right when looking towards the opponent goal?

Tom0Brien: Looking towards our own goal, as the current convention is our goal is in the positive x direction in field space.

ysims: That will definitely need documenting here and where ever relevant
Old convention was that towards opponents goal is forward (positive)
I checked NUbook and we've got the old convention documented on the maths page.

ysims (approved):

PR #1262, opened 2023-10-24
Add simple method for goal detection

Tom0Brien:
This PR adds a new method for detecting the bottom of the goal posts. It simply grabs the closest point in a cluster and assumes this is the bottom of the goal post. Addionally, the float types of the goal detector are converted to double.

### DIFF
--- a/module/vision/GoalDetector/src/GoalDetector.cpp
+++ b/module/vision/GoalDetector/src/GoalDetector.cpp
@@ -52,26 +52,26 @@ namespace module::vision {
     GoalDetector::GoalDetector(std::unique_ptr<NUClear::Environment> environment) : Reactor(std::move(environment)) {
 
         // Trigger the same function when either update
-        on<Configuration>("GoalDetector.yaml").then([this](const Configuration& cfg) {
-            log_level = cfg["log_level"].as<NUClear::LogLevel>();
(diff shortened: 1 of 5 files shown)


module/vision/GoalDetector/src/GoalDetector.cpp
```diff
@@ -293,12 +293,25 @@ namespace module::vision {
...
+                            std::min_element(cluster.begin(), cluster.end(), [&](const int& a, const int& b) {
+                                return rays.col(a).z() < rays.col(b).z();
+                            });
+                        Eigen::Vector3d closest_uPCw = Eigen::Vector3d(rays.col(*closest_uPCw_idx).cast<double>());
+                        // Project goal point onto the field plane
+                        auto Hwc = horizon.Hcw.inverse();
+                        Eigen::Vector3d rGWw =
+                            closest_uPCw * std::abs(Hwc.translation().z() / closest_uPCw.z()) + Hwc.translation();
+                        goals->rGWw.push_back(rGWw);
+                        log<NUClear::DEBUG>("Goal at: ", rGWw.transpose());
```
ysims, line 308: How come this is all separate to the old method? Shouldn't this method be replacing the goal measurement in the individual Goal message?

PR #1263, opened 2023-10-24
Update GoalDetector method

Tom0Brien:
This PR  does the following:

- Simplifies `GoalDetector` to detect goals by just requiring a large cluster count and selecting bottom point as the closest point in the cluster
- Adds visualisation in NUsight localisation view

I have removed the maximum distance throwout as we tend to only use a "small" mesh size anyway.

image


module/vision/GoalDetector/src/GoalDetector.hpp
```diff
@@ -28,13 +28,13 @@ namespace module::vision {
...
+            double confidence_threshold                = 0.0f;
             int cluster_points                         = 0;
-            float disagreement_ratio                   = 0.0f;
-            Eigen::Vector3f goal_projection_covariance = Eigen::Vector3f::Zero();
+            double disagreement_ratio                  = 0.0f;
+            Eigen::Vector3d goal_projection_covariance = Eigen::Vector3d::Zero();
             bool use_median                            = false;
-            float max_goal_distance                    = 0;
-        } config{};
+            double max_goal_distance                   = 0;
```
JohanneMontano, line 36: Just curious - why do we need the extra floating point accuracy here?

Tom0Brien: It's convention as `double` is best to use unless performance becomes an issue

ysims (dismissed):
Looks fine. Checking if it intersects the hull would be good but I can have a look at adding that back in and testing later.

ysims (dismissed):

ysims (approved):

PR #1264, opened 2023-11-11
Update NUClear to latest version

Tom0Brien:
This PR updates NUClear to latest commit and makes required changes to get everything compiling.

TODO:
- [x] Fix logs always displaying
- [x] Fix NatNet
- [x] Fix NetworkForwarder
- [ ] Test NatNet still works...


module/input/NatNet/src/NatNet.cpp
```diff
@@ -64,22 +64,23 @@ namespace module::input {
                         .then("NatNet Data", [this](const UDP::Packet& packet) {
                             // Test if we are "connected" to this remote
                             // And if we are we can use the data
-                            if (remote == packet.remote.address && version != 0) {
+                            auto address = static_cast<uint32_t>(std::stoul(packet.remote.address));
```
TrentHouliston, line 67: This isn't valid, the string is an IP address (something like `10.1.1.10` or maybe an IPv6 address) and using stoul will just extract the first number component and use it which is not correct here since it'll be just assuming that any IP address that starts with the same number is the same remote.
Instead remote should be a string and just store it in there.

module/input/NatNet/src/NatNet.cpp
```diff
@@ -441,14 +442,15 @@ namespace module::input {
...
+    void NatNet::process(const std::vector<uint8_t>& input) {
         // Get our packet
         const Packet& packet = *reinterpret_cast<const Packet*>(input.data());
 
         if (cfg.dump_packets) {
             static int i = 0;
+            std::vector<char> input_char(input.begin(), input.end());
             std::ofstream output(fmt::format("natnet_{:06d}.bin", i++));
-            output.write(input.data(), input.size());
+            output.write(input_char.data(), input_char.size());
```
TrentHouliston, line 453: No need for the copy, just write out the buffer but you'll need a cast

module/platform/Webots/src/Webots.cpp
```diff
@@ -525,7 +525,7 @@ namespace module::platform {
                                 for (uint32_t length = read_length(buffer); buffer.size() >= length + sizeof(length);
                                      length          = read_length(buffer)) {
                                     // Decode the protocol buffer and emit it as a message
-                                    char* payload = reinterpret_cast<char*>(buffer.data()) + sizeof(length);
+                                    uint8_t* payload = reinterpret_cast<uint8_t*>(buffer.data()) + sizeof(length);
```
TrentHouliston, line 528: If it doesn't cause too many problems elsewhere, you can also make the buffer a uint8_t to avoid the casting

module/support/logging/ConsoleLogHandler/src/ConsoleLogHandler.cpp
```diff
@@ -98,19 +105,20 @@ namespace module::support::logging {
...
 
                 // This is our source
                 source = reactor + " "
-                         + (message.task->identifier[0].empty() ? "" : "- " + message.task->identifier[0] + " ");
+                         + (message.task->identifiers.name.empty() ? "" : "- " + message.task->identifiers.name + " ");
             }
 
             // Output the level
             switch (message.level) {
+                case NUClear::UNKNOWN: std::cerr << source << ": "; break;
```
TrentHouliston, line 121: I'd fold this in with fatal as if you get unknown something weird/wrong has happened

Bidski (dismissed):

module/input/NatNet/src/NatNet.cpp
```diff
@@ -104,9 +106,14 @@ namespace module::input {
...
-            address.sin_port        = htons(cfg.command_port);
-            address.sin_addr.s_addr = htonl(remote);
+            address.sin_family = AF_INET;
+            address.sin_port   = htons(cfg.command_port);
+
+            // Convert IP address from string to binary form
+            if (inet_pton(AF_INET, remote.c_str(), &address.sin_addr) <= 0) {
+                log<NUClear::ERROR>("Invalid remote IP address");
+                return;
+            }
```
TrentHouliston, line 116: Rather than all this and the sendto, you can just emit<UDP> instead then you can just pass the address in without having to do the conversion yourself

TrentHouliston (changes requested):

module/input/NatNet/src/NatNet.cpp
TrentHouliston: Or leave it, when I first wrote this I had dreams of making this natnet stuff a separate library so others could use it too

PR #1280, opened 2023-11-28
Add a module to convert NUClear Logs and ReactionStatistics to messages for recording

JosephusPaye:
This PR adds a new module `support::logging::MessageLogHandler`  that listens for `LogMessage` and `ReactionStatistics` messages from NUClear and recodes them into protobuf messages. This make it possible to record those messages to NBS files for later playback and analysis.

I'm hoping to get this in to use it for my NUsight workshop on Friday.

### PR Checklist:

### DIFF
--- a/module/support/logging/MessageLogHandler/data/config/MessageLogHandler.yaml
+++ b/module/support/logging/MessageLogHandler/data/config/MessageLogHandler.yaml
@@ -1,0 +1,8 @@
+# Controls the minimum log level that NUClear log will display from this module
+log_level: INFO
+
+log_settings:
+  # Sets the minimum log level that this module will handle for logs coming in
+  min_level: TRACE
+  # If true, will ignore the display level of the log and recode everything
+  log_all: true
--- a/module/support/logging/MessageLogHandler/src/MessageLogHandler.cpp
+++ b/module/support/logging/MessageLogHandler/src/MessageLogHandler.cpp
@@ -1,0 +1,120 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2023 NUbots
+ *
+ * This file is part of the NUbots codebase.
+ * See https://github.com/NUbots/NUbots for further info.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+#include "MessageLogHandler.hpp"
+
+#include "extension/Configuration.hpp"
+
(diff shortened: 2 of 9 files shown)


module/support/logging/MessageLogHandler/src/MessageLogHandler.cpp
```diff
@@ -0,0 +1,91 @@
...
+            msg->timestamp     = NUClear::clock::now();
+            msg->level         = to_message(log_msg.level);
+            msg->display_level = to_message(log_msg.display_level);
+            if (log_msg.task) {
+                msg->reaction_statistics = to_message(*log_msg.task);
+            }
+            emit(msg);
+        });
+
+        on<Trigger<NUClear::message::ReactionStatistics>>().then(
```
Tom0Brien, line 84: Why do we need to pass the message into the `to_message` function here?

Do the NUClear library messages not get included in the NUsight protobut stuff?

module/support/logging/MessageLogHandler/src/MessageLogHandler.cpp
```diff
@@ -0,0 +1,91 @@
...
+namespace module::support::logging {
+
+    using extension::Configuration;
+
+    namespace {
+        message::nuclear::ReactionStatistics to_message(const NUClear::message::ReactionStatistics& in) {
+
+            message::nuclear::ReactionStatistics out;
+
+            out.identifiers.name     = in.identifiers.name;
```
Tom0Brien, line 17: This looks like it will require latest NUClear, which I have a PR for here: https://github.com/NUbots/NUbots/pull/1264

module/support/logging/MessageLogHandler/src/MessageLogHandler.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+    class MessageLogHandler : public NUClear::Reactor {
+    private:
+        /// The configuration variables for this reactor
+        struct {
+            /// The minimum log level to forward
+            NUClear::LogLevel min_level{NUClear::TRACE};
+            /// If the display level should be ignored and all messages logged
+            bool log_all{true};
+        } config{};
```
Tom0Brien, line 16: Our current convention is to use `cfg` rather then `config`

module/support/logging/MessageLogHandler/src/MessageLogHandler.cpp
JosephusPaye: I think the current setup for NUbots and NUsight doesn't work with messages outside the `shared/message` directory.

@TrentHouliston or @Bidski can say definitively.

module/support/logging/MessageLogHandler/data/config/MessageLogHandler.yaml
```diff
@@ -0,0 +1,8 @@
+# Controls the minimum log level that NUClear log will display from this module
+log_level: DEBUG
```
Tom0Brien, line 2:
```suggestion
log_level: INFO
```

module/support/logging/MessageLogHandler/src/MessageLogHandler.cpp
JosephusPaye: Good catch. What's left for that to get merged?

Tom0Brien (changes requested):

module/support/logging/MessageLogHandler/src/MessageLogHandler.cpp
TrentHouliston: The protobuf stuff is separate from the c++ NUClear stuff.
I believe the enums are probably the same value and you could just directly cast it rather than going through this process.
The advantage of this process is any mismatch is a compile error, not a future being confused why things are different error.

module/support/logging/MessageLogHandler/src/MessageLogHandler.cpp
Tom0Brien: Just needs a review

module/support/logging/MessageLogHandler/src/MessageLogHandler.cpp
```diff
@@ -0,0 +1,91 @@
...
+                default: return message::nuclear::LogLevel::UNKNOWN;
+            }
+        }
+
+    }  // namespace
+
+    MessageLogHandler::MessageLogHandler(std::unique_ptr<NUClear::Environment> environment)
+        : Reactor(std::move(environment)), cfg{} {
+
+        on<Configuration>("MessageLogHandler.yaml").then([this](const Configuration& cfg) {
```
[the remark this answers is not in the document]
JosephusPaye: Oh, so the class member is `cfg` but the param here is `config`.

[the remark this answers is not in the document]
Tom0Brien: Yeah, rationale is that `cfg` is used throughout the code and shorter variable name is preferable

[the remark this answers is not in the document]
JosephusPaye: Fixed!

Tom0Brien (dismissed):
LGTM!

ysims (approved):

PR #1281, opened 2023-11-28
Fix the green horizon detector such that the convex hull and clustering are done correctly

ysims:
The old method had issues

- Clustering after getting the boundary caused issues, not all the points would make it into the cluster.
- The convex hull algorithm did not always work. A lot of `O(n)` convex hull algorithms end up being incorrect in some edge cases, not sure if that's what was going on.

This PR

- Breaks up the steps of getting field points and getting boundary points into two steps rather than one function call. Adds in a function to the vision utilities to just get the boundary points.
- Changes the order of steps so we (1) get all field points, (2) cluster and get the closest, then (3) get the boundary points, and finally (4) calculate the convex hull. The main point of this is clustering then boundary to make sure clustering works.
- Switches from the existing convex hull algorithm to using Chan's convex hull algorithm, a reliable `O(nlog(h))` algorithm. 
- Adds in automated tests for calculating the convex hull
- `double` is used in the new convex hull code, and so I went and changed from `float` to `double` in the GreenHorizonDetector. This ended up spreading to the other utilities and then to other detectors.
- Some other changes to make the code more readable

## Checklist

- [x] Updated NUbook if necessary (add link to NUbook PR here) *NUbook currently is general enough that it is still correct with this PR.

Tested in Webots and the real robot. Only thing on the real robot is that odometry needs to settle before it works smoothly.


module/vision/BallDetector/src/BallDetector.cpp
```diff
@@ -175,10 +175,10 @@ namespace module::vision {
                     // largest angular radius possible from the edge points available. Equal to cos(theta), where theta
                     // is the angle between the central ball axis (uBCw) and the edge of the ball. This helps to find
                     // the approximate distance to the ball.
-                    float radius = 1.0f;
+                    float radius = 1.0;
```
Tom0Brien, line 178:
```suggestion
                    double radius = 1.0;
```

shared/utility/math/geometry/ConvexHull.hpp
```diff
@@ -24,346 +24,405 @@
...
  * SOFTWARE.
  */
-
 #ifndef UTILITY_MATH_GEOMETRY_CONVEXHULL_HPP
 #define UTILITY_MATH_GEOMETRY_CONVEXHULL_HPP
 
 #include <Eigen/Core>
+#include <Eigen/Dense>
+#include <algorithm>
+#include <iostream>
```
Tom0Brien, line 33: Is `<iostream>` needed?

module/vision/GreenHorizonDetector/src/GreenHorizonDetector.cpp
```diff
@@ -36,173 +36,172 @@
...
+                msg->vision_ground_truth = mesh.vision_ground_truth;
+            }
+
+            // Find the convex hull of the cluster
+            msg->horizon.reserve(hull_indices.size());
+            for (const auto& idx : hull_indices) {
+                const Eigen::Vector3d ray      = uPCw.col(idx);
+                const double d                 = mesh.Hcw(2, 3) / ray.z();
+                Eigen::Vector3d ray_projection = ray * d;
+                msg->horizon.emplace_back(ray_projection.normalized());
```
Tom0Brien, line 199: Since you are normalizing here, wouldn't `uPCw.col(idx)` be the same here.

Tom0Brien: It looks like before this was done to add the offset before normalizing

Tom0Brien (changes requested):
Quick review

Tom0Brien (approved):
One small thing is to change Spherical Reciprocal Coordinates to Cartesian in the detectors, as we don't exploit this anywhere and just requires casting between the two representations. I think it can be left to another PR though.

LGTM

PR #1284, opened 2023-11-29
Update format.py to try to better track renames

TrentHouliston:
For formatting in the licence git was not doing the best job of tracking the dates of changes as it was tracking through what it considered copies not just through actual renames.

The result was lots of files were considered to be older than they were which made PRs sometimes fail as when they were squashed into the merge queue as some of the files would go from being copies to not etc.

That may still happen but it should happen less now?

### Reviewers, Note

I don't like how this works, and it makes things much slower.
Please find a better option thanks

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -346,6 +346,7 @@ RUN install-from-source https://github.com/Tom0Brien/tinyrobotics/archive/refs/t
 
 # Install python libraries
 RUN pip install \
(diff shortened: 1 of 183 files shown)


ysims (approved):
getting more issues with dates in PRs failing in merge queue so would like to get this in if it helps. might need to turn off until it's definitely fixed if it keeps happening tho...

PR #1285, opened 2023-12-02
Adds a skeleton NUSense as a module with arduino based firmware

TrentHouliston:
Add code to compile a NUSense compatible binary using the Ardiuno framework.
It adds the necessary.

This PR is just here as a reference of what is possible. There is no expectation that you use this for NUSense itself.

I'm not the biggest fan of the cmake code I stole to integrate the arduino code into a cmake project
If I get bored I might rewrite it better.


Bidski:
> @Bidski does not like the idea at all

The idea doesn't even like this idea lol

PR #1287, opened 2023-12-12
Reduce default gains in various scripts and ScriptTuner

Tom0Brien:
This PR reduces arm gains in walk engine as intermediate step to reduce the erratic movement when starting walk engine, however, we should figure out a better solution for this going forward.

Also reduces default gains in script tuner and various other scripts to prevent oscillations from overshoot.


Crafty15 (approved):
LGTM

PR #1290, opened 2023-12-13
Fix scripts cutting out when running / explicitly disable the model watchdog

ysims:
I'm not really sure why this works, maybe unbind doesn't actually disable it? But the problem with running scripts was that the model watchdog was enabled, and at some point in the script the packet watchdog ran and brought up that 'cannot run packet watchdog when model watchdog is enabled' log. At that point the hardware io loop is broken and the robot won't move again.
Adding the disable stops this from happening and the scripts all run fine (ran many times on two different robots).


Crafty15 (approved):
Fixed forever! LGTM

PR #1293, opened 2023-12-29
Fix Convex Hull Side and Some Cleanup/Optimisations

ysims:
- Remove the spherical coordinates from ball as we do not use them, and updated NUsight accordingly
- Visual Mesh message has both uPCw and rPWw for use in detectors rather than calculating rPWw multiple times in different detectors
- rPWw calculation has been changed from a for loop to an Eigen-powered one-liner. Using the nusight profiling tool, this is noticably faster
- The hull side function assumed full vectors to the ground (rPWw), but we were instead passing in uPCw in the detectors. Now the easily accessible rPWw from the Visual Mesh message is used to determine side. 
- Green horizon hull vectors changing required an update in NUsight so they were drawn correctly

Tested in Webots. Created many balls and moved them around the scene and consistently worked.


Tom0Brien (approved):
LGTM.

PR #1300, opened 2024-01-23
Add NSGA-II optimiser with Webots and onboard robot functionality

ysims:
This has been waiting a long time and I think can be merged as it is now and built upon in the future. It is a solid mostly-self-contained research project so is good for FYPs and WIL, and we don't want to lose current progress on it but rather build upon it.

This version has

- NSGA-II optimiser (core files are adapted from a github repo)
-  NSGA2Optimiser and NSGA2Evaluator, NUClear reactors that use the NSGA-II algorithm
- There are some changes scattered through the Webots module and message proto for using the optimiser with Webots
- An OnboardWalkOptimiser reactor that handles the interfacing with the real robot
- Utilises the Director for walking
- Currently uses the QuinticWalk

### Future Work

- General code style, documentation directives, could do with a big refactor too
- The walk engine it uses is hardcoded and it would be way better to generalise this so you could just point to the module name and it reads the yaml and does whatever is needed. Simply switching over to Walk rather than QuinticWalk is a little tedious but simple enough as a small project for someone.
- The code is very walk specific. Generalising for any optimisation problem would be way better. I could see this being a course-credited research project aiming to generalise it for use in tuning the IMU as well (overkill but useful and gives a purpose to the generalisation).
- Use of our existing falling module/s. ie just track the Stability state and include falling in the role (don't need the skill layer if we don't want it to react to it).

### Reviewers

Feel free to ask me anything specific if you don't want to trawl through this PR. 

I don't want to spend longer cleaning the code up, but there are no plans currently for anyone to work on it so if it's left it will probably go out of date and be harder to get back into. I've cleaned up things that I could do quickly but I can definitely see a lot more that could be done. If we do run this for WIL again I think it would be beneficial for this to be merged and the student start their own branch working on building upon this.

It does work in sim and on the real robot fine, but the real robot version needs some mocap ground truth for better evaluation.


module/support/optimisation/NSGA2Evaluator/src/tasks/RotationEvaluator.cpp
```diff
@@ -0,0 +1,237 @@
...
+        fitness_scores->generation                         = generation;
+        fitness_scores->obj_score                          = scores;
+        fitness_scores->constraints                        = constraints;
+        return fitness_scores;
+    }
+
+    std::vector<double> RotationEvaluator::calculate_scores() {
+
+        return {
+            max_field_plane_sway,  // For now, we want to reduce this
```
jpptm, line 212: Should there be a TODO here somewhere for this?

jpptm (commented):
Partial review 10/57 files viewed

module/support/optimisation/NSGA2Evaluator/src/NSGA2Evaluator.cpp
```diff
@@ -0,0 +1,318 @@
...
+                if (event.value == Event::Value::FITNESS_SCORES_SENT) {
+                    return State(State::Value::FINISHED);
+                }
+                else {
+                    return transition_events(event);
+                }
+
+            default: return State(State::Value::UNKNOWN);
+        }
+    }
```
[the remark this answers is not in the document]
ysims: Lol I didn't notice that they were all the same thing, yeah can be simplified more.

module/support/optimisation/NSGA2Evaluator/src/tasks/RotationEvaluator.cpp
ysims: hmm the commenting is maybe hard to follow here. the idea of tracking the maximum sway is to feed it into the evaluator. NSGA2 is specifically for optimising multiple values, and torso sway was one of the values optimised along with distance. The comment is probably eluding to the fact that reducing this value might not be necessary for a good walk but they just needed to come up with a second value to optimise. I think it's an okay metric and probably is too high level to make a todo from, but I'll have a look at the comments and see if anything can be clarified.

module/support/optimisation/NSGA2Evaluator/data/config/NSGA2Evaluator.yaml
```diff
@@ -0,0 +1,5 @@
+gravity:
+  min: 0.5
+  max: 9.2
+
+fallen_angle: 0.95 # ~55 degrees
```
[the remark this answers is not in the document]
TrentHouliston: Following from this, if you're gonna put angles in config, use angles and get the cos of the angle in code. Config files are for humans

module/support/optimisation/NSGA2Evaluator/src/NSGA2Evaluator.cpp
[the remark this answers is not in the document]
TrentHouliston: As for breaks, you don't put breaks in switch statements when you have returns instead. 

But in this case yes seems like an if would work great :P

module/support/optimisation/NSGA2Evaluator/src/tasks/WalkEvaluator.cpp
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    std::vector<double> WalkEvaluator::constraints_not_violated() {
+        return {
+            0.0,  // Robot didn't fall
+            0.0   // Second constraint unused, fixed to 0
+        };
+    }
+
+}  // namespace module::support::optimisation
```
jpptm, line 241: Same comment - walk, strafe and rotation have very similar code structure so we might be able to think of a smart way to condense it all

module/support/optimisation/NSGA2Evaluator/src/tasks/WalkEvaluator.hpp
```diff
@@ -0,0 +1,115 @@
...
+        /// @brief The walk command rotation.
+        double walk_command_rotation = 0.0;
+
+        /// @brief Indicates that a fall has occured.
+        bool fallen = false;
+    };
+
+}  // namespace module::support::optimisation
+
+#endif  // MODULE_SUPPORT_OPTIMISATION_WALKEVALUATOR_HPP
```
jpptm, line 115: The similarity applies to the config files as well

module/support/optimisation/NSGA2Optimiser/data/config/NSGA2Optimiser.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+# Fixed position of the arm joints during walking
+arms:
+  shoulder_pitch: [1.7, 0.5, 2.5]
+  elbow: [-0.7, -2, 0]
+
+# The following values are setting the initial values and limits for the walk engine command.
+walk_command:
+  velocity: [0.15, 0.01, 0.25] #  [-0.2, -1.00, -0.01]
+  rotation: [0.15, -1.57, 0.1] # [0.15, 0.1, 1.57]
```
jpptm, line 85: Same follows with the configs, we could have a yaml that has the following fields:

Walk:
    population: x
    seed: x

Strafe:
    population: x
    ...

and so on, so we can condense code further

module/support/optimisation/NSGA2Optimiser/src/nsga2/Individual.cpp
```diff
@@ -0,0 +1,259 @@
+/* Adapted from https://github.com/dojeda/nsga2-cpp
```
jpptm, line 1: Can we install from source via Docker to get the code as a package instead?

jpptm (commented):
Partial review pt.2, 25/57 files viewed

module/support/optimisation/NSGA2Optimiser/src/nsga2/Individual.cpp
ysims: it's not directly used afaik, it's just adapted from, and the repo has a lot more in it than what we use when I looked.

module/support/optimisation/NSGA2Evaluator/src/tasks/StrafeEvaluator.cpp
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    void StrafeEvaluator::set_up_trial(const NSGA2EvaluationRequest& current_request) {
+        // Set our generation and individual identifiers from the request
+
+        trial_duration_limit = std::chrono::seconds(current_request.trial_duration_limit);
+
+        // Set our walk command
+        walk_command_velocity.x() = 0.0;
+        walk_command_velocity.y() = current_request.parameters.real_params[21];
```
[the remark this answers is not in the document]
ysims: a one liner would be making an eigen object and setting it equal to that but it'll still go over two lines with formatting and would probably be more overhead?

module/support/optimisation/NSGA2Optimiser/data/config/NSGA2Optimiser.yaml
ysims: if I'm understanding you correctly, I think having separate configs is good for modularising the different optimisation tasks. you only run one at a time. this is probably also refactor-y as I also think there could be large improvements to the config situation.

module/support/optimisation/NSGA2Evaluator/data/config/NSGA2Evaluator.yaml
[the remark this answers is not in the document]
ysims: I can't see where gravity is ued in the code so I'm going to remove it. the name doesn't make sense anyway with a min and max.

module/support/optimisation/NSGA2Evaluator/src/NSGA2Evaluator.cpp
```diff
@@ -0,0 +1,318 @@
...
+        }
+        else if (last_eval_request_msg.task == "strafe") {
+            task = std::make_unique<StrafeEvaluator>();
+        }
+        else if (last_eval_request_msg.task == "rotation") {
+            task = std::make_unique<RotationEvaluator>();
+        }
+        else {
+            log<NUClear::ERROR>("Unhandled task type:", last_eval_request_msg.task);
+        }
```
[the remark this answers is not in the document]
ysims: while using an enum makes it a little better I think this whole thing needs generalising so you don't have to keep adding code in so many places every time you make a new problem for the optimiser. also going through this I dont really get why there are all these different tasks since there should be only one task with a configurable walk command?

module/support/optimisation/NSGA2Evaluator/src/tasks/RotationEvaluator.cpp
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    void RotationEvaluator::set_up_trial(const NSGA2EvaluationRequest& current_request) {
+        // Set our generation and individual identifiers from the request
+
+        trial_duration_limit = std::chrono::seconds(current_request.trial_duration_limit);
+
+        // Set our walk command
+        walk_command_velocity.x() = -0.01;
+        walk_command_velocity.y() = 0.01;
```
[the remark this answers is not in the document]
ysims: I don't get why this isn't zero anyway, this looks like something you'd put in for tuning but this is an optimiser so I don't see the point. shouldn't need to have any x-y for turning on the spot.

module/support/optimisation/NSGA2Evaluator/src/tasks/RotationEvaluator.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+        evaluator->walk(Eigen::Vector3d(walk_command_velocity.x(), walk_command_velocity.y(), walk_command_rotation));
+        evaluator->schedule_trial_expired_message(0, trial_duration_limit);
+    }
+
+    bool RotationEvaluator::has_fallen(const Sensors& sensors) {
+        // omega           = sensors.gyroscope.z();
+        // deltaT          = sim_time - old_time;
+        // old_time        = sim_time;
+        // theta += omega * deltaT / 1000;
```
jpptm, line 155: should these be here still?

module/support/optimisation/NSGA2Optimiser/src/tasks/MultiPathOptimiser.cpp
```diff
@@ -0,0 +1,165 @@
...
+namespace module::support::optimisation {
+    using message::support::optimisation::NSGA2EvaluationRequest;
+    using utility::support::Expression;
+
+    void MultiPathOptimiser::setup_nsga2(const ::extension::Configuration& config, nsga2::NSGA2& nsga2_algorithm) {
+        NUClear::log<NUClear::INFO>("Multipath Optimiser Setting up NSGA2");
+        std::vector<std::vector<double>> population_data;
+
+        std::string input_file_name =
+            "../NUbots/module/support/optimisation/NSGA2Optimiser/data/config/Supplied_Pop_Reals.csv";
```
jpptm, line 51: yikes

jpptm: An on<Trigger<CommandLineArguments>> would probably be ideal when used in the relevant reactor? Or just a config coming from some yaml file during on<Configuration>

module/support/optimisation/NSGA2Optimiser/src/tasks/MultiPathOptimiser.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+        population_data = read_in_population_file(input_file_name);
+        NUClear::log<NUClear::INFO>("Data file size is", population_data.size());
+
+
+        // bool supp_pop =
```
jpptm, line 57: Scar tissue most likely

module/support/optimisation/NSGA2Optimiser/src/tasks/MultiPathOptimiser.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+        for (const auto& element : std::vector<std::string>({std::string("rotation")})) {
+            param_initial_values.emplace_back(walk_command[element][0].as<Expression>());
+            param_limits.emplace_back(walk_command[element][1].as<Expression>(),
+                                      walk_command[element][2].as<Expression>());
+        }
+
+        walk_config_path     = config["task_config_path"].as<std::string>();
+        trial_duration_limit = config["trial_duration_limit"].as<int>();
+        // supplied
```
jpptm, line 81: supplied?

module/support/optimisation/NSGA2Optimiser/src/tasks/MultiPathOptimiser.cpp
```diff
@@ -0,0 +1,165 @@
...
+                }
+
+                data.push_back(record);
+            }
+        }
+
+        if (!inputFile.eof()) {
+            // cerr << "Could not read file " << inputFileName << "\n";
+            NUClear::log<NUClear::INFO>("Could not read file", file_name);
+            // std::__throw_invalid_argument("File not found.");
```
jpptm, line 159: Address commented code

module/support/optimisation/OnboardWalkOptimisation/README.md
```diff
@@ -0,0 +1,23 @@
...
+- `message::input::Sensors` to determine if the robot is stable.
+
+## Emits
+
+- `message::support::optimisation::OptimisationResetDone` to tell the optimisation when the robot has been reset for a new trial run.
+- `message::support::optimisation::OptimisationRobotPosition` to tell the optimiser where the robot is in the world.
+
+## Dependencies
+
+- An optimiser
```
jpptm, line 23: Might be useful to provide the link for the source code of the adapted optimiser here

jpptm:
I've skimmed through the code and they look fine, at least talking from a code structure perspective. I did not pay too much attention to the technicality and implementation but as @ysims said we can do this after the next person who is going to be working on this addresses the aforementioned technical/reformatting/reimplementation comments

module/support/optimisation/OnboardWalkOptimisation/README.md
ysims: I'm inclined to keep this module separate to any specific optimiser, as it should be generic enough to be used by any.

jpptm (approved):
LGTM, this pr thread should prove a good resource for the next person to work on this feature in the future.

PR #1302, opened 2024-02-02
Fix asan errors on main

ysims:
Main currently has three asan errors that occur during normal `robocup` play. 

- FieldLocalisation: the occupany map rectangles were sometimes being created with negative length/width
- Vision: size of the matrices in the one-liner for rPWw were wrong
- WalkToFieldPosition: a 3d vector was initialised with only two values, leaving the last value uninitialised


JesseWilliamson (approved):
lgtm

PR #1303, opened 2024-02-02
More thoroughly reset odometry when resetting localisation 

ysims:
When the robot gets penalised, it's localisation becomes very off because it is moved in simulation or physically moved to a new position. We have a localisation reset to fix this, but it doesn't fully reset odometry when it resets localisation.

This PR resets the fiilters and odometry to the initial values from config.


jpptm (approved):
lgtm

PR #1304, opened 2024-02-03
Add a working StandUpFront for Frankie

ysims:
Very reliable get up front for the new legs on Frankie. I expect we will see this working nicely on any other upgraded robots.


ysims:
Superceded by #1307

PR #1307, opened 2024-02-09
Tunes Frankie (nugus4) and small code changes for quality of life and fixes for the real robot and lab

ysims:
- Frankie's nuc was swapped for Kevin's, so now Frankie is `nugus4`. As such, Frankie's HardwareIO, Walk and Camera tunes have been moved to `nugus4`. The `nugus1` nuc isn't in a robot so nothing to change it to. 
- Adjusted the camera to make it brighter
- Tuned the `nugus4` Left.yaml pitch and yaw so field lines in localisation look correct
- Added back in the logic to field occupancy map to not draw the penalty box and crosses for the lab field
- Changed NetworkForwarder from `Buffer<10>` to `Single`. The reaction will only run one at a time and therefore should only be using one thread. This appears to fix our issues with the robot not walking if NetworkForwarder is in the role, and allows us to use nusight with the real robot while its moving.
-  Tuned kick planner a little bit to get the robot kicking at the right time
- Tuned the green horizon to be more picky about what is a field point
- Tuned the ball detector. Basically took out the distance disagreement and then some small tuning. Robot chases the ball, if it does lose it when it's close it finds it again pretty quick. 
- Added in the StandUpFront for Frankie which was in another PR but now that the robot number has changed it's out of date so just including it in here.
-  Added in networking things and the message log handler in scripttuner so that we can see logs in NUsight when running script tuner.

Overall, this PR results in Frankie correctly seeing the ball, walking to it, and attempting to kick it when close. Ran it from the side of the field and it walked around kicking the ball a few times around the field, with correct localisation. After about the third kick localisation had drifted perpendicular of what it should have been. It appeared that when the robot was unsteady (ie would have falled over but I held it up) localisation got worse. 

### Next Steps

Not part of this PR, but noticeable problems to fix are

- Kick tune, as it is unsteady
- Walk planning tuning for turning, as it is a bit unsteady when turning dramatically
- Walk tuning for turning while facing the ball (pivoting), which is needed for aligning to the goal
- Tuning/improvements to localisation to keep it steady and not drift
- Walk fixes to linearly interpolate some of the positions rather than using splines as the foot moves in an awkward curve on the y-axis
- Testing/tuning for ready positioning, aligning to goal and kicking to goal (localisation may be good enough for small tests)
- StandUpBack has not been looked at. This should be done FROM SCRATCH. Old scripts have issues on the robots as positions overextend the limbs. Going from scratch with StandUpFront had no issues and no servos were broken.

### DIFF
--- a/module/input/Camera/data/config/Cameras/Left.yaml
+++ b/module/input/Camera/data/config/Cameras/Left.yaml
@@ -7,9 +7,9 @@ settings:
   OffsetX: 0
   OffsetY: 0
   ExposureAuto: Continuous # Options: Off, Once, Continuous
-  AutoExposureExposureTimeUpperLimit: 4000 # Lower time reduces motion blur
+  AutoExposureExposureTimeUpperLimit: 6000 # Lower time reduces motion blur
   GainAuto: Off # Options: Off, Once, Continuous
-  Gain: 18
+  Gain: 10
   BalanceWhiteAuto: Off # Options: Off, Once, Continuous
   AcquisitionFrameRateEnable: true
   DeviceLinkThroughputLimit: 60000000 # Throughput can limit the framerate
--- a/module/input/Camera/data/config/nugus4/Cameras/Left.yaml
+++ b/module/input/Camera/data/config/nugus4/Cameras/Left.yaml
@@ -10,7 +10,7 @@ lens:
   # focal_length / (camera_pixel_size * width)
   # 1.98mm / (4.8µm * 1280px) = 0.322265625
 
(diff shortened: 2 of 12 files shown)


Crafty15 (approved):
LGTM

jpptm:
> Added back in the logic to field occupancy map to not draw the penalty box and crosses for the lab field

What was the reason for this?

ysims:
> > Added back in the logic to field occupancy map to not draw the penalty box and crosses for the lab field
> 
> What was the reason for this? Everything LGTM, just curious about this one

We don't have penalty boxes and penalty/center crosses in the lab

PR #1310, opened 2024-02-16
Swap NUSense over to using a stream reactor

TrentHouliston:
Swaps the NUSense C++ code to using a StreamReactor and use message reflection to forward all messages coming from NUsense to the general NUClear system.

Based on what was there before this will also require a change on the NUSense side so it sends hash/timestamp/length etc like an NBS packet.

Also I have not tested this at all :P


module/platform/NUsense/HardwareIO/src/NUSenseParser.cpp
```diff
@@ -0,0 +1,61 @@
...
+     * Read a little endian value from a byte array
+     *
+     * @tparam T The type of the value to read
+     *
+     * @param ptr The pointer to the first byte of the value
+     *
+     * @return The value read from the byte array
+     */
+    template <typename T>
+    T read_le(const uint8_t* ptr) {
```
Bidski, line 15: This needs to include a max array length.

Also, just use htons/ntohs (or whatever variant is appropriate) and implement a [64-bit version](https://stackoverflow.com/questions/3022552/is-there-any-standard-htonl-like-function-for-64-bits-integers-in-c)

module/platform/NUsense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -1,118 +1,60 @@
...
-            command_list->targets.push_back(command);
-
-            // Emit it so it's captured by the reaction above
-            emit<Scope::DIRECT>(std::move(command_list));
+        // Send servo targets to the device
+        on<Trigger<ServoTargets>>().then("ServoTargets", [this](const ServoTargets& packet) { send_packet(packet); });
 
-            log<NUClear::INFO>("Servo target received");
+        // Emit any messages sent by the device to the rest of the system
+        on<Trigger<NBSFrame>>().then("From NUSense", [this](const NBSFrame& packet) {
```
Bidski, line 56:
```suggestion
        on<Trigger<NUSenseFrame>>().then("From NUSense", [this](const NUSenseFrame& packet) {
```

Bidski (changes requested):

module/platform/NUsense/HardwareIO/src/NUSenseParser.cpp
TrentHouliston: Including max array length for this is the responsibility of the caller which is why it's a private cpp only function not a public one for anyone to use. All the usages of it in the functions below have had appropriate length checks on the buffer to ensure no out of bounds memory access happens.

htons and ntohs only work for dealing with big endian values, not little endian values since they interact with "native" and "network" byte order only. So reading little-endian values form the network and converting them to native doesn't work with them.

That said, after putting this code through godbolt, it seems that the compiler isn't as smart as I thought when it comes to converting to appropriate instructions.

I made a bunch of implementations and it seems to only recognise what is going on here in a few cases.
I think either the bitwise version of the recursive template would be best given that it recognises those two correctly. 
https://godbolt.org/z/1Wof68xGq

Bidski: Just dump this function and use a reinterpret cast, the likelihood of ever running on a big endian system is so slim that it doesn't bare accounting for

Bidski (changes requested):

module/platform/NUsense/HardwareIO/src/HardwareIO.hpp
```diff
@@ -2,28 +2,45 @@
...
-
-        /// @brief Manage desired port for NUSense
-        utility::io::uart nusense;
+        template <typename T>
+        void send_packet(const T& packet) {
+            // Serialize the packet
+            auto payload = NUClear::util::serialise::Serialise<T>::serialise(packet);
+
+            // Create the nbs packet
+            std::vector<uint8_t> nbs;
```
JohanneMontano, line 22:
```suggestion
            std::vector<uint8_t> bytes;
```

module/platform/NUsense/HardwareIO/src/HardwareIO.hpp
```diff
@@ -2,28 +2,45 @@
...
+#include "utility/reactor/StreamReactor.hpp"
 
-    class HardwareIO : public NUClear::Reactor {
+namespace module::platform::NUsense {
 
+    class HardwareIO : public utility::reactor::StreamReactor<HardwareIO, NBSParser, 5> {
     public:
-    /// @brief Called by the powerplant to build and setup the HardwareIO reactor.
-    explicit HardwareIO(std::unique_ptr<NUClear::Environment> environment);
+        /// @brief Called by the powerplant to build and setup the GPS reactor.
```
JohanneMontano, line 12: Maybe GPS wasn't supposed to be here?

PR #1311, opened 2024-03-11
Add localisation ground truth method

JesseWilliamson:
Adds graphs for debugging localisation error using Webots ground truth data, like #1132 does for odometry.

Relevant changes to NUWebots:
https://github.com/NUbots/NUWebots/pull/111

### DIFF
--- a/module/localisation/FieldLocalisation/src/FieldLocalisation.cpp
+++ b/module/localisation/FieldLocalisation/src/FieldLocalisation.cpp
@@ -33,6 +33,9 @@
 #include "message/behaviour/state/Stability.hpp"
 #include "message/platform/RawSensors.hpp"
 
+#include "utility/math/euler.hpp"
+#include "utility/nusight/NUhelpers.hpp"
+
 namespace module::localisation {
 
     using extension::Configuration;
@@ -42,7 +45,10 @@ namespace module::localisation {
     using message::localisation::ResetFieldLocalisation;
     using message::vision::FieldLines;
 
+    using utility::math::euler::MatrixToEulerIntrinsic;
+    using utility::nusight::graph;
     using utility::support::Expression;
+
 
     using namespace std::chrono;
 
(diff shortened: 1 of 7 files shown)


ysims:
Update PR name. Also LGTM otherwise. One comment though is I'm not sure I like set up we have currently with raw sensors holding the ground truth data. Maybe it can be its own message. But that could be another PR.

module/localisation/FieldLocalisation/src/FieldLocalisation.cpp
```diff
@@ -122,13 +128,37 @@ namespace module::localisation {
                     filter.time(dt);
 
                     auto field(std::make_unique<Field>());
-                    field->Hfw        = compute_Hfw(filter.get_state());
+                    field->Hfw = compute_Hfw(filter.get_state());
```
Tom0Brien, line 131: Let's add a configuration option where you can select between using filter or ground truth

Tom0Brien: Something like
```
if(cfg.use_ground_truth){
        field->Hfw = raw_sensors.localisation_ground_truth.Hfw;
} else {
        field->Hfw = compute_Hfw(filter.get_state());
}
```

Tom0Brien (changes requested):

module/localisation/FieldLocalisation/src/FieldLocalisation.cpp
```diff
@@ -112,29 +113,37 @@ namespace module::localisation {
...
                     auto field(std::make_unique<Field>());
-                    field->Hfw = compute_Hfw(filter.get_state());
-                    if (log_level <= NUClear::DEBUG && raw_sensors.localisation_ground_truth.exists) {
-                        debug_field_localisation(field->Hfw, raw_sensors);
-                    }
-                    field->covariance = filter.get_covariance();
+                    field->Hfw = raw_sensors.localisation_ground_truth.Hfw;
                     emit(field);
                 }
+                else {
```
ysims, line 122: I would recommend something like

```
// Emit Field message using ground truth
if (cfg.use_ground_truth_localisation) {
    ...
    return;
}

// Otherwise calculate using field lines
if (!fallen && field_lines.rPWw.size() > cfg.min_observations ... etc etc
...
});
    
Because every layer of if/for/etc adds a cognitive load to the reader, so by returning after the if we can complete the rest without the extra `else` nesting the code.

JesseWilliamson: I think semantically if-else makes more sense when choosing between two roughly equivalent options for the same task like this. But I see what you mean about the second being easier to read.

ysims (approved):

PR #1315, opened 2024-03-18
Add ORBSLAM3

Tom0Brien:
TODO:

- [ ] Build ORBSLAM3 as library in Docker
- [ ] Map output from ORBSLAM3 coordinates to our coordinate system (Htw)
- [ ] Figure out how to scale position to best match real translation units
- [ ] Test on real robot


module/input/ORBSLAM/src/ORBSLAM.cpp
```diff
@@ -0,0 +1,90 @@
...
+                    img_in  = cv::Mat(image.dimensions.y(),
+                                     image.dimensions.x(),
+                                     CV_8UC1,
+                                     const_cast<uint8_t*>(image.data.data()));
+                    img_out = img_in.clone();
+                    break;
+                default:
+                    log<NUClear::WARN>(
+                        fmt::format("Image format not supported: {}", utility::vision::fourcc(image.format)));
+                    powerplant.shutdown();
```
KipHamiltons, line 70: If we are shutting down, shouldn't it log fatal? Must be higher than warning? 😂

module/vision/Yolo/src/Yolo.cpp
```diff
@@ -0,0 +1,314 @@
...
+                double norm_factor = img.dimensions.x();
+
+                // Helper lambda to simplify unprojection calls
+                auto unproject_point = [&](double x, double y) {
+                    return unproject(Eigen::Matrix<double, 2, 1>(x / norm_factor, y / norm_factor),
+                                     img.lens,
+                                     normalized_dim);
+                };
+
+                Eigen::Matrix<double, 3, 1> top_left_ray     = unproject_point(box_x, box_y);
```
KipHamiltons, line 199: Vector3d??

PR #1317, opened 2024-03-19
Add robot localisation module

Tom0Brien:
This PR adds:

- A robot localisation module which estimates the position and velocity of other robots on the field.
- NUsight visualisation of tracked robots in localisation view.

image

The module works by tracking multiple robots using a UKF filter for each. Vision measurements are associated with each tracked robot using global nearest neighbor and an acceptance radius. Improving the filtering method and data association is definitely possible, but this acts as a good baseline to improve upon.


module/localisation/RobotLocalisation/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Description
+
+## Usage
+
+## Consumes
+
+## Emits
+
+## Dependencies
```
ysims, line 11: fill readme

module/localisation/RobotLocalisation/src/RobotLocalisation.cpp
```diff
@@ -0,0 +1,171 @@
...
+                }
+
+                // If the tracked robot has not been seen, increment the consecutively missed count
+                tracked_robot.missed_count = tracked_robot.seen ? 0 : tracked_robot.missed_count + 1;
+            }
+
+            // Make a vector to store kept robots
+            std::vector<TrackedRobot> new_tracked_robots;
+            // Only keep robots that are not missing or too close to others
+            for (const auto& tracked_robot : tracked_robots) {
```
KipHamiltons, line 96: Could consider replacing this with a ranges::filter. An erase remove idiom would be most efficient actually, since you want it on the same vector after the fact.

module/localisation/RobotLocalisation/README.md
```diff
@@ -0,0 +1,24 @@
+# RobotLocalisation
+
+## Description
```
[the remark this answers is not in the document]
Tom0Brien: Not any resource in particular, it's a well established idea though

module/localisation/RobotLocalisation/src/RobotLocalisation.cpp
Tom0Brien: I agree. I believe this requires a bit of effort though in setting up TrackedRobot struct to work nicely with this, I could work on it but not super critical imo

ysims (approved):

PR #1318, opened 2024-03-23
Some changes to the robot localisation code [base is not main!]

ysims:
- data association function: removed tracked robots as it's a class variable, updated docs
- Changed the part where the current tracked robots are checked over to make it a bit less indented/possibly easier to read, merged the two steps as well, and also stopped using the remove_if as it `move`s the tracked robots, which is why the `id` being `const` was a problem as it basically makes new tracked robot and copies the original contents
- Made the `id` in tracked robot `const`, also made the `next_id` an unsigned long cause I'm lowkey worried about overflow
- Got rid of the make new robot function, moved most of the stuff for setting up the ukf into the constructor, and did an `emplace_back` so we can add to the vector and make a robot at the same time
- Added a `get_rRWw` function because it was annoying with those `RobotModel<double>::StateVec`s everywhere haha
- Also got rid of the unused yaml fields in initial mean and brought the config struct in one layer, I was gonna go further cause it seems we don't need so much nesting then realised the velocity wasn't used either so I'm gonna leave that up to you to deal with as I'm not sure if you wanted to expand with velocity (but we don't want unused variables if you're not gonna add it in this PR imo)
- Initialisation, more debug logs, moved usings and includes as they're not used in the hpp, made a ternary

### DIFF
--- a/module/localisation/RobotLocalisation/data/config/RobotLocalisation.yaml
+++ b/module/localisation/RobotLocalisation/data/config/RobotLocalisation.yaml
@@ -10,13 +10,9 @@ ukf:
       position: [1e-1, 1e-1]
       velocity: [1e-2, 1e-2]
 
-  initial:
-    mean:
-      position: [5.0, 0.0]
-      velocity: [0.0, 0.0]
-    covariance:
-      position: [1e-3, 1e-3]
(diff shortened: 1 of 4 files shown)


Tom0Brien:
Awesome!

PR #1321, opened 2024-03-26
Update Frankie getups and arm offsets

ysims:
- Updates `scripttuner.role` to not have ConsoleLogHandler (must have been too copy-paste happy in the roles update)
- Create a new Stand script for the NUgus, equal to the at-rest walk position
- Add a reliable StandUpBack for Frankie
- Fix up the ankle pitch on the StandUpFront script (caused by motor not being connected when locking)
- Add an offset to the arms to match the physical change on nugus4 - horn should be straight up then one position towards the back (counterclockwise). The problem with having it straight up is that the servo's physical 180-degree position is within the range of movement we want the arm to be able to do. This causes issues when using the arms for getting up from the back, ie it will spin around the opposite way to get to the position after the 180-degree mark.


Tom0Brien (approved):
LGTM.

PR #1326, opened 2024-04-06
Send duration to the NUSense device for correct velocity

ysims:
Otherwise the NUSense and servos don't know how fast we want to move. Time-based velocity just wants a milliseconds value. 

Also some cleaning of comments.

Also removed the NUSense role as it **breaks the build**. Use `test/sensor` with the NUSense flag for the subcontroller.


Claegtun (approved):

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -31,7 +33,7 @@ namespace module::platform::NUSense {
 
             nusense = utility::io::uart(cfg.nusense.port, cfg.nusense.baud);
 
-            log<NUClear::INFO>("PORT OPENED");
+            log<NUClear::INFO>("Port to NUSense opened.");
```
jpptm, line 36:
```suggestion
            log<NUClear::INFO>(fmt::format("Port {} successfully opened.", cfg.nusense.port));
```

PR #1328, opened 2024-04-06
Cleanup code in NUSense HardwareIO

ysims:
Lots of things

- Fill in readme
- Initialise variables
- Documentation directives on comments
- PacketHandler files names to PascalCase as it is a class
- Lowered some nesting (ifs in ifs)
- Ran format with the license code so we get the right license headers
- Removed the empty test file, which isn't needed anymore and isn't generated for new modules
- Removed a lot of commented out stuff
- Removed some includes that didn't look to be needed, moved some over to the cpp as they're only needed there


Tom0Brien (approved):

TrentHouliston:
Just use the stream reactor rather than continuing to work on this implementation!

jpptm:
lgtm, although yes it will eventually be replaced by stream reactor when I get a good portion of my assignments done

PR #1332, opened 2024-04-20
Clean up foot controller and tune gains

Tom0Brien:
This PR does the following:

- Moves foot controller functionality into a templated function to make general
- Add ability to configure each servo with different proportional gain value
- Adds goal position to RawSensors when using Webots

_Note:_ This removes keep level mode, which makes `FootController` redundant but I intend to add additional modes back in later PR's

Additionally,  currently if walk state doesn't exist or the walk phase is double support, the component of odometry using kinematics for torso z height results in 0... This PR fixes this by making the WalkState message not optional. On startup a WalkState message is emitted to ensure odometry still runs for roles which don't have walk.

Note, we should really change this logic to use foot down information which I will do in later PR.


module/platform/Webots/src/Webots.cpp
```diff
@@ -159,6 +159,71 @@ namespace module::platform {
...
+        }
+        if (name == "neck_yaw_sensor") {
+            return 18;
+        }
+        if (name == "head_pitch_sensor") {
+            return 19;
+        }
+
+        throw std::runtime_error(fmt::format("Unable to translate unknown NUgus.proto servo name: {}", name));
+    }
```
jpptm, line 225:
```suggestion
    [[nodiscard]] uint32_t translate_name_to_id(const std::string& name) {
       
        std::vector<std::string> names{};
        
        // Fill names below, or alternatively declare somewhere else
        
        for (size_t i = 0; i < names.size(); ++i) {
            if (name == names[i]) {
                return i;
            }
        }
        throw std::runtime_error(fmt::format("Unable to translate unknown NUgus.proto servo name: {}", name));
    }
```
something like this instead? Alternatively do as a switch statement or use a map so it could just be one declaration within Webots.hpp and this function does not need to exist anymore as one can just do `uint8_t id = names[name];`

ysims: or turn clang format off and have each if on one line

TrentHouliston: Or if it's one specific file/folder that's got a lot of these you can put another `.clang-format` file in here with 
```yaml
BasedOnStyle: InheritParentConfig
# rules you want to change
```
and it'll then apply to everything in the folder the .clang-format is and all folders below it.

ysims:
who was the walk tuned on

jpptm (approved):
lgtm

PR #1333, opened 2024-04-22
Add field type selector to localisation view

Tom0Brien:
This PR adds a dropdown in localisation menu to switch between `Robocup` and `Lab` hardcoded field dimensions

Using field description message to update dimensions dynamically is a good idea in principle but isn't usually practical because the `FieldDescription` message is emitted only once on startup and NUsight won't reliably be open or pick it up, so this method is better for now,


TrentHouliston:
Upgrading network forwarder and adding a request RPC call to get the latest copy of a specific message type was the longer term solution to this problem. But for now this is a good idea

ysims (approved):

PR #1334, opened 2024-04-22
Add dynamic soccer position choosing

A-Herfel:
Allows robots to choose and determine their position dynamically. Uses a leader-election system.  

### Pre-PR Checklist:

Have you

### Reviewers, Note

The logic could be simplified somewhat (i.e. by distibuted decision making, only needing startup_time), but would be less easily adaptable to future changes, especially those reliant on real-world variables, e.g. ball position. The development of this feature is ongoing. 

We should always have only 1 striker. The striker is currently our oldest active robot (alive and unpenalised). Where robots became active at the same time, the player with the lowest robot_id will become the leader. 

Test cases passed on Webots: 
- [x] Single robot
- [x] Initial startup for multiple robots
- [x] Leader Robot "dies"/loses connection. Then boots back up. Leadership transfers to follower robot. 
- [x] Follower Robot "dies"/loses connection. Then boots back up. Leadership does not transfer. 
- [ ] Above test cases but for penalised, then unpenalised robots
- [x] Non-dynamic robot is not considered in active robots (Goalie)   
- [ ] Leader Robot "dies"/loses connection, but boots back up before other robots detect it is gone. Leadership transfers to follower robot. 

### Things left to do

- [x] Fix segfault in non-dynamically decided robot positions at "set" step
- [ ] Unpenalisation case doesn't work 
- [x] Add the purpose information to nusight
- [ ] Test all cases again on robots, with the improved visibility from nusight


module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -86,17 +100,40 @@ namespace module::purpose {
             emit<Task>(std::make_unique<FallRecovery>(), 2);
```
A-Herfel, line 100: emit empty robocup msg

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -27,13 +27,15 @@
 #include "Soccer.hpp"
```
A-Herfel, line 28: import localisation/ball etc. and find the half of the field the ball is on

With will grab last version of the msg

module/network/RobotCommunication/data/config/RobotCommunication.yaml
```diff
@@ -1,5 +1,5 @@
 # Controls the minimum log level that NUClear log will display
-log_level: INFO
+log_level: DEBUG
```
ysims, line 2: when you've finished testing change back to INFO

```suggestion
log_level: INFO
```

module/purpose/Soccer/data/config/webots/Soccer.yaml
```diff
@@ -1,9 +1,9 @@
 # Controls the minimum log level that NUClear log will display
-log_level: INFO
+log_level: DEBUG
```
ysims, line 2:
```suggestion
log_level: INFO
```

module/purpose/Soccer/data/config/webots2/Soccer.yaml
```diff
@@ -1,9 +1,9 @@
 # Controls the minimum log level that NUClear log will display
-log_level: INFO
+log_level: DEBUG
```
ysims, line 2:
```suggestion
log_level: INFO
```

module/purpose/Soccer/data/config/webots3/Soccer.yaml
```diff
@@ -1,9 +1,9 @@
 # Controls the minimum log level that NUClear log will display
-log_level: INFO
+log_level: DEBUG
```
ysims, line 2:
```suggestion
log_level: INFO
```

module/purpose/Soccer/data/config/webots4/Soccer.yaml
```diff
@@ -1,5 +1,5 @@
 # Controls the minimum log level that NUClear log will display
-log_level: INFO
+log_level: DEBUG
```
ysims, line 2:
```suggestion
log_level: INFO
```

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -62,16 +69,26 @@ namespace module::purpose {
...
             cfg.force_playing = config["force_playing"].as<bool>();
 
             // Get the soccer position, if not valid option then default to striker
             cfg.position = Position(config["position"].as<std::string>());
+
+            // initialise robot storage
+            active_robots = std::vector<RobotInfo>();
+
+            // decide which player we are
+            PLAYER_ID = global_config.player_id;
```
ysims, line 90:
```suggestion
            player_id = global_config.player_id;
```
not really a const variable so snake_case

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -62,16 +69,26 @@ namespace module::purpose {
...
+            .then([this](const Configuration& config, const GlobalConfig& global_config) {
             // Use configuration here from file Soccer.yaml
             this->log_level   = config["log_level"].as<NUClear::LogLevel>();
             cfg.force_playing = config["force_playing"].as<bool>();
 
             // Get the soccer position, if not valid option then default to striker
             cfg.position = Position(config["position"].as<std::string>());
+
+            // initialise robot storage
+            active_robots = std::vector<RobotInfo>();
```
ysims, line 87: you can get rid of this and initialise in the hpp

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -62,16 +69,26 @@ namespace module::purpose {
...
+
+            // decide which player we are
+            PLAYER_ID = global_config.player_id;
+            log<NUClear::DEBUG>("Configure Player ID ", int(PLAYER_ID));
```
ysims, line 91: You could keep this log in if it would help in the future with debugging. It'll only run if the user sets log level to debug.

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -88,32 +105,53 @@ namespace module::purpose {
...
             // The robot should always try to recover from falling, if applicable, regardless of purpose
             emit<Task>(std::make_unique<FallRecovery>(), 2);
+            // Trigger Find purpose by emitting robocup
+            auto msg = std::make_unique<RoboCup>();
+            msg->current_pose.player_id = PLAYER_ID;
+            emit(msg);
         });
 
-        on<Provide<FindPurpose>>().then([this] {
+        on<Provide<FindPurpose>(), Trigger<RoboCup>>().then([this] (const RoboCup robocup) {
```
ysims, line 114:
```suggestion
        on<Provide<FindPurpose>, Trigger<RoboCup>>().then([this] (const RoboCup robocup) {
```

module/purpose/Soccer/src/Soccer.hpp
```diff
@@ -29,19 +29,26 @@
...
 #include "extension/Behaviour.hpp"
+#include "message/input/RoboCup.hpp"
 
 namespace module::purpose {
 
+    using message::input::RoboCup;
+
     class Soccer : public ::extension::behaviour::BehaviourReactor {
     private:
+        uint PLAYER_ID;
```
ysims, line 43:
```suggestion
        uint player_id = 0;
```

module/purpose/Soccer/src/Soccer.hpp
```diff
@@ -68,6 +76,36 @@ namespace module::purpose {
             Position position{};
         } cfg;
 
+        struct RobotInfo {
+            uint8_t robot_id;
```
ysims, line 80:
```suggestion
            uint8_t robot_id = 0;
```

module/purpose/Soccer/src/Soccer.hpp
```diff
@@ -68,6 +76,36 @@ namespace module::purpose {
...
+            std::chrono::steady_clock::time_point last_heard_from;
```
ysims, line 81:
```suggestion
            std::chrono::steady_clock::time_point last_heard_from = NUClear::clock::now();
```

module/purpose/Soccer/src/Soccer.hpp
```diff
@@ -68,6 +76,36 @@ namespace module::purpose {
...
+
+            bool operator<(const RobotInfo& other) const {
+                return robot_id < other.robot_id;
+            }
+        };
+
+        /// @brief Add and update active robots
+        void manage_active_robots(const RoboCup& robocup);
```
ysims, line 89: add `@param` after the brief for the `robocup` parameter

module/purpose/Soccer/src/Soccer.hpp
```diff
@@ -68,6 +76,36 @@ namespace module::purpose {
...
+
+        /// @brief Add RobotInfo ordered by id
+        void add_robot(RobotInfo new_robot);
```
ysims, line 92: add `@param` after the brief for the new_robot parameter

module/purpose/Soccer/src/Soccer.hpp
```diff
@@ -68,6 +76,36 @@ namespace module::purpose {
...
+
+        /// @brief Return the index of the striker
+        uint8_t find_striker();
+
+        /// @brief Decide the correct soccer position
+        void give_directions();
+
+        /// @brief emit purpose based on leader's instructions
+        void follow_directions(const RoboCup& robocup, const uint8_t incoming_robot_id);
```
ysims, line 101: also `@params` for both robocup and incoming robot id

module/purpose/Soccer/src/Soccer.hpp
```diff
@@ -68,6 +76,36 @@ namespace module::purpose {
...
+
+        /// @brief leader determines its own soccer position
```
ysims, line 103: `@param`

module/purpose/Soccer/src/Soccer.hpp
```diff
@@ -68,6 +76,36 @@ namespace module::purpose {
...
+        void self_direct(uint8_t striker_idx);
+
+        /// @brief Store robots that can currently play
+        std::vector<RobotInfo> active_robots;
```
ysims, line 107:
```suggestion
        std::vector<RobotInfo> active_robots{};
```
initialises it empty

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -88,32 +105,53 @@ namespace module::purpose {
...
                 emit<Task>(std::unique_ptr<FindPurpose>(nullptr));
             }
+
+            give_directions();
         });
 
         on<Trigger<Unpenalisation>>().then([this](const Unpenalisation& self_unpenalisation) {
             // If the robot is unpenalised, stop standing still and find its purpose
             if (!cfg.force_playing && self_unpenalisation.context == GameEvents::Context::SELF) {
+                // TODO: check if we still need the FindPurpose thing
```
ysims, line 150: yeah you'll want the FindPurpose Task to run as this robot is unpenalised and needs to find out what its position is now.

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -88,32 +105,53 @@ namespace module::purpose {
...
                 emit<Task>(std::make_unique<FindPurpose>(), 1);
+                auto msg = std::make_unique<RoboCup>();
+                msg->current_pose.player_id = PLAYER_ID;
```
ysims, line 153:
```suggestion
                msg->current_pose.player_id = player_id;
```

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -124,8 +162,131 @@ namespace module::purpose {
                 log<NUClear::INFO>("Force playing started.");
                 cfg.force_playing = true;
                 emit<Task>(std::make_unique<FindPurpose>(), 1);
+                auto msg = std::make_unique<RoboCup>();
+                msg->current_pose.player_id = PLAYER_ID;
```
ysims, line 166:
```suggestion
                msg->current_pose.player_id = player_id;
```

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -124,8 +162,131 @@ namespace module::purpose {
...
+                emit(msg);
+            }
+        });
+
+        // check if we have heard from robots periodically, and remove them if not
+        on<Every<2, Per<std::chrono::seconds>>>().then([this] {
+            auto now = std::chrono::steady_clock::now();
+            constexpr auto timeout = std::chrono::seconds(8); //TODO: test with robot or get advice
```
ysims, line 174: make the timeout number a config variable rather than hardcoding

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -124,8 +162,131 @@ namespace module::purpose {
...
+
+            // remove inactive robots
```
ysims, line 176:
```suggestion
            // Remove inactive robots
```

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -124,8 +162,131 @@ namespace module::purpose {
                 log<NUClear::INFO>("Force playing started.");
                 cfg.force_playing = true;
                 emit<Task>(std::make_unique<FindPurpose>(), 1);
+                auto msg = std::make_unique<RoboCup>();
+                msg->current_pose.player_id = PLAYER_ID;
+                emit(msg);
+            }
+        });
+
+        // check if we have heard from robots periodically, and remove them if not
```
ysims, line 171:
```suggestion
        // Check if we have heard from robots periodically, and remove them if not
```

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -124,8 +162,131 @@ namespace module::purpose {
...
+            give_directions();
+        }
+
+        follow_directions(robocup, incoming_robot_id);
+    };
+
+    // find which robot should be our striker
+    uint8_t Soccer::find_striker() {
+        return 0;
+        // TODO: ideas- furthest ahead on field could be striker, or closest to ball
```
ysims, line 230: also want to keep in mind that we don't want the striker changing when it doesn't need to. 
ie player 1 is penalised, player 2 is striker
player 2 is close to scoring
player 1 is unpenalised and becomes the striker
player 2 moves back to defense when it could have scored

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -124,8 +162,131 @@ namespace module::purpose {
...
+        // A robot may emit leader messages mistakenly. Only listen to real leader.
+        bool other_is_leader = active_robots.front().robot_id == incoming_robot_id;
+        bool purposes_exist = !robocup.purposes.purposes.empty();
+
+        if (other_is_leader && purposes_exist) {
+            for (const auto& purpose : robocup.purposes.purposes) {
+                if (purpose.player_id == PLAYER_ID ) {
+                    if (purpose.purpose == SoccerPosition::DEFENDER) {
+                        emit<Task>(std::make_unique<Defender>(cfg.force_playing));
+                        log<NUClear::DEBUG>("made defender");
```
ysims, line 280: I like these logs and I'd recommend making them `INFO` level so we can see when the robot changes position

ysims (commented):
these are only really small things, the only big thing would be to prevent unnecessary changes to who is the striker

module/purpose/Soccer/src/Soccer.cpp
A-Herfel: The only issue is that they'd be "made defender" every 2 secs but I could wrap a guard to not have the task if they already are?

ysims: ah I see yeah that'd be good, I think it'd be useful to see the change

module/purpose/Soccer/src/Soccer.cpp
A-Herfel: Think it's this way in R-2-R Comms

module/purpose/Soccer/src/Soccer.cpp
A-Herfel: oops forgot to ask about this. I think that emitting the Robocup msg should now trigger the FindPurpose code, so wondering if we need 

emit<Task>(std::make_unique<FindPurpose>(), 1);

module/purpose/Soccer/src/Soccer.cpp
ysims: it might be good to get this merged though then add this on, as this PR will work well for single-robot scenarios and it might be a bit more work to add in this functionality.

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -269,22 +283,37 @@ namespace module::purpose {
...
+                }
+            }
+
+            // Store the new positions. Bit messy, but we should not reassign the striker if we become the leader later
+            // We cannot do this by index/same loop as our active_robots and our leader's active_robots may differ
+            for (int i = 0; i < int(active_robots.size()); ++i) {
+                if (active_robots[i].robot_id == striker_id) {
+                    active_robots[i].position = Position("STRIKER");
+                } else {
+                    active_robots[i].position = Position("DEFENDER");
```
ysims, line 316:
```suggestion
            for (int i = 0; i < int(active_robots.size()); ++i) {
                active_robots[i].position = (active_robots[i].robot_id == striker_id) ? Position("STRIKER") : active_robots[i].position = Position("DEFENDER");
            }
```

module/purpose/Soccer/src/Soccer.hpp
```diff
@@ -74,11 +74,14 @@ namespace module::purpose {
             bool force_playing = false;
             /// @brief The soccer position of the robot
             Position position{};
+            /// @brief The number of seconds to wait before assuming a teammate is inactive
+            uint8_t timeout = 5;
```
ysims, line 78:
```suggestion
            uint8_t timeout = 0;
```
init config to 0

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -269,22 +283,37 @@ namespace module::purpose {
...
+            } else {
+                if (soccer_position != Position::DEFENDER) {
+                    emit<Task>(std::make_unique<Defender>(cfg.force_playing));
+                    log<NUClear::INFO>("Robot made defender");
+                }
+            }
+
+            // Store the new positions. Bit messy, but we should not reassign the striker if we become the leader later
+            // We cannot do this by index/same loop as our active_robots and our leader's active_robots may differ
+            for (int i = 0; i < int(active_robots.size()); ++i) {
```
ysims, line 312:
```suggestion
            for (size_t i = 0; i < active_robots.size(); ++i) {
```
something like this should work? or make `i` just whatever type `.size()` gives

module/purpose/Soccer/data/config/webots2/Soccer.yaml
ysims: this still needs updating, and the others

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -84,36 +101,47 @@ namespace module::purpose {
...
-            // This emit starts the tree to play soccer
-            emit<Task>(std::make_unique<FindPurpose>(), 1);
+            // Emit find purpose and create robocup message with necessary info
+            find_purpose();
             // The robot should always try to recover from falling, if applicable, regardless of purpose
             emit<Task>(std::make_unique<FallRecovery>(), 2);
         });
 
-        on<Provide<FindPurpose>>().then([this] {
+        on<Provide<FindPurpose>, Trigger<RoboCup>>().then([this](const RoboCup robocup) {
```
ysims, line 110:
```suggestion
        on<Provide<FindPurpose>, Trigger<RoboCup>>().then([this](const RoboCup& robocup) {
```

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -124,8 +152,153 @@ namespace module::purpose {
                 log<NUClear::INFO>("Force playing started.");
                 cfg.force_playing = true;
                 emit<Task>(std::make_unique<FindPurpose>(), 1);
```
ysims, line 154:
```suggestion
```
since the function does it

ysims (commented):
My recent PR has been merged and will cause minor issues for you. It changes the behaviour flow so that instead of using `Every<x per seconds>` on each Provider to make modules run, we just use one Every in Soccer, specifically on the FindPurpose reaction. 
This might make things nicer though in the end, as you don't need the find_purpose() function anymore. You don't need to make it run cause it'll run anyway 10 times a second. 
Instead of a Trigger<RoboCup>, do a With<RoboCup>. With will just use the last RoboCup message. You won't get every message. If you need every message, make another reaction that triggers on RoboCup and save the information for use in the find purpose.

A-Herfel:
> My recent PR has been merged and will cause minor issues for you. It changes the behaviour flow so that instead of using `Every<x per seconds>` on each Provider to make modules run, we just use one Every in Soccer, specifically on the FindPurpose reaction. This might make things nicer though in the end, as you don't need the find_purpose() function anymore. You don't need to make it run cause it'll run anyway 10 times a second. Instead of a Trigger, do a With. With will just use the last RoboCup message. You won't get every message. If you need every message, make another reaction that triggers on RoboCup and save the information for use in the find purpose.

We don't need every message but I assume that Nuclear doesn't distinguish between the origin. I will need to receive every message, but only save the message from a "leader"

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -127,9 +168,174 @@ namespace module::purpose {
...
+                    it = active_robots.erase(it);
+                    give_directions();
+                } else {
+                    ++it;
+                }
             }
+
+            log<NUClear::DEBUG>("active robots length ", int(active_robots.size()));
         });
+
```
A-Herfel, line 193: ugh

module/purpose/Soccer/data/config/Soccer.yaml
```diff
@@ -10,3 +10,6 @@ position: DYNAMIC
 
 # The number of seconds to wait before assuming a teammate is inactive
 timeout: 8
+
+# The largest robot id number
+max_robots: 4
```
ysims, line 15: I was looking a lot at the game controller yesterday and it appears this is a variable sent by gamecontroller, so we could use that here

PR #1339, opened 2024-04-25
Update odometry and walk engine to use foot down information

Tom0Brien:
This PR updates odometry and allows walk engine to use foot down information.

1. Anchor point for odometry is updated on planted foot phase switch based on sensors
2. Adds option to walk engine to only allow a switch in planted foot if the proceeding foot is planted (based on sensors)


module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -333,24 +341,29 @@ namespace module::input {
...
+                    // Set the z translation, roll and pitch of the anchor frame to 0 as assumed to be on field plane
+                    Hwp.translation().z() = 0;
+                    Hwp.linear() = rpy_intrinsic_to_mat(Eigen::Vector3d(0, 0, mat_to_rpy_intrinsic(Hwp.linear()).z()));
                 }
-                else if (current_walk_phase.value == WalkState::Phase::RIGHT) {
+                else if (current_planted_foot_phase.value == WalkState::Phase::RIGHT) {
                     Hwp = Hwp * sensors->Htx[FrameID::L_FOOT_BASE].inverse() * sensors->Htx[FrameID::R_FOOT_BASE];
+                    // Set the z translation, roll and pitch of the anchor frame to 0 as assumed to be on field plane
+                    Hwp.translation().z() = 0;
+                    Hwp.linear() = rpy_intrinsic_to_mat(Eigen::Vector3d(0, 0, mat_to_rpy_intrinsic(Hwp.linear()).z()));
```
jpptm, line 360: Would it be possible to turn this to a switch case instead? Given that we already use enum values set up, using switch should hopefully make it much faster

module/input/SensorFilter/src/SensorFilter.cpp
```diff
@@ -333,24 +341,29 @@ namespace module::input {
...
                 }
-                // Set the z translation, roll and pitch of the anchor frame to 0 as assumed to be on field plane
-                Hwp.translation().z() = 0;
-                Hwp.linear() = rpy_intrinsic_to_mat(Eigen::Vector3d(0, 0, mat_to_rpy_intrinsic(Hwp.linear()).z()));
             }
 
             // Compute torso pose using kinematics from anchor frame (current support foot)
             Eigen::Isometry3d Hpt = Eigen::Isometry3d::Identity();
-            if (current_walk_phase.value == WalkState::Phase::RIGHT) {
+            if (current_planted_foot_phase.value == WalkState::Phase::RIGHT) {
```
jpptm, line 366: Same as above

module/input/SensorFilter/src/SensorFilter.hpp
```diff
@@ -107,8 +107,8 @@ namespace module::input {
         /// @brief tinyrobotics model of NUgus used for kinematics
         tinyrobotics::Model<double, n_servos> nugus_model;
 
-        /// @brief Current walk phase of the robot
-        WalkState::Phase current_walk_phase = WalkState::Phase::LEFT;
+        /// @brief Current planted foot phase of the robot
+        WalkState::Phase current_planted_foot_phase = WalkState::Phase::LEFT;
```
jpptm, line 111: It might make more sense to make this `WalkState::Phase::DOUBLE` and consider changing to `WalkState::Phase::FEET_DOWN` for more expressiveness as the robot would, by default, be standing on its feet during stand I assume

jpptm (commented):
Some comments

module/input/SensorFilter/src/SensorFilter.hpp
Tom0Brien: This variable is used for tracking and updating the anchor frame, and should never be double, I have renamed things and added comments to try make it clearer, there is probably a nicer way to track and name all of this ...

ysims (approved):

PR #1341, opened 2024-04-27
Add improved playback module

Tom0Brien:
This PR adds a module which allows fast playback of `.nbs` files by emitting messages and skipping idle times between
them. The fast playback module uses TimeTravel to jump the clock forward to either the nearest chrono task or time point of next message in the NBS file.
All the current modes available are:

- `FAST` - Playback as fast as possible
- `REALTIME` - Playback as fast as the messages were recorded
- `SEQUENTIAL`-   Playback one message at a time, waits for all message sub tasks to complete before continuing

The core playback module and the use of the command line are split up into two modules nbs/Player and tools/NBSPlayback

The following steps are taken to start playback from the tools/NBSPlayback

1. Set Playback Mode: Emit a message of type SetModeRequest to establish the playback mode using configuration settings.
2. Load the Files: Emit a message of type LoadRequest to load the files specified in the command line arguments into the playback system and enable message types which want to be played back.
3. Start Playback: Emit a message of type PlayRequest to initiate playback of the loaded files immediately.

This will allow other modules to request the playback of NBS files in a generic manner. For example, for camera calibration or optimization routines.


module/nbs/Player/src/Player.hpp
```diff
@@ -0,0 +1,115 @@
...
+#include "utility/nbs/Decoder.hpp"
+
+namespace module::nbs {
+
+    class Player : public NUClear::Reactor {
+    private:
+        /// The configuration variables for this reactor
+        struct {
+        } config{};
+
```
ysims, line 46:
```suggestion
```

module/support/logging/DataLogging/data/config/DataLogging.yaml
```diff
@@ -2,23 +2,34 @@ log_level: INFO
 
 output:
   directory: recordings
-  split_size: 5 * GiB
+  split_size: 50 * GiB
```
ysims, line 5: Do we want to increase it by this much? The smaller chunks might be good for GDrive but also we could just change this for practice games locally.

TrentHouliston: Definitely keep chunks smaller is better, Honestly I think 5GiB is larger than it should be

Tom0Brien: If we keep this then, I think it would be great if we update recordings to have subfolders for every time its ran, so it's easy to figure out what set of recordings belong together

ysims (approved):

PR #1342, opened 2024-04-29
Fix running with ASan by fixing the directory name

ysims:
ngl kinda confused as to when this broke and why this wasn't a problem before (did we rename the user?)

In the docker container, the user is `nubots`, not `NUbots`, so the log_path to save ASan logs to was wrong.

### DIFF
--- a/tools/run.py
+++ b/tools/run.py
@@ -64,7 +64,7 @@ def run(args, use_gdb, **kwargs):
 
     # Make sure default log path exists
     if use_asan:
-        log_path = os.path.join(os.sep, "home", "NUbots", "NUbots", "log")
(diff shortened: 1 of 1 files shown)


Bidski (approved):
Looks like @TrentHouliston screwed that up :stuck_out_tongue: 
https://github.com/NUbots/NUbots/blame/cf822b8a3ed731032eb926ca3aa81233944e99fc/tools/run.py

ysims:
> Looks like @TrentHouliston screwed that up 😛 https://github.com/NUbots/NUbots/blame/cf822b8a3ed731032eb926ca3aa81233944e99fc/tools/run.py

Oh lol I was confused cause that was 10 months ago and I've used ASan since then, but I must have had the log folder then so the error didn't come up.

PR #1344, opened 2024-05-03
Add udev rule that symlinks all instances of /dev/ttyACM to /dev/nusense and use this symlink in HardwareIO/NUSense

JohanneMontano:
Sometimes, nusense will toggle from `/dev/ttyACM0` to `/dev/ttyACM1` and since the port config value in the code cannot be changed during run time, it will break the communication line between nusense and the nuc.

This pull request adds code in the arch install script to make the udev rule file. It aims to match all instances of `/dev/ttyACM `and creates a symlink `/dev/nusense`. This way, the cpp code can just have `/dev/nusense` as its port config and the communications line between the nusense and the nuc will not break when ACM toggles from whatever the config value was to something else.


JohanneMontano:
Since we don't want to reflash all the NUCs with new arch installs, this should do for the time being:

```
cat << EOF | sudo tee -a /etc/udev/rules.d/10-nusense.rules
KERNEL=="ttyACM*", SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ATTRS{manufacturer}=="STMicroelectronics", ATTRS{idProduct}=="5740", ATTRS{idVendor}=="0483", SYMLINK+="nusense", GROUP="uucp"
EOF
```

or make the file using `sudo nano`. At this point in time I'm not sure whether the idVendor and idProduct fields are specific to a certain nusense device and I can't confirm it today since I am at home, hence I marked the pull request as draft.

ysims:
on Sarah who has a nusense board:
```
idVendor           0x0483 STMicroelectronics
idProduct          0x5740 Virtual COM Port
```

TrentHouliston:
Do not put this in the install script. Unless you like reinstalling everything any time you have a change it should go in system configuration. That's what it's for!

TrentHouliston:
Also if each NUSense is being different it's likely something you need to set in the NUSense firmware

jpptm:
> on Sarah who has a nusense board:
> 
> ```
> idVendor           0x0483 STMicroelectronics
> idProduct          0x5740 Virtual COM Port
> ```

I probably swapped it around

JohanneMontano:
@ysims what command did you use? According to https://superuser.com/questions/1106247/how-can-i-get-the-vendor-id-and-product-id-for-a-usb-device it is shown in pairs of VID:PID and my output is this:

`Bus 003 Device 002: ID 0483:5740 STMicroelectronics Virtual COM Port`

JohanneMontano:
```
nubots@nugus2 ~ ❯❯❯ ls -lah /dev/nusense                                   ✘ 2 
lrwxrwxrwx 1 root root 7 May 16 10:36 /dev/nusense -> ttyACM0
nubots@nugus2 ~ ❯❯❯ ls -lah /dev/ttyACM0
crw-rw-rw- 1 root uucp 166, 0 May 16 10:36 /dev/ttyACM0
nubots@nugus2 ~ ❯❯❯ 
```

JohanneMontano:
Tested and the symlink appears, keyboardwalk works and HwIO StreamReactor says that it connected successfully to /dev/nusense.

doc/ArchInstall/arch-chroot_install.sh
```diff
@@ -184,6 +184,13 @@ systemctl enable systemd-resolved.service
 systemctl enable wpa_supplicant
 systemctl enable wpa_supplicant@${WIFI_INTERFACE}
 
+#####################
+# SERIAL PORT RULES #
+#####################
+cat << EOF > /etc/udev/rules.d/10-nusense.rules
+KERNEL=="ttyACM*", SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ATTRS{manufacturer}=="STMicroelectronics", ATTRS{idProduct}=="5740", ATTRS{idVendor}=="0483", SYMLINK+="nusense", GROUP="uucp", MODE="0666"
+EOF
+
```
Bidski, line 193: Remove this. You are adding a config file to `SystemConfiguration`, that is a much better option than the OS install script

module/tools/SystemConfiguration/data/system/default/etc/udev/rules.d/10-nusense.rules
```diff
@@ -0,0 +1 @@
+KERNEL=="ttyACM*", SUBSYSTEM=="tty", SUBSYSTEMS=="usb", ATTRS{manufacturer}=="STMicroelectronics", ATTRS{idProduct}=="5740", ATTRS{idVendor}=="0483", SYMLINK+="nusense", GROUP="uucp", MODE="0666"
```
Bidski, line 1: What is the mode before you set it here?

Bidski (changes requested):

module/tools/SystemConfiguration/data/system/default/etc/udev/rules.d/10-nusense.rules
JohanneMontano: it did not have a mode specified, I added this in as a new thing

Bidski: I mean when you just plug the device in without the udev rules file, what is the mode that is automatically assigned to `/dev/ttyACM0`?

JohanneMontano: tried to mimic what `10-nubots.rules` had

JohanneMontano: not sure what the c maps to but 

```
nubots@nugus2 ~ ❯❯❯ ls -lah /dev/ttyACM0                                                                                                                                                             ✘ 130 
crw-rw-rw- 1 root uucp 166, 0 May 16 11:04 /dev/ttyACM0
```
which is something - 666 afaik?

TrentHouliston: Well if you have that udev rule in now it'll always be 666

I'd imagine 660 was what it was before but can't be sure

Bidski: sounds like you dont need to set the mode at all then?

TrentHouliston: Given that previously it was accessible by root and the uucp group and not everyone that would mean the last number must have been 0 but the first two must have been 6s

Bidski: And if it was 660 you also dont need to set the mode

JohanneMontano: works :+1:

Bidski (approved):

TrentHouliston (approved):
This is a good length PR

PR #1345, opened 2024-05-03
Add RoboCup Configuration Tool

ysims:
We were after a tool that makes things easy before a RoboCup game to set the wifi network and anything else. 

Last year we had issues with the robot position not being set (eg Kevin thinking he was a  goalie and diving everywhere) and the robot set to wrong field (although we were mostly saved by @TrentHouliston here). We also didn't take advantage of the robocup service, and we probably left our robots connected to a field network when they shouldn't have been. 

A lot of this is because people aren't aware of what needs to be done (robocup service was unknown) or how to do the things (how to configure networks???), and also _stress_. 

So this tool is an ncurses interface that shows you at a glance what your setup is for network and some game config so you can easily see if anything is wrong. It will make it easy to switch over to the right field, and set the robot up with the right player number. The readme covers a lot of the functionality and how to use.

### Things left to do

- [x] Add functionality to disconnect from wifi networks, so we don't remain connected to a field when not playing
- [ ] NUbook page


module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
```diff
@@ -0,0 +1,489 @@
...
+                "be the player number. ";
+        }
+
+        // Write the new ip address to the file
+        std::ofstream(fmt::format("{}/30-wifi.network", systemd_folder))
+            << fmt::format("[Match]\nName={}\n\n[Network]\nAddress={}/16\nGateway={}.{}.3.1\nDNS=8.8.8.8",
+                           wifi_interface,
+                           ip_address,
+                           ip_parts[0],
+                           ip_parts[1]);
```
ysims, line 267: I copied the existing files so idk if there's things here that need to be dynamically changed too.

module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
```diff
@@ -0,0 +1,489 @@
...
+        system(("mkdir -p " + wpa_supplicant_dir).c_str());
+
+        std::filesystem::copy_file(
+            fmt::format("system/default/etc/wpa_supplicant/wpa_supplicant-{}.conf", wifi_interface),
+            wpa_supplicant_file,
+            std::filesystem::copy_options::overwrite_existing);
+
+        // Append to the wpa_supplicant file with a high priority
+        std::string command =
+            "wpa_passphrase " + ssid + " " + password + " | sed 's/}/\tpriority=9999\\n}/' >> " + wpa_supplicant_file;
```
ysims, line 287: So this priority thing is yikes but I'm not sure how to make it nicer. Maybe parse the file and find the highest priority number. Practically this is probably fine I doubt anyone is setting a higher priority than this but still.

ysims:
Ah also this uses config from systemconfiguration, soccer, striker/defender/goalie but they're not in the role bc well systemconfiguration would start running, so I've just said in the readme to build those too. It's unlikely to be a problem because systemconfig and robocup role will be built at some point to the robot before this is used.

module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
```diff
@@ -121,6 +121,14 @@ namespace module::tools {
                     system("systemctl stop robocup");
                     log_message = "RoboCup service stopped!";
                     break;
+                case 'E':  // enables wifi services
+                    system("systemctl enable wpa_supplicant");
```
JesseWilliamson, line 125: `enable` enables a service at startup. Use `enable --now` to enable on startup and start immediately. Same with disable.

module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
```diff
@@ -0,0 +1,506 @@
...
+                case 'S':  // starts the robocup service
+                    system("systemctl start robocup");
+                    log_message = "RoboCup service started!";
+                    break;
+                case 'D':  // stops the robocup service
+                    system("systemctl stop robocup");
+                    log_message = "RoboCup service stopped!";
+                    break;
+                case 'E':  // enables wifi services
+                    system("systemctl enable wpa_supplicant");
```
ysims, line 125:
```suggestion
                    system("systemctl enable --now wpa_supplicant");
```

module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
```diff
@@ -0,0 +1,506 @@
...
+                    system("systemctl enable systemd-networkd");
```
ysims, line 126:
```suggestion
                    system("systemctl enable --now systemd-networkd");
```

module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
```diff
@@ -0,0 +1,506 @@
...
+                    break;
+                case 'W':  // disables wifi services
+                    system("systemctl disable wpa_supplicant");
+                    system("systemctl disable systemd-networkd");
```
ysims, line 130:
```suggestion
                    system("systemctl disable --now wpa_supplicant");
                    system("systemctl disable --now systemd-networkd");
```

module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
```diff
@@ -0,0 +1,506 @@
...
+        system("systemctl restart systemd-networkd");
+        system("systemctl restart wpa_supplicant");
+        system(("systemctl restart wpa_supplicant@" + wifi_interface).c_str());
+
+        log_message = "Network configured!";
+    }
+
+    void RoboCupConfiguration::set_config_values() {
+        /* GAME CONFIG */
+        {  // Write the robot's position to the soccer file
```
Tom0Brien, line 217: Why scope?

module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
```diff
@@ -0,0 +1,506 @@
...
+                case 3:  // team_id
+                    team_id = team_id == 33 ? 1 : team_id + 1;
+                    break;
+            }
+            return;
+        }
+        // Game configuration column
+        if (row_selection == 0) {
+            ++robot_position;
+            {
```
Tom0Brien, line 317: Why scope?

shared/utility/support/yaml_expression.hpp
```diff
@@ -321,6 +321,52 @@ namespace YAML {
             return true;
         }
     };
+
+    template <int Rows, int Cols>
+    struct convert<Eigen::Matrix<double, Rows, Cols>> {
```
Tom0Brien, line 326: Are these being used anywhere?

Tom0Brien (changes requested):
Small things

module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
Tom0Brien: Probs fine

module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
ysims: so the config and file variables are contained and don't exist outside

module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
```diff
@@ -0,0 +1,506 @@
...
+                // Get ready position for this position
+                std::string ready_file = get_config_file(robot_position.get_config_name());
+                YAML::Node config      = YAML::LoadFile(ready_file);
+                ready_position         = config["ready_position"].as<Eigen::Vector3d>();
+            }
```
ysims, line 322:
```suggestion

            // Get ready position for this position
            std::string ready_file = get_config_file(robot_position.get_config_name());
            YAML::Node config      = YAML::LoadFile(ready_file);
            ready_position         = config["ready_position"].as<Eigen::Vector3d>();
```

shared/utility/support/yaml_expression.hpp
ysims: encoding and decoding the ready position

Tom0Brien (approved):

Tom0Brien (approved):

PR #1346, opened 2024-05-03
Add aliases for nugusX and robot names to associate with hardware

ysims:
Some config isn't so much NUC-specific as it is hardware-specific. Kevin can have it's NUC swapped so it goes from nugus2 to nugus1, but the config will no longer match as it should really be Kevin's config, not nugus2's config. This means anytime NUCs are swapped around, we need to swap all the configs around (which can be a bit confusing). 

This PR creates aliases for the hostnames. I.e. nugus1 can be mapped to "Kevin". These aliases/robot names can then be used in the config and script system, so we can load in config and scripts from folders with robots names. When NUCs are swapped around, only the `robot_names.yaml` needs to be updated, which is much less confusing and quicker, particularly if it happens at RoboCup.

I've also changed folder names to robot names in cases where the config is tied to the hardware. This isn't always the case, ie system config files and purpose config is pretty NUC-specific not hardware-specific. 

I removed Sarah's walk so she'll use the default since Sarah's walk seems to be the x-series legs walk, which probably won't work on her. I don't think she has a tuned walk, or if she has it's back in the GitHub history. 

Also changed some camelCase to snake_case.

- [ ] Updated NUbook (add link to NUbook PR here) - needs doing


jpptm:
build is failing because fileName hasnt been changed to file_name in Webots, but otherwise this lgtm.

ysims:
@JohanneMontano CI passing now

JohanneMontano (approved):
lgtm

PR #1348, opened 2024-05-04
Slight tune to Kevin's HardwareIO and tunes SplineKick on Kevin

ysims:
Kevin's arms have been out for a while which affects the getup. Hip change was to align the feet better to improve the mirroring on the kick.


Crafty15 (approved):
LGTM

PR #1352, opened 2024-05-13
Fix name config (again)

ysims:
And simplify the function, as it was originally designed to return the map but we don't need that.


JohanneMontano (approved):
lgtm

PR #1359, opened 2024-05-25
Shuffle around roles

Tom0Brien:
This PR does the following:

- moves `hardwareio` to bottom of the roles to improve connection reliability. 
- adds `yolo` to `robocup` role
- creates new role `test/robocup` with networking stuff useful for testing robocup in the lab


roles/test/robocup.role
```diff
@@ -0,0 +1,69 @@
+# This role is used for running the full autonomous soccer set up with a real robot. It should be used with a computer
```
ysims, line 1: whats the difference between this and robocup?

Tom0Brien: Robocup doesn't have Networking stuff which we always need to add when testing, maybe we could just have all of that in the main robocup

ysims (dismissed):

PR #1364, opened 2024-06-10
Add error and fault handling for NUSense.

Claegtun:
Adds some more tailored handling and raising of errors for NUSense.

- Logs the count of packet-errors, CRC-errors, and timeouts.
- Logs a dispatch as a message from NUSense for fleeting errors; this is a generic error-field.

### Pre-PR Checklist:

Have you

### Reviewers, Note

Work in progress.

### Things left to do

- [x] Fix the null-terminator in the string.

### DIFF
--- a/module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
+++ b/module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
@@ -119,6 +119,7 @@ namespace module::platform::NUSense {
             // Battery data
             sensors->battery = 0;  // not yet implemented
 
+            // Log the latest dispatch from NUSense; only very rare exceptions should be dispatched.
             log<NUClear::WARN>("Dispatch from NUSense: " + std::string(data.dispatch.data(), 5) + " £££££££\n");
 
             // Servo data
@@ -148,6 +149,7 @@ namespace module::platform::NUSense {
                 servo.profile_acceleration  = 0;  // not present in NUSense message
                 servo.profile_velocity      = 0;  // not present in NUSense message
 
+                // Log any errors and timeouts from the servo.
                 if (val.num_errors != 0) {
                     log<NUClear::WARN>(fmt::format("{} packet-error(s) from ID {} ({})",
                                                    val.num_errors,
@@ -155,6 +157,8 @@ namespace module::platform::NUSense {
                                                    nugus.device_name(static_cast<NUgus::ID>(val.id))));
                 }
                 if (val.num_crc_errors != 0) {
+                    // For now, the CRC is set to debug until terminating-resistors are gotten since there are many when
+                    // the robot is walking.
                     log<NUClear::DEBUG>(fmt::format("{} CRC-error(s) from ID {} ({})",
                                                     val.num_crc_errors,
                                                     val.id,
--- a/module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
+++ b/module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
@@ -120,7 +120,7 @@ namespace module::platform::NUSense {
             sensors->battery = 0;  // not yet implemented
 
             // Log the latest dispatch from NUSense; only very rare exceptions should be dispatched.
-            log<NUClear::WARN>("Dispatch from NUSense: " + std::string(data.dispatch.data(), 5) + " £££££££\n");
+            log<NUClear::WARN>("Dispatch from NUSense: " + data.dispatch);
 
             // Servo data
             for (const auto& [key, val] : data.servo_map) {
--- a/module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
+++ b/module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
@@ -120,7 +120,9 @@ namespace module::platform::NUSense {
             sensors->battery = 0;  // not yet implemented
 
             // Log the latest dispatch from NUSense; only very rare exceptions should be dispatched.
-            log<NUClear::WARN>("Dispatch from NUSense: " + data.dispatch);
+            if (data.dispatch.size() > 3) {
+                log<NUClear::WARN>("Dispatch from NUSense: " + data.dispatch);
+            }
 
             // Servo data
             for (const auto& [key, val] : data.servo_map) {
--- a/module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
+++ b/module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
@@ -119,6 +119,8 @@ namespace module::platform::NUSense {
             // Battery data
             sensors->battery = 0;  // not yet implemented
 
+            log<NUClear::WARN>("Dispatch from NUSense: " + std::string(data.dispatch.data(), 5) + " £££££££\n");
+
             // Servo data
             for (const auto& [key, val] : data.servo_map) {
                 // Get a reference to the servo we are populating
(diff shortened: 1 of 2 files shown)


module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -119,6 +119,9 @@ namespace module::platform::NUSense {
             // Battery data
             sensors->battery = 0;  // not yet implemented
 
+            // Log the latest dispatch from NUSense; only very rare exceptions should be dispatched.
+            log<NUClear::WARN>("Dispatch from NUSense: " + std::string(data.dispatch.data(), 5) + " £££££££\n");
```
Claegtun, line 123: Ignore this for now. Testing a few things.

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -119,6 +119,9 @@ namespace module::platform::NUSense {
             // Battery data
             sensors->battery = 0;  // not yet implemented
 
+            // Log the latest dispatch from NUSense; only very rare exceptions should be dispatched.
+            log<NUClear::WARN>("Dispatch from NUSense: " + data.dispatch);
```
Claegtun, line 123: Would this be better as `DEBUG` instead?

Tom0Brien: WARN or INFO is best, unless you want to specify log level on the NUsense side and display accordingly here. WARN is probably what we want though for anything coming from subcontroller

Tom0Brien (approved):

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
jpptm, line 123: Does this log every time we trigger on NUSense?

shared/message/platform/NUSenseData.proto
```diff
@@ -102,4 +110,5 @@ message NUSense {
      */
     map<uint32, Servo> servo_map = 1;
     IMU                imu       = 2;
+    string             dispatch  = 3;
```
jpptm, line 113: Can I have a little bit of docs here please? IMO servo_map and imu is quite intuitive but what is dispatch supposed to be? Do we really want to log this message for every single NUSense packet

shared/message/platform/NUSenseData.proto
```diff
@@ -62,6 +62,14 @@ message Servo {
     float voltage = 12;
     /// The last read temperature of the servo
     float temperature = 13;
+    /// The number of successes.
+    uint32 num_successes = 14;
```
jpptm, line 66: I don't see this used anywhere in these changes. Does this get used in the nusense firmware? Also, do we specifically need a proto field for the number of successes? This can still be tracked by saying that a packet that did not return any error means that it succeeded sending the commands to the servos and it can just be tracked by a uint32 variable in the header file of hwio.

Claegtun: Sorry, I should have linked this which is the sister PR. This won't be merged until the other PR is approved as well. https://github.com/NUbots/NUcontroller/pull/16

The number of successes is just in case we want to infer the total number of packets, i.e. `num_success + num_crc_errors + num_errors`, in order to maybe use a percentage, e.g. "5% of packets received had a CRC error". 

Also, keep in mind that HardwareIO does not know anything about individual dynamixel packets. For every `NUSenseData` packet sent along USB, there could be at least a dozen dynamixel packets for each servo. So, the only way to infer packet-statistics is to log them on NUSense and send them over.

shared/message/platform/NUSenseData.proto
Claegtun: We could make it its own standalone packet, but for now, I want it to at least work with what we have now which is only the one NUSense packet.

The dispatch a generic string message for uncommon errors and exceptions, e.g. nanopb.

shared/message/platform/NUSenseData.proto
Claegtun: Another thing for that it could be used is to infer and to monitor the dynamixel throughput.

shared/message/platform/NUSenseData.proto
Claegtun: Added a comment.

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
Claegtun: Fixed so that it only logs when it has at least three characters.

ysims (dismissed):
lgtm

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -119,6 +119,11 @@ namespace module::platform::NUSense {
             // Battery data
             sensors->battery = 0;  // not yet implemented
 
+            // Log the latest dispatch from NUSense; only very rare exceptions should be dispatched.
+            if (data.dispatch.size() > 3) {
+                log<NUClear::WARN>("Dispatch from NUSense: " + data.dispatch);
+            }
+
```
TrentHouliston, line 126: I think this should just be a separate message, rather than baked into the mega message. It's making that message very hard to work with.

For any exceptional error, just send a "log" message of some kind, then you can on trigger it in here and do something with it

shared/message/platform/NUSenseData.proto
TrentHouliston: Do you want to count the number of successes? or the total number of packets?
Also I'd make these comments for this and errors that it is just the number of errors since the last packet, so it doesn't grow over time.

shared/message/platform/NUSenseData.proto
```diff
@@ -62,6 +62,14 @@ message Servo {
...
+    /// The number of timeouts.
+    uint32 num_timeouts = 15;
+    /// The number of CRC-errors.
+    uint32 num_crc_errors = 16;
+    /// The number of packet-errors.
+    uint32 num_errors = 17;
```
TrentHouliston, line 72: Might be a good idea to wrap these four fields in a submessage, they're all closely related and doing one task together

shared/message/platform/NUSenseData.proto
TrentHouliston: I wouldn't muddy this with an extra field for this. If you want to have specific log messages, make a new log message type and send that

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
Claegtun: Will remove the `dispatch` field and move it to another message in a different PR.

shared/message/platform/NUSenseData.proto
Claegtun: Will act on these suggestions soon.

shared/message/platform/NUSenseData.proto
Claegtun: Will remove the `dispatch` field and move it to another message in a different PR.

shared/message/platform/NUSenseData.proto
Claegtun: Will change to total number instead.

shared/message/platform/NUSenseData.proto
```diff
@@ -34,6 +34,18 @@ package message.platform;
 // Message mainly for NUSense <-> NUC communication. NUSense will encode this message via nanopb after it queries all
 // servo states and send it to the NUC.
 message Servo {
+
+    message PacketCounts {
+        /// The number of successes.
```
jpptm, line 39: does this comment need an update?

shared/message/platform/NUSenseData.proto
```diff
@@ -34,6 +34,18 @@ package message.platform;
...
+        uint32 total = 1;
+        /// The number of timeouts.
+        uint32 timeouts = 2;
+        /// The number of CRC-errors.
+        uint32 crc_errors = 3;
+        /// The number of packet-errors.
+        uint32 errors = 4;
```
jpptm, line 46: What's the difference between a crc error and a packet error?

```suggestion
        /// The number of CRC-errors.
        uint32 crc_errors = 3;
        /// The number of packet-errors.
        uint32 packet_errors = 4;
```

jpptm (commented):
Last few nitpicks but other than this, this lgtm

shared/message/platform/NUSenseData.proto
Claegtun: A CRC error is when the CRC of the received packet does not match the computed CRC, i.e. the received packet has been corrupted.

A packet-error is when the received packet has an error-code in its error-field which could mean a few things depending on the context, out-of-range data-length, inaccessible address, attempting to write to a read-only register, a mismatched CRC on the servo's side, etc.

shared/message/platform/NUSenseData.proto
Claegtun: Done.

jpptm (dismissed):
lgtm

jpptm (approved):

PR #1370, opened 2024-06-17
Ball intercept goalie behaviour

Crafty15:
Bounded box strategy added to goalie behaviour. Basic pattern is that it will try to stay close to the ball while remaining in the goalie box. If the ball comes into the box it tries to walk it out.  [PR#1378](https://github.com/NUbots/NUbots/pull/1378) adjusts the bounded box strategy so that the robots return to a configured start position if they lose the ball.

- Adds the bounded box strategy to the goalie.
- Adds walk to ball strategy to the goalie.
- Changes `LookAround` to `LookForBall`. 
   - Otherwise robot could get stuck looking around if the ball ends up behind it.
- Removes dive to ball. 
  - I think this is a lot of potential damage for little gain considering the dive is pretty slow to react anyway.
- Made distance behind ball configurable. Reducing this for the goalie helps it stay between the ball and the goal when the ball is beside the goalie box.

### Reviewers, Note

Thoughts on if we should kick the ball when we can or play it safe? The real goalie robot will likely be the one without the x-series leg upgrade. I suppose it will depend on if we can get it kicking reliably.

### Things left to do

- [ ] Test on real robot


ysims:
Probably won't get Sarah kicking reliably. It isn't hard to add in the kick if we do put an x series robot in goalie. Could also make it a config variable as this is a common thing we change.

PR #1372, opened 2024-06-18
Update OpenCR config to use 3Mbps baudrate

ysims:
- Changed config for NUSense and OpenCR to have 3Mbps baud rate, although the value for NUSense isn't used
- Removed general fields from robot-specific config
- Copied OpenCR config to NUSense so the offsets are the same on both, and so NUSense uses robot-name config rather than hostname config.
- Switched sarah and frankie around on the name config as the nucs were swapped
- Fixed compile error in NUSense HardwareIO (not currently in CI, needs adding)

NUbots/NUcontroller#17


TrentHouliston (commented):
Nusense doesn't care about your baud rate settings! The baud rate for it is set by the firmware. That said. Maybe it would be a good idea for nusense to set the baud rate to the servos based on the baud rate set on the ACM device (even though it doesn't change the communication rate to the nusense itself)

ysims:
> Nusense doesn't care about your baud rate settings! The baud rate for it is set by the firmware. That said. Maybe it would be a good idea for nusense to set the baud rate to the servos based on the baud rate set on the ACM device (even though it doesn't change the communication rate to the nusense itself)

True lol maybe I should go through and remove any of those we don't need.

ysims:
I'll leave the baud rate in as it is used in the code, although I don't think the value matters.

Claegtun (approved):
Seems good to me.

PR #1374, opened 2024-06-22
Add a soft start module so the robot doesn't jerk to position dangerously

ysims:
The first thing the robot does generally is to run a zero walk command, causing it to jump very quickly to the standing position. This is a bit dangerous and does startle people who aren't expecting it. 

This PR adds in a module that takes control of the servos to begin with and moves the servos to the stand position over a few seconds at a low gain. The module will emit Done when it is happy, and as it is a root Task this will remove it from the tree. Unless it's requested again, it won't be active again, and as it's only in Startup reactions this won't happen.

Could be made a bit neater with scripts if the servos providers used position to determine done, but had some issues with it before.


module/strategy/StartSafely/src/StartSafely.cpp
```diff
@@ -0,0 +1,92 @@
...
+            cfg.servo_targets[10] = config["right_hip_pitch"].as<double>();
+            cfg.servo_targets[11] = config["left_hip_pitch"].as<double>();
+            cfg.servo_targets[12] = config["right_knee"].as<double>();
+            cfg.servo_targets[13] = config["left_knee"].as<double>();
+            cfg.servo_targets[14] = config["right_ankle_pitch"].as<double>();
+            cfg.servo_targets[15] = config["left_ankle_pitch"].as<double>();
+            cfg.servo_targets[16] = config["right_ankle_roll"].as<double>();
+            cfg.servo_targets[17] = config["left_ankle_roll"].as<double>();
+            cfg.servo_targets[18] = config["head_yaw"].as<double>();
+            cfg.servo_targets[19] = config["head_pitch"].as<double>();
```
JohanneMontano, line 51:
```
for (uint8_t i  = 0; i<config["servo_targets"].size(); ++i) {
    const auto& servo_name = config[i].first; // For the sake of clarity
    const auto& servo_position = config[i].second;
    cfg.servo_targets[i] = servo_position;
}
```

then servo_targets is just a sequence of map entries, or alternatively something that will make this code a lot shorter

TrentHouliston (commented):
I would suggest just running the stand script instead.
That way the "default" position for the safe start will always be the same as any other stand in the system.

So even if you were not running a walk and instead were just doing stand this would still be safe

JohanneMontano (approved):
Changes have been addressed afaik, lgtm. Might be worth checking with @TrentHouliston  again just to be sure.

Tom0Brien:
We should probably remove the soft start code in the `FootController` alongside this PR

ysims:
Webots is running fine but I'll need to test again on a real robot as I was having issues last time with seg faulting on `keyboardwalk`.

roles/webots/demo.role
```diff
@@ -41,6 +41,7 @@ nuclear_role(
   planning::GetUpPlanner
   planning::FallingRelaxPlanner
   # Strategies
+  trategy::StartSafely
```
ysims, line 44:
```suggestion
  strategy::StartSafely
```

ysims:
This should be good to go, no issues when I was running it in webots last night or the real robot this morning.

JohanneMontano (approved):

PR #1375, opened 2024-06-22
Revise OpenCR's response handler for byte unstuffing and formal function declarations

JohanneMontano:
This pull request adds byte unstuffing for when we read data from opencr. When the sequence 0xFF, 0xFF, 0xFD is seen in the read buffer and the next byte after the sequence is 0xFD, the script ignores this byte (because it is the stuffing as per the dynamixel protocol) and does not add it to the response buffer.


module/platform/OpenCR/HardwareIO/src/hardwareio/handle_response.cpp
```diff
@@ -167,12 +177,59 @@ namespace module::platform::OpenCR {
...
                     }
                     else {
                         current_phase = Phases::TIMEOUT;
                     }
                     break;
 
+                // Going to the case below means that we saw a 0xFF in the data field
+                // If the next byte is another 0xFF, then we go to UNSTUFF_2
+                case Phases::UNSTUFF_1:
+                    current_phase = buf[i] == packet_header[1] ? Phases::UNSTUFF_2 : Phases::DATA;
```
dex-k, line 193: I think we might need the timeout check here (and in other unstuffing phases)

module/platform/OpenCR/HardwareIO/src/hardwareio/handle_response.cpp
```diff
@@ -167,12 +177,59 @@ namespace module::platform::OpenCR {
...
+                    }
+                    break;
+
+                // Going to the case below means that we've seen our 0xFFFFFD in the data field
+                // If the next byte is 0xFD, ignore it and go back to the DATA phase
+                case Phases::UNSTUFF_3:
+                    if (buf[i] != packet_header[2]) {
+                        response.push_back(buf[i]);
+                    }
+                    // If we decide to ignore a packet we have to decrement the packet length
```
Claegtun, line 220:
```suggestion
                    // If we decide to ignore a packet we have to decrement the packet length
                    // If we decide to ignore a stuffed byte, we have to decrement the packet length.
```

JohanneMontano: Added

module/platform/OpenCR/HardwareIO/src/hardwareio/handle_response.cpp
JohanneMontano: I revised the fsm so we dont have to write as much code. Now we only have to check for timeouts in one data case instead of many like in here

Tom0Brien (approved):

ysims (approved):

PR #1377, opened 2024-06-25
Have root set the frequency update for behaviour providers

ysims:
Our behaviour providers need to run at some frequency to check if things change, often time based things. Currently we've been doing this per-Provider but then if a parent and child are both running eg 10 times a second, this will compound for the child if the parent is triggering the child each time. The PR just makes the root update 10 times/second. 

Things wouldn't run if idle'd, but thats only really used lower down. Main motivation is for walk path planner to have reasonable dt for new method.

Tested in webots.


Tom0Brien (approved):
LGTM

PR #1378, opened 2024-06-25
Improve WalkInsideBoundedBox strategy and set areas of the field for players to stay within

Crafty15:
This PR adds the following

### Ball search timeout
- Added a ball search timeout to the bounded box strategy so that the robot will return to a configured field position if it loses track of the ball. 
- Most of the time it will find the lost ball while returning to the field position. 
- If the robot doesn't find the ball it can remain lost behind the robot toward own goal. This is due to the robot trying to face the opponents goal, so doesn't always turn enough to face the ball, but rather strafes back to the position. 

### Allow pass walk path planning
If the ball is between the bounding box and the robots own goal, an offset is added in the y direction to allow team mate to pass ball 

### Adds bounding box to all purposes
Striker, defender and goalie are all bounded to a box on the field to avoid self collisions


module/purpose/Goalie/src/Goalie.cpp
```diff
@@ -144,15 +155,19 @@ namespace module::purpose {
     }
 
     void Goalie::play() {
-        // Stop the ball!
+        // Walk to the ball and kick!
```
ysims, line 158:
```suggestion
        // Walk to the ball if within our section of the field
```

module/purpose/Goalie/src/Goalie.cpp
```diff
@@ -144,15 +155,19 @@ namespace module::purpose {
     }
 
     void Goalie::play() {
-        // Stop the ball!
+        // Walk to the ball if within our section of the field
```
JohanneMontano, line 158:
```suggestion
        // Walk to the ball if it is within our section of the field
```

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -316,6 +334,72 @@ namespace module::purpose {
         else {
             emit<Task>(std::make_unique<Defender>(cfg.force_playing));
         }
+
+        // Update bounding box based on the position of the robot and existence of other robots
+        bool defender_exist = std::count_if(robots.begin(),
+                                            robots.end(),
+                                            [](const RobotInfo& robot) { return robot.position == Position::DEFENDER; })
+                              == 1;
```
JohanneMontano, line 342:
```suggestion
        bool defender_exist = std::count_if(robots.begin(),
                                            robots.end(),
                                            [](const RobotInfo& robot) { return robot.position == Position::DEFENDER; })
                              >= 1;
```

module/purpose/Striker/README.md
```diff
@@ -6,7 +6,7 @@ Play soccer in the striker position, which is an offensive position. Either bypa
 
 In the ready state, makes a Ready subtask.
 
-In the playing state (for both normal game mode and penalty shootout), finds the ball, fixates on the ball, walks to the ball, and kicks the ball.
+In the playing state (for both normal game mode and penalty shootout), finds the ball, fixates on the ball, walks to the ball, and kicks the ball. It will not approach the ball if it is not in the front half, and instead will stand nearby.
```
JohanneMontano, line 9:
```suggestion
In the playing state (for both normal game mode and penalty shootout), finds the ball, fixates on the ball, walks to the ball, and kicks the ball. It will not approach the ball if it is not in its own bounding box, and instead will stand nearby.
```

JohanneMontano (changes requested):

JohanneMontano (approved):

PR #1379, opened 2024-06-26
Vision heuristics with YOLO + VMesh (mostly balls)

ysims:
- Tune the VMesh Ball Detector to focus more on seeing balls further away. Upped the min distance so it ignores balls close by. Upped the cluster size so small 'balls' don't get through, as small balls are thought to be far away.
- Tune the YOLO detector to allow through balls at a much lower confidence (this helped with balls on field lines and paler balls)
- Added in heuristic to YOLO detector to not allow balls, robots and field intersections if they are outside of the green horizon. 
- Tuned the green horizon to allow lower confidence field points, as it was not covering the whole field sometimes. Helped a lot.
- Removed the check for if the robot is in the green horizon in the RobotLocalisation module as this is already done in the Yolo module. Had to keep in one check and the horizon message as it is used in the logic for if robots go out of view. A new method for this, ie based on fov and image size as per todo comment, will allow us to remove the greenhorizon here entirely.
- Updated GreenHorizonDetector to not just take the closest cluster. Instead it merges clusters if they are close enough to the robot. We try to set this value quite low, as any viable cluster is going to have its closest point very close to the robot. The field is unlikely to be split up horizontally, but more so vertically, therefore the base of the field cluster should be at the bottom of the camera. Tested also with looking as far up as the robot can look, and it still keeps the clusters. There are some strong green patches in the mesh around the field and these stay excluded. Tested with some chairs in the way to vertically split up the field as if robots were there, and it created a good convex hull over the field.

### DIFF
--- a/module/localisation/RobotLocalisation/src/RobotLocalisation.cpp
+++ b/module/localisation/RobotLocalisation/src/RobotLocalisation.cpp
@@ -67,11 +67,8 @@ namespace module::localisation {
                     // Position of robot {r} in world {w} space
                     auto rRWw = Hwc * vision_robot.rRCc;
 
-                    // Only consider vision measurements within the green horizon
-                    if (point_in_convex_hull(horizon.horizon, rRWw)) {
-                        // Data association: find tracked robot which is associated with the vision measurement
-                        data_association(rRWw);
-                    }
+                    // Data association: find tracked robot which is associated with the vision measurement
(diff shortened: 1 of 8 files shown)


Tom0Brien (dismissed):
We may want an option to turn off green horizon check, but lgtm!

Tom0Brien (approved):

PR #1389, opened 2024-07-02
NUsight Director View

JesseWilliamson:
This PR adds an NUsight page which displays a visual overview of director for debugging.
image


ysims:
It might be good to make active green and I always think of done as being blue for some reason but it doesn't matter too much. I was also thinking done could be a different shape, like a triangle? or maybe the border could be it's activeness and the infill the doneness.

JesseWilliamson:
image
Something like this? I had a look at sending parent with each item so I could render them as a tree, but I'm worried about the performance cost of sending more info than we are already. I'm also not sure what Director information is most useful.

TrentHouliston:
The director tree shouldn't be that much information compared to everything else that's being sent. You should be able to send almost everything and still come close to causing a bandwidth/rendering issue.

That is so long as you only send information when the director tree changes

PR #1393, opened 2024-07-02
Add Idle state to Soccer which can be toggled with back panel buttons

Tom0Brien:
This PR adds an Idle state to the Soccer module which can be enabled by pressing the left black button. Regular soccer mode can be resumed with the green button. When resuming (green button), localisation is reset and play will begin after a configurable delay.

This PR is an extension of @JohanneMontano's work on getting the reset localisation buttons working.

The buzzer functionality is not working but we can investigate further in later PR.

### DIFF
--- a/module/purpose/Soccer/src/Soccer.cpp
+++ b/module/purpose/Soccer/src/Soccer.cpp
@@ -53,6 +53,7 @@ namespace module::purpose {
     using Unpenalisation = message::input::GameEvents::Unpenalisation;
     using message::behaviour::state::Stability;
     using message::behaviour::state::WalkState;
+    using message::input::ButtonLeftDown;
     using message::input::ButtonMiddleDown;
     using message::input::GameEvents;
     using message::localisation::ResetFieldLocalisation;
@@ -75,6 +76,8 @@ namespace module::purpose {
 
             // Get the soccer position, if not valid option then default to striker
             cfg.position = Position(config["position"].as<std::string>());
+
+            cfg.disable_idle_delay = config["disable_idle_delay"].as<int>();
         });
 
         // Start the Director graph for the soccer scenario!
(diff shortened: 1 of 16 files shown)


module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -121,13 +129,39 @@ namespace module::purpose {
...
+            idle = true;
+        });
+
+        on<Trigger<ButtonLeftUp>, Single>().then([this] { emit<Scope::DIRECT>(std::make_unique<Buzzer>(0)); });
+
+        on<Trigger<EnableIdle>, Single>().then([this] {
+            // Stop all tasks and stand still
+            emit<Task>(std::unique_ptr<FindPurpose>(nullptr));
+            emit(std::make_unique<Stability>(Stability::UNKNOWN));
+            log<NUClear::INFO>("Idle mode enabled");
```
JohanneMontano, line 146:
```suggestion
            log<NUClear::DEBUG>("Idle mode enabled");
```

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -121,13 +129,39 @@ namespace module::purpose {
...
+            idle = false;
+        });
+
+        on<Trigger<ButtonMiddleUp>, Single>().then([this] { emit<Scope::DIRECT>(std::make_unique<Buzzer>(0)); });
+
+        on<Trigger<DisableIdle>, Single>().then([this] {
+            // If the robot is not idle, restart the Director graph for the soccer scenario!
+            if (!idle) {
                 emit<Task>(std::make_unique<FindPurpose>(), 1);
+                log<NUClear::INFO>("Idle mode disabled");
```
JohanneMontano, line 164:
```suggestion
                log<NUClear::DEBUG>("Idle mode disabled");
```

JohanneMontano (commented):
LGTM, but imo these should be debug logs

module/purpose/Soccer/src/Soccer.cpp
Tom0Brien: I think it's nice to see that button has been pressed and certain mode is enabled

JohanneMontano (approved):
lgtm

module/purpose/Soccer/src/Soccer.cpp
ysims: I think these should be INFO

module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -121,13 +129,39 @@ namespace module::purpose {
...
+            emit<Scope::DELAY>(std::make_unique<DisableIdle>(), std::chrono::seconds(cfg.disable_idle_delay));
+            emit<Scope::DIRECT>(std::make_unique<Buzzer>(1000));
+            idle = false;
+        });
+
+        on<Trigger<ButtonMiddleUp>, Single>().then([this] { emit<Scope::DIRECT>(std::make_unique<Buzzer>(0)); });
+
+        on<Trigger<DisableIdle>, Single>().then([this] {
+            // If the robot is not idle, restart the Director graph for the soccer scenario!
+            if (!idle) {
```
ysims, line 162: how come we have this idle variable?

Tom0Brien: Since we have a delay, you can press the disable idle then the enable idle button, but since we have the delay the enable idle won't prevent this trigger from running, so I added flag which is updated instantly

PR #1394, opened 2024-07-03
Remove Idle Stand Still from RoboCup

ysims:
and the commented out one in keyboardwalk.

I don't think it makes sense to have stand still and start safely as they are competing functionalities. If either of them run, it makes the other obsolete. The idle stand still in the background is only really useful for the start when nothing is telling the motors to move. After that, the motors are always going to be in some position. Stand still was added because in penalised state, the robot wouldn't stand if the binary was rerun, but start safely will do this instead.


ysims:
superceded kinda

PR #1404, opened 2024-07-05
Add logic for whether to approach ball in kickoff 

ysims:
In kickoff, we now keep walking to the ready position if it's the opponent's kickoff. If the ball moves enough away from the center of the field, we assume they have touched the ball and we can play.

Secondary time is used to determine if kickoff time is finished as this time shouldn't be used for anything else unless it's in a mode like direct free kick etc, but in that case we'd be in a different Provider and it wouldn't matter. 

Kickoff applies not just to the start of a half, but also whenever a robot scores a goal. These cases were all tested in Webots. 

The distance for the ball to move to assume kickoff is over seems to be large in webots so hopefully is okay even if localisation has a bit of error.

Only applies to Striker as the others wouldn't attempt to touch the ball in this position.

### DIFF
--- a/module/purpose/Striker/src/Striker.cpp
+++ b/module/purpose/Striker/src/Striker.cpp
@@ -142,14 +142,12 @@ namespace module::purpose {
                         play();
                         return;
                     }
-                    log<NUClear::INFO>("Waiting for kickoff.");
                     // Walk to ready so we are ready to play when kickoff finishes
                     emit<Task>(std::make_unique<WalkToFieldPosition>(
                         pos_rpy_to_transform(Eigen::Vector3d(cfg.ready_position.x(), cfg.ready_position.y(), 0),
                                              Eigen::Vector3d(0, 0, cfg.ready_position.z()))));
                     return;
                 }
-                log<NUClear::INFO>("Playing.");
                 play();
             });
 
--- a/module/purpose/Striker/src/Striker.cpp
+++ b/module/purpose/Striker/src/Striker.cpp
@@ -118,7 +118,20 @@ namespace module::purpose {
         });
 
         // Normal PLAYING state
-        on<Provide<NormalStriker>, When<Phase, std::equal_to, Phase::PLAYING>>().then([this] { play(); });
+        on<Provide<NormalStriker>, When<Phase, std::equal_to, Phase::PLAYING>, With<GameState>>().then(
+            [this](const GameState& game_state) {
+                // If it's not our kickoff and timer is going, stand still
+                // The secondary timer will only happen for kickoff here
+                log<NUClear::INFO>(game_state.data.our_kick_off,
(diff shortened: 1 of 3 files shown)


module/purpose/Striker/src/Striker.cpp
```diff
@@ -118,7 +123,35 @@ namespace module::purpose {
...
+                        return;
+                    }
+                    log<NUClear::INFO>("Waiting for kickoff.");
+                    // Walk to ready so we are ready to play when kickoff finishes
+                    emit<Task>(std::make_unique<WalkToFieldPosition>(
+                        pos_rpy_to_transform(Eigen::Vector3d(cfg.ready_position.x(), cfg.ready_position.y(), 0),
+                                             Eigen::Vector3d(0, 0, cfg.ready_position.z()))));
+                    return;
+                }
+                log<NUClear::INFO>("Playing.");
```
Tom0Brien, line 152: This will spam console when playing right?

module/purpose/Striker/src/Striker.cpp
```diff
@@ -118,7 +123,35 @@ namespace module::purpose {
...
+                if (!game_state.data.our_kick_off
+                    && (game_state.data.secondary_time - NUClear::clock::now()).count() > 0) {
+                    // Check if the ball has moved, if so start playing
+                    if (ball != nullptr && field != nullptr
+                        && (field->Hfw * ball->rBWw).norm() > cfg.ball_kickoff_outside_radius) {
+                        log<NUClear::INFO>("Ball has moved, starting to play.");
+                        play();
+                        return;
+                    }
+                    log<NUClear::INFO>("Waiting for kickoff.");
```
ysims, line 145:
```suggestion

```

module/purpose/Striker/src/Striker.cpp
ysims, line 152:
```suggestion
```

module/purpose/Striker/src/Striker.cpp
```diff
@@ -118,7 +123,33 @@ namespace module::purpose {
...
+                         const std::shared_ptr<const Ball>& ball,
+                         const std::shared_ptr<const Field>& field) {
+                // If it's not our kickoff and timer is going, stand still
+                // The secondary timer will only happen for kickoff here
+                if (!game_state.data.our_kick_off
+                    && (game_state.data.secondary_time - NUClear::clock::now()).count() > 0) {
+                    // Check if the ball has moved, if so start playing
+                    if (ball != nullptr && field != nullptr
+                        && (field->Hfw * ball->rBWw).norm() > cfg.ball_kickoff_outside_radius) {
+                        log<NUClear::INFO>("Ball has moved, starting to play.");
```
ysims, line 141:
```suggestion
```

JohanneMontano (approved):

PR #1410, opened 2024-07-06
Remove unnecessary Defender configs

A-Herfel:
Remove extra yaml's as we only have one Defender, soon to be dynamically decided

### DIFF
--- a/module/purpose/Defender/src/Defender.cpp
+++ b/module/purpose/Defender/src/Defender.cpp
@@ -150,7 +150,7 @@ namespace module::purpose {
     void Defender::play() {
         // Walk to the ball and kick!
         // Second argument is priority - higher number means higher priority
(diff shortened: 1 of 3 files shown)


module/purpose/Defender/src/Defender.cpp
```diff
@@ -150,8 +150,7 @@ namespace module::purpose {
     void Defender::play() {
         // Walk to the ball and kick!
         // Second argument is priority - higher number means higher priority
-        emit<Task>(std::make_unique<FindBall>(),
-                   1);                                  // if the look/walk to ball tasks are not running, find the ball
+        emit<Task>(std::make_unique<FindBall>(), 1);  // if the look/walk to ball tasks are not running, find the ball
```
ysims, line 153: revert change? probs just run format to fix this

ysims (approved):

PR #1414, opened 2024-07-07
Remove Single to fix buttons when running robocup

dex-k:
Button presses were getting skipped due to Single in each button trigger. Removing single seems to have fixed the issue.


jpptm:
Not sure how relevant this would be, but should the lambdas that trigger on button related messages be in a sync group? Just to make sure that the up phase and down phase cannot run concurrently

Tom0Brien (approved):

PR #1420, opened 2024-07-08
Add ability to backspace when editing fields in robocupconfiguration

Tom0Brien:
This PR adds the ability to backspace when editing fields in robocupconfiguration


module/tools/RoboCupConfiguration/src/RoboCupConfiguration.cpp
```diff
@@ -388,18 +388,34 @@ namespace module::tools {
...
         while (true) {
             auto ch = getch();
             switch (ch) {
-                case 27: return "";
+                case 27: return "";  // Escape key
                 case '\n':
-                case KEY_ENTER: return chars.str(); break;
+                case KEY_ENTER: return input;
+                case KEY_BACKSPACE:
+                case 127:  // ASCII code for backspace
```
JohanneMontano, line 402: Why are we doing key backspace then the ascii for backspace via fallthrough? Should it be just `KEY_BACKSPACE`?

ysims (approved):

PR #1424, opened 2024-07-08
Add debug information for purposes in localisation view

Tom0Brien:
This PR adds the player id and assigns a different colour to some elements in localisation view to help distinguish between bounding boxes when running multiple robots.

### DIFF
--- a/module/network/RobotCommunication/src/RobotCommunication.cpp
+++ b/module/network/RobotCommunication/src/RobotCommunication.cpp
@@ -211,6 +211,10 @@ namespace module::network {
                     msg->purpose = *purpose;
                 }
 
+
+                // Emit to the local network for debugging
+                emit(std::make_unique<RoboCup>(*msg));
+
                 emit<Scope::UDP>(msg, cfg.broadcast_ip, cfg.send_port);
             });
     }
--- a/module/network/RobotCommunication/src/RobotCommunication.cpp
+++ b/module/network/RobotCommunication/src/RobotCommunication.cpp
@@ -211,10 +211,6 @@ namespace module::network {
                     msg->purpose = *purpose;
                 }
 
-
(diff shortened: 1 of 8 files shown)


module/network/RobotCommunication/src/RobotCommunication.cpp
```diff
@@ -211,6 +211,10 @@ namespace module::network {
                     msg->purpose = *purpose;
                 }
 
+
+                // Emit to the local network for debugging
+                emit(std::make_unique<RoboCup>(*msg));
+
```
ysims, line 217: This may cause unwanted behaviour as we assume robocup messages come from other robots. You could use global config and change it to emit regularly. But would need to test that this doesn't break things too.

nusight2/src/client/components/localisation/network.ts
```diff
@@ -76,10 +76,27 @@ export class LocalisationNetwork {
...
+
+    robot.player_id = robocup.currentPose?.playerId!;
+
+    // Update colour based on player id
+    if (robot.player_id === 1) {
+      robot.color = "blue";
+    } else if (robot.player_id === 2) {
+      robot.color = "purple";
+    } else if (robot.player_id === 3) {
+      robot.color = "black";
```
ysims, line 92: Since default is black, maybe orange or another colour

nusight2/src/client/components/localisation/robot_model.ts
```diff
@@ -137,7 +137,7 @@ export class BoundingBox {
 export class LocalisationRobotModel {
   @observable private model: RobotModel;
   @observable name: string;
-  @observable color?: string;
+  @observable color: string;
```
ysims, line 140:
```suggestion
  @observable colour: string;
```

ysims (approved):

PR #1429, opened 2024-07-13
Add NBS recording and playback roles

Tom0Brien:
This PR adds roles for NBS recording and playback. These can be used  for tuning localisation and vision


jpptm (commented):
Lgtm, but the data logging one doesnt seem to use data playback. Fancy changing thr path to robocup/datalogging or something similar?

ysims:
maybe the folder could be called `data`

JohanneMontano (approved):
lgtm

PR #1430, opened 2024-07-14
Fix conflict with idle and penalisation

ysims:
Going out of idle was ignoring penalisation so I've done some stuff to fix it. also moved around the penalisation active robot/position update down lower to prevent race conditions and also made it so get up will be lower priority when in idle and penalisation so it definitely stops moving.


module/purpose/Soccer/src/Soccer.cpp
```diff
@@ -207,8 +207,8 @@ namespace module::purpose {
         on<Trigger<EnableIdle>>().then([this] {
             // Stop all tasks and stand still
             emit<Task>(std::unique_ptr<FindPurpose>(nullptr));
+            emit<Task>(std::unique_ptr<FallRecovery>(nullptr));
```
TrentHouliston, line 210: No fall recovery on idle? Is idle like really idle?

ysims: it's a state we have for making the robot stop entirely when we pick it up

JohanneMontano (approved):

PR #1437, opened 2024-07-14
Simplify the dynamic purpose function a little

ysims:
Moving over from NUbots#1378. Just simplifying things a little.


JohanneMontano (approved):
Since this was taken from a bigger branch where things were properly tested, this lgtm.

PR #1441, opened 2024-07-14
Add HardwareIO threadpool and improve watchdog behaviour

dex-k:
This PR:
- Adds a one thread threadpool to HardwareIO, replacing `Sync` groups and `MainThread`.
  - This is a workaround due to the bug in Sync groups.
  - Is more reliable than `MainThread` because other stuff can be running on `MainThread`, but the HardwareIO pool will be unique to HardwareIO.
- Adds a `flush()` function to `utility::io::uart`
- Refactors the Watchdogs a bit:
  - Add functions to cleanly create them at later points in the code.
  - Move packet watchdog creation to after unbinding of the model watchdog to ensure they aren't both running at once, because we don't need the packet watchdog until after `startup`.
  - Flush the `opencr` uart buffer when the packet watchdog times out, to ensure we can never have a residual full packet left after clearing the packet queue, which previously had the potential to be unpredictable.


TrentHouliston (commented):
It's possible the reason this is happening is that the watchdog is being made during setup, and triggering too early because the main system hasn't started up yet

module/platform/OpenCR/HardwareIO/src/HardwareIO.cpp
```diff
@@ -169,10 +107,68 @@ namespace module::platform::OpenCR {
             // The model watchdog is started, which has a longer time than the packet watchdog
             // The packet watchdog is disabled until we start the main loop
             model_watchdog.enable();
-            packet_watchdog.disable();
 
             // The startup function sets up the subcontroller state
             startup();
+
+            packet_watchdog =
```
Tom0Brien, line 114: Would it make sense to move the model watchdog creation here, then create the packet watchdog when we get model information so that no overlap can ever occur

module/platform/OpenCR/HardwareIO/src/HardwareIO.cpp
```diff
@@ -284,7 +280,7 @@ namespace module::platform::OpenCR {
                         // Start the packet watchdog since the main loop is now starting
                         model_watchdog.disable();
                         model_watchdog.unbind();
-                        log<NUClear::WARN>("Packet watchdog enabled");
+                        log<NUClear::INFO>("Packet watchdog enabled");
                         packet_watchdog.enable();
```
Tom0Brien, line 284: Here is where I would think the packet watchdog could be created, after the model watchdog is unbound

JohanneMontano:
Tested on Frankie with `test/sensor`, `test/localisation` and `robocup`, with and without the fix. Dropped packets on startup disappear with the fix.

dex-k:
Tested successfully on kevin with keyboardwalk and robocup. Occasional dropped packets but from different ID's, and it recovers gracefully without falling over etc - wouldn't know there was dropped packets without the logs.

TrentHouliston (approved):

PR #1442, opened 2024-07-14
Add servo temperature monitor tab

Tom0Brien:
image


ysims:
maybe highest temp would be useful to display

JohanneMontano:
I think displaying the highest temp together with what servo it is with is more useful than the average temp

ysims:
I'd also recommend restoring the branch after merge and making a PR into main so we keep this

PR #1447, opened 2024-07-15
Fix VSCode dev container integration

dex-k:
Corrects the path to the docker image to allow for opening in dev container in vscode


dex-k:
this decided to include 30 commits and it was easier to create a new PR than fix this

PR #1450, opened 2024-07-15
Cleanup watchdog creation into functions

dex-k:
Splits out watchdog refactors (not incl uart flush) from #1441


JohanneMontano:
For clarification, this was meant to fix the dropped packets issue on startup. This should not have broken hwio so i think lgtm

JohanneMontano (approved):

PR #1451, opened 2024-07-15
Flush uart buffer on packet watchdog

dex-k:
Split out from #1441.

Only merge after #1450.


module/platform/OpenCR/HardwareIO/src/hardwareio/create_watchdogs.cpp
```diff
@@ -0,0 +1,106 @@
...
+                                "chain will cause all later packets (of higher ID) to be dropped. Consider "
+                                "checking cables for ID {} ({})",
+                                int(first_dropped_packet),
+                                nugus.device_name(first_dropped_packet)));
+            }
+
+            // Send a request for all servo packets, only if there were packets dropped
+            // In case the system stops for some other reason, we don't want the watchdog
+            // to make it automatically restart
+            if (num_packets_dropped > 0) {
```
Claegtun, line 96: Just a naïve question that won't stop approval, but would `num_packets_dropped` ever be zero here? I know that it _can_, but what I am really driving at is what scenario would make that happen. I am just thinking aloud. It may be worth at least thinking about this hypothetical scenario in the wider context of these elusive OpenCR bugs.

JohanneMontano: I'd be more worried about possibly flushing parts of or whole valid dynamixel packets that would lead to more dropped packets

module/platform/OpenCR/HardwareIO/src/HardwareIO.cpp
```diff
@@ -284,7 +211,10 @@ namespace module::platform::OpenCR {
                         // Start the packet watchdog since the main loop is now starting
                         model_watchdog.disable();
                         model_watchdog.unbind();
-                        log<NUClear::WARN>("Packet watchdog enabled");
+
+                        log<NUClear::INFO>("Packet watchdog enabled");
+
+                        packet_watchdog = create_packet_watchdog();
                         packet_watchdog.enable();
```
Claegtun, line 218: Just for my own sake, this should be the only spot where the `packet_watchdog` is first enabled, right? Is it disabled before?

Claegtun (dismissed):
Seems well to me as long as it works on the real robot.

module/platform/OpenCR/HardwareIO/src/HardwareIO.cpp
dex-k: correct, it isn't enabled anywhere else.

This did make me realise the `.enable()` does nothing, so I'll remove it for both the packet and model watchdogs.

Claegtun (dismissed):
Seems well to me.

Claegtun (approved):
Seems well to me.

PR #1452, opened 2024-07-15
Clean up useless shutdown in hwio

dex-k:
Split from #1441.

Only merge after #1451.

This is just junk hanging around because the uart is closed on shutdown anyway. Not really important but may as well simplify stuff.


JohanneMontano:
Merge #1450  first so the diffs decrease, then wait for @Claegtun to finish revieiwing #1451. This lgtm.

JohanneMontano (approved):

PR #1454, opened 2024-07-16
Thread pool changes split from #1441

dex-k:
Don't merge without further testing until we're sure what broke things.

Worth testing removing threadpool from things that weren't `Sync`ed or `MainThread`ed before #1441

Only merge after #1450, #1451, #1452


dex-k:
Already in robocup2024 and works (or rather, hasn't broken yet)

ysims:
what's going on with this PR @dex-k

dex-k:
> what's going on with this PR @dex-k 

Oh shit idk. I think this went into `robocup` and then whatever working state we got ended up in `main`? Weird it's leftover though.

PR #1455, opened 2024-07-16
Remove unnecessary enable calls on watchdogs

dex-k:
This messed up between #1450 and #1451 and is just part of cleaning up #1451


Claegtun (approved):
Seems well to me.

PR #1460, opened 2024-07-17
fix buttons not working when battery voltage is low

dex-k:
Duplicate of #1458 but for `main` instead of `robocup`.

subcontroller_error masked in all dynamixel packet errors from all devices. this was done before masking out the alert bit for battery voltage errors, which meant that we had a nonzero subcontroller_error if the battery was fresh (because servos would report a voltage error).

then button debouncing ignored buttons when we had a subcontroller error, causing the problem.

we now keep the subcontroller error check, but don't consider servo errors as subcontroller errors anymore. which is better.


Tom0Brien (approved):
With button debouncing, removing the error check in sensor filter would be reasonable as well imo. but this lgtm to start with

PR #1461, opened 2024-07-17
Some localisation changes

ysims:
- Reordered camera settings to make it easier to read
- Tune Frankie's camera
- Weight intersections more than field lines
- Don't associate two field intersection features as the same landmark
- Remove the centre cross from the landmarks


ysims:
think this is scattered in other PRs now

PR #1465, opened 2024-07-17
We did it! The buttons will never break again! It's the last button fix!

dex-k:
This makes us not ignore button presses if we have subcontroller errors, just let the debouncing and laws of probability help us there. Also lower debounce threshold slightly, kinda just on vibes. Slight refactor for readability too.

Tested and working on lil Kev :)


ysims:
@dex-k can you get a PR up for this to go into `main` pls ty

PR #1472, opened 2024-08-15
Add IP, SSID and password toggles to the RoboCup configuration tool

ysims:
This PR aims to make it simpler to change IPs and networks by toggling in the robocup configuration tool. This is mostly useful at RoboCup. The IPs in the config shouldn't need changing, but the SSID and passwords of the fields may change next year. I'll keep the ones from this year in as placeholders.

The toggle for IP will look for any X placeholder and replace it with the current player ID for ease.

### DIFF
--- a/module/tools/RoboCupConfiguration/README.md
+++ b/module/tools/RoboCupConfiguration/README.md
@@ -39,4 +39,4 @@ N/A
 
 ## Dependencies
 
-N/A
+- ncurses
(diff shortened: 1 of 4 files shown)


Tom0Brien (approved):
lgtm, I think we should look at updating our lab network to 192.168.x.x so we don't need to switch Ethernet profiles when connecting to game router/robot

PR #1482, opened 2024-08-27
Tweaks and improvements for ScriptTuner

dex-k:
This isn't really very important at all, but I was bored and was in the mood to learn a bit about `ncurses`, so here we are.

I've added:
- A new alert message which shows up if you have unsaved changes
  - Currently, a flag is just set for every editing action. I considered trying to keep a hash of the last saved `Script` and comparing to that, so that any net-zero-change action like, say, mirroring and un-mirroring the script, doesn't show up as a change, but that ended up being harder than I thought and constantly hashing the script every input might have a performance overhead that isn't worth it? idk. 
- Autosave functionality
  - Autosave every `AUTOSAVE_INTERVAL` seconds (default `30`) if there are unsaved changes
  - Alerts you that an autosave has taken place and where it was saved to
  - Deletes autosaves from _the current session only_ if the script is manually saved (to avoid deleting an autosave file if ScriptRunner is frantically re-opened after realising the script wasn't saved)
  - Autosave locally to `NUbots/.scripttuner/autosave` if running in the docker container 
- Use of colour in the UI to emphasise important things and reduce the visual clutter
  - Unselected servos are dimmed
  - Locked and unlocked indicators are red/green
  - Frame number list has colour and dimming for emphasis too
  - New alert messages have colour 
- An ascii art NUgus body showing which motors are where on the body
  - Shows if they're locked/unlocked
  - Allows intuitive navigation around the body (rather than just the servo list) by using `wasd`.
  - Responsive to terminal size to not clutter up display if there's no space.
- An option (press `D`) to display angles in degrees (with indicator to avoid confusion)
  - Also allows for entering angles in degrees, but all angles are internally stored in radians still, so entering an angle in degrees sometimes means the displayed value is rounded (i.e. entering `150` will then be stored as `149.9` due to being internally converted to radians)
- Option for `hjkl` vim-style navigation
- Press `/` to switch selection between left/right servo pairs.
- Feedback for illegal inputs like navigating past the end of a list, entering invalid values etc.
- Allow for backspacing within input fields!
  - You can keep backspacing past the field though, which breaks things. That's considered user error if you do that, so don't do that.
- Makes `getch()` input non-blocking without breaking things, which makes stuff play a bit nicer.
- Some small typo fixes and readability tweaks.

Comparison of old and new UI:
image

image


.gitignore
```diff
@@ -80,6 +80,9 @@ log
 log/*
 nsga2-output/*
 
+# ScriptTuner stuff
```
ysims, line 83:
```suggestion
# ScriptTuner autosaves
```

module/purpose/ScriptTuner/src/ScriptTuner.cpp
```diff
@@ -141,25 +158,114 @@ namespace module::purpose {
             emit(std::move(waypoint));
         });
 
+        // Debugging to make sure logs are working
+        /// TODO: Remove before merging with main
+        on<Every<5, std::chrono::seconds>>().then("Heartbeat", [this] {
+            static size_t counter = 0;
+            log<NUClear::DEBUG>("Beep Boop ", counter++);
+        });
```
ysims, line 166: remove?

module/purpose/ScriptTuner/src/ScriptTuner.cpp
```diff
@@ -141,25 +158,114 @@ namespace module::purpose {
...
+
+        on<Every<AUTOSAVE_INTERVAL, std::chrono::seconds>, Single>().then("Autosave", [this] {
+            log<NUClear::DEBUG>("Autosaving script if required...");
+
+            // Only autosave if autosaving is enabled and we have unsaved changes
+            if (!autosave_enabled || !unsaved_changes)
+                return;
```
ysims, line 173:
```suggestion
            if (!autosave_enabled || !unsaved_changes) {
                return;
            }
```

module/purpose/ScriptTuner/src/ScriptTuner.cpp
```diff
@@ -141,25 +158,114 @@ namespace module::purpose {
...
+
+            // Before we start, save the temp save the old path so we can delete it
+            // once we have a new autosave file.
+            auto last_autosave_path = this->autosave_path;
```
ysims, line 177: grammar seems off in the comment, should 'save the temp' be removed

module/purpose/ScriptTuner/src/ScriptTuner.cpp
```diff
@@ -141,25 +158,114 @@ namespace module::purpose {
...
+
+            // Split the script name into the path and filename
+            auto [autosave_dir, filename] = utility::file::pathSplit(script_path);
```
ysims, line 180: use the standard library filesystem::path functions, we have a PR up somewhere to remove `fileutil.cpp`.

module/purpose/ScriptTuner/src/ScriptTuner.cpp
```diff
@@ -141,25 +158,114 @@ namespace module::purpose {
...
+            if (autosave_on_robot) {
+                // If we're autosaving on the robot, then save to an `autosave/` subdirectory
+                // in the same folder of the current script.
+                autosave_dir += "/autosave/";
+            }
+            else {
+                // Otherwise, if we're running in the docker container, then replace the
+                // autosave directory with a subdirectory in the mounted `../NUbots/` folder.
+                autosave_dir = "/home/nubots/NUbots/.scripttuner/autosave/";
+            }
```
ysims, line 192: I feel like it would be better to have the config variable be the directory rather than a bool and then hardcoding the respective paths

module/purpose/ScriptTuner/src/ScriptTuner.hpp
```diff
@@ -46,6 +46,17 @@ namespace module::purpose {
      */
     class ScriptTuner : public NUClear::Reactor {
     private:
+        /// @brief The autosave interval in seconds
+        /// @note Set to 0 to disable autosave
+        static constexpr size_t AUTOSAVE_INTERVAL = 30;
+
+        /// @brief Should we autosave to the local script directory, or to the
+        /// mounted ../NUbots folder in the docker container
+        bool autosave_on_robot;
```
ysims, line 55: generally we initalise all the variables in the .hpp then we can be sure everything is initalised, and you don't have to use up lines in the cpp initialiser list

```suggestion
        bool autosave_on_robot = false;
```

module/purpose/ScriptTuner/src/ScriptTuner.cpp
dex-k: lol whoops

module/purpose/ScriptTuner/src/ScriptTuner.cpp
dex-k: done

ysims (approved):

PR #1485, opened 2024-09-22
Fix R2R seg fault by checking if GameState is null first

ysims:
Noticed when GameController didn't exist but RobotComms did, it seg faults. Optional GameState is used without checking for nullptr first. Fixed up so there's a check.

### DIFF
--- a/module/network/RobotCommunication/src/RobotCommunication.cpp
+++ b/module/network/RobotCommunication/src/RobotCommunication.cpp
@@ -138,6 +138,7 @@ namespace module::network {
                 msg->timestamp = NUClear::clock::now();
 
                 // State
+                // If there is game state information, then process 
                 if (game_state) {
                     int penalty_reason = game_state->data.self.penalty_reason;
                     switch (penalty_reason) {
--- a/module/network/RobotCommunication/src/RobotCommunication.cpp
+++ b/module/network/RobotCommunication/src/RobotCommunication.cpp
@@ -138,7 +138,7 @@ namespace module::network {
                 msg->timestamp = NUClear::clock::now();
 
                 // State
-                // If there is game state information, then process 
+                // If there is game state information, then process
                 if (game_state) {
                     int penalty_reason = game_state->data.self.penalty_reason;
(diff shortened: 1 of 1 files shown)


module/network/RobotCommunication/src/RobotCommunication.cpp
```diff
@@ -138,11 +138,13 @@ namespace module::network {
                 msg->timestamp = NUClear::clock::now();
 
                 // State
-                int penalty_reason = game_state->data.self.penalty_reason;
-                switch (penalty_reason) {
-                    case 0: msg->state = 0; break;
-                    case 1: msg->state = 1; break;
-                    default: msg->state = 2; break;
+                if (game_state) {
```
jpptm, line 141:
```suggestion
// Check if game state is null or not to avoid seg fault
                if (game_state) {
```

jpptm: or something similar. Otherwise lgtm

jpptm (dismissed):

module/network/RobotCommunication/src/RobotCommunication.cpp
ysims:
```suggestion
                // If there is game state information, then process 
                if (game_state) {
```

ysims: everything in this reaction is optional and needs to be checked, so I don't think it needs much explanation

jpptm (dismissed):

TrentHouliston:
Did you want to make it explicit what you do when you don't have one?

ysims:
> Did you want to make it explicit what you do when you don't have one?

Default for state is unknown_state, also if we want to do this then it should probably be done for all of the if statements rather than just this one, but then it seems like it'd be a bit cluttered when protobuf already does defaults? And out of scope of this PR.

TrentHouliston:
Fair enough

JohanneMontano (approved):

PR #1487, opened 2024-09-27
Add rerun as plotjuggler alternative

Tom0Brien:
This PR adds [Rerun](https://www.rerun.io/) support for data visualization in real time. This is a potential drop in replacement of plotjuggler which can be embedded in nusight. The benefit of a tool like this is to allow easy visualization of general time series data, 3d points etc without the need to develop a new nusight module for adhoc debugging. 

Notes:
1. This does motivate the idea of just improving our chart view so we don't need this and can use `.nbs` recordings etc
2. Rerun doesn't currently support the `serve` api in the rerun C++ SDK (https://rerun.io/docs/reference/sdk-operating-modes#serve) so I am not sure if the embeded app can connect to our code, @JosephusPaye do you know more about how this works?  

image

Embeded rerun app:

image


JosephusPaye:
Looks like the serve API starts a websocket server that clients like the embedded view can connect to to display data in real time. Unfortunately like you mentioned, they haven't implemented it for C++ yet, so you can't use it from a C++ NUbots module.

This would be a great time to use a Python module if NUClear had support for that, since they've got Python support for the serve API.

I'm not sure if their serve API in Python has a way to receive data, in addition to just serving/sending it. If so, you could set up a `./b` Python tool that starts the rerun server and prints out its address/port. Then have the C++ module send data to that (like we do for PlogJuggler), and point the embedded web viewer in NUsight to the same address to see the data.

PR #1489, opened 2024-10-05
Onboarding task

vincenttumminello:
Created Pong module and updated ping module to increment a counter every time a pong msg is received.


module/onboarding/Ping/src/Ping.cpp
```diff
@@ -0,0 +1,34 @@
+#include "Ping.hpp"
+
+#include "extension/Configuration.hpp"
+
+#include "message/onboarding/Ping.hpp"
+#include "message/onboarding/Pong.hpp"
+
+int32_t pongcount = 0;
```
ysims, line 8: is this meant to  be here? wouldn't it come from the Pong message?

module/onboarding/Ping/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+
+## Consumes
+
+
+## Emits
+
+
+## Dependencies
+
```
ysims, line 16: always good to fill out your readmes!

.vscode/settings.json
```diff
@@ -24,7 +24,7 @@
   "python.formatting.provider": "black",
   "[python]": {
     "editor.codeActionsOnSave": {
-      "source.organizeImports": true
+      "source.organizeImports": "explicit"
```
ysims, line 27: This can sometimes get accidentally committed when we don't want it committed - checking the files changed tab in your PR can help identify that you've committed it and then you can revert it

shared/message/onboarding/Ping.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package message.onboarding;
+
+message Ping{
+    // TODO: Fill this message with an int32 field named "count". What should its tag be?
```
ysims, line 6: Change this comment to explain what the count field is

ysims (commented):
nice

module/onboarding/Ping/src/Ping.cpp
JohanneMontano: You can't use a global variable to cache your results. Only the protobuf messages can store these things

module/onboarding/Ping/src/Ping.cpp
```diff
@@ -25,7 +23,7 @@ namespace module::onboarding {
 
         on<Trigger<Pong>>().then([this](const Pong& pong_msg) {
             auto ping_msg = std::make_unique<Ping>();
-            ping_msg->count = pongcount++; // TODO: Assign counter value.
+            ping_msg->count = pong_msg.count + 1; // Assign counter value.
```
JohanneMontano, line 26: pong_msg is declared as const &, so you've promised to not touch that argument but this line is altering the state of that message. This probably wont compile. Even if it did, it still is wrong. You should be creating a new message and emitting that one instead.

ysims:
Cleaning up. 
Looks like you addressed the comments successfully, nice work! There is one CI (continuous integration) check that failed - formatting! To fix this in future, run `./b format` in your terminal and it will automatically format your files. Then commit and push to fix the check. If you download clang-format locally and install the clang-format vs code extension, it will do it automatically for you (for cpp files).

PR #1500, opened 2024-11-25
[NUsight] Format code

BrendanAnnable:
Prettier was upgraded in #1497 but somehow passed without needing to rerun prettier.

This PR fixes that.


BrendanAnnable:
Turns out to be more painful then I could ever imagine, reverted prettier in #1499 instead.

PR #1502, opened 2024-11-26
[NUsight] Move to ES modules

BrendanAnnable:
This basically moves away from JS scripts and defaults all JS files as proper ES modules. 

Also deprecates the `require()` import syntax, which unfortunately protobufjs still needs to function (although I hear we are planning to move to [Protobuf ES](https://github.com/bufbuild/protobuf-es)!).

https://nodejs.org/api/packages.html#type
https://blog.logrocket.com/es-modules-in-node-today/


JosephusPaye:
This is gonna be great to have. Last I tried was about a year ago and many of our dependencies were incompatible with an ESM-first approach then.

TrentHouliston (approved):

PR #1504, opened 2024-11-27
[NUsight] Remove disposeOnUnmount usages

BrendanAnnable:
The `disposeOnUnmount` function is deprecated in MobX 6 and its invocation will throw an error. This PR replaces all usages with a manual dispose method to prepare for such a day (#1503).

Converting classes to functional components would likely be a cleaner solution, but would come with a much greater scope of work.

### DIFF
--- a/nusight2/src/client/components/localisation/view.tsx
+++ b/nusight2/src/client/components/localisation/view.tsx
@@ -89,7 +89,6 @@ export class LocalisationView extends React.Component<LocalisationViewProps> {
   private dispose?: () => void;
 
   componentDidMount(): void {
-    this.dispose?.();
     this.dispose = compose([
       addDocListener("pointerlockchange", this.onPointerLockChange, false),
       addDocListener("mousemove", this.onMouseMove, false),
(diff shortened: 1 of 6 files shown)


nusight2/src/client/components/localisation/view.tsx
```diff
@@ -75,29 +76,34 @@ export class FieldDimensionSelector extends React.Component<FieldDimensionSelect
...
 
   componentDidMount(): void {
-    document.addEventListener("pointerlockchange", this.onPointerLockChange, false);
-    document.addEventListener("mousemove", this.onMouseMove, false);
-    document.addEventListener("keydown", this.onKeyDown, false);
-    document.addEventListener("keyup", this.onKeyUp, false);
-    document.addEventListener("wheel", this.onWheel, false);
-    disposeOnUnmount(
-      this,
+    this.dispose?.();
```
TrentHouliston, line 92: Do you want to remove these if the docs are clear that it's not required?

TrentHouliston (approved):

PR #1516, opened 2024-12-13
[NUsight] Implement Kinematics Tab

KimKhangHoang:
This PR implements the setup for the Kinematics tab. It visualises the robot's kinematics view, including:

- Robot torso representation
- Rotation axes and coordinate space
- Joints data display

The implementation follows a structure similar to the Localisation tab.

Screenshot 2025-04-08 101656


nusight2/src/client/components/localisation/r3f_components/text_geometry_helper/tmp7qhfputt/TextGeometryHelper.tsx
```diff
@@ -0,0 +1,23 @@
...
+        font: font,
+        size: 0.1,
+        height: 0,
+      });
+      setGeometry(newGeometry);
+    });
+  }, [text]);
+
+  return geometry || new TextGeometry("");
+};
```
ysims, line 23: this looks like a temp file generated from the formatting tool

nusight2/src/assets/fonts/roboto/tmpajzd1jaw/Roboto Medium_Regular.json
```diff
@@ -0,0 +1,5182 @@
+{
```
ysims, line 1: this looks like a temp file generated from the formatting tool

nusight2/src/client/components/kinematics/r3f_components/coordinates/view.tsx
```diff
@@ -0,0 +1,65 @@
...
+    return () => {
+      observer.disconnect();
+      window.removeEventListener("focus", handleFocus);
+    };
+  }, []);
+
+  // Load the font and create the text geometry
+  useEffect(() => {
+    const loader = new FontLoader();
+    loader.load("/fonts/roboto/Roboto Medium_Regular.json", (font: any) => {
```
ysims, line 40: can we change this filename so the space is an underscore?

ysims (commented):
Nice!
CI will fail until #1546 is merged, once that's in you can merge main and this should hopefully pass.

nusight2/src/client/components/localisation/r3f_components/text_geometry_helper/tmp7qhfputt/TextGeometryHelper.tsx
KimKhangHoang: oh yeah, I'll delete it

nusight2/src/client/components/kinematics/r3f_components/coordinates/view.tsx
KimKhangHoang: Sure, I'll rename it to Roboto_Medium_Regular.json

ysims:
@Tom0Brien can you check if this is what you're after in terms of functionality? I think there was talk of showing axis on all the joints, but doing that in a way to prevent clutter might be tricky and could be another PR.

nusight2/src/client/components/kinematics/r3f_components/nugus/view.tsx
```diff
@@ -0,0 +1,74 @@
...
+      }
+    });
+  }, [nugusUrdfPath]);
+
+  const rotation = model.Hft.decompose().rotation;
+  const motors = model.motors;
+
+  React.useEffect(() => {
+    // Update robot's pose
+    if (robotRef.current) {
```
Tom0Brien, line 29: I don't think we should have the robot orientation changing, just keep the pose fixed and update joints

Tom0Brien (changes requested):
small thing, then LGTM!

nusight2/src/client/components/kinematics/r3f_components/nugus/view.tsx
KimKhangHoang: Thanks, Tom — I’ve removed the line that updated the robot’s quaternion.

I also noticed that some joints (like hip yaw or ankle roll) have static values (always 0 or fixed), at least in the sample data I’m running. I’ve kept them in for now. Would it make sense to treat these as redundant and remove them from the rendering and UI display, or do they have semantic meaning in other contexts?

nusight2/src/client/components/kinematics/r3f_components/axes/view.tsx
JosephusPaye, line 1: For these `r3f_components/XYZ/view.tsx` files, you can simplify by just having `r3f_components/XYZ.tsx`, instead of having each in its own folder with just the one file.

Generally speaking, the separation into model/view/controller is something we want to keep only at the page/tab-level going forward. Plain React components can be stored in a file named after them, without extra folders if not necessary.

nusight2/src/client/components/localisation/r3f_components/text_geometry_helper/TextGeometryHelper.tsx
JosephusPaye, line 1: Related to my other comment, this can be moved up on level to be `r3f_components/text_geometry_helper.tsx` (file names should be in camel case, even ones with components).

JosephusPaye (commented):
Nice work 👍 

Just two suggestions...

nusight2/src/client/components/kinematics/r3f_components/axes/view.tsx
KimKhangHoang: Thanks, Josephus — I’ve flattened the R3F components under the kinematics tab and renamed the files to camelCase as suggested.

nusight2/src/client/components/localisation/r3f_components/text_geometry_helper/TextGeometryHelper.tsx
KimKhangHoang: I noticed other parts of the codebase with similar one-file-per-folder setups or non-camelCase names. I’ve left those as-is to keep this PR focused, but it might be something worth cleaning up in a separate PR down the line.

nusight2/src/client/components/kinematics/r3f_components/nugus/view.tsx
Tom0Brien: They should definitely be kept.

Tom0Brien (approved):
LGTM!

nusight2/src/client/components/localisation/r3f_components/text_geometry_helper/TextGeometryHelper.tsx
Tom0Brien: Good point. The localisation view has these unneeded folders, should clean that up.

JosephusPaye: Oh damn, I meant snake case 😅

KimKhangHoang:
I’ve updated the file names to snake_case and kinda cleaned up R3F components (since I think it was a quick fix), as well as merged with the latest changes from main.

ysims (approved):

PR #1519, opened 2025-01-16
Adds a `Wait` special Task that will run the Provider after a given time and optionally `Wait` on Servo Providers

ysims:
This PR adds in a new special `Wait` Task, similar to how we have `Done` and `Idle`. It will get the Director to rerun the Provider on its current active task.

Alongside this, some optimisations to the Servo Providers is made. `ServoCommand`s now has an enum for what kind of `Done` you want from Servos, that being time-based `Done`, no `Done` and maybe in the future we could have position-based `Done`. Scripts use the time-based `Done` with `Wait`. The Walk is different because we don't wait for `Done`s, it just keeps sending requests, and so we no longer check `Done` from it. 

- [x] Updated NUbook if necessary (NUbots/NUbook#321)
- [x] Use NUClear time travel in test to properly test the Wait functionality
- [x] Check for anywhere else to use Wait
- [x] Fix script bug


module/extension/Director/src/Director.cpp
```diff
@@ -269,6 +269,21 @@ namespace module::extension {
...
+
+        on<Trigger<WaitDelay>, Sync<Director>, Pool<Director>, Priority::HIGH>().then(
+            "Wait Delay",
+            [this](const WaitDelay& w) {
+                // Get the provider that we are waiting on
+                auto provider = providers.at(w.provider->id);
+
+                // If the provider is still active, then we can run it
+                if (provider == provider->group.active_provider) {
+                    run_task_on_provider(provider->group.active_task, provider, RunReason::WAIT);
```
TrentHouliston, line 285: Should this just fold in like any other task, and rather than having a special RunReason, just say the wait task was Done (so SubtaskDone run reason)

module/extension/Director/src/Director.hpp
```diff
@@ -56,6 +56,11 @@ namespace module::extension {
         // Only use a single thread at a time for the Director
         static constexpr int concurrency = 1;
 
+    private:
+        struct WaitDelay {
+            std::shared_ptr<component::Provider> provider;
+        };
+
     private:
```
TrentHouliston, line 64: Back to back private sections!

module/actuation/Servos/src/Servos.hpp
```diff
@@ -60,14 +60,17 @@ namespace module::actuation {
...
                                                            servo.command.position,
                                                            servo.command.state.gain,
                                                            servo.command.state.torque));
+                        // Rerun the Provider when the servo is done moving
+                        emit<Task>(std::make_unique<Wait>(servo.command.time));
                     }
-                    // If the time to reach the position is over, then stop requesting the position
-                    else if (NUClear::clock::now() >= servo.command.time) {
+                    // When the servo is done moving, emit Done
+                    else if (run_reason == RunReason::WAIT) {
```
TrentHouliston: You don't actually emit a wait here!

module/actuation/Servos/src/Servos.hpp
TrentHouliston, line 1: You didn't get rid of the every 90 per second

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -181,6 +181,51 @@ namespace module::extension {
             }
         }
 
+        // Check if a Wait command was emitted and schedule to run the Provider again
+        //
+        for (const auto& t : requested_tasks) {
+            if (t->type == typeid(::extension::behaviour::Wait)) {
+                if (requested_tasks.size() > 1) {
+                    log<WARN>("Wait task was emitted with other tasks, the other tasks will be ignored");
```
TrentHouliston, line 189: I think you can wait and do other tasks tbh, wait is kinda just another task

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -181,6 +181,51 @@ namespace module::extension {
...
+                emit(std::make_unique<NUClear::dsl::operation::ChronoTask>(
+                    [this, provider, weak_task](const NUClear::clock::time_point&) {
+                        // Check if the task still exists
+                        if (weak_task.expired()) {
+                            return false;
+                        }
+
+                        // Get the task
+                        auto task = weak_task.lock();
+                        emit(std::make_unique<WaitDelay>(provider));
```
TrentHouliston, line 215: Better to just auto task = weak_task.lock() and then do the emit if(task);

Technically the task could expire between you calling expired and locking
```suggestion
                        // Check if the task still exists
                        auto task = weak_task.lock();
                        if (weak_task) {
                            emit(std::make_unique<WaitDelay>(provider));
                        }
```

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -181,6 +181,51 @@ namespace module::extension {
...
+
+                        // We don't renew, remove us
+                        return false;
+                    },
+                    NUClear::clock::now() + delay,
+                    -1));  // Our ID is -1 as we will remove ourselves
+                // emit(msg);
```
TrentHouliston, line 222: scar tissue
```suggestion
```

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -181,6 +181,51 @@ namespace module::extension {
...
+
+                // We don't do anything else on wait
+                return;
```
TrentHouliston, line 225: I think you can totally do other things along with wait! I think this should be handled elsewhere to make sure that the task pack is runnable (with wait tasks always being runnable)

shared/extension/behaviour/RunReason.hpp
```diff
@@ -49,6 +49,8 @@ namespace extension::behaviour {
         STOPPED,
         /// A subtask has finished and emitted a Done message
         SUBTASK_DONE,
+        /// The provider emitted a Wait message and the time has elapsed, causing the provider to run
+        WAIT,
```
TrentHouliston, line 53: I think wait can come in like subtask_done

module/extension/Director/tests/TestBase.hpp
```diff
@@ -101,14 +101,18 @@ class TestBase : public extension::behaviour::BehaviourReactor {
             if constexpr (I > 1) {  // Check the next step
                 next_step<I - 1>(v);
             }
-            else {  // Shutdown after the last step
+            else if (auto_shutdown) {  // Shutdown after the last step
```
TrentHouliston, line 104: Just make this a defaulted constructor argument

module/actuation/Servos/src/Servos.hpp
```diff
@@ -51,26 +52,30 @@ namespace module::actuation {
...
-                                       servo.command.state.gain,
-                                       servo.command.state.torque));
-                        }
-                        emit(std::make_unique<ServoTarget>(servo.command.time,
-                                                           ID,
-                                                           servo.command.position,
-                                                           servo.command.state.gain,
-                                                           servo.command.state.torque));
+            on<Provide<Servo>, Priority::HIGH>().then([this](const Servo& servo, const RunReason& run_reason) {
+                if (run_reason == RunReason::NEW_TASK) {
```
TrentHouliston, line 56: I think the logic in this function can be simplified. Just emit on new task, but also if time has passed and waiting in time then emit done, otherwise emit wait

module/extension/Director/src/Director.cpp
```diff
@@ -271,6 +271,15 @@ namespace module::extension {
                 reevaluate_group(g);
             });
 
+        on<Trigger<WaitFinished>, Sync<Director>, Pool<Director>, Priority::REALTIME>().then(
+            "Wait Delay",
+            [this](const WaitFinished& w) {
```
TrentHouliston, line 276: Technically the same provider could start and then leave and come back. But in that case the task would have been nulled out I think.

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -229,10 +229,50 @@ namespace module::extension {
         group.done = false;
         group.update_data();
 
+        // Check if a Wait command was emitted and schedule to run the Provider again
+        // Other tasks can run with Wait
+        // Wait should be removed and then readded at the end
+        TaskList non_wait_tasks;
+        TaskList wait_tasks;
+        for (const auto& t : requested_tasks) {
```
TrentHouliston, line 237: If a wait is optional or not optional should it happen with the other tasks? Or is wait special enough that it ignore that. Probably always expect wait to wait?

module/extension/Director/src/director/run_task_pack.cpp
```diff
@@ -229,10 +229,50 @@ namespace module::extension {
...
+            else {
+                non_wait_tasks.push_back(t);
+            }
+        }
+
         // Remove null data tasks from the list, this allows root tasks to be cleared
         TaskList tasks;
-        tasks.reserve(requested_tasks.size());
-        for (const auto& t : requested_tasks) {
+        tasks.reserve(non_wait_tasks.size());
```
TrentHouliston, line 274: I would leave the names as requested tasks in case we add more categories in the future. Or maybe just filter out runnable tasks, but at the end just use the original vector

module/extension/Director/tests/wait_task_cancel.cpp
```diff
@@ -0,0 +1,137 @@
...
+                    events.push_back("task executed, no wait");
+                }
+            });
+
+            /**************
+             * TEST STEPS *
+             **************/
+            on<Trigger<Step<1>>, Priority::LOW>().then([this] {
+                // Freeze time
+                emit(std::make_unique<NUClear::message::TimeTravel>(NUClear::clock::now(),
```
TrentHouliston, line 71: Maybe time travel to absolute times in this test and then include now in the log lines. Then you can test that it's waiting not just running in order

TrentHouliston (commented):
Didn't review two of the tests yet

module/extension/Director/src/director/run_task_pack.cpp
ysims: I'm not sure I get what you mean
do you mean instead of `tasks`, make `requested_tasks` empty and use that?
or not have the `non_wait_tasks` and instead leave `requested_tasks` as-is, assuming Wait will disappear later?

module/actuation/Servos/src/Servos.hpp
ysims: this is what it there currently

module/extension/Director/src/Director.cpp
ysims: I feel like the task should be gone in that case

module/extension/Director/src/director/run_task_pack.cpp
ysims: as in, do we want it to only run when all subtasks can run? I think that could maybe be good, but could be something to add in later if someone wants that kind of functionality.

module/actuation/Servos/src/Servos.hpp
TrentHouliston: I think I was reviewing before that commit happened

module/extension/Director/src/director/run_task_pack.cpp
TrentHouliston: More the names of the variables, really the original requested tasks list should already have the wait tasks in it still... so you only need a filtered list of runnable tasks which has the wait tasks removed and is used for the rest. then at the end you just assign the original requested tasks to current tasks

module/extension/Director/tests/TestBase.hpp
```diff
@@ -64,8 +64,11 @@ class TestBase : public extension::behaviour::BehaviourReactor {
      * @param timeout The time to wait for the test to complete before failing
      */
     explicit TestBase(std::unique_ptr<NUClear::Environment> environment,
+                      bool auto_shutdown_                         = true,
                       std::chrono::steady_clock::duration timeout = std::chrono::milliseconds(1000))
         : BehaviourReactor(std::move(environment)) {
+        // Set whether or not to automatically shutdown the powerplant after the steps are finished
+        auto_shutdown = auto_shutdown_;
```
TrentHouliston, line 71: Dont' need to set this, just only have it in the constructor

TrentHouliston (approved):

PR #1520, opened 2025-01-19
Add a b script to automate message generation for NUSense and update the relevant guides for it

jpptm:


ysims (approved):

module/platform/NUSense/HardwareIO/README.md
```diff
@@ -21,3 +21,49 @@ This module is built when the subcontroller CMake flag is set to `NUSense`. Usin
...
+
+For example, if a user wants to generate nanopb messages and specify that certain properties of messages must have max counts, the command below should put the generated files in the `recordings` directory.
+
+```bash
+./b nusense generate_nanopb_message recordings/ServoTarget.proto \
+    --selections ServoTargets:targets:max_count:20 \
+                 SubcontrollerServoTargets:targets:max_count:20
+```
+
+The `recordings` directory is convenient as it is symlinked to the docker container. It is also imperative to specify maximum counts for repeated fields as encoding a message without it takes significantly longer and introduces a lot of overhead that is undesired in terms of performance.
```
Bidski, line 39: Don't do this as a b script. Integrate into the build pipeline of `nuclear_message` so that the generated files are automagically updated whenever the proto files or options files are updated. Also just commit the options files next to the proto files, these can then be dependencies of the build process as well

This also means that when you get around to the `./b install` side of things you wont need to do anything special

Bidski (changes requested):

module/platform/NUSense/HardwareIO/README.md
jpptm: I will address this in a different pr, I will be using this tool a lot later on for a task. I will delete it after I've finished said task.

PR #1522, opened 2025-01-25
Automate nanopb message generation properly using the nuclear_message pipeline

jpptm:
I have tested this and it generates the source files to the right directory correctly.

This will only invoke the `nanopb_generator` binary if a .options file for a corresponding message is seen in `~/NUbots/shared/message/nanopb_options`. Note that a typo in the filename bit of the .options file will result to a failed build i.e., the wrong source files will likely be generated for NUSense. This is guaranteed for repeated fields. They are wrong in the sense that the type will correspond to a `pb_size_t` instead of an array with a defined size which will result to terrible computational overheads that we do not want to deal with (relating to decoding and especially encoding).

The custom command depend on the proto and option files and the custom target depends on the output files so when these files are edited in some way, cmake regenerates the headers automatically when `./b build` is invoked.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -373,6 +373,9 @@ RUN install-from-source https://github.com/Tom0Brien/tinyrobotics/archive/refs/t
 # lame
 RUN install-from-source https://downloads.sourceforge.net/lame/lame-3.100.tar.gz
 
+# Nanopb for generating messages for NUSense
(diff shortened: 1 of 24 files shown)


TrentHouliston (commented):
Why are you generating nanopb in the NUbots codebase? Isn't it a NUSense thing?

jpptm:
closing this until we have a more solid idea for a proper pipeline

PR #1523, opened 2025-01-26
Delayed Get up to improve side detection

vincenttumminello:
- Slightly delayed get up task emission to improve side detection
- It's not a perfect solution but from the testing I did in simulation it seemed to improve on the issue

### Reviewers, Note

The getup side detection issue can be recreated in simulation and I found it happened most often when the robot fell over not directly forwards or backwards but to the side a bit. 

### Things left to do

- Working on a better fix using accelerometers to determine when the robot has actually settled on the ground. 

### Labels Info (delete the rest of this once you have read it)

- G-Motion
- L-C++


.vscode/settings.json
```diff
@@ -18,7 +18,83 @@
...
+    "stdexcept": "cpp",
+    "stop_token": "cpp",
+    "streambuf": "cpp",
+    "thread": "cpp",
+    "cfenv": "cpp",
+    "cinttypes": "cpp",
+    "typeindex": "cpp",
+    "typeinfo": "cpp",
+    "valarray": "cpp",
+    "variant": "cpp"
```
jpptm, line 97:
```suggestion
```

.vscode/settings.json
```diff
@@ -29,5 +105,6 @@
   },
   "[jsonc]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
-  }
+  },
+  "C_Cpp.errorSquiggles": "disabled"
```
jpptm, line 109:
```suggestion
```

module/planning/GetUpPlanner/README.md
```diff
@@ -3,7 +3,8 @@
 ## Description
 
 The GetUpPlanner module is used to automatically trigger a get up when the robot has fallen over.
-It monitors the robot to see when it is lying down and has been relatively still for the past n frames before triggering a getup action.
+It monitors the angle of the robot to see when it falls past a set threshold angle (the "fallen_angle") before triggering a delay which then triggers a getup action once elapsed.
+The purpose of the delay is to allow the robot to settle to a final fallen position before any sensor readings are taken to more accurately determine which side it has fallen onto.
```
jpptm, line 7:
```suggestion
It monitors the angle of the robot with the world z axis to see when it falls past a set threshold angle (the "fallen_angle") before triggering a delay which then triggers a getup action once elapsed.
The purpose of the delay is to allow the robot to settle to a final fallen position before any sensor readings are taken to more accurately determine which side it has fallen onto.
```

module/planning/GetUpPlanner/src/GetUpPlanner.cpp
```diff
@@ -49,6 +49,8 @@ namespace module::planning {
         on<Configuration>("GetUpPlanner.yaml").then([this](const Configuration& config) {
             this->log_level  = config["log_level"].as<NUClear::LogLevel>();
             cfg.fallen_angle = config["fallen_angle"].as<float>();
+            cfg.start_delay  = config["start_delay"].as<double>();
+            cfg._is_fallen   = config["_is_fallen"].as<bool>();
```
jpptm, line 53:
```suggestion
            cfg._is_fallen   = config["is_fallen"].as<bool>();
```
why the `_`?

jpptm: also it looks like this key is nowhere to be found in your YAML file. YAML will probably throw a `Bad key` runtime error if you dont add this key in

module/planning/GetUpPlanner/src/GetUpPlanner.hpp
```diff
@@ -39,8 +39,13 @@ namespace module::planning {
         struct Config {
             /// @brief Threshold angle for executing getup, between torso z axis and world z axis
             float fallen_angle = 0.0f;
+            double start_delay = 0.0;    // seconds
+            bool _is_fallen    = false;  // To prevent against multiple instances
```
jpptm, line 43: Should this really be a config value? This looks more like a state that you should be getting somewhere else. Something like `message::behaviour::state::Stability` perhaps? And then using `With` inside `on` or something similar to this effect

tools/run.py
```diff
@@ -68,24 +68,24 @@ def run(args, use_gdb, trace, trace_output, **kwargs):
...
+    #     # Append log_path option if other options have been set
+    #     if "ASAN_OPTIONS" in env:
+    #         # Only append log_path if it hasn't already been set
+    #         if "log_path" not in env["ASAN_OPTIONS"]:
+    #             env["ASAN_OPTIONS"] = f"{env['ASAN_OPTIONS']}:log_path={os.path.join(log_path, 'asan.log')}"
+    #         # Check for shadow gap flag (needed to let cuda operations run correctly)
+    #         if "protect_shadow_gap" not in env["ASAN_OPTIONS"]:
+    #             env["ASAN_OPTIONS"] = f"{env['ASAN_OPTIONS']}:protect_shadow_gap=0"
+    #     else:
+    #         env.update({"ASAN_OPTIONS": f"log_path={os.path.join(log_path, 'asan.log')}:protect_shadow_gap=0"})
```
jpptm, line 88: Not part of this pr

tools/utility/dockerise/run.py
```diff
@@ -51,12 +51,12 @@ def _is_wsl1():
...
-    major = search.group(1)
-    minor = search.group(2)
+    # search = re.search("^(\d+)\.(\d+)\.\d+", kernel_release)
+    # major = search.group(1)
+    # minor = search.group(2)
 
-    if major is None or minor is None:
-        return False
+    # if major is None or minor is None:
+        # return False
```
jpptm, line 59: Revert, not part of this pr

tools/utility/dockerise/run.py
```diff
@@ -243,8 +243,8 @@ def _run(**kwargs):
 
         # If we are running in WSL 1 we need to translate our path to a windows one for docker.
         # Docker with WSL 2 doesn't need this as it supports binding paths directly from WSL into a container.
-        if _is_wsl1():
-            bind_path = subprocess.check_output(["wslpath", "-m", b.project_dir])[:-1].decode("utf-8")
+        # if _is_wsl1():
+        #     bind_path = subprocess.check_output(["wslpath", "-m", b.project_dir])[:-1].decode("utf-8")
```
jpptm, line 247: Revert, not part of this pr

module/planning/GetUpPlanner/src/GetUpPlanner.cpp
```diff
@@ -67,13 +70,36 @@ namespace module::planning {
...
-                // // Check if angle between torso z axis and world z axis is greater than config value
-                // Only emit if we're not already requesting a getup
+                // Check if angle between torso z axis and world z axis is greater than config value, check if already
+                // fallen
                 if (angle > cfg.fallen_angle && getup.run_state == RunState::NO_TASK) {
-                    emit<Task>(std::make_unique<GetUp>());
-                    log<DEBUG>("Execute getup");
+
+                    // If not already fallen
+                    if (!cfg._is_fallen) {
```
jpptm, line 78: I have a feeling that this shouldnt be a config value. You could also probably move some of this logic out of this block so its not too nested

jpptm (dismissed):

jpptm:
It would be nice to test on the real robot soon when we get the chance to do so

jpptm:
Also for detection, maybe we can use the other axes of rotations as well. From my understanding, the code currently only checks the angle between the world's z axis and the torso axis. Can we check the angle between the other axes as well to get a better idea of our orientation programatically?

module/planning/GetUpPlanner/src/GetUpPlanner.hpp
ysims: This looks like it should be a normal class variable. 
The GetUpPlanner holds the logic for if it's fallen or not, so it wouldn't take that information in externally.
It could be good to investigate the Stability state and if this module isn't setting it to fallen, then why not and should it be doing so.

tools/run.py
ysims: @vincenttumminello did you have issues running the code?

module/planning/GetUpPlanner/src/GetUpPlanner.hpp
ysims: Nevermind, the Stability is set to `FALLEN` in the GetUp module

ysims:
> Also for detection, maybe we can use the other axes of rotations as well. From my understanding, the code currently only checks the angle between the world's z axis and the torso axis. Can we check the angle between the other axes as well to get a better idea of our orientation programatically?

The GetUpPlanner (which has been changed in this PR) is just used to determine if the robot needs to get up/has fallen. The GetUp module checks the orientation to determine which way it should get up. I'm not sure if there is any benefit to using other axis in this stage.

ysims:
> It would be nice to test on the real robot soon when we get the chance to do so

I would agree with testing on the robot. While waiting is likely to fix the issue we were having with the robots, I'm not convinced what you are seeing in simulation is the same scenario as on the real robot. The real robot doesn't get up from the correct side when falling normally. I've not seen this happen in simulation, and if there is a problem in simulation around falling on the side it is more likely to be because [we tell the robot to get up from the front if it falls on its side due to a lack of any scripts to handle the side](https://github.com/NUbots/NUbots/blob/main/module/skill/GetUp/data/config/GetUp.yaml#L13-L18).

ysims:
I feel like there used to be some stuff where the accelerometer was checked to see if the robot was at rest. I can't see that in the code anymore. Maybe that would also help this issue.

tools/run.py
vincenttumminello: No I didn't run into any issues, at least not on my PC. I had issues running it in webots on the laptop in the lab but I'm not certain on the cause of that. It might've been this causing it

vincenttumminello:
No worries, I will work on the requested changes. Once done do I have to make a new PR or is there some way to just essentially re-upload the new code to this PR?

jpptm:
> No worries, I will work on the requested changes. Once done do I have to make a new PR or is there some way to just essentially re-upload the new code to this PR?

Edit your code then once youve addressed the changes just commit it again here, no need to make a new pr

module/planning/GetUpPlanner/src/GetUpPlanner.hpp
vincenttumminello: Yeah the stability is set to `FALLEN` in the GetUp module but that doesn't happen until after the delay elapses which is why I made a new variable which only prevents against multiple instances of the delay starting which ultimately prevented the robot from actually getting up. If you want me to, I could probably change where the stability is being set to `FALLEN` to be inside the GetUpPlanner module instead? Then I could use that instead of my own bool as suggested.

ysims: sounds like a good idea

TrentHouliston: Sounds like a perfect use case for the new Wait task right @ysims?

tools/run.py
TrentHouliston: it's because you have a cmakecache etc file in your main folder, vscode makes them sometimes when it tries to run cmake (which doesn't work) Then the b script gets confused as to where the build folder is, and it shows up as this incredibly strange ASAN key not found error

module/planning/GetUpPlanner/src/GetUpPlanner.hpp
```diff
@@ -39,8 +39,13 @@ namespace module::planning {
         struct Config {
             /// @brief Threshold angle for executing getup, between torso z axis and world z axis
             float fallen_angle = 0.0f;
+            double start_delay = 0.0;    // seconds
+            bool is_fallen    = false;  // To prevent against multiple instances
+            std::chrono::system_clock::time_point fall_time;
         } cfg;
 
+        struct StartGetUp {};
```
Tom0Brien, line 47: Shouldn't need this anymore

tools/run.py
Tom0Brien: Ahh yes, that error. This was a quick hack to fix that error showing up.

module/planning/GetUpPlanner/README.md
```diff
@@ -3,7 +3,8 @@
 ## Description
 
 The GetUpPlanner module is used to automatically trigger a get up when the robot has fallen over.
-It monitors the robot to see when it is lying down and has been relatively still for the past n frames before triggering a getup action.
+It monitors the angle of the robot with the world z axis to see when it falls past a set threshold angle (the "fallen_angle") before triggering a delay (inside the getup module) which then triggers a getup action once elapsed.
+The purpose of the delay is to allow the robot to settle to a final fallen position before any sensor readings are taken to more accurately determine which side it has fallen onto.
```
ysims, line 7: remove delay info

module/planning/GetUpPlanner/src/GetUpPlanner.cpp
```diff
@@ -58,6 +58,7 @@ namespace module::planning {
                     log<DEBUG>("Idle");
                     return;
                 }
+
```
ysims, line 61:
```suggestion
```

module/planning/GetUpPlanner/src/GetUpPlanner.cpp
```diff
@@ -67,13 +68,15 @@ namespace module::planning {
...
                 if (angle > cfg.fallen_angle && getup.run_state == RunState::NO_TASK) {
+                    // Emit GetUp task
                     emit<Task>(std::make_unique<GetUp>());
-                    log<DEBUG>("Execute getup");
+                    log<DEBUG>("Execute Getup");
+                }
+                else {
+                    // Otherwise do not need to get up so emit no tasks
                 }
-                // Otherwise do not need to get up so emit no tasks
```
ysims, line 76: revert all this stuff too

module/planning/GetUpPlanner/src/GetUpPlanner.hpp
```diff
@@ -49,4 +49,5 @@ namespace module::planning {
 
 }  // namespace module::planning
 
+
```
ysims, line 52:
```suggestion
```

module/skill/GetUp/src/GetUp.cpp
```diff
@@ -60,67 +61,73 @@ namespace module::skill {
             cfg.getup_upside_down = config["scripts"]["getup_upside_down"].as<std::vector<std::string>>();
         });
 
-        on<Provide<GetUpTask>, Needs<BodySequence>, With<Sensors>>().then(
-            [this](const RunReason& run_reason, const Sensors& sensors) {
+        on<Provide<GetUpTask>, Uses<BodySequence>, With<Sensors>>().then(
+            [this](const RunReason& run_reason, const Uses<BodySequence>& body, const Sensors& sensors) {
```
ysims, line 65: We should still have the `Needs` in here. Prevents it all running if it doesn't have priority to control the motors, and will run once it can take control.

module/skill/GetUp/src/GetUp.cpp
```diff
@@ -60,67 +61,73 @@ namespace module::skill {
...
                 switch (run_reason) {
```
ysims, line 66: I feel like this might look nicer if it is something like

```
if (run_reason == RunReason::NEW_TASK) {
  // Wait
}
else if (run_reason == RunReason::SUBTASK_DONE && !body.done) {
  // side check and script emit
  }
else if (run_reason == RunReason::SUBTASK_DONE) {
  // the Done emit etc
}
else {
  Continue
}
  ```
  
  The more nesting, the higher the cognitive load for someone reading

File: module/skill/GetUp/src/GetUp.hpp [hunk 17]
```diff
@@ -39,6 +39,7 @@ namespace module::skill {
     private:
         /// @brief Stores configuration values
         struct Config {
+            int delay_time; // milliseconds
```
ysims, line 42:
```suggestion
            /// @brief The time to delay before getting up, in milliseconds, to settle the sensors for accurate side detection 
            int delay_time = 0; // milliseconds
```
or something like that. good habits to get into, initialising variables and adding documentation directives (the `@brief`)

module/skill/GetUp/src/GetUp.cpp
```diff
@@ -60,67 +61,73 @@ namespace module::skill {
...
+                                emit<Task>(load_script<BodySequence>(cfg.getup_left));
+                            }
+                            else if (upright) {
+                                log<INFO>("Getting up from upright");
+                                emit<Task>(load_script<BodySequence>(cfg.getup_upright));
+                            }
+                            else if (upside_down) {
+                                log<INFO>("Getting up from upside_down");
+                                emit<Task>(load_script<BodySequence>(cfg.getup_upside_down));
+                            }
```
jpptm, line 122: It might be nicer to move this to a function while using an enum for the orientations and use a switch block format instead. In terms of condensing the logic a bit, would converting into a roll-pitch-yaw configuration allow us to simplify this?

module/skill/GetUp/src/GetUp.cpp
vincenttumminello: Not sure why exactly but it won't compile when I change Uses to Needs here, which is how I ended up with Uses.

ysims: I'd put in both

vincenttumminello: Do you mean like this:
`on<Provide<GetUpTask>, Needs<BodySequence>, With<Sensors>>().then([this](const RunReason& run_reason,
                                                                                 const Uses<BodySequence>& body,
                                                                                 const Sensors& sensors)`
                                                                                 
or like this:
`on<Provide<GetUpTask>, Needs<BodySequence>, Uses<BodySequence>, With<Sensors>>().then([this](const RunReason& run_reason,
                                                                                 const Uses<BodySequence>& body,
                                                                                 const Sensors& sensors)`
                                                                                 
Both compile and seem to work just fine??

module/skill/GetUp/src/GetUp.cpp
ysims: maybe, but outside the scope of this PR

module/skill/GetUp/src/GetUp.cpp
ysims: after talking to @TrentHouliston it should work having on Needs<BodySequence> and the Uses argument
I can have a look at it sometime to see why it isn't compiling but it should be possible. we want Needs otherwise the walk algorithm will keep running when something else takes over, such as the get up or kick, and when it gets control again it won't restart from the beginning.

ysims: I just switched `on<Provide<GetUpTask>, Uses<BodySequence>` to `on<Provide<GetUpTask>, Needs<BodySequence>` on your branch locally and it compiles.

vincenttumminello: Yeah cool, I just wasn't sure if it was going to be an issue having on<Needs> and then the Uses argument after. I thought they might've needed to match up. I will make that change and recommit

module/skill/GetUp/src/GetUp.cpp
```diff
@@ -60,72 +61,76 @@ namespace module::skill {
...
+            }
+            else if (run_reason == RunReason::SUBTASK_DONE) {
+                // When the subtask is done, we are done
+                log<INFO>("Finished getting up");
+                // We are standing
+                emit(std::make_unique<Stability>(Stability::STANDING));
+                emit<Task>(std::make_unique<Done>());
+            }
+            else {
+                // These shouldn't happen but if they do just continue doing the same thing
```
ysims, line 130:
```suggestion
                // This shouldn't happen but if it does just continue doing the same thing
```

module/skill/GetUp/src/GetUp.cpp
```diff
@@ -60,72 +61,76 @@ namespace module::skill {
...
-                    default: emit<Task>(std::make_unique<Continue>()); break;
+        on<Provide<GetUpTask>, Needs<BodySequence>, With<Sensors>>().then([this](const RunReason& run_reason,
+                                                                                 const Uses<BodySequence>& body,
+                                                                                 const Sensors& sensors) {
+            if (run_reason == RunReason::NEW_TASK) {
+                // Wait
+                log<DEBUG>("Delaying getup...");
+                emit<Task>(std::make_unique<Wait>(NUClear::clock::now() + std::chrono::milliseconds(cfg.delay_time)));
+            }
+            else if (run_reason == RunReason::SUBTASK_DONE && !body.done) {
```
ysims, line 72:
```suggestion
            // The initial Wait has completed
            else if (run_reason == RunReason::SUBTASK_DONE && !body.done) {
```

module/skill/GetUp/src/GetUp.cpp
```diff
@@ -60,72 +61,76 @@ namespace module::skill {
...
-                        emit(std::make_unique<Stability>(Stability::STANDING));
-                        emit<Task>(std::make_unique<Done>());
-                    } break;
-                    // These shouldn't happen but if they do just continue doing the same thing
-                    default: emit<Task>(std::make_unique<Continue>()); break;
+        on<Provide<GetUpTask>, Needs<BodySequence>, With<Sensors>>().then([this](const RunReason& run_reason,
+                                                                                 const Uses<BodySequence>& body,
+                                                                                 const Sensors& sensors) {
+            if (run_reason == RunReason::NEW_TASK) {
+                // Wait
```
ysims, line 68:
```suggestion
                // Wait so that sensors have time to settle
```

ysims (commented):
Click the resolve comment button when you finish addressing it

ysims (dismissed):
assuming it all works okay

module/skill/GetUp/data/config/GetUp.yaml
```diff
@@ -1,6 +1,8 @@
 # Controls the minimum log level that NUClear log will display
 log_level: INFO
 
+delay_time: 500 #  milliseconds
+
 scripts:
   getup_upright:
     - "Stand.yaml"
```
ysims, line 8: Just below this is a getup left and getup right with just stand up front script running. It might be better to change the logic to just wait in these instances rather than run any script, as we do end up on front or back when it falls to the side.

ysims (approved):

ysims:
Code formatting failed

ysims (approved):

PR #1525, opened 2025-01-31
Skip Continues in the Behaviour extension rather than the Director

ysims:
Instead of emitting the task pack to Director and getting `run_task_pack` to check for `Continue`s, just do the check in the Behaviour extension before it reaches the Director. 

The purpose of this is to reduce the number of things the Director thread needs to do, as it is running on one thread with a lot to do.

### DIFF
--- a/shared/extension/Behaviour.hpp
+++ b/shared/extension/Behaviour.hpp
@@ -38,7 +38,7 @@
 #include "behaviour/RunReason.hpp"
 #include "behaviour/TaskData.hpp"
 #include "behaviour/commands.hpp"
-#include "behaviour/unique_tasks.hpp"
+#include "behaviour/task_types.hpp"
 
 namespace extension::behaviour {
 
--- a/shared/extension/Behaviour.hpp
+++ b/shared/extension/Behaviour.hpp
@@ -38,6 +38,7 @@
 #include "behaviour/RunReason.hpp"
 #include "behaviour/TaskData.hpp"
 #include "behaviour/commands.hpp"
+#include "behaviour/unique_tasks.hpp"
 
 namespace extension::behaviour {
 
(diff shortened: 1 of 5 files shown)


shared/extension/behaviour/ProviderScope.hpp
```diff
@@ -59,6 +60,21 @@ namespace extension::behaviour {
...
+            for (const auto& task : tasks) {
+                if (task.type == typeid(::extension::behaviour::Continue)) {
+                    // Tell the user if there is a Continue task with other tasks
+                    if (tasks.size() > 1) {
+                        NUClear::log<NUClear::LogLevel::WARN>(
+                            "Continue task was emitted with other tasks, the other tasks will be ignored");
+                    }
+
+                    // Do nothing else
+                    return;
```
TrentHouliston, line 74: You need to also clear the tasks list so that next time there isn't still a continue in there (also maybe add a unit test to make sure that doesn't happen)

shared/extension/Behaviour.hpp
```diff
@@ -38,6 +38,7 @@
 #include "behaviour/RunReason.hpp"
 #include "behaviour/TaskData.hpp"
 #include "behaviour/commands.hpp"
+#include "behaviour/unique_tasks.hpp"
```
TrentHouliston: maybe just call it tasks or task types?

shared/extension/behaviour/unique_tasks.hpp
```diff
@@ -0,0 +1,77 @@
...
+         * Create a new Wait task with the time at which the Provider should run again.
+         *
+         * @param time the time at which the Provider should run again
+         */
+        explicit Wait(const NUClear::clock::time_point& time) : time(time) {}
+    };
+
+}  // namespace extension::behaviour
+
+
```
TrentHouliston:
```suggestion
```

TrentHouliston (approved):

PR #1526, opened 2025-02-03
Filter robot-to-robot messages by team 

ysims:
Currently we don't filter out other team's messages, which could cause problems if the other team is using the protocol. It is particularly a problem when our robots play each other. The protocol doesn't hold a team number, instead it holds a colour. 

This PR also renames all CYAN to BLUE and MAGENTA to RED per the changes to the rules on the colours, and removes the unnecessary `Data` message nested inside `GameState`.

### DIFF
--- a/module/input/GameController/src/GameController.cpp
+++ b/module/input/GameController/src/GameController.cpp
@@ -44,19 +44,18 @@ namespace module::input {
     using message::input::GameEvents;
     using message::input::GameState;
     using message::support::GlobalConfig;
-    using TeamColour      = message::input::GameEvents::TeamColour::Value;
-    using Score           = GameEvents::Score;
-    using GoalScored      = GameEvents::GoalScored;
-    using Penalisation    = GameEvents::Penalisation;
-    using Unpenalisation  = GameEvents::Unpenalisation;
-    using Coach_message   = GameEvents::CoachMessage;
-    using HalfTime        = GameEvents::HalfTime;
-    using BallKickedOut   = GameEvents::BallKickedOut;
-    using KickOffTeam     = GameEvents::KickOffTeam;
-    using GamePhase       = GameEvents::GamePhase;
-    using GameMode        = GameEvents::GameMode;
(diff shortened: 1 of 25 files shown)


module/network/RobotCommunication/src/RobotCommunication.cpp
```diff
@@ -87,31 +87,54 @@ namespace module::network {
...
-                            RoboCup incoming_msg = NUClear::util::serialise::Serialise<RoboCup>::deserialise(payload);
+                                // Check if the incoming message is from the same team
+                                bool own_team_message = team_colour == incoming_msg.current_pose.team;
 
-                            // filter out own messages
-                            if (global_config.player_id != incoming_msg.current_pose.player_id) {
-                                emit(std::make_unique<RoboCup>(std::move(incoming_msg)));
-                            }
-                        });
+                                // Filter out messages from ourself and from other teams
```
JohanneMontano, line 133:
```suggestion
                                // Filter out messages from ourselves and from other teams
```

module/input/GameController/src/GameController.cpp
```diff
@@ -44,19 +44,18 @@ namespace module::input {
...
+    using GoalScored     = GameEvents::GoalScored;
+    using Penalisation   = GameEvents::Penalisation;
+    using Unpenalisation = GameEvents::Unpenalisation;
+    using Coach_message  = GameEvents::CoachMessage;
+    using HalfTime       = GameEvents::HalfTime;
+    using BallKickedOut  = GameEvents::BallKickedOut;
+    using KickOffTeam    = GameEvents::KickOffTeam;
+    using GamePhase      = GameEvents::GamePhase;
+    using GameMode       = GameEvents::GameMode;
+    using PenaltyReason  = GameState::PenaltyReason;
```
JohanneMontano, line 58: Could pass as game states as well

JohanneMontano (commented):
Some small comments but otherwise lgtm.

JohanneMontano (approved):
Sorry for being annoying HAHA lgtm

PR #1527, opened 2025-02-17
Falling Relax

vincenttumminello:
- Uncommented the emit line in FallRecovery.cpp so that the falling relax script is run
- Made 3 scripts with differing degrees of relaxation (in case quick on the fly changes were needed/wanted): 
        - Relax.yaml will relax everything
        - RelaxHead.yaml will relax only the head
        - RelaxUpperBody.yaml will relax the upper body

### Reviewers, Note

This hasn't been tested on the robots yet. 

### Things left to do

Probably should be tested

G-Motion

### DIFF
--- a/shared/utility/skill/scripts/nugus/RelaxHead.yaml
+++ b/shared/utility/skill/scripts/nugus/RelaxHead.yaml
@@ -1,0 +1,10 @@
+[
+  {
+    "duration": 25,
+    "targets":
+      [
+        { "id": "HEAD_YAW", "position": 0, "gain": .NaN, "torque": 0 },
+        { "id": "HEAD_PITCH", "position": 0, "gain": .NaN, "torque": 0 },
+      ],
+  },
+]
--- a/shared/utility/skill/scripts/webots/RelaxUpperBody.yaml
+++ b/shared/utility/skill/scripts/webots/RelaxUpperBody.yaml
@@ -1,0 +1,34 @@
+- duration: 25
+  targets:
+    - id: R_SHOULDER_PITCH
+      position: 0
+      gain: .NaN
+      torque: 0
+    - id: L_SHOULDER_PITCH
+      position: 0
+      gain: .NaN
+      torque: 0
+    - id: R_SHOULDER_ROLL
+      position: 0
+      gain: .NaN
+      torque: 0
+    - id: L_SHOULDER_ROLL
+      position: 0
+      gain: .NaN
+      torque: 0
+    - id: R_ELBOW
+      position: 0
+      gain: .NaN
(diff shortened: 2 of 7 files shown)


shared/utility/skill/scripts/nugus/RelaxHead.yaml
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    "duration": 25,
+    "targets":
+      [
+        { "id": "HEAD_YAW", "position": 0, "gain": .NaN, "torque": 0 },
+        { "id": "HEAD_PITCH", "position": 0, "gain": .NaN, "torque": 0 },
+      ],
+  },
+]
```
ysims, line 10: change this from json to yaml

shared/utility/skill/scripts/webots/RelaxUpperBody.yaml
```diff
@@ -0,0 +1,34 @@
...
+      gain: .NaN
+      torque: 0
+    - id: HEAD_YAW
+      position: 0
+      gain: .NaN
+      torque: 0
+    - id: HEAD_PITCH
+      position: 0
+      gain: .NaN
+      torque: 0
```
ysims, line 34: does this work in webots? I seem to recall issues sending nan but can't recall exactly.

ysims (commented):
this would be good to get in and check out during practice games. 
definitely test on the robot first

shared/utility/skill/scripts/webots/RelaxUpperBody.yaml
vincenttumminello: I'm not sure, it's hard to tell what a relaxed joint looks like in webots. Although unlike with the main branch when I ./run this, the robot immediately falls over so maybe?

ysims: Webots might give you console output - something about nan not being valid. The robot shouldn't fall over immediately.

vincenttumminello: I'm not getting any errors on the console in webots so I'm unsure what's going on.

ysims (approved):
Code all looks good and looks like it should do the job. This is good to merge, given it works as expected.

ysims (changes requested):
There were issues with this when we ran it, I think one problem might be that the falling relax runs the script as a BodySequence, which takes control of the whole body. So when it tries to relax at the wrong time, the whole robot freezes.

Claegtun (approved):
Seems good to me, unless anyone can spot anything else in the next few hours.

PR #1528, opened 2025-02-23
Implement NUC-NUSense handshaking

jpptm:
This pull request makes the NUC send a configurable message to NUSense during the start of a binary. This change is to also make NUSense turn the dxl_power on only when the two have confirmed that they can communicate back to back.


module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -85,11 +86,74 @@ namespace module::platform::NUSense {
...
+                    send_packet(servo_targets);
+                });
+
+            NUC_packet_catcher.disable();
+        });
+
+        on<Startup>().then([this] {
+            // Start the handshake watchdog
+            emit<Scope::WATCHDOG>(ServiceWatchdog<HandshakeWatchdog>());
+        });
```
ysims, line 146: Watchdogs are active as soon as they're defined, you don't emit a service message to start them or anything. They just run once the time has elapsed and no service has happened. This would prolong the watchdog from running, which doesn't seem to be what we want to do.

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -85,11 +86,74 @@ namespace module::platform::NUSense {
...
+                            target.id,
+                            (target.position - nugus.servo_offset[target.id]) * nugus.servo_direction[target.id],
+                            target.gain,
+                            target.torque);
+                    }
+
+                    send_packet(servo_targets);
+                });
+
+            NUC_packet_catcher.disable();
```
ysims, line 140:
```suggestion
            NUC_packet_catcher =
                on<Trigger<ServoTargets>, Sync<HardwareIO>>().then([this](const ServoTargets& commands) {
                    // Copy the data into a new message so we can use a duration instead of a timepoint
                    // and take the offsets and switch the direction.
                    auto servo_targets = SubcontrollerServoTargets();

                    // Change the timestamp in servo targets to the difference between the timestamp and now
                    // Take away the offset and switch the direction if needed
                    for (auto& target : commands.targets) {
                        servo_targets.targets.emplace_back(
                            target.time - NUClear::clock::now(),
                            target.id,
                            (target.position - nugus.servo_offset[target.id]) * nugus.servo_direction[target.id],
                            target.gain,
                            target.torque);
                    }

                    send_packet(servo_targets);
                }).disable();
```
right?

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -85,11 +86,74 @@ namespace module::platform::NUSense {
             // Tell the stream reactor to connect to the device
             emit(std::make_unique<ConnectSerial>(device, baud));
 
+            // TODO: Apply the servo offsets within NUSense instead
```
ysims, line 89: so this PR just sends the information right, it doesn't actually have NUSense using it yet?

jpptm: hence its a draft atm. Will put it up when ive got the code for the firmware

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -85,11 +86,67 @@ namespace module::platform::NUSense {
...
+                    // and take the offsets and switch the direction.
+                    auto servo_targets = SubcontrollerServoTargets();
+
+                    // Change the timestamp in servo targets to the difference between the timestamp and now
+                    // Take away the offset and switch the direction if needed
+                    for (auto& target : commands.targets) {
+                        servo_targets.targets.emplace_back(
+                            target.time - NUClear::clock::now(),
+                            target.id,
+                            (target.position - nugus.servo_offset[target.id]) * nugus.servo_direction[target.id],
```
Claegtun, line 132: If you are going to send the offsets and the directions in the handshake message and deal with them on the subcontroller-side -- which is the intended vision --, then you should remove them here.

module/platform/NUSense/HardwareIO/src/HardwareIO.cpp
```diff
@@ -210,27 +267,6 @@ namespace module::platform::NUSense {
...
-                servo_targets.targets.emplace_back(
-                    target.time - NUClear::clock::now(),
-                    target.id,
-                    (target.position - nugus.servo_offset[target.id]) * nugus.servo_direction[target.id],
-                    target.gain,
-                    target.torque);
-            }
-
-            send_packet(servo_targets);
-        });
```
Claegtun, line 232: Why is this removed?

Claegtun: Just wondering.

Claegtun: Never mind.

Claegtun (approved):
Seems good to me.

PR #1529, opened 2025-03-02
Add ability to use VisualMesh with GPU in Docker + Change webots image format to RGBA

Tom0Brien:
This PR make required changes to use opencl engine with the VisualMesh which increases runtime performance in webots a considerable amount. The user will need to have a gpu/igpu which supports opencl.

This requires the following PR on the webots side : https://github.com/NUbots/NUWebots/pull/103

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -435,6 +435,8 @@ RUN chmod 440 /etc/sudoers.d/user
 ARG user_uid=1000
 ARG user_gid=$user_uid
 ARG dialout_gid=20
+ARG video_gid=44
+ARG render_gid=110
 
 # useradd -G is not sufficient to add the nubots user to the dialout
 # group as docker manages to lose supplementary groups when changing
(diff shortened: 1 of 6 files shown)


ysims:
The GPU change doesn't need the format change does it? Maybe add it into the title if you want it in one PR so the change is easier to find.

This would still have the issue where we not longer support the RoboCup world right? We do have our own fork [NUbots/hlvs_webots](https://github.com/NUbots/hlvs_webots) where we could update this as well.

Tom0Brien:
> The GPU change doesn't need the format change does it? Maybe add it into the title if you want it in one PR so the change is easier to find.
> 
> 
> 
> This would still have the issue where we not longer support the RoboCup world right? We do have our own fork [NUbots/hlvs_webots](https://github.com/NUbots/hlvs_webots) where we could update this as well. 

The image format change is needed for VM to work with opencl engine. We should form and update the image format for the Robocup world/controller.

ysims:
Everyone should already be using the fork, it has some ease of use changes and is in the nubook guide. It's probably fine to change.

jpptm:
This looks goated, testing with `nvidia-smi` and confirming we're getting gpu activity would be great. I can try soon when uni stops strangling me perhaps

ysims (approved):
When you merge these can you also merge 
https://github.com/NUbots/hlvs_webots/pull/1

PR #1530, opened 2025-03-05
Added restriction to not check renamed files

Amadeus0917:
This PR excludes renamed files from the formatting process by implementing a filter that prevents them from being checked.

### DIFF
--- a/tools/format.py
+++ b/tools/format.py
@@ -30,6 +30,9 @@ import datetime
 import multiprocessing
 import os
 import shutil
+
+# Added for formatting
+import subprocess
 import sys
 import tempfile
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


Crafty15:
Nice! Is this ready for review? If so, you can edit the main comment at the top to include an explanation of what the PR does, as well as anything you think might be useful for reviewers to know. You can look at other pull requests to get an idea. You can also remove the template text. 
If it's not ready yet but you just wanted the PR in so that the scripts run, you can convert it to a draft PR using the link just under where there reviewers are listed on the top right. 
I'm not really sure how we can test this, but i don't think it will hurt anything to just get it in and see what happens. Maybe @ysims will have an idea? 
Let me know if you need a hand with anything :)

Amadeus0917:
It is ready for review, though I only tested it to see if it runs. It didn't give me an error but I'm not really sure how to test if the code I added is working as intended.

tools/format.py
```diff
@@ -48,42 +51,78 @@
...
+        ["git", "diff", "--name-only", "--diff-filter=ACMRTUXB", "HEAD"], stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    )
+
+    if result.returncode == 0:
+        # Split output into lines, each line is a file path
+        modified_files = result.stdout.decode().splitlines()
+        return modified_files
+    else:
+        # If the diff command failed, return an empty list
+        return []
```
jpptm, line 67:
```suggestion
return result.stdout.decode().splitlines() if not result.returncode else []
```

tools/format.py
```diff
@@ -48,42 +51,78 @@
...
+
+
+# Filter the modified files against the include patterns
+def filter_files(modified_files, include_patterns):
+    included_files = []
+    for file in modified_files:
+        for pattern in include_patterns:
+            if fnmatch(file, pattern):  # Use fnmatch correctly
+                included_files.append(file)
+                break  # No need to check further patterns once it's matched
```
jpptm, line 77:
```suggestion
            # Use fnmatch correctly
            if fnmatch(file, pattern):
                included_files.append(file)
                # No need to check further patterns once it's matched
                break
```

tools/format.py
```diff
@@ -48,42 +51,78 @@
...
+    "**/Dockerfile",
+]
+
+# Get the list of modified files
+modified_files = get_modified_files()
+
+# Filter modified files against the include patterns
+files_to_format = filter_files(modified_files, include_patterns)
+
 # TODO: Fix this
```
jpptm, line 103: If this is fixed now, this should just be removed
```suggestion
```

tools/format.py
```diff
@@ -48,42 +51,78 @@
...
+            "{path}",
+        ]
+    ],
+    "include": files_to_format,
+    "exclude": [
+        "shared/utility/motion/splines/*",
+        "shared/utility/platform/models/nugus/nugus.proto",
+        "nusight2/src/assets/robot-models/nugus/nugus.proto",
+        "module/support/optimisation/NSGA2Optimiser/src/nsga2",
+    ],  # TODO exclude files that are not ours
```
jpptm, line 124:
```suggestion
    ],  # TODO exclude files that are not ours
```
Do this TODO if possible and simple enough. Can we do a naive exclusion check? For example, files that are not ours are not going to be in the `include_patterns`

jpptm: if it is done, remove it

jpptm:
> It is ready for review, though I only tested it to see if it runs. It didn't give me an error but I'm not really sure how to test if the code I added is working as intended.

You could rename a file, run the formatter and check if it is being included or not. I think it should print what files are being formatted iirc

ysims (changes requested):
I'm not sure if this fixes the problem we have, and maybe breaks some functionality.

One of the functionalities of the format script is a command line argument `-a` to run the formatter on all of the files. There are two versions because formatting all of the files takes a long time, and for most cases only new/changed files need to be fixed.  

From what I can tell, this PR changes the license formatting to only format modified (and not renamed) files, so the `-a` functionality wouldn't work anymore for license formatting? We also still want to format renamed files, otherwise the tool might miss some formatting.

The problem we have isn't that files are formatted when they shouldn't be - it's that when a file is formatted, sometimes it gets the wrong year, and CI before and after squashing has differing years. This PR wouldn't stop that from happening - if a file is modified where the year fails to resolve correctly across CI checks, the problem still is there. 

The `_get_history_dates` function is where you want to make changes. Ignoring renaming of files when getting the year is a good idea. Even if it's not the most correct, it'll at least be approximately correct and won't have issues between squashing. 

Actually testing if it works will be tricky, and as @Crafty15 said, we probably just want to merge and see if it works. 
You should also run `./b format -a` to apply the new license formatting. Since we have not had it running in CI for a while, I expect there to be some files that need the license. Modifying the way the dates are resolved will probably change some existing dates. 

Feel free to DM me or come chat in the lab @Amadeus0917 if you have any questions.

tools/format.py
Crafty15: We are trying to figure out what it means by "files that are not ours". @ysims or @jpptm, any ideas here?

Crafty15:
> _get_history_dates

Where is this function? I can't find it :(

ysims:
> > _get_history_dates
> 
> Where is this function? I can't find it :(

https://github.com/NUbots/NUbots/blob/main/tools/format.py#L119
or in this PR in the files changed tab https://github.com/NUbots/NUbots/pull/1530/files#diff-6d4e44640b9ec14c2be71e355057398f8af988c5ec70af2af77d7826fde4f7efR154

tools/format.py
ysims: Anything in the include files list that shouldn't be included, because they're not our files to license. eg the nsga2 algorithm files aren't ours, so we exclude them. Same with Bit-Bot's spline files, as they are Bit-Bots owned. It looks to me like it's already done, it might just be left over from when @TrentHouliston first wrote this and wouldn't have known what should be excluded.

TrentHouliston: Correct, anything that we copy and paste from somewhere else into our code from somewhere else with another licence shouldn't be excluded

TrentHouliston:
It's still possible to get the code to actually show the correct date if anyone is interested. 

The big problem with the current code is that it doesn't track rename files properly.
Specifically it has an issue where if you have

`a.hpp`

and then in a commit you add
`b.hpp`

and b is similar to a it'll treat b as a rename of a even when a still exists.

Git itself is able to handle this and know when it's a rename because the other file still exists. It's just that whatever method I chose didn't do that and I didn't realise until after I committed this.

Fixing it means updating finding the history of the file to not include a rename if the other file still exists!

PR #1531, opened 2025-03-10
Remove the QuinticWalk

ysims:
QuinticWalk was superceded by Walk a while ago. They are very similar so there is no need to keep QuinticWalk in.
This PR just deletes the module.


Tom0Brien:
There should be a lot of utilities which can be removed as well alongside this

Tom0Brien (approved):

PR #1532, opened 2025-03-10
Director-fy the old Inverse Kinematics Kick (IKKick)

ysims:
This PR gets the IKKick compiling again, in preparation for CI having unused modules building. Note that it doesn't necessarily get it working.

In doing this, it needed to be changed from the old Controller system to the newer Director behaviour system. 
Old camelCase has been changed to snake_case, and config structs have been added. 

This definitely doesn't aim to make it actually usable, just compiling. It could be good to check it out. Some things to work on would be

- Fix up variable names to our conventions (see NUbook maths page)
- Verify the maths makes sense
- Separate the classes out into separate files so it's not so cluttered
- Check the director logic makes sense for ending the kick
- Fill in the readme
- Test in Webots


Tom0Brien:
IMHO this module can be removed

TrentHouliston (approved):
If the goal is to get this compiling then this is progress. But I agree with Tom that deleting this module might be better. I think with the director and our spline code we could make something better with about as much effort as fixing this one

ysims:
A team member is tasked with investigating the IK Kick, so I can't delete it. But I need something done about it so that I can finish off the unused modules in CI.

PR #1533, opened 2025-03-10
Allow parent folder data files and move HardwareIO.yaml to be shared

ysims:
The purpose of this PR is to facilitate getting unused modules compiling in CI.

This PR modifies the CMake functionality for getting data files in a module, to look for data files in parent folders as well (stopping at the module folder). 
This is so that we can make config files that are shared, in a way, across multiple modules. It only really works for sharing across siblings, which is probably a nice enough restriction to have.

This is then used to make `HardwareIO.yaml` a shared config file across OpenCR and NUSense HardwareIO modules. Benefits:

- We can build both modules at the same time, without getting a duplicate yaml error
- Don't need to update both config files when we tune the robot's zero positions

The data folder looks a little ugly in the platform folder but oh well. I also decided that the overhead of rechecking folders is minimal and better than trying to track what folders have been visited, to keep things simple. 

`NUSense.yaml` and `OpenCR.yaml` have been made to contain values relevant to the specific subcontroller.


Tom0Brien:
Not really super related or a big deal but I'm curious, do we actually use the data folder anywhere other than to store the config folder ?

TrentHouliston (commented):
Not a fan of this change, the idea of modules was to prevent them from leaking and being able to access parts of other systems without having modules in between.

Having data files in parent folders somewhat breaks the modular nature of the systems. You could change a yaml file for one modules configuration and not worry about breaking another systems one.

Moving to a message based configuration with a common configuration message making module would be better since then the configuration message itself would type check that nothing is broken

PR #1536, opened 2025-03-23
Switch to use mahony filter yaw estimate

Tom0Brien:
This PR switches the odometry to use mahony filter. 

Mahony filter yaw in webots is much better... someone should test on the real robot plz.

image


ysims:
I've just had a look at it, and while yaw looks okay when moving, there is a constant drift.

Tom0Brien:
> I've just had a look at it, and while yaw looks okay when moving, there is a constant drift.

Hmm okay, I will investigate when I have access to real hardware. Once potential simple solution could be to the fuse both the anchor point method and the mahony filter yaw estimates.

ysims:
Yep, happens when stationary.

Tom0Brien:
Tested this on Frankie and I couldn't see any problems with drift

ysims (approved):
Should keep an eye on it then, might be a robot with a dodgy imu.

PR #1540, opened 2025-03-30
Add goalpost as obstacle for walk path planner

nmthais:
Currently we have a  module for avoiding obstacles which was made for avoiding robots, so adding the goal post as an obstacle partly resolves robots running into goal posts (when robot is running into balls and that lines up with goalpost, it WONT dodge it)


Tom0Brien:
Rather than using the goal post vision measurement, I think it would be better to transform the known goal post positions into the robot space (using localisation) and add them as obstacles. This way we don't rely on seeing/detecting the goal posts to avoid them.

module/planning/PlanWalkPath/src/PlanWalkPath.cpp
```diff
@@ -108,11 +107,14 @@ namespace module::planning {
...
            With<Sensors>,
-           Optional<With<Goals>>>()
+           With<Field>,
+           With<FieldDescription>>()
             .then([this](const WalkTo& walk_to,
                          const std::shared_ptr<const Robots>& robots,
                          const Sensors& sensors,
-                         const std::shared_ptr<const Goals>& goals) {
+                         const Field& field,
+                         const FieldDescription& fieldDesc
```
Tom0Brien, line 116:
```suggestion
                         const FieldDescription& field_desc
```

module/planning/PlanWalkPath/src/PlanWalkPath.cpp
```diff
@@ -246,6 +242,26 @@ namespace module::planning {
...
+
+        list_goalposts.emplace_back(Eigen::Vector3d(fieldDesc.goalpost_own_l.x(),fieldDesc.goalpost_own_l.y() , 0));
+        list_goalposts.emplace_back(Eigen::Vector3d(fieldDesc.goalpost_own_r.x(),fieldDesc.goalpost_own_r.y() , 0));
+        list_goalposts.emplace_back(Eigen::Vector3d(fieldDesc.goalpost_opp_l.x(),fieldDesc.goalpost_opp_l.y() , 0));
+        list_goalposts.emplace_back(Eigen::Vector3d(fieldDesc.goalpost_opp_r.x(),fieldDesc.goalpost_opp_r.y() , 0));
+
+        // Loop through this list, converting all of the goalpost position from field space to robot space, then add them back in all_obstacles list
+        for(const auto& goalpost_pos : list_goalposts){
+            auto rFWw = field.Hfw.inverse() * goalpost_pos;        // Convert goalpost position from field space to world space
+            auto rWRr = sensors.Hrw * rFWw;                        // Convert that to robot space
```
Tom0Brien, line 257:
```suggestion
            auto rFRr = sensors.Hrw * rFWw;                        // Convert that to robot space
```

Tom0Brien (changes requested):
Looks good! Just some small things

nmthais:
thanks Tom !!!

module/planning/PlanWalkPath/data/config/PlanWalkPath.yaml
```diff
@@ -30,4 +30,4 @@ backwards_vector: [-1.0, 0.001] # Vector to walk backwards
 acceleration_multiplier: 0.3
 
 # Radius around obstacles to avoid
-obstacle_radius: 0.6
+obstacle_radius: 1
```
Tom0Brien, line 33: This seems really large. Has this been tested?

PR #1541, opened 2025-03-30
Use CPU in vision configs for Webots

ysims:
Fix to main, currently having these set to GPU gives two separate (but I assume related) errors.

Visual Mesh
```
VisualMesh (ノಠ益ಠ)ノ彡┻━┻: Error selecting an OpenCL device: Invalid device
```

YOLO
```
Yolo (ノಠ益ಠ)ノ彡┻━┻: Check 'false' failed at src/inference/src/core.cpp:131:
Check 'm_default_contexts.find(device_id) != m_default_contexts.end()' failed at src/plugins/intel_gpu/src/plugin/plugin.cpp:248:
[GPU] Context was not initialized for 0 device
```

I assume there is extra set up needed to get this working which is not documented. It might be good to have it set to GPU since I think the user should have a usable GPU for this if they can run Webots? 

Until then, it would be good to revert these so team members can develop without getting these errors.

### DIFF
--- a/module/vision/VisualMesh/data/config/webots/VisualMesh.yaml
+++ b/module/vision/VisualMesh/data/config/webots/VisualMesh.yaml
@@ -10,7 +10,7 @@ cameras:
   left_camera:
     concurrent: 2
     network: config/networks/WebotsNetwork.yaml
-    engine: opencl
+    engine: cpu
     cache_directory: "./vision_cache/"
     classifier:
       height: [0.5, 1.0]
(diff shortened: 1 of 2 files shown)


Tom0Brien:
Hmm, it should just work out of the box. Can you try running `clinfo` inside the container?

ysims:
> Hmm, it should just work out of the box. Can you try running `clinfo` inside the container?

[nubots@docker NUbots]$ clinfo
Number of platforms                               0

Tom0Brien (approved):

PR #1544, opened 2025-04-02
Update the Dockerfile to November 2024

ysims:
This PR updates the Dockerfile to use 17/11/2024 version of Arch Linux.

Benefits of this PR
- With a more recent version, shouldn't see it failing to download from the mirrors
- Keeping libraries up to date fixes bugs (eg the infamous Visual Mesh/Image Compressor memory leak)
- On that note... OpenVINO doesn't seg fault in debug anymore
- Easier to update in the future if anything comes up where this is more urgent

Changes
- Some libraries updated, older gcc is used because of some compile issues
- Finding Python in CMake is a bit different now, `find_package(PythonInterp 3 REQUIRED)` is deprecated and `find_package(Python3 REQUIRED)` is in
- Module loaders in `pkgutil` are deprecated, `importlib` used instead
- Anything using fixed size integer types needs `#include <cstdint>` now
- OpenVINO's inference engine seems to be gone/consolidated, and `runtime` folder doesn't exist anymore
- Regex needs to be specified specifically as a regex in the string

Tested using `webots/robocup` in release and debug. 

Some things I want to check
- [x] Can we leave LLVM on `14.0.6`? Yes!
- [x] Do we need the warning suppression `-Wno-stringop-overflow`?
- [x] SyncWrite??


module/platform/OpenCR/HardwareIO/src/dynamixel/v2/SyncWrite.hpp
```diff
@@ -65,15 +66,20 @@ namespace dynamixel::v2 {
...
             , size(sizeof(T))
-            , data(data)
-            , checksum(calculate_checksum(this)) {}
+            , data{}
+            , checksum(0) {
+            for (size_t i = 0; i < N; ++i) {
+                const_cast<SyncWriteData<T>&>(data[i]) = data_arr[i];
+            }
+            const_cast<uint16_t&>(checksum) = calculate_checksum(this);
+        }
```
ysims, line 82: :worried:

docker/Dockerfile
```diff
@@ -122,7 +122,7 @@ RUN --mount=type=bind,source=usr/local/toolchain,target=/usr/local/toolchain \
     python /usr/local/toolchain/generate_${platform}_toolchain.py --prefix ${prefix}
 
 # LLVM and Clang
-RUN install-from-source https://github.com/llvm/llvm-project/archive/llvmorg-14.0.6.tar.gz \
+RUN install-from-source https://github.com/llvm/llvm-project/archive/refs/tags/llvmorg-14.0.5.tar.gz \
```
ysims, line 125: :thinking:

CMakeLists.txt
```diff
@@ -107,6 +107,8 @@ if(CI_BUILD)
   # -warnings-as-errors=*) * set(CMAKE_C_CLANG_TIDY ${CMAKE_C_CLANG_TIDY} -warnings-as-errors=*)
 endif()
 
+add_compile_options(-Wno-stringop-overflow)
```
ysims, line 110: :confused:

module/vision/Yolo/CMakeLists.txt
```diff
@@ -25,13 +25,13 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 ]]
 find_package(OpenCV REQUIRED)
-set(OpenVINO_DIR /usr/local/runtime/cmake)
-set(InferenceEngine_DIR /usr/local/runtime/cmake)
+set(OpenVINO_DIR /usr/local/lib/cmake)
+# set(InferenceEngine_DIR /usr/local/lib/cmake)
```
Tom0Brien, line 29: I think we can remove this InferenceEngine stuff if deprecated now

ysims:
CI passes :tada:

Bidski (approved):
Consider updating catch2 to allow use of automatic test registration and assigning labels to registered tests

ysims:
I'm gonna do the catch update in another PR because I'm having docker caching issues and need to push my local built image for this PR to main

PR #1546, opened 2025-04-06
Update NUsight GitHub Actions workflow to use the Docker image

ysims:
The purpose of this PR is to fix NUsight's CI as it's currently broken. 

The CI issue does not come up locally, as it works in the Docker environment. However, the CI workflow does not use Docker and instead installs node and runs NUsight from the `nusight2` folder. Since intended workflow is to run through Docker with the `b` script, rather than directly running NUsight, this PR updates the NUsight workflow to use Docker and the `b` script. 

CI passes and all looks good :shipit:


JosephusPaye (approved):
LGTM!

PR #1547, opened 2025-04-11
Update catch2 version

ysims:
Updated the version number to the latest and fixed compile issue with random not being included.


ysims:
Hm, CI fails because we've run out of memory. We do have a Jira task for auditing our libraries - we might be able to get rid of ones we don't need. I'll investigate.

Tom0Brien (approved):
LGTM!

ysims:
I guess this is now grabbing the image I pushed up for this PR. Not sure why it didn't do it when I initially pushed it up and reran checks, may some delay. The out of memory issue should be fixed in https://github.com/NUbots/NUbots/pull/1552.  Checks pass now in this PR.

Tom0Brien (approved):

PR #1550, opened 2025-04-20
Command line args for setting webots port, player id and team id to facilitate quicker team development

ysims:
I started working on team functionality and needed to test with more than one robot, and the `webots build` tool is very annoying, tedious and mostly just good for the competition with the docker image ready to push up. It kept downloading the image every time I rebuilt and overall is doing way too much when we have an image ready to go (the generic image) and we just need the containers to have different config. 

The most simple solution I could find was to pass in arguments and edit the config files if the arguments are used. Nothing extra really needs to be done now to run multiple robots, just add in the extra arguments like

`./b run webots/robocup --webots_port 1002 --player_id 2 --team_id 12`

With us increasingly needing to do multi robot development, this seems super important to streamline the process.

Will look into NUbook PR tomorrow. Just tested with two robots and looked good.


tools/run.py
```diff
@@ -115,6 +118,46 @@ def run(args, use_gdb, trace, trace_output, **kwargs):
...
+        return
+
+    # Update the config
+    config[key] = value
+
+    # Write the updated config back to the file
+    with open(yaml_file, "w") as file:
+        yaml.dump(config, file)
+
+def update_configs(webots_port, player_id, team_id):
```
Tom0Brien, line 145: Rename to something like `update_webots_configs`

Tom0Brien (changes requested):
Maybe also would be good to add a new tool to easily start multiple at once. And then having a NUbook page would be good :D

tools/run.py
```diff
@@ -115,6 +118,46 @@ def run(args, use_gdb, trace, trace_output, **kwargs):
...
+    # Change into the config directory
+    os.chdir("./config")
```
Tom0Brien, line 147: nit, but is changing directory really needed? Could just add `config/` to start of yaml files right?

TrentHouliston: I would generally avoid doing a change of directory in a script

TrentHouliston (commented):
If what you are trying to do is to configure something using the command line, I would say that using command line arguments in the binary might be a better idea

ysims:
WIP while trying to get the multi robot tool working

ysims:
This should be good now. There was talk of switching to docker compose but since it doesn't directly spin up the containers, it might not be straightforward and this script as-is does work. 

The build volume is shared across containers, so just editing the config files doesn't work. I've read them in C++ instead.

The multi robot tool uses different colours for each robot, can specify number of robots on each team, can specify role, can specify if you want just one team.

NUbots/NUbook#326 compliments this.

Tom0Brien (approved):

PR #1552, opened 2025-04-21
Fix GitHub actions runner out of memory during docker image build

ysims:
A couple of PRs the last few days have encountered this issue where our Docker image is using too much disk space. We run CI on Ubuntu, and there are many packages that can be cut down to gain disk space. 

This tool https://github.com/marketplace/actions/free-disk-space-ubuntu seems to be a nice addition to do this, and cuts almost 30GB leaving us with room for growth.

Tested with adding libraries, observing it fail from disk space error, and applied the workflow update which resulted in the checks passing. Reverted the added libraries.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -410,7 +410,11 @@ RUN pip install \
     Pillow \
     protobuf==4.21.2 \
     mycroft-mimic3-tts \
-    tqdm
+    tqdm \
+    optuna \
+    ruamel.yaml \
+    plotly \
+    scikit-learn
 
 # Install tools needed for building individual modules as well as development tools
 RUN install-package \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -410,7 +410,11 @@ RUN pip install \
     Pillow \
     protobuf==4.21.2 \
     mycroft-mimic3-tts \
-    tqdm
+    tqdm \
+    optuna \
+    ruamel.yaml \
+    plotly \
+    scikit-learn
```
ysims, line 417: this can be removed once CI has passed with the changes

ysims:
```suggestion
    tqdm
```

Tom0Brien (approved):

PR #1553, opened 2025-04-23
Adds robot-to-robot communication for ball team assisted ball localisation

BaileyJoe:
- Gives robots ability to receive ball localisation information from teammates
- Built into the ball localisation module
- Split into two checks: a step added into the regular ball localisation loop process and a fallback when we haven't seen a ball for a while
- Settings are customizable via yaml including turning the function off entirely

### TODO

More testing, especially on real robots. The functionality is difficult to test properly as it requires close to a full team of robots for a robust test and most applies in niche scenarios, like when a robot is facing in the opposite direction of the ball.


Tom0Brien (commented):
There is a slight issue here with the math.

`team_guesses` is storing each robots estimate of the ball in the field space `rBFf` but this ball localisation module is only dealing with estimating the position in the robots world space `rBWw`. I would suggest to rename the variables `average` to something like `average_rBFf` and function `team_guess_average` to `get_average_team_rBFf`, this will help make clearer the issues with the frames/spaces.

What you most likely want to do here is have a `With<Field>` added to the main trigger, which will provide you with the current robots estimated `Hfw` transform, then use this to transform the teams average to the current robots world space, with something like `average_rBWw = Hfw.inverse()*average_rBFf`

What also may help to debug is to add a new field to the `message::localisation::Ball` message like `average_rBWw` and add a new coloured ball visualisation in the localisation view using this

module/localisation/BallLocalisation/data/config/BallLocalisation.yaml
```diff
@@ -23,3 +23,18 @@ acceptance_radius: 2
 
 # Maximum number of ball rejections before acceptance
 max_rejections: 25
+
+# Robot to robot communication parameters
+use_r2r_balls: true
+
+# Number of robots on our team
+max_robots: 3
```
ysims, line 31: I'd probably do this dynamically. If the vector isn't of size player ID, then resize it. Then we don't need to encode it specifically.

module/localisation/BallLocalisation/src/BallLocalisation.cpp
```diff
@@ -131,12 +146,24 @@ namespace module::localisation {
                     log<DEBUG>("Accept ball: ", accept_ball);
 
                     // Data association: if we have rejected too many balls, accept the closest one
+                    bool low_confidence = false;
```
ysims, line 149:
```suggestion
                    bool low_confidence = rejection_count > cfg.max_rejections ? true : false;
```

module/localisation/BallLocalisation/src/BallLocalisation.cpp
```diff
@@ -131,12 +146,24 @@ namespace module::localisation {
...
-                    if (accept_ball) {
+                    bool accept_team_guess;
+                    Eigen::Vector3d average_rBFf = Eigen::Vector3d::Zero();
+
+                    if (cfg.use_r2r_balls) {
+                        accept_team_guess = get_average_team_rBFf(average_rBFf);
+                    }
+                    else {
+                        accept_team_guess = false;
+                    }
```
ysims, line 164:
```suggestion
                    Eigen::Vector3d average_rBFf = Eigen::Vector3d::Zero();
                    bool accept_team_guess = cfg.use_r2r_balls ? get_average_team_rBFf(average_rBFf) : false;
```

module/localisation/BallLocalisation/src/BallLocalisation.cpp
```diff
@@ -155,22 +182,117 @@ namespace module::localisation {
...
+                            log<DEBUG>("average rBWw: ",
+                                       ball->average_rBWw.x(),
+                                       ball->average_rBWw.y(),
+                                       ball->average_rBWw.z());
                             emit(graph("rBWw: ", ball->rBWw.x(), ball->rBWw.y(), ball->rBWw.z()));
                             emit(graph("vBw: ", ball->vBw.x(), ball->vBw.y(), ball->vBw.z()));
+                            log<DEBUG>("average_rBWw: ",
+                                       ball->average_rBWw.x(),
+                                       ball->average_rBWw.y(),
+                                       ball->average_rBWw.z());
```
ysims, line 211:
```suggestion
```
duplicate?

module/localisation/BallLocalisation/src/BallLocalisation.cpp
```diff
@@ -155,22 +182,117 @@ namespace module::localisation {
...
+                    auto ball        = std::make_unique<Ball>();
+
+                    ball->rBWw                = average_rBFf;
+                    ball->vBw                 = Eigen::Vector3d::Zero();
+                    ball->time_of_measurement = last_time_update;
+                    ball->Hcw                 = last_Hcw;
+                    emit(ball);
+                }
+            }
+        });
```
ysims, line 254: I wonder if this check would be better in the RoboCup trigger. Then we're only checking if anyone is talked to us, and we update when theres new information.

module/localisation/BallLocalisation/src/BallLocalisation.hpp
```diff
@@ -80,6 +80,18 @@ namespace module::localisation {
...
+
+            bool use_r2r_balls = false;
+
+            int max_robots = 0;
+
+            double team_ball_recency = 0.0;
+
+            double team_guess_error = 0.0;
+
+            double team_guess_default_timer = 0.0;
```
ysims, line 93: Add `@brief`s

module/localisation/BallLocalisation/src/BallLocalisation.hpp
```diff
@@ -94,6 +106,18 @@ namespace module::localisation {
...
+
+        struct TeamGuess {
+            NUClear::clock::time_point last_heard = NUClear::clock::now();
+
+            Eigen::Vector3d rBFf = Eigen::Vector3d::Zero();
+        };
+
+        std::vector<TeamGuess> team_guesses{};
+
+        Eigen::Isometry3d last_Hcw = Eigen::Isometry3d::Identity();
```
ysims, line 119: add documentation directives

ysims (commented):
It would be good to get this in, when testing behaviour I noticed that defending players would lose the ball when it was further away with the attacking robots blocking the view, and they'd start looking for the ball. But the attacking robot should be able to tell it where it is. Also I was considering making robots who had lost the ball inactive but really they should just find out who has the ball from their teammates. 

It would be good if we maybe have a confidence for a team mates ball position (maybe from localisation confidence and how long its been since they've seen it, added to the RoboCup message) then we can just add the ball as a measurement to the filter if its a good measurement.

module/localisation/BallLocalisation/src/BallLocalisation.cpp
```diff
@@ -95,82 +102,196 @@ namespace module::localisation {
...
+            ball->time_of_measurement = last_time_update;
+            ball->Hcw                 = balls.Hcw;
+            ball->average_rBWw        = field.Hfw.inverse() * average_rBFf;
+            if (log_level <= DEBUG) {
+                log<DEBUG>("rBWw: ", ball->rBWw.x(), ball->rBWw.y(), ball->rBWw.z());
+                log<DEBUG>("vBw: ", ball->vBw.x(), ball->vBw.y(), ball->vBw.z());
+                log<DEBUG>("average rBWw: ", ball->average_rBWw.x(), ball->average_rBWw.y(), ball->average_rBWw.z());
+                emit(graph("rBWw: ", ball->rBWw.x(), ball->rBWw.y(), ball->rBWw.z()));
+                emit(graph("vBw: ", ball->vBw.x(), ball->vBw.y(), ball->vBw.z()));
+                log<DEBUG>("average_rBWw: ", ball->average_rBWw.x(), ball->average_rBWw.y(), ball->average_rBWw.z());
```
ysims, line 187:
```suggestion
```
removing duplicate

module/localisation/BallLocalisation/src/BallLocalisation.cpp
```diff
@@ -95,82 +102,195 @@ namespace module::localisation {
...
+            // This occurs when the ball position is an echo - ignore echos
+            if (robocup.ball.confidence == 0.0) {
+                // If the ball has no confidence, then we don't care about it
+                return;
+            }
+
+            Eigen::Vector3d rBFf = robocup.ball.position.cast<double>();
+
+            // Resize the vector of guesses if it is not large enough
+            if (team_guesses.size() < robocup.current_pose.player_id) {
```
dex-k, line 207:
```suggestion
            if (team_guesses.capacity() < robocup.current_pose.player_id) {
```

I think? `size` gives the number of current elements, not themaximum before needing to be reallocated

module/network/RobotCommunication/src/RobotCommunication.hpp
```diff
@@ -46,6 +46,8 @@ namespace module::network {
             std::string udp_filter_address = "";
             /// @brief A delay before the first message is sent, to allow reasonable data to be collected
             int startup_delay = 0;
+            /// @brief The timeout for the ball position to be used
+            std::chrono::seconds ball_timeout{0};
```
dex-k, line 50:
```suggestion
            std::chrono::seconds ball_timeout = 0;
```
consistency?

shared/message/localisation/Ball.proto
```diff
@@ -42,9 +42,15 @@ message Ball {
     /// Covariance of state (rBWw, vBw)
     mat4 covariance = 3;
 
+    /// The ball's confidence in the measurement
```
dex-k, line 45:
```suggestion
    /// Confidence in the ball measurement
```

module/network/RobotCommunication/src/RobotCommunication.hpp
ysims: This doesn't work

BaileyJoe:
@ysims 

This current version looks good to me. I just gave it a quick test and it ran fine on my end too.

ysims (approved):

PR #1556, opened 2025-04-27
Upgrade b script to support asynchronous execution

TrentHouliston:
This PR upgrades the `./b` script so that it can handle async functions.

After this you can make your `register` and `run` functions async and they will be handled appropriately.

This will let you write async ./b commands without having to handle the outer async logic yourself.


Tom0Brien (approved):
lgtm

PR #1557, opened 2025-04-27
Swap to using poetry for python dependency management and automatical…

TrentHouliston:
This PR swaps from requiring people to manually manage the dependencies to using poetry for dependency management.

Now when you have a python dependency for the ./b script you can add it to pyproject.toml, and when you run `./b command` poetry will make sure that the dependencies are all up to date.

This also will happen in the docker file with poetry installing dependencies to the build directory so it is cached between builds.

I likely have the dependencies wrong for the NUbots ./b commands, so if someone wants to go through and work out what versions of what are required then put them in the pyproject.toml file.

Also in order to run the ./b code after this you'll have to install poetry https://python-poetry.org/docs/#installation

Therefore, this will really need a NUbook page explaining how to set up to run the codebase.

### Things left to do

- [ ] Update NUbook
- [ ] Get the correct dependencies 
- [ ] Add them here


pyproject.toml
```diff
@@ -2,5 +2,55 @@
...
+pycryptodomex = "3.21.0"
+natsort = "8.4.0"
+jsonschema = "4.23.0"
+
+# Image Processing
+Pillow = "11.1.0"
+opencv-contrib-python-headless = "4.11.0.86"
+
+# Machine Learning
+tensorflow = "2.18.0"
```
Tom0Brien, line 56:
```suggestion
tensorflow = "2.18.0"

# Optimisation
optuna  = "0.12.0"
```

docker/Dockerfile
```diff
@@ -487,6 +473,9 @@ RUN install -d -m 0755 /home/nubots/NUbots \
     && install -d -m 0777 /home/nubots/build \
     && install -d -m 0777 /home/nubots/NUbots/nusight2/node_modules
 
+# Configure poetry to use the build directory for storage for the current user
+RUN poetry config cache-dir /home/fourtel/build/poetry-cache
```
Tom0Brien, line 477:
```suggestion
RUN poetry config cache-dir /home/nubots/build/poetry-cache
```

docker/Dockerfile
```diff
@@ -400,36 +404,15 @@ RUN install-from-source https://github.com/nanopb/nanopb/archive/refs/tags/0.4.9
 ### ADD NEW PROGRAMS/LIBRARIES HERE ###
 #######################################
 
-# Install python libraries
```
Tom0Brien, line 403: Is poetry used inside the container now for python deps?

docker/Dockerfile
```diff
@@ -400,36 +404,15 @@ RUN install-from-source https://github.com/nanopb/nanopb/archive/refs/tags/0.4.9
...
-RUN pip install \
-    pygit2==1.13.3 \
-    xxhash \
-    pylint \
-    termcolor \
-    stringcase \
-    Pillow \
-    protobuf==4.21.2 \
-    mycroft-mimic3-tts \
```
Tom0Brien, line 412: `mycroft-mimic3-tts` probably needs to stay for now because it's called from C++ using a `system` call and will most likely bypass the poetry venv

Tom0Brien: We should replace this with the piper C++ library `https://github.com/rhasspy/piper` but can be done in a later PR

TrentHouliston: It all works with the venv stuff, when you run anything via ./b it's all set up in a poetry run command so the paths all work

Tom0Brien: Okay cool

docker/Dockerfile
TrentHouliston: Sort of, for anything that is used for our bids it is. Anything else is just installed on demand before the install using pip (which is inside a venv in /usr/local)

Tom0Brien (approved):
I can't really test this easily without a cached docker build, so honestly probably fine to merge and we can fix any issues with missing dependencies in follow up if needed

Tom0Brien:
@TrentHouliston we may need some help from the goddess of compute to get the docker file to build

ysims:
```
/home/nubots/build/poetry-cache/virtualenvs/non-package-mode-VkqyJB1f-py3.12/lib/python3.12/site-packages/stringcase.py:247: SyntaxWarning: invalid escape sequence '\W'
  return re.sub("\W+", "", string)
```
This warning comes up a lot when building now.

PR #1558, opened 2025-04-27
Add sensor filter optimisation tools

Tom0Brien:
Similar to recent tools added for optimising localisation parameters, this PR adds tools for optimising the parameters for odometry, currently only the `Kp` and `Ki` parameters of the Mahony filter and optimises the roll and pitch error only.

image

Future PR will include yaw once we update to use Mahony filter estimate.

### DIFF
--- a/module/tools/OdometryBenchmark/src/OdometryBenchmark.cpp
+++ b/module/tools/OdometryBenchmark/src/OdometryBenchmark.cpp
@@ -1,0 +1,178 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2025 NUbots
+ *
+ * This file is part of the NUbots codebase.
+ * See https://github.com/NUbots/NUbots for further info.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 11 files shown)


module/tools/OdometryBenchmark/src/OdometryBenchmark.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+            // Start playback
+            emit<Scope::INLINE>(std::make_unique<PlayRequest>());
+        });
+
+        on<Trigger<Sensors>, With<RawSensors>>().then([this](const Sensors& sensors, const RawSensors& raw_sensors) {
+            // Check if ground truth odometry data exists
+            if (!raw_sensors.odometry_ground_truth.exists) {
+                return;
+            }
```
ysims, line 115: might be good to have an error unless this is expected to happen sometimes?

Tom0Brien: Good idea. Added.

ysims (approved):

PR #1560, opened 2025-05-04
Replace poetry with uv

Tom0Brien:
This PR updates python package management to use UV over poetry. It is much faster and has some additional features/benefits such as automatically detecting updates to `pyproject.toml` avoiding need to run `poetry lock` every time a dependency is changed.

image


.github/workflows/nubots.yaml
```diff
@@ -166,15 +166,16 @@ jobs:
...
+          ./b configure --build-dir build \
+          -- \
+          -DCMAKE_BUILD_TYPE=${{ inputs.build-type }} \
+          -DCMAKE_TOOLCHAIN_FILE=/usr/local/toolchain.cmake \
+          -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+          -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+          -DCI_BUILD=ON \
+          -DBUILD_TESTS=ON \
+          -DENABLE_COVERAGE=${{ inputs.coverage }} \
+          -DENABLE_CLANG_TIDY=${{ inputs.clang-tidy }}
```
ysims, line 178: ?

.github/workflows/nubots.yaml
```diff
@@ -183,19 +184,19 @@ jobs:
...
```
ysims, line 196: ?

ysims (approved):

PR #1568, opened 2025-05-11
Improve ps3walk walk command output

Tom0Brien:
This PR does the following:

1. Adds acceleration limit to joystick commands to improve controlling robot with ps3 controller sticks
2. Adds scripts/button mapping to configuration file (disables scripts by default, toggled with R2)
3. Adds new limits to the head pitch/yaw to prevent head servos reaching limits
4. Lock head by default
5. Adds nusight stuff to both ps3walk and keyboardwalk

### DIFF
--- a/module/purpose/PS3Walk/src/PS3Walk.cpp
+++ b/module/purpose/PS3Walk/src/PS3Walk.cpp
@@ -119,7 +119,6 @@ namespace module::purpose {
                                 if (moving) {
                                     log<INFO>("Stop walking");
                                     emit<Task>(std::make_unique<Walk>(Eigen::Vector3d::Zero()));
-                                    // emit<Task>(std::unique_ptr<Walk>(nullptr));
                                 }
                                 else {
                                     log<INFO>("Start walking");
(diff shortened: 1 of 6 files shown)


module/purpose/PS3Walk/src/PS3Walk.cpp
```diff
@@ -134,69 +139,86 @@ namespace module::purpose {
                                 head_locked = !head_locked;
                             }
                             break;
-                            // dance moves here:
-                        case BUTTON_DPAD_UP:
+                        // case BUTTON_DPAD_UP:
```
ysims, line 142: It'd be nice to have this set up with a yaml list of button/script pairs, then we won't have this large block of commented out code.

Tom0Brien: Added, I've also disabled by default because I found myself accidentally triggering scripts when walking robot around

ysims (approved):
I was thinking more dynamically generated cases based on the yaml map so that removing cases was as simple as commenting out the yaml file lines, but that can be something for the future.

PR #1570, opened 2025-05-11
Add in team association to robot localisation

ysims:
This PR adds functionality to robot localisation to associate robots as team mates or opponents. It also processes robot communication messages as robot measurements. NUsight is updated to assign red or blue to the robot cylinders. 

N1 in NUgus is the red defender at the back. Looking pretty good! Went through ready state and playing for a bit, mostly was good, the only issues were: sometimes got the multiple tracked robots spawning in the wake of a robot; and robot just not looking to see where robots were.

Screenshot from 2025-05-11 14-49-40

It seems to work well in the full robocup set up. We may want to add in a custom field to the robot to robot communication message for the team ID, as the only way currently to know if a robot is our team mate is from the colour. But we only know the colour from the GameController. This means without GameController, the robots won't listen to each other.

- [ ] Updated NUbook if necessary (add link to NUbook PR here) (should write something up, might just make a Jira task since I think we don't have any robot localisation info written up)
- [x] Updated relevant module READMEs - todo


module/localisation/RobotLocalisation/data/config/RobotLocalisation.yaml
```diff
@@ -14,6 +14,7 @@ ukf:
     position: [1e-3, 1e-3]
     velocity: [1e-3, 1e-3]
 
-association_distance: 0.4
+association_distance: 0.5
+teammate_association_distance: 2.0
```
Tom0Brien, line 18: This seems quite high.. Was there a reason it needed to be larger than standard association distance 0.5. This would mean that if two teammates are close together, they would need to move 2m apart to be correctly initialized in the tracked robot list

Tom0Brien (changes requested):
Needs formatting.

Looks good! 

I do have some concerns but they can addressed in a later PR. Currently we don't take into consideration the uncertainty of the teammates self localisation pose, so a poorly localized robot would be considered a perfectly fine measurement currently. I think what we should probably do is add a "uncertainty/confidence" metric to the field localisation message (already there but not set as exactly what we want). With this we can ignore messages where the robot is not confident in its pose

module/localisation/RobotLocalisation/data/config/RobotLocalisation.yaml
ysims: yeah can probably reduce it to not cause issues when there are robots next to each other, but we are getting error from both the other robot's localisation error and vision object position error. If the teammate says it is in view, then the closest vision robot must be the teammate. It might need to be implemented, but ideally it's only really the larger value when it's a team mate being compared against itself or a 'opponent' that might be that team mate.

ysims:
> Needs formatting.
> 
> Looks good!
> 
> I do have some concerns but they can addressed in a later PR. Currently we don't take into consideration the uncertainty of the teammates self localisation pose, so a poorly localized robot would be considered a perfectly fine measurement currently. I think what we should probably do is add a "uncertainty/confidence" metric to the field localisation message (already there but not set as exactly what we want). With this we can ignore messages where the robot is not confident in its pose

Does much need to be done to get a confidence from field localisation? And do we have any idea how reasonable the value might be (I assume difficulty fitting to the vision features would influence the confidence?).

Tom0Brien:
> > Needs formatting.
> > Looks good!
> > I do have some concerns but they can addressed in a later PR. Currently we don't take into consideration the uncertainty of the teammates self localisation pose, so a poorly localized robot would be considered a perfectly fine measurement currently. I think what we should probably do is add a "uncertainty/confidence" metric to the field localisation message (already there but not set as exactly what we want). With this we can ignore messages where the robot is not confident in its pose
> 
> Does much need to be done to get a confidence from field localisation? And do we have any idea how reasonable the value might be (I assume difficulty fitting to the vision features would influence the confidence?).

We have both the final cost value in the nl solver and the confidence from the kalman filter, so a combination of these could give a nice metric to measure confidence

module/localisation/RobotLocalisation/data/config/RobotLocalisation.yaml
Tom0Brien: I think what may be better for teammate data association is rather than using distance, just take advantage of the fact each teammate has a unique identifier already. So something like: 

```
void RobotLocalisation::data_association(const std::vector<Eigen::Vector3d>& robots_rRWw,
                                         uint teammate_id) {

    // ---------- 1. Teammate messages -------------------------------------------------
    if (teammate_id != 0) {

        // A. Is there already a track with this teammate_id?
        auto track_itr = std::find_if(tracked_robots.begin(), tracked_robots.end(),
                         [teammate_id](const TrackedRobot& t){ return t.teammate_id == teammate_id; });

        if (track_itr == tracked_robots.end()) {
            // B. No → create a new track that is *guaranteed* to own that ID
            tracked_robots.emplace_back(TrackedRobot(robots_rRWw.front(), cfg.ukf, next_id++));
            tracked_robots.back().teammate_id = teammate_id;
            tracked_robots.back().seen        = true;
        }
        else {
            // C. Yes → just UKF-update that exact track
            track_itr->ukf.measure(robots_rRWw.front().head<2>(),
                                   cfg.ukf.noise.measurement.position,
                                   MeasurementType::ROBOT_POSITION());
            track_itr->seen = true;
        }

        return;   // finished with this message
    }

    // ---------- 2. Vision-only detections (same as today) ----------------------------
    for (const auto& rRWw : robots_rRWw) {
        ...              // keep the existing nearest-neighbour / gating code
    }
}
```

Tom0Brien (approved):
LGTM!

PR #1575, opened 2025-05-18
Receive the button-data from NUSense.

Claegtun:
### Brief

Adds fields for the buttons' states in the `NUSenseData` protobuf message and conveys the received states to the `RawSensor` data in `HardwareIO`.

### BackPanel

A submessage is also created for the back-panel as an entity. At the moment, it only holds the buttons. The intention would be to have LEDs so that the NUC can poll the LEDs' states. However, there is no urgent need to poll the LEDs at this stage. Therefore, the `BackPanel` message only has the buttons for now.

### Reviewers, Please Note

This branch can only be merged once [NUcontroller/PR-27](https://github.com/NUbots/NUcontroller/pull/27) has been approved.


shared/message/platform/NUSenseData.proto
```diff
@@ -91,6 +91,19 @@ message IMU {
     uint32 temperature = 3;
 }
 
+message BackPanel {
+
+    // For now, I am only including the two buttons since I see no current need to poll the state of the LEDs, and I
+    // want to conserve as much bandwidth as possible. If such a need arises, then one can add it under the BackPanel
+    // message.
```
ysims, line 98: probably just simplify this to "LEDs not added to conserve bandwidth, but can be added under the BackPanel if needed". Or attribute yourself to the "I"s.

Claegtun: Done.

ysims (dismissed):
lgtm

shared/message/platform/NUSenseData.proto
```diff
@@ -91,6 +91,17 @@ message IMU {
     uint32 temperature = 3;
 }
 
+message BackPanel {
+
+    // LEDs are not yet added in order to conserve bandwidth. However, they can be added later once the need arises.
+    message Buttons {
+        bool left   = 1;
+        bool middle = 2;
+    }
```
jpptm, line 100: Might be better if these were enums instead

Claegtun: Instead of `bool`?

jpptm: If we only allow single presses yes. But now that I think about it, if we do it this way we can handle button combinations as well.

Claegtun: It is also this way because it is like this in `RawSensors`, which has a boolean for each button.

Claegtun: It is nested in case we need to add the LEDs, and it would be nice to have all of them nested in case there are other LEDs outside the backpanel.

jpptm: then do it as needed and add that in another pr that adds the LEDs or do it in here. Also, does the NUC really need to know about the LEDs?

Claegtun: It doesn't, but the LEDs are in `RawSensors`.

jpptm: In that case, I think the LEDs are only worth sending through if we plan to log that information somewhere, maybe emit it to nusight so a tab can use it and it might be useful during a game when we are sshed into the robot. But then again a human can see the LEDs. I'll leave the decision to you but I'd prefer it if NUSense didn't have to encode messages that aren't important

PR #1580, opened 2025-05-24
Add field positioning utility functions

ysims:
Two main functions

- An optimisation function used by the defender (and probably support players too) to find the best position based on where the other robots are, the rough position we want to be in, and keeping moderately close to the opponent robots. 
- Function to find a ready position for the robot, based on the configuration of teammates. The front robot will position for kick off, the others will stick to their side and make rows.

Ready position examples:

Screenshot from 2025-05-24 15-34-17

Screenshot from 2025-05-24 15-07-23

Screenshot from 2025-05-24 15-13-05

Ready position could do with some added logic to make the formations a little nicer, e.g., when there is just one robot in a row it could be closer to the center (but trying to offset so they're not directly behind), and for 3x0 situations we could even it out so that its not fully weighted to one side. But for now I think this is working as intended.

### DIFF
--- a/shared/utility/strategy/positioning.hpp
+++ b/shared/utility/strategy/positioning.hpp
@@ -1,0 +1,264 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2025 NUbots
+ *
+ * This file is part of the NUbots codebase.
+ * See https://github.com/NUbots/NUbots for further info.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


shared/utility/strategy/positioning.hpp
```diff
@@ -0,0 +1,264 @@
...
+                                 all_robot_positions,
+                                 clearance,
+                                 equidist_weight,
+                                 neutral_weight,
+                                 avoid_weight);
+                grad[i]   = (f1 - f2) / (2 * eps);
+            }
+
+            // Update the position
+            X -= step_size * grad;
```
Tom0Brien, line 163: I think using the nlopt library would be better suited here rather than implementing this from scratch

It can easily support adding constraints to ensure the optimal position is not outside the field, too far away etc

Also has better algorithms and debugging features out of the box

PR #1581, opened 2025-05-24
Change name for FindFeature to FindBall

ysims:
I think anyone who has worked in these files knows that it is annoying to find this haha. Originally I thought this would get added to but so far we haven't, might as well be `FindBall`.


yippyy:
i agree with this, much easier to find lol

yippyy (approved):
good job :)

PR #1584, opened 2025-05-25
Add localisation uncertainty and reset when uncertain

ysims:
## Changes

- Adds a `cost` data field to `Field.proto`, using the output of the cost function in FieldLocalisationNLopt for the current position 
- Adds a `uncertainty_reset` function that  will find a new optimal field position by first using a local search and then undertaking a global field grid search if the local search fails, both isolated to the half of the field the robot was last known to be on.
- Averages the cost of field line points and field intersection features
- Adds a constant penalty to out of field points
- Optimised parameters for webots with cost changes
- Reaction handle on the main loop to disable and enable it while a reset occurs
- Removed unused function declarations that exist instead in OccupanyMap (`setup_fieldline_distance_map` and `setup_field_landmarks`)
- Moved functions in FieldLocalisationNLopt to be private instead of public
- Renamed `normalizeAngle` to `normalise_angle` in `utility/math/angle.hpp`
- Use `UncertaintyResetFieldLocalisation` to pause the robot in `Soccer`

## Todos

- [x] Add documentation directives for new variables and functions
- [x] Move hardcoded variables into config variables
- [x] General clean up
- [x] More testing during normal play (tested by manually moving robot in webots so far)
- [x] Testing on the real robot and tuning of the threshold
- [x] Issues with NUsight not showing updates (I think this was from bad values, things seem stable now)

## Future

Planning to use this to ignore uncertain teammates and also stop playing if uncertain.


Tom0Brien:
I'm not so sure NIS is all we want here, from my understanding, in essence it's a metric measuring the difference in our latest measurement (in our case the optimisation result) and the current estimate, which I suspect does become large when kidnapping the robot or moving it. This may indicate the robot is lost in some cases but in others it may just be that the optimiser has found a better solution and should be driving the estimate toward the optimal solution. What I would suggest exploring is something like this plus the final cost of the optimisation routine. Nominally the cost should be low and if it goes above a certain threshold and remains there for X amount of time and doesn't converge we should reset.

ysims:
Checked Frankie and Kevin with the current parameters. Looked good and thresholding seems to be better now. Bad situations are noticeably in a different cost range than while stationary and while walking.

PR #1585, opened 2025-05-25
Mocap ground truth localisation 

Tom0Brien:
This PR updates ground truth pose logic to work with both webots and optitrack motion capture data. Also some minor changes and instructions for obtaining data from the motive software.

TODO:
- Make same changes to the `FieldLocalisation` module

Associated webots PR https://github.com/NUbots/NUWebots/pull/118

### DIFF
--- a/module/localisation/Mocap/src/Mocap.hpp
+++ b/module/localisation/Mocap/src/Mocap.hpp
@@ -1,0 +1,22 @@
+#ifndef MODULE_LOCALISATION_MOCAP_HPP
+#define MODULE_LOCALISATION_MOCAP_HPP
+
+#include <nuclear>
+
+namespace module::localisation {
+
+    class Mocap : public NUClear::Reactor {
+    private:
+        /// @brief Stores configuration values
+        struct Config {
+            uint32_t robot_rigid_body_id = 0;
+        } cfg;
+
+    public:
+        /// @brief Called by the powerplant to build and setup the Mocap reactor.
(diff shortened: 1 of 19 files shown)


module/localisation/Mocap/src/Mocap.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include <nuclear>
+
+namespace module::localisation {
+
+    class Mocap : public NUClear::Reactor {
+    private:
+        /// @brief Stores configuration values
+        struct Config {
+            uint32_t robot_rigid_body_id = 0;
```
ysims, line 12: `@brief`

roles/webots/localisation.role
```diff
@@ -19,7 +19,6 @@ nuclear_role(
   network::PlotJuggler
   # Platform
   input::SensorFilter
-  input::GameController
```
ysims, line 22: This is needed for team robot localisation. I have a PR (#1583) that stops it from spamming when theres no game controller instance.

shared/message/platform/webots/messages.proto
```diff
@@ -1,6 +1,6 @@
 // MIT License
 //
-// Copyright (c) 2021 NUbots
+// Copyright (c) 2025 NUbots
```
ysims, line 3: I think we keep this as the year it was on creation? then it's implictily [creation year - present year]

ysims (commented):
New files are missing license headers.

PR #1586, opened 2025-05-27
Fix video extract

Tom0Brien:
This PR fixes some minor issues with the `./b nbs video` tool

It was breaking when trying to use GPU and for our use cases I don't think we really need the GPU so have just opted to force CPU use


ysims:
revert uv lock? might need to gitignore it if it keeps coming up.

Tom0Brien:
I think if the lock file changes it should be committed, but I'm not entirely sure on what and why it changes without changing deps

ysims:
formatter failed

PR #1588, opened 2025-06-02
Fix seg fault in RobotLocalisation from invalid constructor param

ysims:
Hard to show here but [this line](https://github.com/NUbots/NUbots/blob/c533d5e2d0a8618b985c54b826dd24db80857a42/module/localisation/RobotLocalisation/src/RobotLocalisation.hpp#L98) was throwing a seg fault in debug mode because the config value its sending through is a 2D vector, whereas RobotModel has only a StateVec constructor wanting a 4D Vector to fill the position and velocity components. This PR adds a constructor to RobotModel so that passing a 2D vector will work.

I didn't encounter this in release, only debug.

### DIFF
--- a/module/localisation/RobotLocalisation/src/RobotModel.hpp
+++ b/module/localisation/RobotLocalisation/src/RobotModel.hpp
@@ -69,6 +69,9 @@ namespace module::localisation {
             StateVec(const Eigen::MatrixBase<OtherDerived>& state)
                 : rRWw(state.template segment<2>(PX)), vRw(state.template segment<2>(VX)) {}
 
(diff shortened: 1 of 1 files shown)


module/localisation/RobotLocalisation/src/RobotModel.hpp
```diff
@@ -69,6 +69,9 @@ namespace module::localisation {
             StateVec(const Eigen::MatrixBase<OtherDerived>& state)
```
Tom0Brien, line 69: minor: Probably best to specify this exact dimension here too
```suggestion
            StateVec(const Eigen::Matrix<Scalar, size, 1>& state)
```

Tom0Brien (dismissed):

PR #1597, opened 2025-06-09
Use a StreamReactor PostConnect hook instead of using a separate WatchDog for sending handshake requests to NUSense

jpptm:
### Brief

This PR
- uses `PostConnect` instead of a watchdog;
- and adds a message-type to handle cases where NUSense is reset during a binary.


Claegtun:
Can be merged once PR-29 in NUcontroller has been approved.

Claegtun (approved):
Seems good to me.

PR #1600, opened 2025-06-12
Fix construction of odometry ground truth Hrw matrix

ysims:
Must have got muddled somewhere along the way. Robot was doing some very funky things in ready! Trying to rotate to face backwards to where it needed to walk to haha.


dex-k (approved):
yew

PR #1601, opened 2025-06-13
Emit localisation in the main update loop rather than on new information updates

ysims:
We seemed to have a bit of repetition in emitting localisation robots and running debug info in the VisionRobots and RoboCup reactions. This PR moves it to just emit run debug info in the main update loop.


A-Herfel (approved):
Checks out

PR #1608, opened 2025-06-20
Check ball for nullptr before proceeding in FieldPlayer

ysims:
I think this got lost in some of the clean up. Seg fault if no ball is seen at all yet in the program.


Crafty15 (approved):
LGTM

PR #1612, opened 2025-06-21
NUsight two teams support - flip red team on the NUsight localisation view, have both teams in NUsight

ysims:
To properly test two teams (nugus vs nugus), we need some updates to NUsight. 

- The first is specific to Webots, when both teams are run on the same computer. NUsight identifies robots from the same device as different if they have different `name`s. `name` comes from hostname by default in the robot's NUClearNet module. By checking for the multi tool's `team_id` flag, the name can be set to add in the team name. So we end up with `webots1_12`, `webots_1_13`, etc. This results in all the robot showing when running the multi tool.
- The second issue is both teams showing up on the same side of the field. The robot thinks it is on the positive x half of the field internally, so we flip any robots with team_colour equal to RED. This will only kick in if GameController is running, because it tells the robot its colour. This is a z rotation around the field of the Hfw matrix that all objects in localisation view are transformed with, so it applies to all objects (field lines, goals, ball, etc).  

Screenshot from 2025-06-21 19-14-03


JosephusPaye (commented):
NUsight stuff looks good to me. The other changes look good too based on the PR description, but I only had a cursory look.

One thing to mention is that variable and class property names in NUsight are meant to be camel cased, but it looks like there are many recently introduced ones in the localisation model that are not.

ysims:
> One thing to mention is that variable and class property names in NUsight are [meant to be camel cased](https://nubook.nubots.net/guides/general/code-conventions#case), but it looks like there are many recently introduced ones in the localisation robot model that are not.

I was wondering what was going on with the casing, but pushed it to the back of my mind hahaha. I'll change teamColour then maybe @KimKhangHoang you can have a look at going through the nusight code and fixing casing?

KimKhangHoang:
@ysims -  yeah, I can go through the casing stuff on my working branch too 👍

Crafty15 (approved):
LGTM

PR #1617, opened 2025-06-24
Consider the goalie in the FieldPlayer

ysims:
This required some set up but in the end looks pretty good with the existing goalie behaviour! 

- Wanted to add goalie info to RobotLocalisation, so just went and changed it to have the full Purpose message.  This changed a few things, like how the `Localisation::Robot` message is used in the upper behaviour, and NUsight.  
- If the ball is in the defensive third, the goalie is included in the check for closest robot, so if it's close to the ball then the FieldPlayers will let it work as the attacker, and if the ball is not close enough the FieldPlayers should come to get it. As long as there is a FieldPlayer, someone will start coming back to be a defending player to take the ball. Goalies are ignored from the defending player check, so it's not mistaken as the defender.
- Goalies are also ignored in ready positioning

### DIFF
--- a/module/purpose/FieldPlayer/src/FieldPlayer.cpp
+++ b/module/purpose/FieldPlayer/src/FieldPlayer.cpp
@@ -95,9 +95,11 @@ namespace module::purpose {
                 // Make an ignore list with the goalie, if they exist
                 std::vector<unsigned int> ignore_ids{};
                 // Add inactive robots to the ignore list
-                for (const auto& robot : (robots ? robots->robots : std::vector<message::localisation::Robot>{})) {
-                    if (!robot.purpose.active) {
-                        ignore_ids.push_back(robot.purpose.player_id);
+                if (robots) {
+                    for (const auto& robot : robots->robots) {
+                        if (!robot.purpose.active) {
+                            ignore_ids.push_back(robot.purpose.player_id);
+                        }
                     }
                 }
 
@@ -118,7 +120,7 @@ namespace module::purpose {
                 Eigen::Vector3d rBFf     = field.Hfw * ball->rBWw;
                 double defending_third_x = fd.dimensions.field_length / 3.0;
                 if (robots && rBFf.x() < defending_third_x) {
-                    for (const auto& robot : (robots ? robots->robots : std::vector<message::localisation::Robot>{})) {
+                    for (const auto& robot : robots->robots) {
                         if (robot.purpose.purpose == SoccerPosition::GOALIE) {
                             log<DEBUG>("Ball not in defending third, ignoring goalie for closest to ball");
                             ignore_ids.push_back(robot.purpose.player_id);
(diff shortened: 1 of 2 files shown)


module/purpose/FieldPlayer/src/FieldPlayer.cpp
```diff
@@ -90,27 +92,50 @@ namespace module::purpose {
...
+                                                                  global_config.player_id,
+                                                                  ignore_ids);
+
                 // If we have robots, determine if we are closest to the ball
                 // Otherwise assume we are alone and closest by default
+                // Only consider the goalie if the ball is in the defending third
+                Eigen::Vector3d rBFf     = field.Hfw * ball->rBWw;
+                double defending_third_x = fd.dimensions.field_length / 3.0;
+                if (robots && rBFf.x() < defending_third_x) {
+                    for (const auto& robot : (robots ? robots->robots : std::vector<message::localisation::Robot>{})) {
```
dex-k, line 121:
```suggestion
                    for (const auto& robot : robots->robots) {
```
The `if` above already checks if `robots` is truthy

module/purpose/FieldPlayer/src/FieldPlayer.cpp
```diff
@@ -90,27 +92,50 @@ namespace module::purpose {
                     return;
                 }
 
+                // Make an ignore list with the goalie, if they exist
+                std::vector<unsigned int> ignore_ids{};
+                // Add inactive robots to the ignore list
+                for (const auto& robot : (robots ? robots->robots : std::vector<message::localisation::Robot>{})) {
```
dex-k, line 98: imo it's less clunky as
```
if (robots)
    for (const auto& robot: robots->robots) {...}
```
rather than iterating over an empty vector

module/purpose/FieldPlayer/src/FieldPlayer.cpp
```diff
@@ -162,14 +187,21 @@ namespace module::purpose {
...
-                                                                            field.Hfw,
-                                                                            sensors.Hrw,
-                                                                            cfg.equidistant_threshold,
-                                                                            global_config.player_id,
-                                                                            std::vector<unsigned int>{closest_to_ball})
-                                         : true;
+                // Add closest_to_ball to the ignore list so we don't consider the attacker as the furthest back
+                ignore_ids.push_back(closest_to_ball);
+                // Add goalies to the ignore list so we don't consider them as the furthest back
+                for (const auto& robot : (robots ? robots->robots : std::vector<message::localisation::Robot>{})) {
```
dex-k, line 193: same here as above

module/purpose/FieldPlayer/src/FieldPlayer.cpp
dex-k: just noticed this is already done on line 224 anyway, so good for consistency too

nusight2/src/client/components/localisation/network.ts
```diff
@@ -134,7 +134,7 @@ export class LocalisationNetwork {
       return {
         id: localisation_robot.id!,
         rRWw: Vector3.from(localisation_robot.rRWw),
-        color: localisation_robot.isBlue ? "blue" : "red",
+        color: localisation_robot.teammate ? robot.teamColour : robot.teamColour === "red" ? "blue" : "red",
```
dex-k, line 137:
```suggestion
        color: localisation_robot.teammate ? robot.teamColour : (robot.teamColour === "red" ? "blue" : "red"),
```

PR #1618, opened 2025-06-25
Create the option to associate field line intersections with land marks using the Hungarian algorithm

jpptm:
Currently, `FieldLineLocalisationNLopt` uses a greedy algorithm to associate field line intersections and land marks. This association is not guaranteed to be optimal and correct every time. The Hungarian algorithm is mathematically proven to yield the optimal one-to-one mapping, with slower speed as a trade off. Even with this caveat, it has been determined that 100x100 cost matrices will take around less than 1 ms to determine an assignment, and 1000x1000 matrices will take slightly longer. With that said, we are not anywhere near these matrix dimensions so solving should take much less time.

- [x] Determine average time it takes to determine association for both methods
- [x] Assess and analyse the quality of the derived solution


jpptm:
Not properly tested yet, but it is working. No conclusive results yet with regards to the quality of localisation, but it feels faster than the greedy method. Time results to follow

jpptm:
At worst, hungarian is just as fast as greedy if not slightly better (talking 10-20 us range)
At best, it is around 2-3 times faster

These numbers were derived using `test/localisation`. I assume that both algorithms will perform worse to a degree when `robocup` is ran instead.

Screenshot From 2025-06-25 17-24-55

Screenshot From 2025-06-25 17-25-04

Screenshot From 2025-06-25 17-25-49

Screenshot From 2025-06-25 17-26-13

jpptm:
Localisation results from both algorithm look pretty bad, but I think other factors are coming into play here as well. For instance. the landmarks at the centre of the field are rarely detected. In addition, notice that yolo is always confidently wrong about an L-intersection, constantly classifying it as a T intersection with significant confidence.

More proper testing can be done with a tidier field and more importantly, determining the source of the false negatives but I think this can be reviewed properly for now.

Screenshot From 2025-06-25 17-58-47
Screenshot From 2025-06-25 17-57-56
Screenshot From 2025-06-25 17-57-44
Screenshot From 2025-06-25 17-56-42
Screenshot From 2025-06-25 17-53-34

jpptm:
update:

Hungarian:

image


Greedy:

image

ysims (approved):

jpptm:
The formatting fails and keeps throwing these no space left errors, im not really sure what I have to do to merge this. Will search for solutions later.

ysims:
Should be good now to add back into the queue, it should merge main automatically in the queue to get the fix.

PR #1619, opened 2025-06-25
Adds kick walk skill

Crafty15:
- This skill emits a modified walk command to the walk engine when a kick task is emitted. The effect is a single large kick step that integrates seamlessly with the robot's normal walking pattern. 
- It has been tested on the real robot with great results. 
- Left and Right leg kicks.
- Some more tuning may be required to better align with the goal before kicking as a kick can sometimes be triggered if the robot gets too close to the ball while trying to align.
- `kick_approach_velocity_x` config value is a useful parameter that makes the robot slow down on approach to the ball.
- kick velocity parameters can be found in the `Walk` config files.
- Removed the `reset()` call from keyboardwalk kick command as it stops the walk, so is incompatible.
- Removed `Stability::STANDING` check in plankick. Also incompatible as we don't want to stop walking.


module/purpose/AllRounder/src/AllRounder.cpp
```diff
@@ -164,7 +164,8 @@ namespace module::purpose {
         // Second argument is priority - higher number means higher priority
         emit<Task>(std::make_unique<FindBall>(), 1);    // if the look/walk to ball tasks are not running, find the ball
         emit<Task>(std::make_unique<LookAtBall>(), 2);  // try to track the ball
-        emit<Task>(std::make_unique<WalkToKickBall>(), 3);  // try to walk to the ball and align towards opponents goal
+        emit<Task>(std::make_unique<WalkToBall>(), 3);  // try to walk to the ball and align towards opponents goal
```
ysims, line 167: Is `WalkToBall` needed? It is a basic linear walk to ball, whereas `WalkToKickBall` positions behind the ball towards the goal.

module/purpose/Attack/src/Attack.cpp
```diff
@@ -35,6 +37,7 @@ namespace module::purpose {
                 log<DEBUG>("We have the ball or it is free, walk to the goal!");
                 // Try to walk to the ball and align towards opponents goal
                 emit<Task>(std::make_unique<WalkToKickBall>(), 3);
+                emit<Task>(std::make_unique<KickTo>(), 4);  // kick the ball towards the goal
```
ysims, line 40: I'd probably make this general to any attack case - ie use it in the Tackle as well. Then we have a better chance of knocking the ball out from the opponent's feet. If you're not sure about it and want it tested first, can skip for now and can check it later.

module/skill/KickWalk/src/KickWalk.cpp
```diff
@@ -0,0 +1,60 @@
...
+            cfg.kick_approach_velocity_x = config["kick_approach_velocity_x"].as<double>();
+            cfg.kick_approach_velocity_y = config["kick_approach_velocity_y"].as<double>();
+        });
+
+        on<Provide<Kick>, Uses<Walk>>().then(
+            [this](const Kick& kick, const RunReason& run_reason, const Uses<Walk>& walk) {
+                if (run_reason == RunReason::NEW_TASK) {
+                    if (NUClear::clock::now() - last_kick_end < std::chrono::seconds(2)) {
+                        log<DEBUG>("KickWalk is waiting for cooldown after last kick.");
+                        NUClear::clock::time_point wait_until = last_kick_end + std::chrono::seconds(2);
```
ysims, line 29: Wait time should be a config variable.

module/skill/KickWalk/src/KickWalk.hpp
```diff
@@ -0,0 +1,39 @@
...
+            double kick_step_height         = 0.08;  // Height of kick step (m)
+            double kick_approach_velocity_x = 0.0;   // Forward velocity during kick (m/s)
+            double kick_approach_velocity_y = 0.0;   // Lateral velocity during kick (m/s)
+            double normal_velocity_x        = 0.1;   // Normal forward walking velocity (m/s)
+            int kick_steps                  = 2;     // Number of steps to execute kick over
+
+            // State tracking
+            utility::input::LimbID current_kick_leg = utility::input::LimbID::UNKNOWN;
+            int kick_step_count                     = 0;
+            bool is_kicking                         = false;
```
ysims, line 27: All except velocity can be removed.

module/skill/KickWalk/src/KickWalk.hpp
```diff
@@ -0,0 +1,39 @@
...
+        } cfg;
+
+        NUClear::clock::time_point last_kick_end = NUClear::clock::now();
```
ysims, line 30: `@brief`

module/skill/Walk/data/config/frankie/Walk.yaml
```diff
@@ -41,3 +41,8 @@ arms:
   left_shoulder_roll: 0.35
   right_elbow: -0.7
   left_elbow: -0.7
+
+  kick:
+  # Step kick parameters
+  kick_velocity_x: 0.5 # Forward velocity during kick (m/s)
+  kick_velocity_y: 0.05 # Lateral velocity during kick (m/s)
```
ysims, line 48: Is this right that Frankie uses the same values as in webots, rather than the default walk? Also indentation. I assume it should be the same as the default, so I'd recommend just removing this.

module/skill/Walk/src/Walk.cpp
```diff
@@ -143,11 +145,54 @@ namespace module::skill {
...
+                    double full_period     = walk_generator.get_step_period();
+                    WalkState::Phase phase = walk_generator.get_phase();
+
+                    // If the current time in the walk generator phase is either the end of the other leg or the start
+                    // of the desired leg, start the kick step and increase the velocity of the walk generator when it
+                    // is currently executing a kick step and it is at the end of the desired leg, stop.
+                    if (!kick_step_in_progress) {
+                        // Check if the conditions allow for the kick to start
+                        // Todo maybe set offset based on the frequency
+                        bool end_of = (current_time + 0.1) >= full_period;
```
ysims, line 163: The offsets should be config variables.

module/skill/Walk/src/Walk.cpp
```diff
@@ -143,11 +145,54 @@ namespace module::skill {
...
+                        // If its the end of the kick foot's support, we can kick
+                        if (end_of && other_support) {
+                            log<INFO>("Kick step started");
+                            kick_step_in_progress = true;
+                            velocity_target       = Eigen::Vector3d(cfg.kick_velocity_x, cfg.kick_velocity_y, 0.0);
+                        }
+                    }
+                    else {
+                        // Check if the step can end
+                        bool end_of = (current_time + 0.1) >= full_period;
```
ysims, line 177: Offset config

module/skill/Walk/src/Walk.hpp
```diff
@@ -58,13 +59,25 @@ namespace module::skill {
...
 
         /// @brief Last time we updated the walk engine
         NUClear::clock::time_point last_update_time{};
 
         /// @brief Generates swing foot and torso trajectories for given walk velocity target
         utility::skill::WalkGenerator<double> walk_generator{};
+
+        /// @brief Used to track the step phase for the step kick
+        bool kick_step_in_progress = false;
+        bool done_sent             = false;
```
ysims, line 77: Looks like this can be removed

module/skill/Walk/src/Walk.hpp
```diff
@@ -58,13 +59,25 @@ namespace module::skill {
...
+        /// @brief Store the kick phase
+        message::behaviour::state::WalkState::Phase initial_kick_phase;
+        NUClear::clock::time_point kick_step_start_time;
```
ysims, line 80: These can be removed.

module/purpose/Attack/src/Attack.cpp
Crafty15: Is that what you meant?

module/skill/KickWalk/src/KickWalk.hpp
```diff
@@ -0,0 +1,33 @@
...
+    using utility::input::LimbID;
+
+    class KickWalk : public ::extension::behaviour::BehaviourReactor {
+    private:
+        /// @brief Stores configuration values
+        struct Config {
+            double kick_approach_velocity_x = 0.0;  // Forward velocity during kick (m/s)
+            double kick_approach_velocity_y = 0.0;  // Lateral velocity during kick (m/s)
+            long new_kick_wait_time         = 0.0;  // Time to wait before the next kick (in seconds)
+
```
ysims, line 20:
```suggestion
             /// @brief Forward velocity during kick (m/s)
            double kick_approach_velocity_x = 0.0;
             /// @brief Lateral velocity during kick (m/s)
            double kick_approach_velocity_y = 0.0;
            /// @brief Time to wait before the next kick (in seconds)
            long new_kick_wait_time = 0.0; 
```

PR #1622, opened 2025-06-26
Disable robocup service with unknown USB devices

dex-k:
This PR adds functionality to ensure the systemd `robocup.service` doesn't auto restart when we pull a robot off the field mid-game, but still auto-restarts when we want to put them back onto the field.

It does this by requiring `robocup.service` to only start _after_ `robocup-usb-check.service` (this is added in a separate `.conf` file, without touching `robocup.service`, so it can be easily disabled by deleting / commenting out the file).

`robocup-usb-check.service` runs `check-usb-devices.sh`, a script which compares the USB devices connected to `usb_allowlist`, a file of known/expected devices listed by `vendor_id:product_id` (output of `lsusb`). If there's an unexpected device present, `robocup-usb-check.service` fails, automatically restarting until success. On success, it re-starts `robocup.service` if it isn't runnning already (because if `robocup-usb-check.service` isn't immediately successful, `robocup.service` fails to start).
- All "normal" devices are in the allowlist (cameras, usb hubs, etc), but this way _any_ external device (keyboard, monitor) can be used to stop the binary from starting.

The robocup binary/service does **not** automatically stop when a device is plugged in, but it won't _restart_, so the binary still has to be manually killed with `ctrl-C`, `kill $(pidof robocup)`, `htop` etc.

Starting/stopping `robocup.service` can be done with the `play`/`stop` aliases, or with (`sudo`) `robocupconfiguration`, but `stop` shouldn't be used to kill the binary because then the service is disabled and won't restart.

--

It works, but I'm not convinced this is the cleanest implementation. If `robocup.service` is started while e.g. a keyboard is plugged in, the service appears to fail
```shell
nubots@nugus1 ~ ❯❯❯ play                                                                   ✘ 130 
Created symlink '/etc/systemd/system/multi-user.target.wants/robocup.service' → '/etc/systemd/system/robocup.service'.
A dependency job for robocup.service failed. See 'journalctl -xe' for details.
```
but in reality the usb check service is running in the background, and the robocup service will auto start once the keyboard is disconnected. This is a product of trying to make the usb-check behaviour be separate from the robocup service so that it doesn't change normal operation. Very open to suggestions.


module/tools/SystemConfiguration/data/system/default/etc/robocup.d/check-usb-devices.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
+ALLOWLIST_FILE="$SCRIPT_DIR/usb_allowlist"
+
+# Extract vendor:product pairs from allowlist, ignoring comments and blank lines, normalize to lowercase
+allowlist=$(grep -Eo '^[0-9a-fA-F]{4}:[0-9a-fA-F]{4}' "$ALLOWLIST_FILE" | tr '[:upper:]' '[:lower:]')
+
+# Track unauthorized devices
+unauthorized=()
```
ysims, line 10:
```suggestion
unauthorised=()
```

module/tools/SystemConfiguration/data/system/default/etc/robocup.d/check-usb-devices.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Loop over each line of lsusb
+while IFS= read -r line; do
+    id=$(awk '{print tolower($6)}' <<< "$line")
+    if ! grep -qx "$id" <<< "$allowlist"; then
+        desc=$(sed 's/.*\(ID [^ ]\+ [^ ]\+.*\)/\1/' <<< "$line")
+        unauthorized+=("$desc")
```
ysims, line 17:
```suggestion
        unauthorised+=("$desc")
```

module/tools/SystemConfiguration/data/system/default/etc/robocup.d/check-usb-devices.sh
```diff
@@ -0,0 +1,29 @@
...
+    fi
+done <<< "$(lsusb)"
+
+# Log and exit if any unauthorized devices found
+if [ ${#unauthorized[@]} -ne 0 ]; then
```
ysims, line 22:
```suggestion
if [ ${#unauthorised[@]} -ne 0 ]; then
```

module/tools/SystemConfiguration/data/system/default/etc/robocup.d/check-usb-devices.sh
```diff
@@ -0,0 +1,29 @@
...
+    for dev in "${unauthorized[@]}"; do
```
ysims, line 23:
```suggestion
    for dev in "${unauthorised[@]}"; do
```

module/tools/SystemConfiguration/data/system/default/etc/robocup.d/usb_allowlist
```diff
@@ -0,0 +1,12 @@
...
+
+# Internal USB hubs
+1d6b:0003 # Linux Foundation 3.0 root hub
+1d6b:0002 # Linux Foundation 2.0 root hub
+# WiFi/Bluetooth adapter
+8087:0033 # Intel Corp.
+# Camera
+1e10:4000 # Point Grey Research, Inc. Blackfly S BFS-U3-13Y3C
+# OpenCR
+0403:6001 # Future Technology Devices International, Ltd FT232 Serial (UART) IC
```
ysims, line 12: Add NUSense

PR #1623, opened 2025-06-26
Add functionality for a support player

ysims:
A few bug fix and setup things in here too, to make things work nicely.

- RobotLocalisation: Added a check to ignore teammates who have a high localisation cost.
- RobotLocalisation: Fixes a bug where the purpose wasn't being overwritten when a new message is received from a tracked teammate.
- RobotCommunication: Made GlobalConfig required, because otherwise there is not team ID associated with the message, which isn't helpful.
- RobotCommunication: Added the cost of localisation, to support first point above.
- RobotCommunication: Purpose player ID is overwritten with the global config player ID. This is to enforce consistency, as the global config is the main truth of player ID, and for empty purpose messages to still contain the correct player ID.
- FieldPlayer: Search if localisation is bad (head search pattern, turn on spot)
- FieldPlayer: Only ignore teammate IDs
- Emit unknown purpose in FindBall if the ball is lost - I don't like emitting the purpose in this layer, but I'm not sure of an alternative to properly catch this without having essentially an identical check in FieldPlayer (with ball timeout config, annoying to track both).
- New Support module that stays in line with the ball, a bit away to the side, ready to take over when needed.
- soccer_strategy: Fix logic in furthest_back to make sense.

### DIFF
--- a/module/localisation/RobotLocalisation/data/config/RobotLocalisation.yaml
+++ b/module/localisation/RobotLocalisation/data/config/RobotLocalisation.yaml
@@ -19,3 +19,5 @@ association_distance: 0.5
 max_missed_count: 25
 
 max_distance_from_field: 1.0
+
+max_localisation_cost: 2.0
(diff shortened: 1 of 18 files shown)


Crafty15 (approved):
LGTM

PR #1632, opened 2025-06-28
Add force playing capability to SoccerNew

ysims:
Forgot to add this before getting the new behaviour in. To avoid having to rewrite the general structure of the FieldPlayer, I think an easy solution is to just emit a playing phase. We don't typically use force playing with a game controller running since the usual use case is not wanting to have to set the game controller up just to do some testing. If there is a game controller running, then the user can just filter out the messages by setting the `udp_filter_address` in the game controller yaml. 

Tested in Webots.


Crafty15 (approved):
LGTM

PR #1635, opened 2025-06-30
Free disk space so formatting passes CI

ysims:
Not sure if this will fix the problem, but we can see on this run https://github.com/NUbots/NUbots/actions/runs/15960508559/job/45015295942 that it's running out of disk space when installing nusight dependencies for formatting. This PR adds a step before installing dependencies for formatting to remove unneeded files.


ysims:
I think we need to move to our own runners.

ysims:
I'm a little concerned about turning off formatting right as I'm about to leave haha.
I removed some libraries but it didn't seem to help.

ysims:
Formatting passed. I removed some libraries to do with firmware and removed the firmware installer.

Crafty15 (approved):
Nice one!

PR #1636, opened 2025-06-30
Add balance control to foot controller

Crafty15:
- Balance code stolen from Tom's old branch [here](https://github.com/NUbots/NUbots/tree/obrien/balance-pid), with the only major change being the use of gyro for roll/pitch rates.
- This PR allows torso correction to be applied to FootControl tasks. 
- So far has been tested on spline kick and the walk. Limited success with spline kick unless gains are turned right up, leading to oscillations. Walk seemed decent with low gains. 
  - More testing might be required on what happens when the robot falls. The delay in fall detection could cause some wild behaviour. I wonder if that could have contributed to Kevin's knees giving out recently. 
   - Update to previous point. Added error check to stop correction if error is too large. It works, but if it occurs while the ankles are at an angle due to correction, they will snap back to their original, non corrected value, which could be just as bad.

P.S. It's probably not critical that we have this in and use it for robocup. I was thinking it might help with the tilting during get ups and starting to walk again when the batteries get low.


module/actuation/FootController/src/FootController.hpp
```diff
@@ -47,10 +51,15 @@ namespace module::actuation {
...
 
     using utility::input::LimbID;
+    using utility::math::euler::mat_to_rpy_intrinsic;
     using utility::nusight::graph;
 
     struct SetGains {};
 
+    // Function declarations
+    void FusedFromQuat(const Eigen::Quaterniond& q, double& fusedPitch, double& fusedRoll);
+    Eigen::Quaterniond QuatFromFused(double fusedPitch, double fusedRoll);
```
ysims, line 61: Documentation directives. These also look like they should be inside the class?

module/actuation/FootController/src/FootController.hpp
```diff
@@ -91,6 +130,110 @@ namespace module::actuation {
...
+                // Hwt quaternion
+                Eigen::Quaterniond Hwt_quat(sensors.Htw.inverse().linear());
+
+                // Get fused roll and pitch
+                double fused_roll;
+                double fused_pitch;
+                FusedFromQuat(Hwt_quat, fused_pitch, fused_roll);
+
+                if (log_level <= DEBUG) {
+                    // graph the correction being applied
```
ysims, line 144:
```suggestion
                    // Graph the correction being applied
```

module/actuation/FootController/src/FootController.hpp
```diff
@@ -79,8 +88,38 @@ namespace module::actuation {
...
+        /// @brief Accumulates the integral of the roll error over time for use in PID control.
+        double integral_roll_error = 0;
+        /// @brief Accumulates the integral of the pitch error over time for use in PID control.
+        double integral_pitch_error = 0;
+
+        /// @brief Stores the timestamp of the last PID control update, used to compute time deltas for integral and
+        /// derivative calculations.
+        NUClear::clock::time_point last_update_time{};
+
+    public:
```
ysims, line 119: All these functions seem like they're just used within the class, so they should be in the private section?

module/actuation/FootController/src/FootController.hpp
```diff
@@ -111,12 +254,45 @@ namespace module::actuation {
...
+        double stheta = 2.0 * (q.y() * q.w() - q.x() * q.z());
+        double sphi   = 2.0 * (q.y() * q.z() + q.x() * q.w());
+        stheta        = (stheta >= 1.0 ? 1.0 : (stheta <= -1.0 ? -1.0 : stheta));  // Coerce stheta to [-1,1]
+        sphi          = (sphi >= 1.0 ? 1.0 : (sphi <= -1.0 ? -1.0 : sphi));        // Coerce sphi   to [-1,1]
+        fusedPitch    = asin(stheta);
+        fusedRoll     = asin(sphi);
+    }
+
+    Eigen::Quaterniond QuatFromFused(double fusedPitch, double fusedRoll)  // Assume: fusedYaw = 0, hemi = true
+    {
```
ysims, line 271: Move implementation to cpp?

module/actuation/FootController/data/config/webots/FootController.yaml
```diff
@@ -17,3 +19,17 @@ servo_gains:
   R_KNEE: 30
   R_ANKLE_PITCH: 30
   R_ANKLE_ROLL: 30
+
+balance:
+  enabled: false
```
ysims, line 24: is there much of a reason to have this? It seems like all the modules (kick and walk) that use the balance control have their own configs that get passed into this module.

shared/utility/skill/WalkGenerator.hpp
```diff
@@ -90,6 +90,9 @@ namespace utility::skill {
 
             /// @brief Option to only switch between planted foot if the next foot is planted
             bool only_switch_when_planted = false;
+
+            /// @brief Option to use balance control when walking
+            bool use_balance_control = false;
```
Tom0Brien, line 95: This is never used for walk generator stuff, so I think it should be moved to a variable on the walk/kick class

module/actuation/FootController/src/FootController.hpp
Tom0Brien, line 1: we should add a check for the amount of error to prevent issues if the robot falls over, so if the robot is X degrees from desired setpoint, pause applying control

Crafty15: I added a check to return once the error has been calculated but it stops the walk entirely and i don't think it should. `ik_task` is set before the correction block, then reassigned with corrected values at the end.

module/actuation/FootController/src/FootController.hpp
```diff
@@ -91,6 +142,117 @@ namespace module::actuation {
...
+                // If the error is too large, we are probably falling over and should pause applying control.
+                if (std::abs(roll_error) > cfg.max_roll_error || std::abs(pitch_error) > cfg.max_pitch_error) {
+                    if (log_level <= DEBUG) {
+                        log<DEBUG>(fmt::format(
+                            "Balance correction disabled due to large error: roll_error = {}, pitch_error = {}",
+                            roll_error,
+                            pitch_error));
+                    }
+                    return;  // Do not apply control if the error is too large
+                }
```
Crafty15, line 187: This check works but it stops the walk entirely :/

Crafty15: Sorted with [c8dec0d](https://github.com/NUbots/NUbots/pull/1636/commits/c8dec0d36d183939cb92e08d6e96cee7347ebf62)

module/skill/SplineKick/data/config/SplineKick.yaml
```diff
@@ -37,3 +37,6 @@ arms:
   left_shoulder_roll: 0.35
   right_elbow: -0.7
   left_elbow: -0.7
+
+# Use balance control. NOTE: correction_enabled must be true in FootController config
```
ysims, line 41:
```suggestion
# Use balance control
```

module/skill/SplineKick/data/config/webots/SplineKick.yaml
```diff
@@ -36,3 +36,6 @@ arms:
   left_shoulder_roll: 0.2
   right_elbow: -0.7
   left_elbow: -0.7
+
+# Use balance control. NOTE: correction_enabled must be true in FootController config
```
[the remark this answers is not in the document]
ysims: or maybe the comment isn't needed anymore

PR #1637, opened 2025-07-02
Adds a new YOLO model trained from YOLO11n

ysims:
This PR adds a new YOLO model. It is trained from YOLO11n, an upgrade from our previous YOLO8n model.

The model is trained using
- Bit-Bots' TORSO-21 real images
- `>40` images collected by the team at Eindhoven
- `>40` images collected by the team a few years ago in the old lab
- `>800` images generated from NUpbr, automatically annotated and then manually accepted or rejected for training (about half of the images were accepted out of 1700).

The intersection accuracy is somewhat similar, but I'd argue that the new model is slightly better, although I did see one misclassification of an X as an L in my tests. Balls is significantly better, with balls now further than like 2 meters seeable. It can definitely see the ball in the center from the starting position at the side of the field, and can usually see it across from one side of the lab field to the other. 

Webots is a bit shaky and it'd be good to train a webots-specific model. It does seem to detect robots a lot better now (particularly close to) but the intersections were not great (false negatives mostly) but I think it's tuned well enough to work reasonably similarly to the old model. The old model is still in here if we want to switch back for Webots, but I think the improved robot detection is a big enough deal to keep the new one set.  

There is some code added for robustness. The input size is not hard coded anymore, and some error checks are added. This was helpful when I was testing various models.

### DIFF
--- a/module/vision/Yolo/src/Yolo.cpp
+++ b/module/vision/Yolo/src/Yolo.cpp
@@ -27,7 +27,6 @@
 #include "Yolo.hpp"
 
 #include <chrono>
-#include <cstring>
 #include <getopt.h>
 #include <iostream>
 #include <vector>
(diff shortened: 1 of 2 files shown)


Tom0Brien (approved):
LGTM!

Tom0Brien (commented):
Would be good to add this dataset used to Huggingface

PR #1638, opened 2025-07-03
Display servo states using data from NUSense

jpptm:


jpptm:
> Is the data sent with every message? It looks like a lot of output. Maybe just have an error log saying which servos have problems. Possibly just one log inside the loop that prints the servo and issue if missing or duplicated, or if there's a present no_id servo?

no, data is only sent once during NUSense's startup routine

Claegtun (approved):
Seems good to me.

PR #1639, opened 2025-07-04
Update the goalie behaviour similarly to the FieldPlayer

ysims:
Open to suggestions on the behaviour. Currently it:

- Freezes in the appropriate penalty states (referee positioning)
- Looks around for the ball if it doesn't know where it is (just the head, doesn't walk)
- If there are no teammates, it plays instead as a FieldPlayer
- If the ball isn't in the defending third, it stays in the goals
- If the ball is in the defending third, but another teammate is closer, it aligns with the ball on the goal line.
- If the ball is in the defending third and it is closest, it attacks. In penalty positioning, it will position behind the ball, ready to play.

Also turned off the high cost localisation search for now. We might want to consider different behaviour - maybe stand on the spot and look around, or just go for it.


jpptm:
What if the goalie is positioned incorrectly and it has some playing space behind it that is not the goal, and the ball happened to be there? Do we care about this scenario? If we do, it might be worth rotating in place. I wouldnt worry about it too much though, this seems to be a niche situation

jpptm:
Also, what happens when there's a horde of robots including goalies near the ball?

module/purpose/Goalie/README.md
```diff
@@ -2,32 +2,41 @@
 
 ## Description
 
-Play soccer in the goalie position. Either bypasses GameController and force playing, or uses GameController information to decide what to do.
+Play soccer in the goalie position.
 
 In the ready state, walks to the goals.
 
-In the playing state (for both normal game mode and penalty shootout), finds the ball, fixates on the ball and dives if the ball is close.
+In the playing state, the goalie will typically stay within the goals unless the ball enters the defending third. In this case, it will either kick it back out (if the closest player) or it will stay between the ball and goals if another played is closer. In penalty states, it will freeze when appropriate and position for attacking if the ball is in the defending third and it is the closest to the ball.
```
jpptm, line 9:
```suggestion
In the playing state, the goalie will typically stay within the goals unless the ball enters the defending third. In this case, it will either kick it back out (if the closest player) or it will stay between the ball and goals if another player is closer. In penalty states, it will freeze when appropriate and position for attacking if the ball is in the defending third and it is the closest to the ball.
```

module/purpose/Goalie/src/Goalie.cpp
```diff
@@ -30,251 +30,221 @@
...
+                // If there are no robots, use an empty vector
+                Who ball_pos = utility::strategy::ball_possession(ball->rBWw,
+                                                                  (robots ? *robots : Robots{}),
+                                                                  field.Hfw,
+                                                                  sensors.Hrw,
+                                                                  cfg.ball_threshold,
+                                                                  cfg.equidistant_threshold,
+                                                                  global_config.player_id,
+                                                                  ignore_ids);
+                // We it's in the defending third, and we are closest to the ball, defend the goals!
```
jpptm, line 198:
```suggestion
                // Since we are in the defending third and we are closest to the ball, defend the goals!
```

jpptm (approved):

PR #1642, opened 2025-07-05
Purge some unused modules and messages

ysims:
- FieldLocalisation
- AllRounder
- Soccer (SoccerNew renamed to Soccer)
- Striker
- Defender
- AlignBallToGoal
- KickToGoal
- Ready
- WalkInsideBoundedBox

And their associated messages. Technically WalkInsideBoundedBox is used in the goalie but that will change once #1639 goes in.


Tom0Brien:
I thought it would be good to get start safely working

Tom0Brien (approved):

PR #1643, opened 2025-07-05
[NUsight] Merge dashboard and localisation tab

KimKhangHoang:
This PR merges the Dashboard tab as a toggle checkbox into the Localisation tab in NUsight. It also restructures the layout a bit to make things cleaner and sets up for further reducing any overlapping functionality.

### Things left to do

- [ ] Apply further functionality reduction based on Webots testing and review feedback.


Tom0Brien:
Looks good! Some small things

The camera type/mode and target info from the localisation view is overlapping with the dashboard overview information box at the bottom in dashboard mode.
image

It would be ideal to keep the dashboard overview information box visible across both the localisation and dashboard modes. 

We should add an indicator in a later PR which shows what side is the opponents/own goal in the localisation view.

KimKhangHoang:
Thanks Tom, I’ve just addressed the first two points. I’ll look into the third point next, in a separate PR/branch once this one is done.

Tom0Brien (approved):
LGTM!

PR #1664, opened 2025-07-08
Localisation tuning

Tom0Brien:
This PR does the following:
### Tuning

- Tunes localisation parameters on real robot (Frankie)
- Temporarily disables reset on high cost, should be re-tuned with new weights and cost term
- Disables hungarian algorithm (doesn't currently support distance throwouts)
- Disables goal base measurements

### Field intersection confidence weighting
Field intersection cost term now incorporates the confidence from vision such that higher confidence measurement are weighted more heavily in the alignment cost


jpptm:
@willburgin has a fix for the distance throwout support in `burgin/hungarian-distance`, this problem will be fixed there instead. Apologies

jpptm (dismissed):

PR #1665, opened 2025-07-08
Update default network forwarder and remove plotjuggler from RoboCup role

Tom0Brien:
Updates the default network forwarder config and removes plotjuggler from RoboCup role. PlotJuggler requires converting all the datapoint emits to JSON which is considerably computationally expensive.


nusight2/src/client/components/localisation/model.ts
```diff
@@ -125,7 +125,7 @@ export class LocalisationModel {
   @observable particlesVisible = true;
   @observable goalsVisible = true;
   @observable fieldLinePointsVisible = true;
-  @observable fieldIntersectionsVisible = true;
+  @observable fieldIntersectionsVisible = false;
```
ysims, line 128: ?

Tom0Brien: Intersection associations show up when sending Field messages which we don't want by default

ysims (approved):

PR #1679, opened 2025-07-14
Add in-step kick

Tom0Brien:
Building upon https://github.com/NUbots/NUbots/pull/1619 this PR adds a new state to the walk engine for generating an in-step kick. This is very similar to before, but rather than just increasing the walk velocity (which just increases the step length essentially), this adds extra tuning parameters for adjusting the trajectory of the step kick.


Crafty15 (dismissed):
LGTM

Crafty15 (dismissed):

PR #1681, opened 2025-07-14
Tune getups with big bumpers

Crafty15:


ysims:
Not sure if we'll need to make the front one frankie-specific since the original getup worked fine on the other robots, but didn't work on frankie. Can check tomorrow.

Crafty15:
> Not sure if we'll need to make the front one frankie-specific since the original getup worked fine on the other robots, but didn't work on frankie. Can check tomorrow.

Will close this for now then

PR #1692, opened 2025-07-15
Adds balance for footcontroller -> robocup2025

Crafty15:
Same as https://github.com/NUbots/NUbots/pull/1636


module/actuation/FootController/data/config/webots/FootController.yaml
```diff
@@ -1,3 +1,5 @@
+log_level: DEBUG
```
Tom0Brien, line 1: INFO

module/purpose/Soccer/data/config/Soccer.yaml
```diff
@@ -2,7 +2,7 @@
 log_level: INFO
 
 # For the robot playing without listening to GameController
-force_playing: false
+force_playing: true
```
Tom0Brien, line 5: DISABLE

PR #1693, opened 2025-07-15
Enhance Camera module with improved configuration and time synchronisation features

TrentHouliston:
Synchronise from 4AI.

- Includes some robustness code which reconnects cameras when they disconnect
- Changes to how the time synchronisation works


jpptm (changes requested):
See comment above

PR #1694, opened 2025-07-15
Update NetworkForwarder for Bidirectional communication

TrentHouliston:
Sync from 4AI

- Updates NetworkFowarder to be able to have bidirectional forwarding
- Moves to a message based configuration
- Moves to a reflection based rather than a python generated based system


jpptm:
This broke stuff for me back at 4AI so i think it is worth thoroughly testing it before merging

PR #1699, opened 2025-07-15
Fix walk oscillations

A-Herfel:
Things the PR does go here

- change how we emit director tasks so that we never stop
- adds a buffer to Attack module so that we don't switch between tasks too often


ysims:
This doesn't cause problems with another player calculating correctly that its attacking and the robot is also still attacking because of the timer?

PR #1700, opened 2025-07-15
Update robot communication broadcast address for the RoboCup fields and add a test role

ysims:


roles/test/robotcommunication.role
```diff
@@ -0,0 +1,18 @@
+# This role is used for running the full autonomous soccer set up with a real robot. It should be used with a computer
+# on the network running GameController, or set force_playing to true in Soccer.yaml to play without GameController.
+# This is used at the RoboCup competition for real games.
```
ysims, line 3:
```suggestion
# A test role to check if robot communication between robots is working. Set RobotCommunication's log level to `DEBUG` to see a log each time a teammate sends a message, with the teammate's ID.
```

PR #1702, opened 2025-07-16
Replace KF with a simple exponential filter

Tom0Brien:
Without a process model, the current KF used in `FieldLocalisationNLopt` is a bit unnecessary. This PR replaces it with a simple exponential filter which is a bit more intuitive to tune by hand.


module/localisation/FieldLocalisationNLopt/src/FieldLocalisationNLopt.cpp
```diff
@@ -309,16 +296,16 @@ namespace module::localisation {
...
+                    field->uncertainty = 0.0;
 
                     emit(field);
                 });
     }
 
     void FieldLocalisationNLopt::debug_field_localisation(Eigen::Isometry3d Hfw) {
         emit(graph("opt state", state.x(), state.y(), state.z()));
-        emit(graph("kf state", kf.get_state().x(), kf.get_state().y(), kf.get_state().z()));
+        emit(graph("filtered state", filtered_state.x(), filtered_state.y(), filtered_state.z()));
```
jpptm, line 308:
```suggestion
        emit(graph("exponentially filtered state", filtered_state.x(), filtered_state.y(), filtered_state.z()));
```
For clarity maybe?

jpptm (approved):
image
lgtm, comment is just an unnecessary change since the class member is documented anyway

PR #1703, opened 2025-07-16
Add a message to the director for visualisation of its state

TrentHouliston:
Adds a message to the director which can be used to visualise the current graph state of the director for debugging.


module/extension/Director/src/director/emit_state.cpp
```diff
@@ -0,0 +1,94 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2024 NUbots
```
ysims, line 4:
```suggestion
 * Copyright (c) 2025 NUbots
```

shared/message/behaviour/Director.proto
```diff
@@ -0,0 +1,106 @@
+// MIT License
+//
+// Copyright (c) 2021 NUbots
```
ysims, line 3:
```suggestion
// Copyright (c) 2025 NUbots
```

ysims (approved):

PR #1716, opened 2025-07-18
Update frankie walk config

Tom0Brien:
Reduce step width to avoid walking over the ball


jpptm:
Change base, and also Sarah will need some retuning as her left ankles seem to be dropping packets and might be in need of a replacement

PR #1719, opened 2025-07-18
Penalty shootout behaviour

ysims:
Penalty shootout behaviour. 

Its own purpose module, and its own roles in webots and the real robot.
Acts in playing, kick off team determines if goalie or the player taking the penalty. 

The goalie will stand still, because the rules say it cant exist the goal area.
The penalty taker will walk to kick ball (walk to ball and align to goals).

Uses localisation, but this requires custom positions. When the middle button is pressed, it resets localisation to either the penalty mark or the middle of the goals.


ysims:
> If this is working, we should probably add the module to the robocup role no?

It's got its own role, it's a root level module. Doesn't work with the main soccer stuff.

PR #1735, opened 2025-07-29
Add role and debug log for testing R2R communication

ysims:
This was useful during RoboCup to make sure things were working.


ZeoNyph (changes requested):
looks good to me! just need to update the branch with main before merging it

PR #1739, opened 2025-08-22
Penalty behaviour (kick taker and goalie)

ysims:
The PR adds in a new module and roles to run penalty shootouts with GameController. 
When taking a penalty kick, the robot will dribble the ball into the goals. When defending, the robot will stand still, since rules require it to not step off the line. This should be further developed. 
In either state, the idea is to press the middle button to trigger localisation to reset, which uses the field description to figure out suitable position candidates.


KimKhangHoang (approved):
LGTM

PR #1740, opened 2025-08-26
Fix Docker images failing to build due to missing OpenVINO patches

ZeoNyph:
## Fix NUC12 Docker Target

- fixes the Docker image not building by adding back the OpenVINO patches that were removed from the AUR at one point as a [master .patch file](https://gist.githubusercontent.com/ZeoNyph/cf8216bc1ffa15b0e44b17899f355c07/raw/42a6a040f41c03714507b86cf257546ebbb6a0cc/openvino_patch.patch)

### Reviewers, Note

It probably isn't the best idea to have all the patches in as one .patch file, but I made it ages ago and I can't be bothered lol


ysims:
Could make the PR commit/title "Fix Docker image failing to build from missing patches" - it's not nuc12 specific. Bit annoying that the patches went missing.

ysims (approved):

PR #1742, opened 2025-10-01
Dyson/ping-pong-summation

dargem:
- Finds the sum of numbers in the range of 1 - 10 inclusive, starting at 1 and ending at 10

### Reviewers, Note

Part of the onboarding task

### Things left to do

- Missing read me for Judge and Pong, ping is not filled out
- Doesn't emit a nod currently


.vscode/settings.json
```diff
@@ -18,7 +18,71 @@
   "files.associations": {
     "*.role": "cmake",
     ".clang-tidy": "yaml",
-    ".Doxyfile": "properties"
+    ".Doxyfile": "properties",
```
jpptm, line 21: `settings.json` should not be committed to main

module/onboarding/Judge/src/Judge.cpp
```diff
@@ -0,0 +1,35 @@
...
+        on<Startup>().then([this] {
+            // pass
+        });
+
+
+        on<Trigger<VerifiedPing>>().then([this](const VerifiedPing& verified_ping) {
+            /*
+            auto nod = std::make_unique<the nod message>();
+            emit(nod);
+            */
```
jpptm, line 30:
```suggestion
            /*
            auto nod = std::make_unique<the nod message>();
            emit(nod);
            */
```
Might wanna look at how to send tasks to director. For instance, `NodYes`

PR #1749, opened 2025-10-20
Merge onboarding task to main onboarding branch

relzp:
This PR would merge the branch containing my onboarding task commits with the main montano/onboarding branch.
My branch implements a Pong message to complement the already-existing Ping message and adds counters to each.
It also sends Ping and Pong messages between the two modules, incrementing and then sending back the counter each time.

### DIFF
--- a/module/onboarding/Ping/src/Ping.cpp
+++ b/module/onboarding/Ping/src/Ping.cpp
@@ -25,7 +25,7 @@ namespace module::onboarding {
 
         on<Trigger<Pong>>().then([this](const Pong& pong_msg) {
             auto ping_msg = std::make_unique<Ping>();
-            log<NUClear::INFO>("Ping");
+            log<INFO>("Ping");
             emit(ping_msg);
         });
     }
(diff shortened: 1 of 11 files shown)


module/onboarding/Ping/src/Ping.cpp
```diff
@@ -25,7 +25,9 @@ namespace module::onboarding {
 
         on<Trigger<Pong>>().then([this](const Pong& pong_msg) {
             auto ping_msg = std::make_unique<Ping>();
-            log<NUClear::INFO>("Ping");
+            ping_msg->count = pong_msg.count + 1;
+            log<INFO>("Ping");
+            log<INFO>(ping_msg->count);
```
ysims: `log` takes any number of arguments, so you can do 
```suggestion
            log<INFO>("Ping", ping_msg->count);
```
which might look neater in the output. It automatically puts spaces between each argument for you, so you don't need to manually add spaces.

ysims (commented):
Nice!

PR #1750, opened 2025-10-23
Completed orientation task

sebastianbrooker2:
finished the implementation of ping / pong modules that includes and emits a counter


module/onboarding/Ping/README.md
```diff
@@ -0,0 +1,16 @@
+# Ping
```
ysims: Have a go looking at other readme files in the repository and fill this and the Pong `README.md` in. These readme files are actually used to generate some of the pages on NUbook!

ysims (commented):
Check what files you are adding before committing - the settings.json should not be committed, and you've got an extra folder (Induction, looks like your recruitment task, don't get personal documents mixed up with the repository!). You can click on the 'files changed' tab to see what has been changed, and see if files have snuck in that shouldn't be here!

Compile the code as you go as well, so you know if the code makes sense. At the moment, I can see some issues that would prevent it compiling, which also show up in the GitHub status checks.

PR #1763, opened 2026-04-08
Add NUSense Fan Controller functionality

miles-p:
This PR is linked to functionality on [NUController PR#38](https://github.com/NUbots/NUcontroller/pull/38)

This PR adds a message to the NUSense IO ProtoBuf framework that reports whether the fans driven by NUSense are spinning fast enough. Use of this error state is not yet implemented, but could be used to throw a warning log in future.

This message, namely, `FanWarning` has two booleans; `fan1_warning` and `fan2_warning` whereby a `false` represents a **functional** fan, spinning fast enough. A `true` would represent a fan that is spinning fast enough. A fan that isn't plugged in will report a fan warning of `true`.

### DIFF
--- a/shared/message/platform/NUSenseData.proto
+++ b/shared/message/platform/NUSenseData.proto
@@ -96,6 +96,9 @@ message Buttons {
     bool middle = 2;
 }
 
+// These two bools hold the warning status of the two fans (J401 & J402) on NUSense.
+// These two fan warnings are triggered by a fan dropping below the threshold speed.
+// (that threshold speed is defined in the NUSense firmware)
 message FanWarning {
     bool fan1_warning = 1;
     bool fan2_warning = 2;
--- a/shared/message/platform/NUSenseData.proto
+++ b/shared/message/platform/NUSenseData.proto
@@ -96,6 +96,11 @@ message Buttons {
     bool middle = 2;
 }
 
+message FanWarning {
+    bool fan1_warning = 1;
+    bool fan2_warning = 2;
+}
+
 message NUSense {
     /* INDEX MAPPING
      *  0  : r_shoulder_pitch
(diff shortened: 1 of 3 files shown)


shared/message/platform/RawSensors.proto
```diff
@@ -99,6 +99,11 @@ message RawSensors {
         float z = 3;
     }
 
+    message FanWarning {
+        bool fan1_warning = 1;
+        bool fan2_warning = 2;
+    }
```
ysims: Rather than duplicating this message, you can import the nusense data proto. You can see an example of this being done in this file, where above there is 
`import "message/platform/webots/messages.proto";`
which then allows the use of the `webots.RobotPoseGroundTruth` message.

shared/message/platform/NUSenseData.proto
```diff
@@ -96,6 +96,11 @@ message Buttons {
     bool middle = 2;
 }
 
+message FanWarning {
+    bool fan1_warning = 1;
+    bool fan2_warning = 2;
+}
```
ysims, line 105: This would benefit from some documentation - if you use three slashes `///` then it'll be seen as a doxygen-style comment. I'd be interested in knowing what these warnings are and are they different warnings or somehow related?

ysims (commented):
Update the PR title to something descriptive :) 
A good way to think up titles is to finish the sentence "when merged, this PR will..."

shared/message/platform/NUSenseData.proto
ysims: I realise now after reading the other PR that these are literally two fans. I thought a single fan had two warning states. Maybe it would be better to make a Fan message, with a warning bool inside. If its possible to know if these are right or left fans that would be even more useful for logging. As it reads, I thought FanWarning was representing one Fan. Or even call it `Fans` if you want to group them as is here. But comments would help this anyway.

miles-p:
@ysims is this one good to go? Thanks :D

ysims (approved):

PR #1766, opened 2026-05-07
fix frankie's right knee offset

miles-p:
Things the PR does go here

Tweaks the right knee of Frankie


Claegtun (approved):
Seems good to me.

PR #1768, opened 2026-05-16
Load the extrinsics offsets for the actual robot

jpptm:
Tuning the extrinsics currently involves this confusing method of copying a robot's offset configs then pasting it to the higher level `Camera.yaml`, tuning that, then pasting it back to the robot config and using the tuned values for the higher level `Camera.yaml`.

This pull request fixes this issue by loading the actual robot configs instead, and using the robot-specific offsets during runtime.


Claegtun (approved):
Assuming that it has been tested and works.

PR #1770, opened 2026-05-19
fix camera config for webots

A-Herfel:
config was loading off robot name folder but webots has no robot name, so reverted code here.


JohanneMontano:
This PR is a duplicate of https://github.com/NUbots/NUbots/pull/1769

PR #1776, opened 2026-05-23
Added variable for NUClear port in NUsight server

ZeoNyph:
This is a minor QoL change; something I noticed with NUsight while developing for the K1 is that while it defaults to port 7447 (as part of NUClearNet.js), it doesn't have the option to set the port yourself. As a result, changing the announce port in the NUClearNet config makes any NUbots binaries undiscoverable to NUsight (port 7447 is unfortunately already bound to an RPC service in the K1, so this was me finding a solution haha)

No need to update `./b yarn`, as its argparse just takes in a single 'args' argument.

NUbook PR [here](https://github.com/NUbots/NUbook/pull/347)

### Things left to do

- [x] Update documentation in NUbook (where necessary)

### DIFF
--- a/nusight2/src/server/dev.ts
+++ b/nusight2/src/server/dev.ts
@@ -17,6 +17,7 @@ const args = minimist(process.argv.slice(2));
 
 const withVirtualRobots = args["virtual-robots"] || false;
 const nuclearnetAddress = args.address || "239.226.152.162";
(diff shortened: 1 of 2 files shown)


ysims:
Nice, I'd add to `dev.ts` as well.

ysims (approved):

PR #1777, opened 2026-05-27
Update robot communication and GameController for 2026 HSL compliance

KimKhangHoang:
This PR implements 2026 HSL communication rule compliance across RobotCommunication and GameController modules.
 
#### RobotCommunication:
- Comment out unused `others` field that was causing packet sizes up to ~800 bytes, exceeding the 512 byte limit
- Add serialised packet size check with WARN log before emit
- Add message counter tracking sends during READY/SET/PLAYING states only, with reset and total log on game FINISHED
- Update team message send/receive port to `10000 + team_number` per 2026 rules
 
#### GameController:
- Migrate from `HlRoboCupGameControlData` version 12 to `RoboCupGameControlData` version 19 to support the 2026 GameController ([release v6.0.0](https://github.com/RoboCup-HumanoidSoccerLeague/GameController/releases/tag/v6.0.0), [header definition](https://github.com/RoboCup-HumanoidSoccerLeague/GameController/blob/v6.0.0/game_controller_msgs/headers/RoboCupGameControlData.h))
- Split old `Mode` enum into `GamePhase` and `SetPlay` to match the new protocol structure
- Remap penalty codes to the new 2026 values
- Remove `coach` message and `drop-in` handling, which were removed from the 2026 protocol
- Update return struct to version 4
- Add `messageBudget` logging for cross-checking against the GC counter

### Pre-PR Checklist:

Have you

### Reviewers, Note

Tested with NUWebots (4 robots) and the new 2026 GameController. Robot connects and shows up in GC interface, no 512 byte packet size warnings, message counter matches GC budget countdown, and total messages well under 12000 for a full game.

### Things left to do

- [ ] Update team colour handling: extend `GameState.proto` TeamColour enum for all v3 colours and review team-based filtering in `RobotCommunication` given port-per-team rule
- [ ] Further penalty and set play behaviour testing
- [ ] Update NUbook GameController setup page for 2026 GC
- [ ] Wire up `fallen`, `pose`, `ball` fields in return struct from localisation if required

### DIFF
--- a/module/input/GameController/src/GameController.cpp
+++ b/module/input/GameController/src/GameController.cpp
@@ -236,11 +236,8 @@ namespace module::input {
         log<DEBUG>("GameController message budget remaining:", new_own_team.message_budget);
 
         // Get colours
-        state->team.team_colour =
-            new_own_team.field_player_colour == gamecontroller::TeamColour::BLUE ? TeamColour::BLUE : TeamColour::RED;
-        state->opponent.team_colour = new_opponent_team.field_player_colour == gamecontroller::TeamColour::BLUE
-                                          ? TeamColour::BLUE
-                                          : TeamColour::RED;
+        state->team.team_colour     = get_team_colour(new_own_team.field_player_colour);
+        state->opponent.team_colour = get_team_colour(new_opponent_team.field_player_colour);
 
         /*******************************************************************************************
          * Process score updates
@@ -400,8 +397,7 @@ namespace module::input {
          * Process our team colour
          ******************************************************************************************/
         if (old_own_team.field_player_colour != new_own_team.field_player_colour) {
-            TeamColour colour = new_own_team.field_player_colour == gamecontroller::TeamColour::BLUE ? TeamColour::BLUE
-                                                                                                     : TeamColour::RED;
+            TeamColour colour = get_team_colour(new_own_team.field_player_colour);
             state_changes.emplace_back([this, colour] { emit(std::make_unique<TeamColour>(colour)); });
         }
 
@@ -638,5 +634,14 @@ namespace module::input {
         std::string addr = inet_ntop(AF_INET, &ip_addr_n, c, sizeof(c));
 
         return addr;
+    }
+
+    GameState::TeamColour::Value GameController::get_team_colour(const gamecontroller::TeamColour& colour) {
+        switch (colour) {
+            case gamecontroller::TeamColour::BLUE: return TeamColour::BLUE;
+            case gamecontroller::TeamColour::RED:  return TeamColour::RED;
+            // TODO: update GameState.proto TeamColour enum to support all v3 colours
+            default: return TeamColour::UNKNOWN;
+        }
     }
 }  // namespace module::input
--- a/module/input/GameController/src/GameController.cpp
+++ b/module/input/GameController/src/GameController.cpp
@@ -54,14 +54,7 @@ namespace module::input {
     using TeamColour     = GameState::TeamColour::Value;
 
     GameController::GameController(std::unique_ptr<NUClear::Environment> environment)
-        : Reactor(std::move(environment))
-        , receive_port(0)
-        , send_port(0)
-        , TEAM_ID(0)
-        , PLAYER_ID(0)
-        , packet()
-        , game_phase()
-        , set_play() {
+        : Reactor(std::move(environment)) {
 
         // Configure
         on<Configuration, Trigger<GlobalConfig>>("GameController.yaml")
--- a/module/input/GameController/src/GameController.cpp
+++ b/module/input/GameController/src/GameController.cpp
@@ -48,13 +48,20 @@ namespace module::input {
     using Unpenalisation = GameEvents::Unpenalisation;
     using HalfTime       = GameEvents::HalfTime;
     using KickOffTeam    = GameEvents::KickOffTeam;
-    using GameEventPhase      = GameEvents::GamePhase;
+    using GameEventPhase = GameEvents::GamePhase;
     using GameMode       = GameEvents::GameMode;
     using PenaltyReason  = GameState::PenaltyReason;
     using TeamColour     = GameState::TeamColour::Value;
 
     GameController::GameController(std::unique_ptr<NUClear::Environment> environment)
-        : Reactor(std::move(environment)), receive_port(0), send_port(0), TEAM_ID(0), PLAYER_ID(0), packet(), game_phase(), set_play() {
+        : Reactor(std::move(environment))
+        , receive_port(0)
+        , send_port(0)
+        , TEAM_ID(0)
+        , PLAYER_ID(0)
+        , packet()
+        , game_phase()
+        , set_play() {
 
         // Configure
         on<Configuration, Trigger<GlobalConfig>>("GameController.yaml")
@@ -127,8 +134,7 @@ namespace module::input {
                       reset_state();
                   });
 
-        on<Every<2, Per<std::chrono::seconds>>>().then("GameController Reply",
-                                                       [this] { send_reply_message(); });
+        on<Every<2, Per<std::chrono::seconds>>>().then("GameController Reply", [this] { send_reply_message(); });
     }
 
     void GameController::send_reply_message() {
@@ -143,13 +149,13 @@ namespace module::input {
 
         // fallen, pose, ball_age and ball should be optional to be sent to the GameController
         // TODO: wire up from localisation if required in the future
-        packet->fallen  = self_penalised ? 1 : 0;
-        packet->pose[0] = 0.f;
-        packet->pose[1] = 0.f;
-        packet->pose[2] = 0.f;
+        packet->fallen   = self_penalised ? 1 : 0;
+        packet->pose[0]  = 0.f;
+        packet->pose[1]  = 0.f;
+        packet->pose[2]  = 0.f;
         packet->ball_age = -1.f;
-        packet->ball[0] = 0.f;
-        packet->ball[1] = 0.f;
+        packet->ball[0]  = 0.f;
+        packet->ball[1]  = 0.f;
 
         if (game_controller_address != "") {
             emit<Scope::UDP>(packet, game_controller_address, send_port);
@@ -159,7 +165,7 @@ namespace module::input {
     void GameController::reset_state() {
 
         game_phase = static_cast<gamecontroller::GamePhase>(-1);
-        set_play = static_cast<gamecontroller::SetPlay>(-1);
+        set_play   = static_cast<gamecontroller::SetPlay>(-1);
 
         std::copy(std::begin(gamecontroller::RECEIVE_HEADER),
                   std::end(gamecontroller::RECEIVE_HEADER),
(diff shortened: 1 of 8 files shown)


module/input/GameController/src/GameController.cpp
```diff
@@ -224,11 +232,15 @@ namespace module::input {
...
+
         // Get colours
         state->team.team_colour =
-            new_own_team.team_colour == gamecontroller::TeamColour::BLUE ? TeamColour::BLUE : TeamColour::RED;
-        state->opponent.team_colour =
-            new_opponent_team.team_colour == gamecontroller::TeamColour::BLUE ? TeamColour::BLUE : TeamColour::RED;
+            new_own_team.field_player_colour == gamecontroller::TeamColour::BLUE ? TeamColour::BLUE : TeamColour::RED;
+        state->opponent.team_colour = new_opponent_team.field_player_colour == gamecontroller::TeamColour::BLUE
+                                          ? TeamColour::BLUE
+                                          : TeamColour::RED;
```
ysims: There are a lot of options for team colour now. You'll want a function to get TeamColour::<colour> per gamecontroller::TeamColour::<colour>.

module/input/GameController/src/GameController.cpp
```diff
@@ -446,88 +399,19 @@ namespace module::input {
         /*******************************************************************************************
          * Process our team colour
          ******************************************************************************************/
-        if (old_own_team.team_colour != new_own_team.team_colour) {
-            TeamColour colour =
-                new_own_team.team_colour == gamecontroller::TeamColour::BLUE ? TeamColour::BLUE : TeamColour::RED;
+        if (old_own_team.field_player_colour != new_own_team.field_player_colour) {
+            TeamColour colour = new_own_team.field_player_colour == gamecontroller::TeamColour::BLUE ? TeamColour::BLUE
+                                                                                                     : TeamColour::RED;
```
ysims: Switch to accept many colours

module/input/GameController/src/GameController.hpp
```diff
@@ -66,13 +66,14 @@ namespace module::input {
         ReactionHandle listen_handle;
 
         gamecontroller::GameControllerPacket packet;
-        gamecontroller::Mode mode;
+        gamecontroller::GamePhase game_phase;
+        gamecontroller::SetPlay set_play;
```
ysims: It would be good to add documentation directives to at least the variables and functions added. 
Additionally, initialise all of the variables here in the .hpp and remove the initialiser list in the .cpp

module/network/RobotCommunication/data/config/RobotCommunication.yaml
```diff
@@ -3,7 +3,9 @@ log_level: INFO
 startup_delay: 3
 ball_timeout: 3
 
-send_port: 3737
-receive_port: 3737
+# Team messages must be sent on port 10000 + team number assigned by TC at competition
+# NUbots team_id is 1 by default in config/teams.yaml of the GameController app, so port is 10001
+send_port: 10001
+receive_port: 10001
```
ysims: I'd grab the team ID from GlobalConfig in the Configuration block in this module and set the port using that. You could have a variable here that is the empty string, and if not empty string then it overrides the port in the code.

module/network/RobotCommunication/src/RobotCommunication.cpp
```diff
@@ -172,7 +182,7 @@ namespace module::network {
                          const std::shared_ptr<const GameState>& game_state,
                          const std::shared_ptr<const Purpose>& purpose,
                          const std::shared_ptr<const WalkTo>& walk_to,
-                         const std::shared_ptr<const Robots>& robot_localisation,
+                         // const std::shared_ptr<const Robots>& robot_localisation,
```
ysims: Is this the field that makes the message too large? I'd try removing covariance from the message. 
This is also a concern for the protocol itself. If filling in the protobuf exceeds the allowed size, then the team should propose a change to the protocol. My understanding is that the rules are in flux right now, and someone could ask the TC about it and make a proposal PR to the RoboCupProtocol repo. 
ie
- Test removing/changing fields (the covariance mats could become a single float value defined in some common way, ie the sum of the diagonal)
- PR to https://github.com/NUbots/RobocupProtocol with the proposed changes
- Message the TC about it, link to the PR
If they're happy, merge the PR and then they can update their fork when they are happy to push the change.

While everyone is already updating their communication files, might as well propose this. 
Although in saying all this, I realise that there may not be an official communication protocol for the new league yet.

module/network/RobotCommunication/src/RobotCommunication.cpp
```diff
@@ -272,45 +282,45 @@ namespace module::network {
...
-
-                        // Add robot information to list of other robots in message
-                        msg->others.push_back(std::move(rc_robot));
-                    }
-                }
+                // // Where the robot thinks the other robots are.
+                // if (robot_localisation && field) {
+
+                //     // Iterate through robots detected by localisation
+                //     for (const auto& local_bot : robot_localisation->robots) {
```
ysims: If you go ahead with just removing this component of the messsage, I wouldn't leave commented out code in here. Anyone can go back in the Git history and retrieve the code if it is reintroduced.

module/network/RobotCommunication/src/RobotCommunication.cpp
```diff
@@ -320,6 +330,26 @@ namespace module::network {
...
+                    const bool is_counting_state = game_state->phase == GameState::Phase::READY
+                                                   || game_state->phase == GameState::Phase::SET
+                                                   || game_state->phase == GameState::Phase::PLAYING;
+                    if (is_counting_state) {
+                        messages_sent++;
+                        log<DEBUG>("Messages sent this game:", messages_sent, "/ 12000");
+                    }
+                }
+
                 emit<Scope::UDP>(msg, cfg.broadcast_ip, cfg.send_port);
```
ysims, line 323: This should be guarded. If the message budget is met, you shouldn't be sending any more messages.

ysims:
Do the NUbook instructions for GameController still work? I was wondering if there's a new League GitHub organisation where they're keeping the GameController.

module/network/RobotCommunication/src/RobotCommunication.cpp
KimKhangHoang: I removed the commented and unused code. 

On the broader protocol point, is NUbots/RobocupProtocol the source of truth that teams follow, or is the protocol still being defined by the TC that needs checking?

KimKhangHoang:
> Do the NUbook instructions for GameController still work? I was wondering if there's a new League GitHub organisation where they're keeping the GameController.

I think the old setup should still mostly work with the newer GameController. The main thing is to reference the newer GC repo and dependency list if potentially needs updating.

module/input/GameController/src/GameController.cpp
KimKhangHoang: I've added `get_team_colour()` helper. Note that `GameState::TeamColour` currently only has BLUE and RED. Full colour support requires updating `GameState.proto`, which should affect some other modules as well. Should that be done in this PR?

module/input/GameController/src/GameController.hpp
KimKhangHoang: I've added `/// @brief` directives to all variables and functions, and moved initialisations to the `GameController.hpp`

module/network/RobotCommunication/data/config/RobotCommunication.yaml
KimKhangHoang: I've updated to compute port dynamically as `10000 + team_id` from `GlobalConfig`. Set to 0 in yaml to use auto-computed value, or override with a specific value when needed.

module/network/RobotCommunication/src/RobotCommunication.cpp
KimKhangHoang: I've added budget guard before emit. Also moved 512 and 12000 limits to yaml config as `max_message_size` and `max_messages_per_game` for easier updates if rules change.

module/input/GameController/src/GameController.cpp
ysims: Hmm it affects enough things that a separate PR makes sense.

module/network/RobotCommunication/src/RobotCommunication.cpp
ysims: I'd bring this up with @Claegtun who has been following the rules discussions. I'm not aware of any chatter around team communication protocol, it's likely fairly low on the list of priorities. 

https://github.com/RoboCup-Humanoid-TC/RobocupProtocol
The Humanoid League has promoted the NUbots-created RoboCup Protocol since it was created. Uptake by teams varied, some used their own thing and others used an older protocol. It was not enforced, just encouraged for coordination in drop-in games. The source of truth was the fork linked here in the RoboCup-Humanoid-TC organisation. 

But now I don't know if there is a new GitHub organisation for the new league and how the protocol from SPL would factor in, assuming they have one.

KimKhangHoang: Thanks for the context! I'll follow up with Clayton and look into whether the new league has an updated protocol repo. Will keep it on the radar as a follow-on task.

Claegtun: @ysims As far as I am aware, no one has talked about the RobocupProtocol within the league, with the notable exception of the joint Australian team between rUNSWift, Redback Bots, and us, only because I strongly advertised it as a starting point for our team-communication. However, it wasn't so much to revitalise the RobocupProtocol itself, but rather to use it as inspiration for our own protocol. In saying that, they haven't been many deviations from the original.

Coincidentally, in our discussions with the other Australian team, someone has raised the idea of removing the covariance in order to lessen the payload.

In the long-term, the three Australian teams definitely do want to make a universal standard for the whole league which may rekindle the original intention behind the RobocupProtocol. RMIT have strongly supported this idea. However, we have all agreed to put that aspiration aside until after this year's Robocup, since we just want to get something that works.

As for the former SPL, I believe that there was no existing equivalent to the RobocupProtocol. One thing to note is that, from what I have heard, the league will start heavily restricting wireless communication after this year, bit by bit every year, to the point of discouraging teams from using it at all. Apparently, the SPL had already started this. The team-communication rules were more strict than this year.

@KimKhangHoang My advice is to raise an issue on the repository for the RobocupProtocol and see whether anyone reacts. Another suggestion is to suggest a fork of the RobocupProtocol repository into the new NSL TC organisation. https://github.com/RoboCup-HumanoidSoccerLeague

ysims: They possibly suggested removing the covariance because of this reason. The thing with the protocol as well is that teams can make their own fields after index 100. So if there's anything you've been using that you don't think would fit in the universal protocol, then I'd stick it back there. 
If SPL doesn't have a protocol - which does make sense given they didn't do drop-in games - then I can't see why anyone would be against using the Humanoid League's old official one. With the allowance dropping, the protocol could change until it's phased out in favour of no wireless communication.
To ensure compatibility of data across the years, the fields that are dropped can be skipped indices. Protobuf is flexible in this regard.
It looks like the repos have no issue or discussion tabs.

ysims (approved):
Looks good. 

Sorting out team colour imo is an important next step. As is, you might have some issues with other robot team designation that would filter to strategy modules if a team has a colour other than red or blue. Robot localisation will need checking. 

Tbh it might be easier to switch to 'own team' or 'opponent team' in the system. Pretty much nothing atm actually cares what the colour is, just whether the robot is on its own team or not. 

Colour might be useful if you can integrate it into vision. Colour is also part of the RoboCup Protocol message, but it might not be needed in there. We use team colour to determine if the message is from a teammate, but with the rules designating ports per team, that's not really a problem. There could be a colour message sent by the GameController module that defines which colour own and opponent are, for use in systems like vision.

module/network/RobotCommunication/src/RobotCommunication.cpp
Claegtun: The problem seems to be that no one is really keeping an eye on that repository anymore, especially given that the parent organisation, i.e. RoboCup-Humanoid-TC, is now defunct. I don't see much point trying to maintain it there. Instead, I think that the best thing would be to get the attention of the new HSL TC and convince them to make a fork and make the needed changes in order for it to be compliant. Now, if only we had someone on the TC who also has experience with this ... @TrentHouliston.

PR #1779, opened 2026-05-30
Look for outdated landmarks

yippyy:
Just a first go at the outdated landmark search


module/strategy/StrategiseLook/src/StrategiseLook.cpp
```diff
@@ -85,6 +90,22 @@ namespace module::strategy {
...
+            [this](const Field& field,
+                   const std::shared_ptr<const Ball>& ball,
+                   const std::shared_ptr<const Goals>& goals) {
+                const bool stale_ball =
+                    !ball || (NUClear::clock::now() - ball->time_of_measurement) > cfg.ball_search_timeout;
+                const bool stale_goals       = !goals || goals->goals.empty()
+                                               || (NUClear::clock::now() - goals->timestamp) > cfg.goal_search_timeout;
+                const bool poor_localisation = field.cost > cfg.max_localisation_cost;
+
+                if ((stale_ball && stale_goals) || poor_localisation) {
```
ysims:
```suggestion
                if ((stale_ball || stale_goals) || poor_localisation) {
```

probably? You would want to look around for the ball even when the goals aren't stale. 
Using goals here might not be what you want - localisation can be good and goals stale, and that's still fine.

module/strategy/StrategiseLook/src/StrategiseLook.cpp
```diff
@@ -85,6 +90,22 @@ namespace module::strategy {
...
+                    emit<Task>(std::make_unique<LookAround>());
```
ysims, line 114: This Provides LookAround, and then it calls LookAround - this is circular? it wouldn't do anything, it would just keep calling itself again. I believe LookAround is already a planning-level Provider/Task that really does look around, so this Provider should be called something other than LookAround. The purpose of this Provider is to look around if features are stale. But find a nicer name than LookAroundIfFeaturesAreStale.

ysims (commented):
Looks good. You'd need to emit LookForStaleFeatures somewhere to add it to the tree, otherwise it'll never run. Also test to make sure this doesn't stop the robot scoring. Close to the goals, localisation or goals can be stale, but you dont necessarily want to stop. It might be fine if its just the head moving and the robot keeps walking to score.

yippyy:
> Looks good. You'd need to emit LookForStaleFeatures somewhere to add it to the tree, otherwise it'll never run. Also test to make sure this doesn't stop the robot scoring. Close to the goals, localisation or goals can be stale, but you dont necessarily want to stop. It might be fine if its just the head moving and the robot keeps walking to score.

i will test soon!

ysims (approved):

PR #1781, opened 2026-06-01
Added --jobs argument to ./b target

ZeoNyph:
This PR adds a `--jobs` argument to the `./b target` command, allowing someone to determine the maximum job count the build process can have (handy if you ever need to rebuild the image from scratch)

[NUbook PR here.](https://github.com/NUbots/NUbook/pull/355)

### DIFF
--- a/tools/target.py
+++ b/tools/target.py
@@ -46,15 +46,18 @@ def register(command):
     command.add_argument(
         "-r", "--reset", default=False, action="store_true", dest="reset", help="Reset buildx instance"
     )
+    command.add_argument(
+        "-j", "--jobs", default=defaults.jobs, help="the number of jobs to use when building the platform image"
+    )
 
 
-def run(target, username, uid, reset, **kwargs):
+def run(target, username, uid, reset, jobs, **kwargs):
     if target is None:
         target = platform.selected(defaults.image, username)
         print(f"Currently selected platform is {target}")
     else:
         # Ensure the platform image is built
-        platform.build(defaults.image, target, username, uid, reset)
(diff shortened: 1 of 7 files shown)


JohanneMontano:
do we need to update CI as well?

tools/target.py
```diff
@@ -46,15 +46,18 @@ def register(command):
     command.add_argument(
         "-r", "--reset", default=False, action="store_true", dest="reset", help="Reset buildx instance"
     )
+    command.add_argument(
+        "-j", "--jobs", default=defaults.jobs, help="the number of jobs to use when building the platform image"
```
JohanneMontano, line 50: more like `workers` imo but `j` is already set for build so eh

ysims:
The docker image is too big to build with GitHub runners. You need to build the image locally and then push it up to Dockerhub.

tools/target.py
ZeoNyph: i feel like while making it `workers` would make more sense, having it be jobs maintains parity with `./b build` and `install-from-source`, both of which take in a jobs argument anyway

ZeoNyph:
> do we need to update CI as well?

we did actually; the `jobs` arg default is 1 so it caused the jobs to time out

should be fixed now

tools/target.py
JohanneMontano: j should be fine, consistent with what we already have.

ZeoNyph:
> The docker image is too big to build with GitHub runners. You need to build the image locally and then push it up to Dockerhub.

@ysims how would I push it to Dockerhub once built?

PR #1783, opened 2026-06-03
Enhance nbs processing scripts

vincenttumminello:
This PR adds/amends 3 things:
- Enhances filter.py to accept a directory as input and sets the default output path a subdirectory called "filtered" inside the provided directory. 
- Adds --no-zeros option to json extractor and keeps zeros by default. 
- Adds --source option to video.py so nbs files containing both message.input.Image and message.output.CompressedImage can successfully be converted into mp4 files without filtering, defaults to CompressedImage. 
- Adds --Unix option to video.py which exports TimeCodes.txt with Unix time stamps instead. 
- also adds `libopenh264` decoder option because this seemed to be the only one to work with my os (Fedora)


tools/nbs/list_messages.py
```diff
@@ -0,0 +1,55 @@
...
+
+    type_names = []
+    for h in set(index["type_hash"]):
+        if h in MessageTypes:
+            type_names.append(MessageTypes[h].name)
+        else:
+            type_names.append(f"<unknown:{h:#018x}>")
+
+    for name in sorted(type_names):
+        print(name)
```
JohanneMontano: Why do this? We already have the `stats` tool for this. `./b nbs stats` also tells you the sizes of the messages in the nbs file.

tools/nbs/filter.py
```diff
@@ -55,26 +56,10 @@ def register(command):
...
-
-    # If we didn't provide any keep arguments we at least need to ignore them
-    if keep is None:
-        keep = []
-
+def _filter_to(input_files, output, keep, remove):
     with Encoder(output) as out:
         for packet in tqdm(
-            LinearDecoder(*files, show_progress=True),
+            LinearDecoder(*input_files, show_progress=True),
```
JohanneMontano, line 68: If you have a directory you wanna sweep, you just glob it like `./b nbs <tool> recordings/*.nbs`

tools/utility/nbs/__init__.py
```diff
@@ -26,7 +26,21 @@
...
+def resolve_nbs_paths(paths):
+    """Expand directories to .nbs files and drop .nbs.idx index files."""
+    result = []
+    for p in paths:
+        p_abs = os.path.abspath(p)
+        if os.path.isdir(p_abs):
+            result.extend(sorted(os.path.join(p_abs, name) for name in os.listdir(p_abs) if name.endswith(".nbs")))
+        elif p_abs.endswith(".nbs"):
+            result.append(p_abs)
+    return result
```
JohanneMontano, line 46: Not convinced why this is needed, the tools should be able to take a globbed path and it will automatically expand all the files that are inside the directory as mentioned, unless I am misunderstanding something here?

JohanneMontano (changes requested):

tools/nbs/filter.py
vincenttumminello: I did not realise you could glob like that. Will fix it up.

tools/nbs/json.py
```diff
@@ -30,19 +30,25 @@
...
-from utility.nbs import LinearDecoder
+from utility.nbs import LinearDecoder, resolve_nbs_paths
 
 
 def register(command):
     command.description = "Decode an nbs file and convert it to json"
 
     # Command arguments
     command.add_argument("files", metavar="files", nargs="+", help="The nbs files to convert to json")
+    command.add_argument(
```
Tom0Brien, line 41: I feel like this should just be default

vincenttumminello: Say less

tools/utility/nbs/__init__.py
vincenttumminello: It's not strictly needed per say. More a quality of life feature I wanted to add. You can still glob all the files in a folder with *.nbs if you prefer that. I just kept forgetting and it would try to filter the .idx files and break

tools/nbs/list_messages.py
vincenttumminello: Removed

tools/nbs/filter.py
vincenttumminello: I realised why I wanted it like this. See below comment

tools/nbs/images/video_recorder.py
```diff
@@ -53,11 +53,12 @@ def pix_fmt_from_fourcc(code):
 
 
 class Recorder:
-    def __init__(self, output_path, dimensions, fourcc, encoder, bitrate, buffer_size=100):
+    def __init__(self, output_path, dimensions, fourcc, encoder, bitrate, buffer_size=100, unix=False):
```
vincenttumminello, line 56: unix to `unix_ts`

tools/nbs/video.py
```diff
@@ -40,11 +40,19 @@
...
 # Configure TensorFlow to use CPU only for multiprocessing compatibility
 tf.config.set_visible_devices([], "GPU")
 
+# The message types each --source option reads.
+source_types = {
+    "compressed": ["message.output.CompressedImage"],
+    "raw": ["message.input.Image"],
+    "both": ["message.output.CompressedImage", "message.input.Image"],
+}
+
```
vincenttumminello, line 56: @run_on_docker
and remove libopenh264

PR #1786, opened 2026-06-13
Improve automated extrinsics calibration by moving the head around

JohanneMontano:


module/tools/ExtrinsicsCalibration/src/ExtrinsicsCalibration.cpp
```diff
@@ -119,11 +142,58 @@ namespace module::tools {
...
         });
 
+        // Stand the robot up so the head servos are powered and it holds a stable posture while calibrating.
+        on<Startup>().then("Calibration stance", [this] {
+            // These seed messages let modules that depend on a Stability / WalkState run.
+            emit(std::make_unique<Stability>(Stability::UNKNOWN));
+            emit(std::make_unique<WalkState>(WalkState::State::STOPPED));
+            // Always try to recover from a fall, and stand still (a zero-velocity walk).
+            emit<Task>(std::make_unique<FallRecovery>(), 5);
+            emit<Task>(std::make_unique<Walk>(Eigen::Vector3d::Zero()), 2);
```
ysims:
```suggestion
            emit<Task>(std::make_unique<FallRecovery>(), 1);
            emit<Task>(std::make_unique<Walk>(Eigen::Vector3d::Zero()));
```
Walk shouldn't interfere with the head, so they can be the same priority, but check.

module/tools/ExtrinsicsCalibration/src/ExtrinsicsCalibration.cpp
```diff
@@ -119,11 +142,58 @@ namespace module::tools {
...
+                && std::chrono::duration<double>(NUClear::clock::now() - last_scan_move).count()
+                       >= cfg.scan_fixation_time) {
+                scan_idx       = (scan_idx + 1) % cfg.scan_positions.size();
+                last_scan_move = NUClear::clock::now();
+            }
+            const Eigen::Vector2d& target = cfg.scan_positions[scan_idx];
+            const Eigen::Vector3d uPCt    = (Eigen::AngleAxisd(target.x(), Eigen::Vector3d::UnitZ())
+                                             * Eigen::AngleAxisd(target.y(), Eigen::Vector3d::UnitY()))
+                                            * Eigen::Vector3d::UnitX();
+            emit<Task>(std::make_unique<Look>(uPCt, true), 1);
```
ysims:
```suggestion
            emit<Task>(std::make_unique<Look>(uPCt, true));
```
this can be 0 (default), lowest priority.

module/tools/ExtrinsicsCalibration/src/ExtrinsicsCalibration.cpp
```diff
@@ -119,11 +142,58 @@ namespace module::tools {
...
+            emit(std::make_unique<WalkState>(WalkState::State::STOPPED));
+            // Always try to recover from a fall, and stand still (a zero-velocity walk).
+            emit<Task>(std::make_unique<FallRecovery>(), 5);
+            emit<Task>(std::make_unique<Walk>(Eigen::Vector3d::Zero()), 2);
+        });
+
+        // Automatically sweep the head through the configured (yaw, pitch) grid, lingering scan_fixation_time at
+        // each position, so detections span the image without any manual input. Re-emitting the Look task each
+        // tick keeps it alive and lets the Look skill smooth toward the target.
+        on<Every<10, Per<std::chrono::seconds>>>().then("Head scan", [this] {
```
ysims: There's a provider that handles head scanning. You can make a role-specific yaml file for the module to set specific positions.

ysims (dismissed):
What kind of performance are you getting with this? Is it looking to be a good alternative to manual tuning?

JohanneMontano:
@ysims 

before:

image

after (running the script once, looking at only one side of the field):
image

We're meant to do it on both sides to make sure it is usable. If run the binary another time on the other side of the field,

image

ysims (approved):
That's very nice - good stuff 😄

jpptm:
Another nice TODO is optimising `Hpc.translation()` along side the offsets, or just optimising Hpc itself. I haven't studied the math closely for this, but I will put a PR up once I'm confident with how viable it is.

PR #1787, opened 2026-06-17
Update NUClear and enable stats on BehaviourTask

ysims:
Updates NUClear to the latest commit, which allows us to turn on stats for BehaviourTasks. The result is that we can now get Director information in traces! 

NUClear has had a few updates, one around the network class required some minor changes in the code for getting a UDP packet address.


docker/Dockerfile
```diff
@@ -336,7 +336,9 @@ RUN install-from-source https://github.com/libuv/libuv/archive/refs/tags/v1.44.2
     -DBUILD_TESTING=OFF
 
 # NUClear!
-RUN install-from-source https://github.com/Fastcode/NUClear/archive/925dca0f31484a7df64fd335de5a6c9335483c7f.tar.gz \
+# RUN install-from-source https://github.com/Fastcode/NUClear/archive/925dca0f31484a7df64fd335de5a6c9335483c7f.tar.gz \
+#     -DBUILD_TESTS=OFF
+RUN install-from-source https://github.com/Fastcode/NUClear/archive/9159be49a330437cc6f74c1f156dadd8769c10f9.tar.gz \
```
ysims:
```suggestion
RUN install-from-source https://github.com/Fastcode/NUClear/archive/30ee1cda31f4b99715e67aff188065b67cb6a1fd.tar.gz \
```

docker/Dockerfile
```diff
@@ -336,7 +336,9 @@ RUN install-from-source https://github.com/libuv/libuv/archive/refs/tags/v1.44.2
     -DBUILD_TESTING=OFF
 
 # NUClear!
-RUN install-from-source https://github.com/Fastcode/NUClear/archive/925dca0f31484a7df64fd335de5a6c9335483c7f.tar.gz \
+# RUN install-from-source https://github.com/Fastcode/NUClear/archive/925dca0f31484a7df64fd335de5a6c9335483c7f.tar.gz \
+#     -DBUILD_TESTS=OFF
```
ysims:
```suggestion
```

module/purpose/Soccer/data/config/Soccer.yaml
```diff
@@ -2,7 +2,7 @@
 log_level: INFO
 
 # For the robot playing without listening to GameController
-force_playing: false
+force_playing: true
```
ysims:
```suggestion
force_playing: false
```

JohanneMontano (dismissed):
Checked the NUClear repo if the pr was merged and it was, lgtm

JohanneMontano:
`The following tests FAILED:
	  2 - TestextensionDirector (Failed)
Errors while running CTest
FAILED: CMakeFiles/test.util 
cd /__w/NUbots/NUbots/build && /usr/bin/ctest --force-new-ctest-process
ninja: build stopped: subcommand failed.
`
Oof

ysims:
> `The following tests FAILED: 2 - TestextensionDirector (Failed) Errors while running CTest FAILED: CMakeFiles/test.util cd /__w/NUbots/NUbots/build && /usr/bin/ctest --force-new-ctest-process ninja: build stopped: subcommand failed. ` Oof

Yes lmao the Director tests seem to have all failed 💀

module/extension/Director/tests/TestBase.hpp
```diff
@@ -70,12 +73,35 @@ class TestBase : public extension::behaviour::BehaviourReactor {
...
+        // system has not shut down cleanly by then.
+        on<Always>().then("Test Timeout", [this, timeout] {
+            if (clean_shutdown) {
+                return;
+            }
+            std::unique_lock<std::mutex> lock(timeout_mutex);
+            timeout_cv.wait_for(lock, timeout);
+
+            if (!clean_shutdown) {
+                emit<Scope::INLINE>(std::make_unique<Fail>("Test did not complete successfully"));
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Good catches on the race and the wording. Fixed in aab87513ca0:

- `clean_shutdown` is now only ever read under `timeout_mutex` — the initial unlocked `if (clean_shutdown) return;` is gone and the check is folded into the `wait_for` predicate (`wait_for(lock, timeout, [this]{ return clean_shutdown; })`), so it no longer races with the `on<Shutdown>` handler that writes it.
- Corrected the misleading "spawn a real thread" comment. It isn't a `std::thread` — it's an `on<Always>` reaction, which NUClear schedules on its own dedicated single-thread pool (`ThreadPoolDescriptor(name, 1, false)`), separate from the default pool. Because of that dedicated pool, blocking here does **not** stall the default pool or other reactions even when `default_pool_concurrency == 1`, so there's no powerplant deadlock; the comment now says so explicitly.

module/extension/Director/tests/TestBase.hpp
```diff
@@ -70,12 +73,35 @@ class TestBase : public extension::behaviour::BehaviourReactor {
...
+                powerplant.shutdown(true);
+            }
+        });
+
         on<Trigger<Fail>, MainThread>().then([this](const Fail& f) {
             INFO(f.message);
             CHECK(false);
-            powerplant.shutdown(true);
         });
```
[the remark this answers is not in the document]
TrentHouliston: :robot: Fixed in aab87513ca0 — restored `powerplant.shutdown(true)` in the `Fail` handler so an emitted `Fail` (from any test) terminates the run promptly rather than hanging until the 1s watchdog timeout. The timeout path still emits `Fail` inline and shuts down itself; the extra shutdown is idempotent.

PR #1788, opened 2026-06-20
Add RFDETR for object detection

jpptm:
| Metric | YOLO (baseline) | RF-DETR small fp32 | RF-DETR small fp16 |
| :--- | :--- | :--- | :--- |
| **mAP50-95** | 0.4424 | 0.4694 | 0.4695 |
| **mAP50** | 0.7034 | 0.7621 | 0.7634 |
| **Best-F1** | 0.7921 | 0.8201 | 0.8197 |
| **Precision** | 0.9395 | 0.9069 | 0.9085 |
| **Recall** | 0.6846 | 0.7484 | 0.7467 |
| **File size** | — | 120.8 MB | 64.1 MB |

TLDR: RFDETR small is better and more confident at detecting than YOLO, and compressing it to fp16 is virtually free. These models were tested using https://huggingface.co/datasets/NUbots/YOLOSoccer.


jpptm:
very nice

image

module/vision/RFDETR/data/config/webots/RFDETR.yaml
```diff
@@ -0,0 +1,27 @@
+# Controls the minimum log level that NUClear log will display
+log_level: INFO
+
+# Path to the model weights (ONNX, loaded directly by OpenVINO)
+model_path: "rfdetr.onnx"
```
ysims, line 5: There are two onnx files and neither have this name.

module/vision/RFDETR/data/config/webots/RFDETR.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+# Global minimum score (sigmoid of the class logit) below which detections are discarded
+score_threshold: 0.3
+
+# Class-channel offset: logit channel = class_id + offset.
+# The shipped RF-DETR export currently behaves correctly with offset 0.
+class_offset: 0
+
+# OpenVino device (CPU, GPU). Webots typically runs on CPU.
+device: CPU
```
ysims, line 27: The only difference here is the device. Might as well remove all other fields?

module/vision/RFDETR/data/config/RFDETR.yaml
```diff
@@ -0,0 +1,28 @@
...
+ball_confidence_threshold: 0.5
+
+# Minimum confidence required for a goal detection to be accepted
+goalpost_confidence_threshold: 0.5
+
+# Minimum confidence required for a robot detection to be accepted
+robot_confidence_threshold: 0.5
+
+# Minimum confidence required for a field line intersection detection to be accepted
+intersection_confidence_threshold: 0.5
```
ysims, line 17: These are all 0.5 - have they been tuned at all? Goals in that image looks like its not doing so well. Maybe need to turn the threshold down?

module/vision/RFDETR/src/RFDETR.hpp
```diff
@@ -0,0 +1,87 @@
...
+        };
+
+        /// @brief The objects that the RF-DETR model can detect. The order is fixed and must match
+        /// the training class order (see ../training/convert_to_coco.py CANONICAL_CLASSES).
+        std::vector<Object> objects = {{"ball", Eigen::Vector4d(1, 1, 1, 1), 0.0},
+                                       {"goal post", Eigen::Vector4d(1, 0, 1, 1), 0.0},
+                                       {"robot", Eigen::Vector4d(1, 0.5, 0, 1), 0.0},
+                                       {"L-intersection", Eigen::Vector4d(1, 0, 0, 1), 0.0},
+                                       {"T-intersection", Eigen::Vector4d(0, 1, 0, 1), 0.0},
+                                       {"X-intersection", Eigen::Vector4d(0, 0, 1, 1), 0.0}};
```
ysims, line 77: this looks like it should be in config?

module/vision/RFDETR/README.md
```diff
@@ -0,0 +1,74 @@
+# RFDETR
+
+## Description
+
+This module integrates an **RF-DETR** (Roboflow Detection Transformer) model to
+identify and classify objects within images. It is the successor to the `Yolo`
+module (YOLOv8-nano) and detects the same classes, in the same order:
```
ysims: It's replacing the version 11 module, not 8. I wouldn't mention YOLO here though, the comparison will become stale very quickly and it's not relevant to the module. You don't need to know anything about the previous model to understand this module. The only usefulness would be to mention the performance improvement over YOLO, to motivate its use.

module/vision/RFDETR/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+- Balls
+- Goals
+- Robots
+- Field Line Intersections (L, T, X)
```
ysims, line 11: The order comment seems brittle. The order isn't fixed in the module, is it? Just the block I commented on and the model need to match?

module/vision/RFDETR/README.md
```diff
@@ -0,0 +1,74 @@
...
+Confidence thresholds for each class can be specified in the config.
+
+Inference runs on the CPU or GPU using OpenVINO
+(https://github.com/openvinotoolkit/openvino), loading the exported ONNX model
+directly.
+
+### How it differs from `Yolo`
+
+RF-DETR is a DETR-style detector, so the network contract differs from YOLOv8 and
+the decoder/preprocessing are not interchangeable:
```
ysims: The current yolo model is v11. As above, I'm not sure if it makes sense to write the readme with this comparative positioning, and instead focus on the module itself.

module/vision/RFDETR/README.md
```diff
@@ -0,0 +1,74 @@
...
+- **Preprocessing** — square resize → scale to `[0, 1]` → ImageNet `mean/std`
+  normalisation → NCHW, RGB. This must match the training/export transform
+  exactly (see `../training/verify_onnx.py`).
+- **Class-channel offset** — the exported model carries an extra (unused)
+  background channel, so `num_classes` is one larger than the number of objects.
+  The module auto-detects this (`class_offset: -1`).
+
+Everything downstream of detection (ray projection onto the field plane, message
+emission, green-horizon filtering) is identical to `Yolo`, so consumers
+(`BallLocalisation`, `RobotLocalisation`, `FieldLocalisationNLopt`) are unchanged.
```
ysims: This reads more as a changelog than a readme for this module. This comment doesn't tell future readers much. This is also pretty brittle to changes in the other modules. NUClear aims to be highly decoupled, so it shouldn't matter what downstream tasks exist. The main thing is the module input and outputs.

module/vision/RFDETR/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+The ONNX model is produced by the offline RF-DETR training pipeline in
+`../training/` (outside the NUbots repo). It fine-tunes a pretrained RF-DETR on
+the [`NUbots/YOLOSoccer`](https://huggingface.co/datasets/NUbots/YOLOSoccer)
+dataset and installs `data/rfdetr.onnx`. See `../training/README.md`.
+
+## Usage
+
+Include this module to detect balls, goals, robots and field line intersections
+in images. Add `vision::RFDETR` to a role in place of `vision::Yolo`.
```
ysims: This is a very brittle comment. If all the roles are using this module and not yolo (particularly if YOLO is removed) then this comment no longer adds value.

roles/test/localisation.role
```diff
@@ -26,8 +26,8 @@ nuclear_role(
   vision::GreenHorizonDetector
   vision::BallDetector
   vision::FieldLineDetector
-  vision::Yolo
-  # Localisation
+  vision::RFDETR
+  # vision::Yolo
```
ysims, line 30: This could cause a lot of confusion. If the test module isn't using the same modules as robocup, it's hard to understand how the robot will perform using the robocup binary. If this is performing better, then swap yolo for RFDETR everywhere? If it isn't tested/tuned enough yet, then I wouldn't be adding it in here - another team member testing a different component would be very confused if suddenly their tests were all failing. 
I also wouldn't be keeping commented out lines in here. Certainly not without a comment, anyway. The localisation comment is also removed?

ysims (commented):
How much of the new module is copied from the YOLO module? If the object detection (post-processing) components can be separated, it makes sense to separate them into different modules. Eg just the bounding boxes get emitted by RFDETR/YOLO and then we still have the BallDetection module right? Make reactions for accepting bounding box and segmentation, and you could have a config variable that disables or enables each.
Then when a new model is integrated, all the rest isn't having to be reintegrated/duplicated.

module/vision/RFDETR/data/config/RFDETR.yaml
JohanneMontano: I dont think bringing the threshold down makes sense, given that both precision and recall on RFDETR using the same dataset is consistently higher than yolo's. These should be fine for now, until the need for them to change comes.

module/vision/RFDETR/README.md
JohanneMontano: this is a reminder to myself, the pr is not complete yet. It works pretty good though, much better in confidence, correctness and range

roles/test/localisation.role
JohanneMontano: Id argue that commenting a module serves as a reminder to everyone, at least for now since it's good for prototyping. Someone who sees this should be asking anyway, and there is usually only 1 to 2 people flashing code to the robot. Im trying to apply changes bit by bit so its easier to decode, instead of introducing bugs everywhere all at once

module/vision/RFDETR/src/RFDETR.hpp
JohanneMontano: Yolo's version of this is also in its header file as well

module/vision/RFDETR/data/config/RFDETR.yaml
ysims: I'd go from what you see on the robot, not the metrics from the dataset - they'll have different distributions. Just wanted to flag that it might be worth tuning, whether that's now or later. YOLO is currently using very low confidence thresholds for most object classes.

module/vision/RFDETR/src/RFDETR.hpp
ysims: I'd change it there too 😆 not big deal tho.

module/vision/RFDETR/README.md
```diff
@@ -0,0 +1,44 @@
+# RFDETR
+
+## Description
+
+This module integrates an **RF-DETR** (Roboflow Detection Transformer) model to
+identify and classify objects within images. It detects the follwing classes, in the same order:
```
ysims:
```suggestion
identify and classify objects within images. It detects the following classes, in the same order:
```

module/vision/RFDETR/data/config/RFDETR.yaml
jpptm: YOLO constantly misdetects X landmarks as balls, RFDETR does not

jpptm:
YOLO is freaking out rn, Im just gonna merge this to robocup 2026 first and test stuff

PR #1795, opened 2026-06-26
Implement Whistle Detection :dancing_women:

miles-p:
Things the PR does go here

- implement the whistle detection emitter
- adds a reactor for the Whistle message that changes the gamestate

image

I wrote a NUbook PR for this change: [NUbook PR #358](https://github.com/NUbots/NUbook/pull/358)


module/input/WhistleDetection/src/WhistleDetection.cpp
```diff
@@ -0,0 +1,312 @@
...
+                    const auto v       = x[i + j + len / 2] * w;
+                    x[i + j]           = u + v;
+                    x[i + j + len / 2] = u - v;
+                    w *= wlen;
+                }
+            }
+        }
+    }
+
+    void WhistleDetection::setup_audio() {
```
ysims, line 49: This looks like something that would go into the util folder. Do we already have something in there? Possibly none of those branches went in, so we might not.

module/input/WhistleDetection/data/config/WhistleDetection.yaml
```diff
@@ -0,0 +1,42 @@
+# Controls the minimum log level that NUClear log will display
+log_level: INFO
+
+use_whistle_detection: true
```
ysims: While this pains me, I understand why you'd want it in. It's quicker to turn off here than reconfig and copy over binaries. I'd put a comment though.

module/input/WhistleDetection/src/WhistleDetection.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+namespace module::input {
+
+    using extension::Configuration;
+    using message::input::Whistle;
+    using utility::nusight::graph;
+
+    // Radix-2 Cooley-Tukey FFT — in-place, N must be a power of two.
+    // Reference: same spectral feature used by NaoDevils' WhistleNet training pipeline.
+    static void fft(std::vector<std::complex<float>>& x) {
```
ysims: this looks like something that definitely should go in a util folder

module/input/WhistleDetection/src/WhistleDetection.cpp
```diff
@@ -0,0 +1,312 @@
...
+            cfg.startup_delay         = config["startup_delay"].as<int>();
+            cfg.use_whistle_detection = config["use_whistle_detection"].as<bool>();
+
+            if (log_level <= NUClear::LogLevel::DEBUG) {
+                emit(graph("Whistle/cfg/freq_min", cfg.freq_min));
+                emit(graph("Whistle/cfg/freq_max", cfg.freq_max));
+                emit(graph("Whistle/cfg/band_ratio_threshold", cfg.band_ratio_threshold));
+                emit(graph("Whistle/cfg/min_band_energy", cfg.min_band_energy));
+                emit(graph("Whistle/cfg/confirm_frames", static_cast<float>(cfg.confirm_frames)));
+            }
```
ysims: Graphs for constants? Maybe just have a terminal print block.

jpptm:
Need to test it first with the actual whistle when we arrive, and make sure it works with game controller. Will review once Im back

module/input/WhistleDetection/src/WhistleDetection.hpp
```diff
@@ -0,0 +1,97 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2024 NUbots
```
ysims:
```suggestion
 * Copyright (c) 2026 NUbots
```

check others too

module/input/WhistleDetection/src/WhistleDetection.hpp
```diff
@@ -0,0 +1,97 @@
...
+            /// Minimum absolute in-band power (filters near-silence)
+            float min_band_energy{1e-6f};
+            /// Number of consecutive FFT frames above threshold required to confirm a whistle
+            int confirm_frames{3};
+            /// Minimum time between consecutive whistle events in milliseconds
+            int cooldown_ms{1500};
+            /// Milliseconds to suppress detections after startup (avoids triggering on boot beep)
+            int startup_delay{1000};
+            /// Whether to use whistle detection or not
+            bool use_whistle_detection{true};
```
ysims: I think it's neater to zero these

module/input/WhistleDetection/data/config/WhistleDetection.yaml
ysims: You could move this to the behaviour module, where the trigger for whistle is. 
My thinking is that it feels wrong to have a bool to essentially turn a whole module off, when you could just remove the module from the program.

module/input/WhistleDetection/data/config/WhistleDetection.yaml
```diff
@@ -0,0 +1,42 @@
...
+# Number of consecutive FFT frames that must exceed the threshold before a Whistle event
+# is emitted.  At 50% overlap each frame represents 32 ms of new audio, so 3 frames ≈
+# 96 ms of sustained whistle-like signal.
+confirm_frames: 3
+
+# Minimum time in milliseconds between successive Whistle events (prevents re-triggering
+# on the tail of the same whistle).
+cooldown_ms: 1500
+
+# it detects the NUsense beep as a whistle, so the first second or two have to be ignored.
```
ysims, line 41:
```suggestion
# It detects the NUsense beep as a whistle, so the first second or two have to be ignored.
```
You could probably tune NUSense's buzzer out, right? It's a lower frequency iirc from a whistle. Can you change NUSense's buzzer frequency? If you can, I'd do that as well to make sure it's definitely not picked up.

module/input/WhistleDetection/src/WhistleDetection.cpp
```diff
@@ -0,0 +1,312 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2024 NUbots
```
ysims:
```suggestion
 * Copyright (c) 2026 NUbots
```

module/input/WhistleDetection/src/WhistleDetection.cpp
```diff
@@ -0,0 +1,312 @@
...
+        }
+    }
+
+    void WhistleDetection::setup_audio() {
+        if (pcm_handle != nullptr) {
+            snd_pcm_close(pcm_handle);
+            pcm_handle = nullptr;
+        }
+
+        // The NUBots-built ALSA looks for its config at /usr/local/share/alsa/ which isn't
```
ysims:
```suggestion
        // The NUbots-built ALSA looks for its config at /usr/local/share/alsa/ which isn't
```

module/input/WhistleDetection/src/WhistleDetection.cpp
```diff
@@ -0,0 +1,312 @@
...
+        // deployed on the robot. Fall back to the system config so device names resolve.
+        setenv("ALSA_CONFIG_PATH", "/usr/share/alsa/alsa.conf", 0);
+
+        int err = snd_pcm_open(&pcm_handle, cfg.device.c_str(), SND_PCM_STREAM_CAPTURE, SND_PCM_NONBLOCK);
+        if (err < 0) {
+
+            // figured this out the hard way - we need to be in the 'audio' group to access the microphone, otherwise we
+            // get a permission denied error. 'sudo usermod -aG audio $USER' worked for me. also make sure we have all
+            // the ALSA packages or it won't work.
```
ysims, line 64: this would be something for the Dockerfile, right?

module/input/WhistleDetection/src/WhistleDetection.hpp
```diff
@@ -0,0 +1,97 @@
...
+        } cfg;
+
+        /// ALSA PCM capture handle
```
ysims: You might want to follow the convention of using `@brief` for all of the variable comments

```suggestion
        /// @brief ALSA PCM capture handle
```

module/input/WhistleDetection/src/WhistleDetection.hpp
```diff
@@ -0,0 +1,97 @@
...
+        NUClear::clock::time_point start_time{};
+
+        /// Handle for the periodic audio-read reaction (allows disabling during reconfiguration)
+        ReactionHandle audio_handle{};
+
+        /// Opens (or re-opens) the ALSA capture device with the current configuration
+        void setup_audio();
+
+        /// Runs the FFT-based whistle detection pipeline on one window of samples
+        void process_frame(const std::vector<float>& samples);
```
ysims:
```suggestion
        /// @brief Runs the FFT-based whistle detection pipeline on one window of samples
        /// @param samples <description>
        void process_frame(const std::vector<float>& samples);
```
When defining functions, use `@param` and `@return` where applicable to define those variables.

shared/message/input/Whistle.proto
```diff
@@ -0,0 +1,39 @@
+// MIT License
+//
+// Copyright (c) 2024 NUbots
```
ysims:
```suggestion
// Copyright (c) 2026 NUbots
```

module/input/WhistleDetection/data/config/WhistleDetection.yaml
miles-p: good point. I added a clarification comment. I get your trepidation, but it should be easy enough to turn off I think. Especially in going to Robocup where I want to be able to turn it off if it doesn't work/picks up too much.

module/input/WhistleDetection/data/config/WhistleDetection.yaml
miles-p: good point. for some reason some of them have different frequencies. I believe this is because they're active buzzers with one frequency, so if they're different parts, they will emit a different frequency.

ysims: Ah that's annoying.

module/input/WhistleDetection/src/WhistleDetection.cpp
miles-p: whoops :D

module/input/WhistleDetection/src/WhistleDetection.cpp
miles-p: i don't think so - it's related more to the host OS I believe. In resolving the issue on Kevin, it all had to be done on the bare metal OS.

module/input/WhistleDetection/src/WhistleDetection.cpp
miles-p: it was for my own view on PJ. it doesn't even need to exist.

module/input/WhistleDetection/src/WhistleDetection.hpp
miles-p: copy

module/input/WhistleDetection/src/WhistleDetection.hpp
miles-p: done :D

module/input/WhistleDetection/src/WhistleDetection.cpp
miles-p: this is fair. I'm gonna put it in a utils folder for this specific module and just send it :D

module/input/WhistleDetection/src/WhistleDetection.cpp
miles-p: nothing uses ALSA. Johanne says the last thing that would have tried this was the ChatGPT stuff, but that was never sent to main. It can probably go somewhere else, but that can be done by the next person who needs the audio stuff somewhere else I think.

module/input/WhistleDetection/src/WhistleDetection.hpp
miles-p: :)

jpptm:
make a test/whistle role/module and test logs so we can see actual emits being triggered/ simulate it as close to an actual game as possible

miles-p:
Added a test/whistle role for you Johanne

PR #1796, opened 2026-06-27
Defensive positioning upon opponent's set play

KimKhangHoang:
This PR makes us emit defend/support during opponent's set play (e.g, throw-in, goal kick) instead of doing nothing. 

### Pre-PR Checklist:

Have you

Below is a demo video with logging, but my machine is too weak that webots cannot show any meaningful behaviour. Hence, further observation is still needed later.

[Screencast from 2026-06-28 12-22-38.webm](https://github.com/user-attachments/assets/c9df4305-0937-490c-830a-50b035eff244)


module/purpose/FieldPlayer/src/FieldPlayer.cpp
```diff
@@ -112,13 +112,6 @@ namespace module::purpose {
                 bool set_play = game_state.mode.value >= GameState::Mode::DIRECT_FREEKICK
                                 && game_state.mode.value <= GameState::Mode::THROW_IN;
 
-                // If it's the opponent's set play, freeze until the ball is free
-                // Ball is free when secondary_time expires or the ball has moved
-                if (set_play && !game_state.our_kick_off) {
-                    log<DEBUG>("Opponent set play, waiting for ball to be free.");
-                    return;
-                }
-
```
ysims, line 121: this is covered by game_state stopped now right? Trying to learn the new rules through NUbots code changes 😆 if so, looks good

KimKhangHoang: yeah, I’m not sure what the rule was before around freezing when it’s not our set play, but I think the game can now be paused arbitrarily via game_state stopped, and we can move if the set play isn’t ours 🙂.

KimKhangHoang:
@ysims I noticed the weird defend positioning might be a localisation/webots issue. In the logs, the field frame has the ball at (-2.41, 0.26), even though it was physically near the centre, with the goal at (4.5, 0). The defend target seems to be computed correctly from those values. Do you reckon this is fine to merge?

ysims:
Yeah you should be good - maybe check the other files changed in here but I think they're all just formatting changes because I assume the robocup branch isn't formatted
Make sure you do a PR into `main` as well at some point

KimKhangHoang:
> Yeah you should be good - maybe check the other files changed in here but I think they're all just formatting changes because I assume the robocup branch isn't formatted Make sure you do a PR into `main` as well at some point

Quick question: after this PR gets merged into the robocup branch, do I need to open a PR into main soon as well, or is that something the team usually handles after robocup?

PR #1798, opened 2026-06-29
Treat goal as a region instead of as a point

jpptm:


Tom0Brien (commented):
Should be tested on robot

PR #1813, opened 2026-07-02
Check if the robot is penalised before starting

jpptm:


jpptm:
Clayton said I could merge it :)

PR #1814, opened 2026-07-02
Comply HSL bandwidth limits for localisation in NUsight

KimKhangHoang:
This PR adds a combined, size-bounded debug message (HSLLocalisationDebug) that bundles field intersections, robot positions, ball, and field localisation into a single packet forwarded via NetworkForwarder. It complies with the 2026 HSL rule limiting each robot to 1 UDP packet/sec of debug information to an external device.

#### Tests
- Ran `webots-robocup` connected to NUsight, sniffed the NUClearNet UDP channel with tcpdump across multiple scenes.
- Rate is consistently below 1/sec both via wall-clock sniffing and directly via NetworkForwarder's own NUClear::clock-based timing.
- Size: tested against the true worst case for this field/division (18 field-line intersections and 8 robots). Initially this fragmented into 2 UDP packets (1942B); dropping `Field.association_lines` brought it to a single packet (1202B) with no fragmentation.

### Pre-PR Checklist:

Have you

#### Things left to do
- Test on real robot hardware/network, not just webots simulation.

Screenshot from 2026-07-03 11-26-36

Screenshot from 2026-07-02 23-01-47


Tom0Brien:
Would it be better to update the Overview packet to include this data, we already had that module for this purpose, and that also gives information on game state robot thinks its in, penalised etc.

KimKhangHoang:
I do think that would be better, but kept it separate mainly because when we cram field intersections/robots/field data in there too, it may run into the same packet-splitting/fragmentation issue from this PR. Also a bit extra work on the nusight client side needs to be bundled to get this data showing up in Overview. So it's basically just the matter of rule-based testing and timing.

jpptm (approved):
I think it LGTM

PR #1816, opened 2026-07-12
Fix Webots clock rtf estimation

jpptm:
Problem
 
On sufficiently fast machines, a robot in Webots cannot turn on the spot (at the very least in my machine). Low commanded yaw rates produce almost no rotation, higher rates reverse the turn direction entirely, and localisation collapses once the behaviour starts correcting. The same code, Webots version (R2022b), world, and configs work perfectly on slower machines (Alienware pop os), which made this look like a machine-specific mystery (OS, physics threading, servo mapping and friction were all ruled out by bisection and instrumentation).

Root cause

The real time factor used to slave the NUClear clock to Webots sim time was estimated as the ratio of the last two sim/real time deltas, exponentially smoothed per message. Sensor messages arrive once per sim step, and a host that computes a sim step faster than the basic time step executes steps in bursts between Webots' real-time pacing sleeps. Burst-rate messages vastly outnumber pacing-sleep messages, so the per-message filter converges to the burst rate (measured 1.37 on a fast machine, flapping 0.96-2.0 with the default smoothing) instead of the true average of ~1.0. The NUClear clock then runs fast relative to the simulation, desynchronising the walk phase and servo speed calculations: the robot cannot turn on the spot (higher commanded yaw rates reverse the turn direction) and downstream localisation collapses. Slow hosts take at least the basic time step of wall time per sim step, never burst, and are unaffected - which made this look like a machine-specific mystery.

This pull request estimates the rtf as the slope of sim time against real time over a sliding 2 second real-time window instead. This weights time rather than messages, is immune to step bursting, and tracks genuine speed changes. The window resets when the simulation time resets. The clock_smoothing config parameter is no longer used by this path.

Verified on the affected machine with a hip yaw command/measurement trace: before the fix rtf sat at ~1.37 and servo tracking lagged ~200ms behind commands; after the fix rtf tracks the Webots speed factor and keyboardwalk turns correctly in both directions, and the robocup role walks to the ready position and holds localisation.

### DIFF
--- a/module/platform/Webots/src/Webots.hpp
+++ b/module/platform/Webots/src/Webots.hpp
@@ -35,11 +35,11 @@
 #include <deque>
 #include <map>
 #include <mutex>
-#include <utility>
 #include <nuclear>
 #include <string>
 #include <tinyrobotics/kinematics.hpp>
 #include <tinyrobotics/parser.hpp>
+#include <utility>
 #include <vector>
 
 #include "message/input/Image.hpp"
--- a/module/platform/Webots/src/Webots.hpp
+++ b/module/platform/Webots/src/Webots.hpp
@@ -32,8 +32,10 @@
 #include <Eigen/Geometry>
 #include <array>
 #include <atomic>
+#include <deque>
 #include <map>
 #include <mutex>
+#include <utility>
 #include <nuclear>
 #include <string>
 #include <tinyrobotics/kinematics.hpp>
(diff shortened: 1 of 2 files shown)


jpptm:
Someone testing this on their machine using webots/robocup and confirming if it works or not would be great. At the very least for me, webots in the current main just does not work when the robot wants to turn (there is no problem when the robot walks straight)

Tom0Brien (dismissed):
Nice, the old approach has always been a bit of problem, especially if the Sim RTF spikes a lot.

module/platform/Webots/src/Webots.hpp
```diff
@@ -93,6 +95,10 @@ namespace module::platform {
         /// @brief Real time factor of the simulation clock
         double rtf = 1.0;
 
+        /// @brief Sliding window of (sim time ticks, real time ms) pairs used to estimate the real
+        /// time factor as a time-weighted slope rather than a per-message average
+        std::deque<std::pair<uint32_t, uint64_t>> clock_window;
```
ysims:
```suggestion
        std::deque<std::pair<uint32_t, uint64_t>> clock_window{};
```

ysims (approved):

PR #1817, opened 2026-07-12
Add PlotJuggler Series for localisation as an XY plot

miles-p:
Things the PR does go here

- adds another PlotJuggler series for localisation as an XY plot.

Example:
image

### DIFF
--- a/module/localisation/FieldLocalisationNLopt/src/FieldLocalisationNLopt.cpp
+++ b/module/localisation/FieldLocalisationNLopt/src/FieldLocalisationNLopt.cpp
@@ -360,10 +360,10 @@ namespace module::localisation {
     void FieldLocalisationNLopt::debug_field_localisation(const Eigen::Isometry3d& Hfw, const Eigen::Isometry3d& Htw) {
         emit(graph("opt state", state.x(), state.y(), state.z()));
         emit(graph("filtered state", filtered_state.x(), filtered_state.y(), filtered_state.z()));
-        Eigen::Isometry3d Htf = Hfw * Htw.inverse();
-        emit(graph("Htf position (x, y)",
-                   Htf.translation().x(),
-                   Htf.translation().y()));  // to view this in PlotJuggler as an XY plot, control click the x and y
+        Eigen::Isometry3d Hft = Hfw * Htw.inverse();
+        emit(graph("Hft position (x, y)",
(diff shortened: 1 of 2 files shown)


module/localisation/FieldLocalisationNLopt/src/FieldLocalisationNLopt.cpp
```diff
@@ -356,9 +357,16 @@ namespace module::localisation {
                 });
     }
 
-    void FieldLocalisationNLopt::debug_field_localisation(Eigen::Isometry3d Hfw) {
+    void FieldLocalisationNLopt::debug_field_localisation(const Eigen::Isometry3d& Hfw, const Eigen::Isometry3d& Htw) {
         emit(graph("opt state", state.x(), state.y(), state.z()));
         emit(graph("filtered state", filtered_state.x(), filtered_state.y(), filtered_state.z()));
+        Eigen::Isometry3d Htf = Hfw * Htw.inverse();
```
jpptm: Isnt this Hft?

jpptm:
wouldnt it be better to do this in nusight rather than emitting it to plot juggler (if nusight isnt doing this already)??

module/localisation/FieldLocalisationNLopt/src/FieldLocalisationNLopt.cpp
miles-p: Ah. Yes, it is.
It should probably plot field-to-torso instead of the other way, right?

miles-p:
> wouldnt it be better to do this in nusight rather than emitting it to plot juggler (if nusight isnt doing this already)??

good point - NUsight would probably be a good place for this. I still think PJ having it is useful especially for possible changes to localisation where we can plot two different localisation methods in one go. Thoughts?

module/localisation/FieldLocalisationNLopt/src/FieldLocalisationNLopt.cpp
JohanneMontano: Hft would describe the torso in the field space axes, which I think is a more natural way to plot it.

miles-p:
This is now done as `Hft` instead of `Htf`.

JohanneMontano (approved):

miles-p:
fiwb

PR #1819, opened 2026-07-12
@Tom0Brien: Fix own/opp goal label and don't flip on team colours

jpptm:


ysims:
This won't display nicely anymore for testing with two teams of robots, right? ie with the `multi` command.

JohanneMontano:
> This won't display nicely anymore for testing with two teams of robots, right? ie with the `multi` command.

Yea, it will be worth doing that in another pr i think, since it would be very nice to have

ysims (approved):
I'd recommend making a Jira task for the case of two teams. I'm not sure of a good solution for it. Maybe NUsight stores the team ID and flips when a robot comes in with a second ID. Or it's manual and part of the robot selector.

PR #1820, opened 2026-07-15
Add NUSim-compatible code with our own policies (merged with K1 robocup2026)

jpptm:
Add NUSim-compatible code that can run behaviour and keyboardwalk in simulation. Allows us to deploy our policies made using https://github.com/Tom0Brien/mujoco_playground.


jpptm:
Disclaimer: Current get-up policy is WIP, it doesnt work atm. Also, the robot shakes its hands vigorously when walking with low velocities which is something that will need fine tuning later on.

jpptm:
wrong repo :(

PR #1821, opened 2026-07-16
Add `retrieve` tool

miles-p:
I'm quite lazy. I also noticed at Robocup that sometimes configs from tuning were lost because there's no easy way to tell when it was taken. The retrieval process of configs and recordings is slightly painful - not only because I have the memory of a goldfish when it comes to how the `scp` command works.

This script can retrieve any specific file, but can also retrieve recordings and configs; the latter it will move into a timestamped folder - this solves my "when did we get that config off the robot" problem.

The timestamp is UTC.

NOTE: This PR adds a new folder, alike the `recordings` folder that is gitignore'd.

### DIFF
--- a/tools/retrieve.py
+++ b/tools/retrieve.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+#
+# MIT License
+#
+# Copyright (c) 2026 NUbots
+#
+# This file is part of the NUbots codebase.
+# See https://github.com/NUbots/NUbots for further info.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+
+import glob
+import os
+import subprocess
+
+from termcolor import cprint
+
+import b
+from utility.dockerise import run_on_docker
+
+
+@run_on_docker
+def register(command):
+    command.description = "Retrieve files from the target system to the local system"
+
+    command.add_argument("host", help="The host or directory to retrieve the files from")
+
+    command.add_argument("target", help="The target directory/files to get")
+
+    command.add_argument("local", help="The local directory to retrieve the files to")
+
+    command.add_argument("--user", "-u", help="The user to retrieve the files with")
+
+@run_on_docker
+def run(host, target, local, user=None, **kwargs):
+    if not os.path.exists(local):
+        os.makedirs(local)
+
+    cprint(f"Retrieving files from {host}:{target} to {local}", "green")
+
+    if user is None:
+        user = "nubots"
+
+    # Use rsync for efficient incremental transfers over SSH
+    subprocess.run([
+        "rsync",
+        "-aP", # partial progression, archive mode
+        "-e",  # specify the remote shell to use
+        "ssh",
+        f"{user}@{host}:{target}",
+        f"{local}",
+    ], check=True)
--- a/tools/retrieve.py
+++ b/tools/retrieve.py
@@ -101,7 +101,6 @@ def run(host, target, user=None, **kwargs):
             cprint(f"Files in '{TEMP_FOLDER}':", "green")
             print(files_in_temp)
 
-        all_matched = True
         for temp_file in files_in_temp:
             matches = []
             for root, _, filenames in os.walk(os.getcwd()):
@@ -137,16 +136,11 @@ def run(host, target, user=None, **kwargs):
                 # tell the user we couldn't find a match
                 if not best_match:
                     print("Unable to match.")
-                    all_matched = False
             else:
                 cprint(f"No local match found for {temp_file}", "yellow")
-                all_matched = False
 
-        # now remove all the temp files, but only if everything found a home
-        if all_matched:
-            shutil.rmtree(TEMP_FOLDER)
-        else:
-            cprint(f"Some files could not be matched — keeping '{TEMP_FOLDER}' for manual handling", "yellow")
+        # now remove all the temp files
+        shutil.rmtree(TEMP_FOLDER)
 
 
     if target == "recordings":
--- a/tools/retrieve.py
+++ b/tools/retrieve.py
@@ -64,7 +64,9 @@ def rsync_from(host, user, remote_folder, local_folder):
             "-aP",  # partial progression, archive mode
             "-e",  # specify the remote shell to use
             "ssh",
-            f"{user}@{host}:{remote_folder}",
+            # trailing slash on the source makes rsync copy the folder's contents,
+            # not the folder itself (avoids recordings/recordings/)
+            f"{user}@{host}:{remote_folder.rstrip('/')}/",
             f"{local_folder}/",
         ],
         check=True,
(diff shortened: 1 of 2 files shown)


vincenttumminello:
Would it be better to use the timestamp of when the file was last modified??

Tom0Brien:
What might motivate this even more is if you add alias's for robot numbers so you can specify robot with `n1` `n2` `n3` or `r1` `r2` for K1's, like is done in the install script https://github.com/NUbots/NUbots/blob/main/tools/install.py

miles-p:
> Would it be better to use the timestamp of when the file was last modified??

excellent point. i will try to implement this.

tools/retrieve.py
```diff
@@ -0,0 +1,108 @@
...
+
+@run_on_docker
+def run(host, target, local, user=None, **kwargs):
+    # Replace hostname with its IP address if the hostname is already known
+    num_robots = 4
+    host = {
+        "{}{}".format(prefix, num): "10.1.1.{}".format(num)
+        for num in range(1, num_robots + 1)
+        for prefix in ("nugus", "n", "i", "igus")
+    }.get(host, host)
```
JohanneMontano, line 161: Python is by reference so this can do weird things. Dont have two `hosts` variables of the same name

miles-p:
> What might motivate this even more is if you add alias's for robot numbers so you can specify robot with `n1` `n2` `n3` or `r1` `r2` for K1's, like is done in the install script https://github.com/NUbots/NUbots/blob/main/tools/install.py

```
def run(host, target, local, user=None, **kwargs):
    # Replace hostname with its IP address if the hostname is already known
    num_robots = 4
    host = {
        "{}{}".format(prefix, num): "10.1.1.{}".format(num)
        for num in range(1, num_robots + 1)
        for prefix in ("nugus", "n", "i", "igus")
    }.get(host, host)
```

If you mean the ability to run `./b retrieve n1 config`, this is already implemented
image

tools/retrieve.py
```diff
@@ -0,0 +1,108 @@
...
+    elif local is None:
+        raise SystemExit("A local directory is required when not using a preset")
+
+    if not os.path.exists(local):
+        os.makedirs(local)
+
+    cprint(f"Retrieving files from {host}:{target} to {local}", "green")
+
+    if user is None:
+        user = "nubots"
```
JohanneMontano: nubots should just be the default user value

tools/retrieve.py
```diff
@@ -0,0 +1,108 @@
...
+
+import b
+from utility.dockerise import run_on_docker
+
+# Preset names that expand to (remote path, default local directory)
+# Trailing slash on the remote path makes rsync copy the directory contents
+PRESETS = {
+    "configs": ("/home/nubots/config/", "configs", True), # append timestamp is True for configs
+    "logs": ("/home/nubots/recordings/", "recordings", False),
+}
```
JohanneMontano: Even if you extracted just the whole configs folder, theres still no easy way to diff these configs with main since the structure is different as the configs per module is combined into one configs folder in the destination. It would be more useful if this can be addressed.

Tom0Brien:
> > What might motivate this even more is if you add alias's for robot numbers so you can specify robot with `n1` `n2` `n3` or `r1` `r2` for K1's, like is done in the install script https://github.com/NUbots/NUbots/blob/main/tools/install.py
> 
> ```
> def run(host, target, local, user=None, **kwargs):
>     # Replace hostname with its IP address if the hostname is already known
>     num_robots = 4
>     host = {
>         "{}{}".format(prefix, num): "10.1.1.{}".format(num)
>         for num in range(1, num_robots + 1)
>         for prefix in ("nugus", "n", "i", "igus")
>     }.get(host, host)
> ```
> 
> If you mean the ability to run `./b retrieve n1 config`, this is already implemented image

Awesome!

tools/retrieve.py
JohanneMontano: also does this snippet really have to be this complicated?

ysims:
I remember the team talking about this problem many many years ago, although the main problem always seemed to come down to people remembering to take the files off at the end of a tuning session. 
You might be able to drop the config into the right directory for direct committing. It does become a little undefined with per-robot config though, as Johanne said.
Edit: I just realised it should be fine because the config files on the robot are in folders respective to per-robot or per-platform configs. 
 
Have you got this set up for scripts too?

JohanneMontano:
Another idea is if we implement the same thing that other teams have where they just have a big box with a switch inside and a bunch of ethernet cables, allowing for globbing options could save us a lot of time to. For me a convenient thing to do would be for the script to get the configs, then diff them with the one you have locally (you can determine what to look for since the configs are named after the modules). If you can apply those changes automatically as well then all wed have to do is:

Connect robots to switch
run this script which should involve:
    get folder with name of robots and their respective configs inside
    diff which configs are different locally
    apply changes locally

then all Id have to do is git add . && git commit -m "config updates for all robots"

tools/retrieve.py
miles-p: This is a direct copy from `install.py`. I understand where you're coming from however.

tools/retrieve.py
miles-p: fixed :)

miles-p:
> I remember the team talking about this problem many many years ago, although the main problem always seemed to come down to people remembering to take the files off at the end of a tuning session. You might be able to drop the config into the right directory for direct committing. It does become a little undefined with per-robot config though, as Johanne said. Edit: I just realised it should be fine because the config files on the robot are in folders respective to per-robot or per-platform configs.
> 
> Have you got this set up for scripts too?

i did test this with robot specific configs because i recalled this was probably going to be relevant. because it copies the folder recursively, it will keep the subfolders intact. i suppose it's not really to do a `git diff` with, it's there to copy across whatever you need as you tune. i dunno :shrug: :

miles-p:
> I remember the team talking about this problem many many years ago, although the main problem always seemed to come down to people remembering to take the files off at the end of a tuning session. You might be able to drop the config into the right directory for direct committing. It does become a little undefined with per-robot config though, as Johanne said. Edit: I just realised it should be fine because the config files on the robot are in folders respective to per-robot or per-platform configs.
> 
> Have you got this set up for scripts too?

we now have scripts as a preset

miles-p:
> Another idea is if we implement the same thing that other teams have where they just have a big box with a switch inside and a bunch of ethernet cables, allowing for globbing options could save us a lot of time to. For me a convenient thing to do would be for the script to get the configs, then diff them with the one you have locally (you can determine what to look for since the configs are named after the modules). If you can apply those changes automatically as well then all wed have to do is:
> 
> Connect robots to switch run this script which should involve: get folder with name of robots and their respective configs inside diff which configs are different locally apply changes locally
> 
> then all Id have to do is git add . && git commit -m "config updates for all robots"

setting up a hardware backed network for robocup 27 is on my list of things to think about (among other things i want to change about our networking)

I will very happily make a loom much alike other teams we saw this year

JohanneMontano:
> > Another idea is if we implement the same thing that other teams have where they just have a big box with a switch inside and a bunch of ethernet cables, allowing for globbing options could save us a lot of time to. For me a convenient thing to do would be for the script to get the configs, then diff them with the one you have locally (you can determine what to look for since the configs are named after the modules). If you can apply those changes automatically as well then all wed have to do is:
> > Connect robots to switch run this script which should involve: get folder with name of robots and their respective configs inside diff which configs are different locally apply changes locally
> > then all Id have to do is git add . && git commit -m "config updates for all robots"
> 
> setting up a hardware backed network for robocup 27 is on my list of things to think about (among other things i want to change about our networking)
> 
> I will very happily make a loom much alike other teams we saw this year

It doesnt look like the code updates the config files locally. We should be doing this, otherwise there is still room for error for people trying to grab configs. Also looking through copied files individually will take a lot of time which no one would ideally want to spend

miles-p:
> > > Another idea is if we implement the same thing that other teams have where they just have a big box with a switch inside and a bunch of ethernet cables, allowing for globbing options could save us a lot of time to. For me a convenient thing to do would be for the script to get the configs, then diff them with the one you have locally (you can determine what to look for since the configs are named after the modules). If you can apply those changes automatically as well then all wed have to do is:
> > > Connect robots to switch run this script which should involve: get folder with name of robots and their respective configs inside diff which configs are different locally apply changes locally
> > > then all Id have to do is git add . && git commit -m "config updates for all robots"
> > 
> > 
> > setting up a hardware backed network for robocup 27 is on my list of things to think about (among other things i want to change about our networking)
> > I will very happily make a loom much alike other teams we saw this year
> 
> It doesnt look like the code updates the config files locally. We should be doing this, otherwise there is still room for error for people trying to grab configs. Also looking through copied files individually will take a lot of time which no one would ideally want to spend

I'll take a look at this today.

JohanneMontano:
A simplification to the git diff thing is to just check whether the two files are byte-per-byte equivalent (I think we already have a utility that does this somewhere, maybe in systemconfiguration?). If not then the robot files have to replace those. Replacing files using ./b (in the container) should carry over edits to local from my understanding.

miles-p:
@JohanneMontano This now works for scripts, configs, and recordings, or the rare 'all' option for all three.
Scripts and configs are pulled into their respective file in the codebase, and recordings go in their normal spots

ysims (approved):
Approving b4 someone asks you to add another feature 😆 
Documentation would fit in here: https://nubook.nubots.net/system/foundations/build-system#utilities
Might be useful to also mention in these pages https://nubook.nubots.net/guides/main/tuning-and-running-scripts
https://nubook.nubots.net/guides/main/maintaining-subsystems

miles-p:
fiwb

PR #1823, opened 2026-07-19
Add SRIF localisation

vincenttumminello:
This PR adds a new localisation estimator: a Bayesian filter carrying the belief as a Gaussian in square-root information form. Prediction is driven by a body twist finite-differenced from odometry; each vision frame applies a MAP (trust-region Newton) update over a robust landmark likelihood, and the Laplace approximation at the optimum gives the posterior square-root information. So it outputs both the most likely pose and a covariance with each measurement. 

### The estimators state (nx=18):
- x, y, z position of robot torso in field space
- Torso attitude as a unit quaternion
- x, y, z velocity of robot (body fixed) 
- Robot body fixed roll, pitch and yaw rate
- Camera bias roll and pitch offsets
- Gyro bias estimate (roll, pitch and yaw) 
Note: Unit quaternion is helpful here to resolve singularity issues with Euler angles when the robot has fallen over and pitch = +-90 degs. 

### Initialisation
Initialisation is handled by a coarse grid-search over the whole field, scored by the same landmark likelihood used for updates. To solve the field symmetry problem we assume that we start on our own half of the field. In any very very rare case where this is not true we should not initialise/re-initialise the estimator. 

### YOLO detections
If the YOLO detection has a confidence value below 50% they are not used, above this they receive scaled weighting based on their confidence. Detections are associated to a known field map with greedy SNN inside a geometric pre-gate that widens the filter's yaw uncertainty.

### Inverse Kinematics
Inverse kinematics are only used for determining the z-height of the robot so that can get scale information from images. In the event of a fall (detected using the reported stability state of the robot) kinematic height is ignored for updates until it is standing again. 

### IMU 
The gyroscope and accelerometer both provide measurement updates. 

### Out of field Landmarks (not wired in yet)
On every frame, until the number of landmarks hits the maximum, the FAST feature detector finds potential candidates for out of field landmarks. First all points not outside the field are rejected using the green-field horizon. Once candidates are added to the map surprisal nearest neighbours (SNN) is used to associate landmarks on subsequent frames and outlier rejection is done to ensure landmarks are not chosen on dynamic bodies (e.g., people in the crowd). The purpose of these out-of-field landmarks is to help the estimator figure out which side of the field it is on post-initialisation. Most of these landmarks will be bearing-only landmarks since they are far away and we won't have much parallax information to determine depth. 

### Performance
The initial grid-search takes the most time at approx ~30-90 ms (on my desktop i5) and every subsequent update takes max ~1ms. These are very cheap updates. 
And it does better than the NLopt module. I would like to do some real-time testing though, currently I have only tested with nbs recordings or in webots. 

## TODO
- Remove some of claude's fancy comments
- Enable it in all/most/some of the role files\
- Actually enable the side disambiguator (but could be in a later PR)
- Fix doxygen math documentation rendering (Pose.hpp)


module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+# Sign of field-x for the robot's starting half (from game context). Breaks the field's 180 deg
+# symmetry that on-field landmarks cannot: +1 => starts in the +x half, -1 => the -x half, 0 => no prior.
+own_half_x_sign: 1.0
+
+# Multi-hypothesis (field-symmetry) Gaussian mixture bank. Off keeps a single Gaussian.
+# Note: This cannot resolve the symmetry problem. On-field landmarks fit a pose and its 180 deg
+# mirror equally well (each component is now scored on its OWN association), so the pair sits at a
+# genuine 50/50 and the reported pose stays with the bootstrapped hypothesis. Breaking the tie needs
+# asymmetric off-field evidence fed in via SystemLocalisation::addSideLogEvidence, which this module
```
vincenttumminello: This is actually not true, off field evidence cannot initially break the symmetry (we need the initialisation prior of "we start on our own side of the field" to break this), but it can help keep it broken.

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,108 @@
...
+# does not yet source (see utility::slam::SideDisambiguator).
+use_hypothesis_bank: false
+
+# Feed Sensors.gyroscope into the body-twist angular velocity. Sensors.gyroscope is a calibrated rad/s
+# torso-frame rate on both hardware and webots. When off, the angular velocity is finite-differenced
+# from the odometry attitude in Htw instead (useful if a platform lacks a trusted gyro).
```
vincenttumminello: Check this: Pretty sure sensors.cpp requires a gyro to fill in Htw information, which makes this unhelpful. In that case just enforce use_gyro: true and remove it from the config

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,108 @@
...
+use_gyroscope: true
+
+# Low-rate corrections
+# Accelerometer gravity-direction measurement of torso roll/pitch, from the calibrated m/s^2
+# Sensors.accelerometer. The gravity-aligned attitude is also estimated upstream by the Mahony filter
+# and delivered via Htw, so this is a secondary anchor and may be turned off if Htw is trusted alone.
+use_gravity: true
```
vincenttumminello, line 19: TODO: experiment if this is at all needed. We trust the Mahony filter...right?

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,108 @@
...
+gravity_sigma: 1.0 # accelerometer gravity-direction noise std dev [m/s^2]
+# The accelerometer reads gravity whenever the torso is not being accelerated - true of a robot lying
+# still on the carpet, false of one in free fall or hitting the ground. Gating on how far the specific
+# force sits from standard gravity tests that directly, which is the right condition during a fall and a
+# better one than "is the robot upright" while walking. Loose on purpose: gait swings the magnitude by a
+# couple of m/s^2 and the model already carries gravity_sigma of noise.
+gravity_quasi_static_tolerance: 3.0 # [m/s^2] from standard gravity
```
vincenttumminello: Is this even needed? Pretty sure fall detection already uses this logic, don't need to do it twice

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+# Fall handling. Measurement updates are NOT suppressed wholesale while the robot is down: the landmark
+# model is plain geometry and stays valid face-down, and a robot that spins while toppling or getting up
+# changes its heading, which only measurements taken during the event can catch.
+fall:
+  # Confidence handed back on standing up again. The mean is kept - a fall and getup move the torso well
+  # under a metre, so the pre-fall position is still the best estimate available - but the confidence
+  # attached to it is not survivable, above all in yaw.
+  recovery_pos_std: 0.5 # horizontal position std restored on recovery [m]
+  recovery_yaw_std: 0.6 # yaw std restored on recovery [rad]
```
vincenttumminello: Thought I made this redundant?

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+# Odometry twist buffer
+max_odometry_gap: 0.1 # max sample spacing to finite-difference across [s]
```
vincenttumminello: This seems too large and also need to decide if we are actually finite-differencing (see comment about use_gyro), if not them remove all this junk

.vscode/settings.json
```diff
@@ -18,7 +18,10 @@
   "files.associations": {
     "*.role": "cmake",
     ".clang-tidy": "yaml",
-    ".Doxyfile": "properties"
+    ".Doxyfile": "properties",
+    "iterator": "cpp",
+    "ios": "cpp",
+    "*.inc": "cpp"
   },
   "python.analysis.extraPaths": ["${workspaceFolder}/tools", "${workspaceFolder}/nuclear"],
```
Tom0Brien, line 23: Remove changes to this file

docker/Dockerfile
```diff
@@ -391,6 +391,16 @@ RUN install-from-source https://downloads.sourceforge.net/lame/lame-3.100.tar.gz
...
 
+# autodiff
+RUN install-from-source https://github.com/autodiff/autodiff/archive/refs/tags/v1.1.2.tar.gz \
+    -DAUTODIFF_BUILD_PYTHON=OFF \
+    -DAUTODIFF_BUILD_TESTS=OFF \
+    -DAUTODIFF_BUILD_EXAMPLES=OFF \
+    -DAUTODIFF_BUILD_DOCS=OFF
+
+# Boost, for the incomplete gamma function and reproducible random number generation in utility::slam::gaussian
+RUN install-package boost boost-libs
```
Tom0Brien: Does this work on the robot? Libs installed with install-package will not be included in libs when doing ./b install -t

module/localisation/FieldLocalisationSRIF/README.md
```diff
@@ -0,0 +1,202 @@
...
+trust it, rather than a bare pose.
+
+### State
+
+The state is nine elements: the 6-DOF torso pose in the field frame `{f}` plus a 2-DOF camera-mount
+attitude bias.
+
+| Index | Symbol   | Meaning                                                       |
+| ----: | -------- | ------------------------------------------------------------- |
+|  0..2 | `rBFf`   | Torso position in the field frame [m]                         |
```
Tom0Brien: Rename B to T

shared/message/input/Sensors.proto
```diff
@@ -157,8 +157,13 @@ message Sensors {
     /// Homogeneous transform: planted foot {p} to world {w}
     iso3 Hwp = 15;
 
+    /// Homogeneous transform: NUbots world {w} to VSLAM world {n}
+    /// To transform a point from VSLAM world to NUbots world: p_w = Hwn.inverse() * p_n
+    /// To transform a point from NUbots world to VSLAM world: p_n = Hwn * p_w
+    iso3 Hwn = 16;
```
Tom0Brien: Probably not needed anymore

module/localisation/FieldLocalisationSRIF/src/srif/FieldMap.cpp
Tom0Brien, line 1: I'd rename the slam directory to something else

shared/utility/slam/FieldMap.hpp
```diff
@@ -0,0 +1,200 @@
+/**
+ * @file FieldMap.hpp
+ * @brief Defines a static map of known landmark positions on a RoboCup humanoid soccer field
+ *
+ * This file provides #FieldDimensions (the physical measurements of the field, matching the
```
Tom0Brien: Clean up this essay 😆

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,132 @@
...
```
Tom0Brien: I don't feel like this odometery twist stuff will be needed when porting to K1, so worth keeping in mind to simplify later

Tom0Brien (changes requested):
Quick review

module/localisation/FieldLocalisationSRIF/data/config/webots/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,7 @@
+# Webots overrides (merged over data/config/FieldLocalisationSRIF.yaml)
+#
+# Webots mirrors the hardware sensor pipeline: the accelerometer and gyroscope are decoded to physical
+# units (m/s^2, rad/s) via the same NUgus.proto lookup table (see Webots.cpp), and SensorFilter runs the
+# real Mahony/anchor odometry (use_ground_truth: false). So there is no sensor difference to override
+# here - the base config applies unchanged.
```
jpptm:
```suggestion
```
Thats allat of words

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,132 @@
+log_level: INFO
+
+# Sign of field-x for the robot's starting half (from game context). Breaks the field's 180 deg
+# symmetry that on-field landmarks cannot: +1 => starts in the +x half, -1 => the -x half, 0 => no prior.
+own_half_x_sign: 1.0
```
jpptm: Cant this be just by convention? Iirc -x is our side by default.

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,132 @@
...
+# asymmetric off-field evidence fed in via SystemLocalisation::addSideLogEvidence, which this module
+# does not yet source (see utility::slam::SideDisambiguator).
+use_hypothesis_bank: false
+
+# Body-rate measurements. The gyroscope and the walk-engine odometry are no longer known inputs to the
+# process model - they are measurements of the body angular and linear velocity states. That is what
+# lets their noise be modelled honestly and, for the gyroscope, what makes its bias an estimated state
+# rather than a heuristic subtracted up front. Sensors.gyroscope is raw (SensorFilter passes the
+# hardware value through and Mahony keeps its bias internal), so there is no double-correction.
+use_gyroscope: true
```
jpptm: The comment before the config field should describe what the config is for and what it does. Should we be using gyro by default, i.e., not make it configurable?

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,132 @@
...
+gyroscope_sigma: 0.02 # gyroscope noise std dev per axis [rad/s]
```
jpptm: Need to be tuned probs? can this be inferred during runtime?

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,132 @@
...
+use_odometry_velocity: true
```
jpptm: Ditto

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,132 @@
...
+# Deliberately loose, comparable to the walk speed itself: the odometry slips on foot contact and
+# reports the gait the engine believes it is executing. As a measurement that is evidence the filter can
+# weigh, never truth it has to accept.
+odometry_velocity_sigma: 0.15 # [m/s]
```
jpptm: If this can be inferred during run time, that would be ideal. Otherwise put a note or a todo. We should aim to not rely too much on configs. Perhaps this could be determined during the ready state, same with the gyro as well?

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,132 @@
...
+# Zero-velocity updates while the robot is down. A robot on the carpet is not travelling anywhere, and
+# saying so is the only thing stopping the pre-fall gait from integrating across the whole fall.
+# Leaving it unmeasured is not the neutral choice it looks like - it asserts the robot may still be
+# moving at whatever it was doing when it fell.
+zupt_sigma: 0.02 # [m/s] FALLEN, lying still
+zupt_dynamic_sigma: 0.30 # [m/s] FALLING or getting up - the torso moves, just not anywhere
```
jpptm: Ditto. These comments are not very direct at describing what the config does.

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,132 @@
...
+
+# Low-rate corrections
+# Accelerometer gravity-direction measurement of torso roll/pitch, from the calibrated m/s^2
+# Sensors.accelerometer. The gravity-aligned attitude is also estimated upstream by the Mahony filter
+# and delivered via Htw, so this is a secondary anchor and may be turned off if Htw is trusted alone.
+use_gravity: true
```
jpptm: Again, should this be a config? Gyro accel and gravity seem like theyre such integral bits of info that they should always be included. If not, we can calculate sigma on the fly and weight their contributions using their uncertainty.

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,132 @@
...
+# Initial-pose grid search
+grid_step_xy: 0.35 # position step [m]
+grid_step_yaw: 18.0 # heading step [deg]
+min_init_associations: 4 # landmark associations needed to trust a solve
+# Initial sqrt-covariance diagonal for the 18-dim state.
+# Attitude is a quaternion, not roll/pitch/yaw: a tangent-space std of s becomes a component std of s/2,
+# so the 0.5 rad of initial yaw doubt the old layout carried is 0.25 across the four components.
+# [x, y, z | qw, qx, qy, qz | vx, vy, vz | wx, wy, wz | bgx, bgy, bgz | cam_roll, cam_pitch]
+initial_sqrt_covariance:
+  [1.0, 1.0, 0.05, 0.25, 0.25, 0.25, 0.25, 0.30, 0.30, 0.10, 0.50, 0.50, 0.50, 0.05, 0.05, 0.05, 0.02, 0.02]
```
jpptm: Im assuming these are hopefully refined by the model later?

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,132 @@
...
+  confidence_reference: 0.7 # confidence that maps to inlier_probability
+  max_inlier_probability: 0.95
+
+# Hypothesis-bank spawning/pruning/merging (only used when use_hypothesis_bank is true)
+hypothesis:
+  min_weight: 0.02 # prune components below this normalised weight
+  max_components: 4 # cap on live components
+  merge_position: 0.30 # merge gate on position separation [m]
+  merge_yaw: 0.20 # merge gate on yaw separation [rad]
+  respawn_pos_std: 0.60 # respawn a mirror when the lone component's horizontal std exceeds this [m]
```
jpptm: Gotta clean these comments up. Can some of these configs go? If there is some calibration experiment that needs to be done to determine these values then they should be separated. Must remove guess work as much as possible

module/localisation/FieldLocalisationSRIF/README.md
```diff
@@ -0,0 +1,202 @@
...
+|  3..6 | `q`      | Torso attitude quaternion `(w, x, y, z)`, `Rfb = quat2rot(q)` |
+|  7..8 | `deltaC` | Camera-mount attitude bias (roll, pitch) [rad]                |
```
jpptm: why no yaw? del yaw should most likely be included in delta C?

module/vision/FeatureDetector/src/FeatureDetector.cpp
```diff
@@ -0,0 +1,150 @@
...
+                // Label the feature with its index number
+                std::string label = std::to_string(i + 1);
+                cv::putText(imgout,
+                            label,
+                            kp.pt + cv::Point2f(5, -5),  // offset text position
+                            cv::FONT_HERSHEY_SIMPLEX,
+                            0.9,                    // font scale
+                            cv::Scalar(255, 0, 0),  // green text
+                            2);                     // thickness
+            }
```
jpptm: Seems unnecessary for the intents and purposes of this pr?

module/vision/FeatureDetector/src/FeatureDetector.cpp
```diff
@@ -0,0 +1,150 @@
+#include "FeatureDetector.hpp"
```
jpptm: Does this have to be a module? Can it be just an inline utility?

module/vision/FeatureDetector/README.md
```diff
@@ -0,0 +1,11 @@
+# FeatureDetector
```
jpptm: Less files and less words if this is a utility, just needs the doxygen compliant comments

shared/utility/slam/camera/Camera.cpp
```diff
@@ -0,0 +1,583 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2025 NUbots
```
jpptm: This is just a camera calibration script for a pinhole model. What if we wanted to use this on the nugus for some reason? Or, what if we modded the k1s so that they had fisheye cameras? I don't see why our camera models should not suffice here. First, if this is valid then it should be a whole pr on its own. Second, are we absolutely sure certain that our camera models cant be used for slam for some reason? I highly doubt this.

jpptm: see `projection.hpp`

shared/utility/slam/camera/Camera.cpp
```diff
@@ -0,0 +1,583 @@
...
+        // Clamp to avoid numerical errors slightly exceeding [-1, 1]
+        cos_angle = std::clamp(cos_angle, -1.0, 1.0);
+
+        double angle = std::acos(cos_angle);  // radians
+
+        // Use the diagonal FOV for a symmetric cone check
+        return angle <= dFOV / 2.0;
+    }
+
+    bool Camera::isWorldWithinFOV(const cv::Vec3d& rPNn, const Pose<double>& Tnb) const {
```
jpptm: camel case?

shared/utility/slam/camera/Camera.hpp
```diff
@@ -0,0 +1,196 @@
+/*
```
jpptm: Our camera model is able to handle pinhole as well. This is most likely not necessary at all.

shared/utility/slam/camera/CameraLens.hpp
```diff
@@ -0,0 +1,211 @@
...
+#include <cmath>
+
+#include "message/input/Image.hpp"
+
+#include "utility/vision/projection.hpp"
+
+namespace utility::slam::camera {
+
+    /**
+     * @brief Camera lens model with radial distortion (NUbots-compatible).
```
jpptm: Is slam really that coupled to the pinhole model? I am 99% sure that there is some hella smart dude out there that has a method that is agnostic of projection model. This sort of progression is very prevalent in the literature.

shared/utility/slam/camera/CameraLens.hpp
```diff
@@ -0,0 +1,211 @@
...
+                    constexpr double max_theta = 89.9 * M_PI / 180.0;
+                    return utility::vision::rectilinear::r<double>(std::clamp(theta, 0.0, max_theta), focalLength);
+                }
+                case Projection::EQUISOLID: return utility::vision::equisolid::r<double>(theta, focalLength);
+                default: return utility::vision::equidistant::r<double>(theta, focalLength);
+            }
+        }
+
+        /// @brief Incidence angle for an undistorted radius, per the projection model.
+        double angleForRadius(double r) const {
```
jpptm: This whole file is literally just copying projection.hpp. This should not be the case

shared/utility/gaussian_filtering/Pose.hpp
```diff
@@ -0,0 +1,214 @@
...
+        /**
+         * @brief Group operation of \f$\mathsf{SE}(3)\f$
+         *
+         * Computes the composition of two poses: \f$\mathbf{T}^a_c = \mathbf{T}^a_b \mathbf{T}^b_c\f$
+         * @f[
+         * \begin{bmatrix} \mathbf{R}^a_c & \mathbf{r}_{C/A}^a \\ \mathbf{0}^\mathsf{T} & 1 \end{bmatrix} =
+         * \begin{bmatrix} \mathbf{R}^a_b & \mathbf{r}_{B/A}^a \\ \mathbf{0}^\mathsf{T} & 1 \end{bmatrix}
+         * \begin{bmatrix} \mathbf{R}^b_c & \mathbf{r}_{C/B}^b \\ \mathbf{0}^\mathsf{T} & 1 \end{bmatrix} =
+         * \begin{bmatrix} \mathbf{R}^a_b \mathbf{R}^b_c & \mathbf{R}^a_b \mathbf{r}_{C/B}^b + \mathbf{r}_{B/A}^a \\
+         * \mathbf{0}^\mathsf{T} & 1 \end{bmatrix}
```
jpptm: Unrendered latex is hard to read for humans

jpptm: Again, I doubt we need a whole header file for this?

shared/utility/gaussian_filtering/gaussian/GaussianBase.hpp
```diff
@@ -0,0 +1,372 @@
...
+            p = std::fma(p, q, 0x1.49c673066c831p-8);    //  5.0319701025945277e-03
+            p = std::fma(p, q, -0x1.0962386ea02b7p-6);   // -1.6197733983519948e-02
+            p = std::fma(p, q, 0x1.3079edf465cc3p-5);    //  3.7167515521269866e-02
+            p = std::fma(p, q, -0x1.0fb06dfedc4ccp-4);   // -6.6330365820039094e-02
+            p = std::fma(p, q, 0x1.7fee004e266dfp-4);    //  9.3732834999538536e-02
+            p = std::fma(p, q, -0x1.9ddb23c3e14d2p-4);   // -1.0103906603588378e-01
+            p = std::fma(p, q, 0x1.16ecefcfa4865p-4);    //  6.8097054254651804e-02
+            p = std::fma(p, q, 0x1.f7f5df66fc349p-7);    //  1.5379652102610957e-02
+            p = std::fma(p, q, -0x1.1df1ad154a27fp-3);   // -1.3962111684056208e-01
+            p = std::fma(p, q, 0x1.dd2c8b74febf6p-3);    //  2.3299511862555250e-01
```
jpptm: Jesus Christ almighty what in tarnation is this abomination

jpptm (changes requested):
There is a lot of work to be done, particularly in decreasing the duplication of code in the camera models. There is almost certainly a SLAM model out there that is already agnostic of the camera model ie., decoupled from pinhole. NUbots already has an almost-generic camera model with radial distortions accounted for using an analytically exact inverse. I think we don't need a whole new pipeline for just pinhole, when we already account for more than just pinhole.

Some of the code is inconsistent as well, i.e., latex in comments and camel case. We should map the whole algorithm and decide which bits can and have to go, and which bits must stay. Must work to decrease code volume and configurable knobs in the code as much as possible.

If the process and measurement stuff need their own calibration pipeline then do so in another pr maybe? Otherwise maybe some method to estimate and refine them during run time would be great. It will not be fun if all the people that knew how to play with this module left then future members in future robocups struggle with too many knobs to play around with.

I am about a third of the way to finishing this initial review. 22/68 files. Parts 2 and 3 will come later.

vincenttumminello:
Thanks for the initial reviews. When I opened this PR I knew there would be lots of things needing changes. These reviews are helpful because they give me specific things to look at which I almost certainly would have overlooked. I'll work on it when I can

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
vincenttumminello: Yeah these are just the initial values (post grid search). They basically say "we are unsure what our starting x, y, position is but we're pretty confidence that we're standing upright and stationary.". They are also standard deviations so we can apply units to them. i.e., the x, and y position estimates have a 1m standard dev. 
Currently its also assigning high certainty to the camera extrinsics (0.02) which I think will need to be loosened in the future. I'll have to put it on a real robot and tune them.

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
vincenttumminello: A lot of these configs need to go and the comments are horrendous...

module/localisation/FieldLocalisationSRIF/README.md
vincenttumminello: Because its only partially observable and not entirely necessary. We can separate the robots heading error with the camera yaw offset only by exciting both head yaw and tilt. It's untested but I don't think there will be enough head movement during a game to stop a camera yaw bias state from wandering.

module/localisation/FieldLocalisationSRIF/data/config/FieldLocalisationSRIF.yaml
```diff
@@ -0,0 +1,87 @@
+log_level: INFO
+
+# The intent for what belongs here: if you would change it because something about the robot, the
```
Tom0Brien: Make all these comments more concise

docker/Dockerfile
```diff
@@ -391,6 +391,17 @@ RUN install-from-source https://downloads.sourceforge.net/lame/lame-3.100.tar.gz
...
 RUN install-from-source https://github.com/nanopb/nanopb/archive/refs/tags/0.4.9.1.tar.gz
 
+# autodiff
+RUN install-from-source https://github.com/autodiff/autodiff/archive/refs/tags/v1.1.2.tar.gz \
+    -DAUTODIFF_BUILD_PYTHON=OFF \
+    -DAUTODIFF_BUILD_TESTS=OFF \
+    -DAUTODIFF_BUILD_EXAMPLES=OFF \
+    -DAUTODIFF_BUILD_DOCS=OFF
+
+# Boost
```
Tom0Brien: Don't think this is actually used anywhere now

shared/utility/gaussian_filtering/Pose.hpp
```diff
@@ -0,0 +1,214 @@
+/*
```
Tom0Brien, line 1: Is this Pose helper actually needed? Eigen Isometry may be all we need

shared/utility/libraries.cmake
```diff
@@ -1,6 +1,9 @@
 find_package(Eigen3 REQUIRED)
 target_link_libraries(nuclear_utility PUBLIC Eigen3::Eigen)
 
+find_package(OpenCV REQUIRED)
```
Tom0Brien: Is opencv needed? I don't see it in dockerfile

Tom0Brien (changes requested):

shared/utility/libraries.cmake
vincenttumminello: Opencv is in the dockerfile, YoloCoco.cpp uses it and I use it here for the out of field landmarks.

shared/utility/gaussian_filtering/Pose.hpp
vincenttumminello: Yeah the main reason to keep it is for autodiff. `MeasurementFieldLandmarks.hpp` and `MeasurementFieldLines.hpp` build `Pose<Scalar>` inside `logLikelihoodImpl<Scalar>` which is called with dual types to get the gradient and the Hessian.

shared/utility/libraries.cmake
vincenttumminello: Turns out opencv was only needed in Pose.hpp for some unused functions that I ported over, I've removed them and removed opencv from shared/utility so now it's only in the cmakelists for the SRIF module.

docker/Dockerfile
vincenttumminello: Boost is only used in `GaussianBase.hpp` in `simulate()` and `chi2inv` but neither of those actually ever get called. It was part of the code that got ported over from the mcha toolchain and I left it because it might be needed in the future. But I probably should just remove anything that doesn't currently get used??

shared/utility/vision/projection.hpp
```diff
@@ -57,7 +58,10 @@ namespace utility::vision {
 
         template <typename T>
         inline T theta(const T& r, const T& f) {
-            return T(2.0) * std::asin(r / (T(2.0) * f));
+            // Past r = 2f this model has no solution (it is the horizon of an equisolid lens), and asin
+            // would be nan. Saturating there keeps a pixel outside the lens circle mapping to a ray that
+            // a bounds check can reject, rather than poisoning everything downstream with nan.
+            return T(2.0) * std::asin(std::clamp(r / (T(2.0) * f), T(-1.0), T(1.0)));
```
jpptm, line 64: Unnecessary comment

shared/utility/vision/projection.hpp
```diff
@@ -178,9 +182,12 @@ namespace utility::vision {
         const Lens& lens,
         const Eigen::Matrix<T, 2, 1>& dimensions) {
 
-        const T f          = lens.focal_length;
-        const T theta      = std::acos(ray.x());
-        const T rsin_theta = T(1) / std::sqrt(T(1) - ray.x() * ray.x());
+        const T f = lens.focal_length;
+        // A ray that has been rotated into the camera frame is only unit length to within rounding, so
+        // clamp before acos rather than let a component a few ulp outside [-1, 1] produce nan
```
jpptm, line 187: Remove

shared/utility/gaussian_filtering/funcmin.hpp
```diff
@@ -0,0 +1,1271 @@
...
+#include <fmt/format.h>
+#include <limits>
+#include <nuclear>
+
+namespace utility::gaussian_filtering::funcmin {
+
+    /**
+     * @brief Solve trust-region subproblem
+     *
+     * This function solves the following trust-region subproblem:
```
jpptm: Remove the latex here. Also NLopt can already solve trust region problems. The usual combo is to use nlopt with autodiff. I doubt you need this whole file

shared/utility/gaussian_filtering/funcmin.hpp
```diff
@@ -0,0 +1,1271 @@
...
+        const int maxIterations = 5000;
+        for (int i = 0; i < maxIterations; ++i) {
+            // Solve trust-region subproblem
+            Vector p;
+            trsEig(Q, v, g, Delta, p);  // minimise 0.5*p.'*H*p + g.'*p subject to ||p|| <= Delta
+
+            Scalar pg       = p.dot(g);
+            Scalar LambdaSq = -pg;  // The Newton decrement squared is g.'*inv(H)*g = p.'*H*p
+
+            // const Scalar LambdaSqThreshold = std::sqrt(std::numeric_limits<Scalar>::epsilon());     // Loose
```
jpptm: Also a bunch of commented code in there

module/localisation/FieldLocalisationSRIF/src/measurement/MeasurementBodyRates.cpp
```diff
@@ -0,0 +1,127 @@
+#include "MeasurementBodyRates.hpp"
+
+#include <Eigen/Core>
+#include <autodiff/forward/dual.hpp>
+#include <autodiff/forward/dual/eigen.hpp>
+
+#include "utility/gaussian_filtering/measurement/Measurement.hpp"
+
+// Both models are linear in the state, so the exact Hessian is constant and the
+// trust-region Newton update converges in one step.
```
jpptm, line 10:
```suggestion
```

Remove

module/localisation/FieldLocalisationSRIF/src/measurement/MeasurementBodyRates.cpp
```diff
@@ -0,0 +1,127 @@
...
+    // =====================================================================
+    // MeasurementGyroscope
+    // =====================================================================
+
+    MeasurementGyroscope::MeasurementGyroscope(double time, const Eigen::Vector3d& gyroscope, double sigma)
+        : Measurement(time), y_(gyroscope), sigma_(sigma) {
+        updateMethod_ = UpdateMethod::NEWTONTRUSTEIG;
+    }
+
+    Eigen::VectorXd MeasurementGyroscope::simulate(const Eigen::VectorXd& x, const SystemEstimator& /*system*/) const {
```
jpptm, line 25: Whats with system being unused? Just remove it in the declaration and here if you don't need it

module/localisation/FieldLocalisationSRIF/src/measurement/MeasurementFieldLandmarks.cpp
```diff
@@ -0,0 +1,281 @@
...
+        // nominal gate.
+        const double gateAngle =
+            std::min(std::max(options_.gateAngle, options_.gateYawScale * std::sqrt(std::max(yawVar, 0.0))),
+                     options_.gateAngleMax);
+        const double cosGate = std::cos(gateAngle);
+
+        // SNN: enumerate all (detection, landmark) pairs of matching type inside the
+        // geometric pre-gate, score each by its surprisal relative to the clutter
+        // crossover, then assign best-scoring pairs first, each detection/landmark at
+        // most once. A negative score means the inlier component explains the pair
```
jpptm, line 122: Comments shouldnt be a record of anyone's thoughts, it should be describing what the code is doing

module/localisation/FieldLocalisationSRIF/src/measurement/MeasurementFieldLandmarks.cpp
```diff
@@ -0,0 +1,281 @@
...
+            const std::vector<Eigen::Vector3d>& lms = map_.landmarks(candidates_[i].type);
+            for (std::size_t j = 0; j < lms.size(); ++j) {
+                const Eigen::Vector3d rel = lms[j] - rCFf;
+                const double range        = rel.norm();
+                if (range < 1e-9) {
+                    continue;
+                }
+                const Eigen::Vector3d uPredF = rel / range;
+                if (uMeasF.dot(uPredF) < cosGate) {
+                    continue;  // Cheap geometric pre-gate: caps how far an association can reach
```
jpptm, line 148:
```suggestion
                    continue;
```

module/localisation/FieldLocalisationSRIF/src/measurement/MeasurementFieldLandmarks.cpp
```diff
@@ -0,0 +1,281 @@
...
+                }
+
+                // Innovation covariance in the tangent plane: bearing noise + camera
+                // position uncertainty across the range + yaw uncertainty.
+                const Eigen::Matrix<double, 3, 2> T = tangentBasis(uPredF);
```
jpptm, line 153: Camel case

module/localisation/FieldLocalisationSRIF/src/measurement/MeasurementFieldLandmarks.cpp
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    Eigen::VectorXd MeasurementFieldLandmarks::simulate(const Eigen::VectorXd& x,
+                                                        const SystemEstimator& /*system*/) const {
+        // Stack predicted rays (noise-free measurement model)
+        Eigen::Matrix<double, 3, Eigen::Dynamic> uPred = predictRays<double>(x);
+        return Eigen::Map<Eigen::VectorXd>(uPred.data(), uPred.size());
+    }
+
+    double MeasurementFieldLandmarks::logLikelihood(const Eigen::VectorXd& x, const SystemEstimator& /*system*/) const {
```
jpptm, line 209: These measurement files seem to have common functions that can be abstracted further and significantly reduce code volume. For example, writing a generalised logLikelihood implementation can be used by everything else

module/localisation/FieldLocalisationSRIF/src/measurement/MeasurementFieldLandmarks.hpp
```diff
@@ -0,0 +1,285 @@
...
+            // The pre-gate above is a hard geometric cap, so on its own it also caps
+            // what the filter can ever recover from: a getup that leaves the robot
+            // yawed by more than gateAngle puts every predicted bearing outside it,
+            // and no amount of covariance inflation helps, because the pre-gate does
+            // not look at the covariance. Widening it with the yaw uncertainty is what
+            // makes an inflated belief actually able to re-associate. The surprisal
+            // score still decides what associates -- this only stops the cheap
+            // geometric filter from throwing the candidates away first. Normal
+            // operation is unaffected: it takes sigma_yaw > gateAngle/gateYawScale
+            // (10 deg at these defaults) before the widened gate exceeds gateAngle.
```
jpptm: So many comments that overexplain, it should just simply describe the code

jpptm: also make it snake case

module/localisation/FieldLocalisationSRIF/src/measurement/MeasurementFieldLandmarks.hpp
```diff
@@ -0,0 +1,285 @@
...
+
+        virtual Eigen::VectorXd simulate(const Eigen::VectorXd& x, const SystemEstimator& system) const override;
+        virtual double logLikelihood(const Eigen::VectorXd& x, const SystemEstimator& system) const override;
+        virtual double logLikelihood(const Eigen::VectorXd& x,
+                                     const SystemEstimator& system,
+                                     Eigen::VectorXd& g) const override;
+        virtual double logLikelihood(const Eigen::VectorXd& x,
+                                     const SystemEstimator& system,
+                                     Eigen::VectorXd& g,
+                                     Eigen::MatrixXd& H) const override;
```
jpptm, line 104: suuuurely we can abstract this into a utility then all we have to do is provide a poiinter to a function defined in the module or something

module/localisation/FieldLocalisationSRIF/src/measurement/MeasurementFieldLines.hpp
```diff
@@ -0,0 +1,176 @@
+/**
```
jpptm, line 1: Should check if the greedy assignment ever fails and if it does, how often

module/localisation/FieldLocalisationSRIF/src/srif/FieldMap.cpp
```diff
@@ -0,0 +1,121 @@
...
+        switch (type) {
+            case LandmarkType::L_INTERSECTION: return landmarksL_;
+            case LandmarkType::T_INTERSECTION: return landmarksT_;
+            case LandmarkType::X_INTERSECTION: return landmarksX_;
+            case LandmarkType::GOAL_POST: return landmarksGoalPost_;
+        }
+        throw std::invalid_argument("FieldMap::landmarks: unknown LandmarkType");
+    }
+
+    void FieldMap::build() {
```
jpptm, line 21: Doesnt this already exist somewhere in the code base?

module/localisation/FieldLocalisationSRIF/src/srif/FieldMap.hpp
```diff
@@ -0,0 +1,153 @@
...
+     *
+     * Field names and defaults mirror
+     * NUbots/module/support/configuration/SoccerConfig/data/config/FieldDescription.yaml and
+     * NUbots/shared/message/support/FieldDescription.proto (message FieldDescription.FieldDimensions).
+     */
+    struct FieldDimensions {
+        // Populated from a message::support::FieldDescription via srif::field_dimensions()
+        // (FieldMapFromDescription.hpp), so FieldDescription.yaml is the single source of these values
+        // and the estimator core stays free of the message library.
+        double lineWidth;             ///< Width of field lines
```
jpptm, line 32: same with this here - this sould already be somewhere

module/localisation/FieldLocalisationSRIF/src/srif/FieldMapFromDescription.hpp
```diff
@@ -0,0 +1,45 @@
...
+     * @brief FieldDimensions from a NUbots FieldDescription message.
+     *
+     * Maps message::support::FieldDescription::FieldDimensions (snake_case, metres) onto the
+     * estimator's FieldDimensions. Fields the map does not use (goalpost cross-section, crossbar,
+     * net, etc.) are ignored.
+     */
+    inline FieldDimensions field_dimensions(const message::support::FieldDescription& fd) {
+        FieldDimensions dims;
+        dims.lineWidth            = fd.dimensions.line_width;
+        dims.fieldLength          = fd.dimensions.field_length;
```
jpptm, line 28: Same here

module/localisation/FieldLocalisationSRIF/src/measurement/MeasurementQuaternionNorm.hpp
```diff
@@ -0,0 +1,86 @@
...
+                                     const SystemEstimator& system,
+                                     Eigen::VectorXd& g) const override;
+        virtual double logLikelihood(const Eigen::VectorXd& x,
+                                     const SystemEstimator& system,
+                                     Eigen::VectorXd& g,
+                                     Eigen::MatrixXd& H) const override;
+
+        /// @brief Templated log-likelihood for autodiff.
+        template <typename Scalar>
+        Scalar logLikelihoodImpl(const Eigen::VectorX<Scalar>& x) const {
```
jpptm, line 71: This should really be abstracted and rewritten into snake case

module/localisation/FieldLocalisationSRIF/src/srif/FieldSamples.hpp
```diff
@@ -0,0 +1,57 @@
...
+        Pose<double> Hcw;  ///< world -> camera
+        std::vector<Detection> detections;
+    };
+
+    /**
+     * @brief A single message.vision.FieldLines sample (field-line points as camera rays)
+     */
+    struct LinePointsSample {
+        double t;          ///< image capture time [s since epoch]
+        int videoFrame;    ///< index into frameTimes / Left.mp4, or -1
```
jpptm, line 56: Should this be here?

shared/utility/gaussian_filtering/gaussian/GaussianBase.hpp
jpptm: This thing still exists!!!

jpptm (changes requested):
Left 22 / 64 files reviewed. Code volume could definitely decreased significantly imo

docker/Dockerfile
vincenttumminello: Boost is gone woohoo lol

PR #1826, opened 2026-08-06
Remove the `/configs` folder I accidentally added

miles-p:
In #1821, I added a `configs` folder by accident and never deleted it before I PR'ed it.

TLDR:
image

Here's the reason why:
1. At first in the `retrieve` tool, it would just copy all the configs back to `/configs`. It was simple.
2. Then I added logic for the configs to magically go back to your codebase so you could commit them straight from there
3. In my infinite wisdom, I never removed that folder
4. It got merged

### DIFF
--- a/configs/.gitignore
+++ b/configs/.gitignore
@@ -1,2 +1,0 @@
-*
-!.gitignore


JohanneMontano:
hey at least `retrieve` works as intended

JohanneMontano (approved):

miles-p:
fiwb

PR #1827, opened 2026-08-07
Make compatible with uv 0.12+

vincenttumminello:
This adds the `nuclear/pyproject.toml` as a member of the main uv project so that uv versions 0.12+ don't ignore the activate python environment and create another in `nuclear/` and then lead to python dependency errors since `nuclear/pyproject.toml` lists no dependencies to install.


jpptm (approved):
Pretty sure @TrentHouliston said this looks good, so lgtm too

PR #1829, opened 2026-09-17
Add a "block diagram" tool

miles-p:
Wouldn't it be cool if we could generate module block diagrams like this:
localisation_FieldLocalisationNLopt

Now we can!

Simply type `./b blackbox FieldLocalisationNLopt` to get an output.

Shoutout to my best friend, Fable 5.1 for writing this.


vincenttumminello:
Nice. I feel like the default output directory should be `recordings/blackbox` instead of `build/blackbox`

vincenttumminello:
This is also a super long script. Perhaps it can be condensed just a little??

miles-p:
> This is also a super long script. Perhaps it can be condensed just a little??

Claude knows best what are you on about

miles-p:
@vincenttumminello largest file is now 441. Claude broke it up into parts for me.

vincenttumminello (approved):
lgtm

